PR #44, opened 2018-09-18
BTRX - 333 - Admin manage Dul Modals

LeoForconesi:


src/components/modals/AddDulModal.js
```diff
@@ -57,87 +107,253 @@ export const AddDulModal = hh(class AddDulModal extends Component {
...
+      return true;
+    } else {
+      this.handleErrors(response);
+    }
+    return false;
+  };
+
+
+  handleErrors(message) {
+    var tle = 'Conflicts to resolve!';
```
dags: what tle means ?

src/pages/AdminManageDul.js
```diff
@@ -19,11 +18,14 @@ class AdminManageDul extends Component {
     this.state = {
       currentPage: 1,
       showModal: false,
+      editMode: false,
```
dags: is this still used ?

src/pages/AdminManageDul.js
```diff
@@ -326,11 +335,14 @@ class AdminManageDul extends Component {
                     button({
                       id: election.consentId + "_btn_editDUL",
                       className: "cell-button hover-color",
-                      onClick: this.editDul(election)
+                      onClick: () => this.editDul(election)
```
dags, line 337: review this, every click invokes a new instance ...

src/pages/DulReview.js
```diff
@@ -44,10 +44,11 @@ class DulReview extends Component {
       .then(data => {
         this.setState({election: data});
       });
-    Consent.ConsentResource(this.props.match.params.consentId)
-      .then(data => {
-        this.setState({consent: data});
-      });
+    this.setState({consent: Consent.ConsentResource(this.props.match.params.consentId)});
+      // .then(data => {
```
dags: remove not used lines

src/components/modals/AddDulModal.js
LeoForconesi: This stands for Title related to the Error alert title. I've change this to a better variable name.

src/pages/AdminManageDul.js
LeoForconesi: Removed

src/pages/AdminManageDul.js
LeoForconesi: This is done this way due to, onClick listener expects ```this.editDul``` be a function, if we pass the  instance value to the function the app enters into an infinite loop.

PR #83, opened 2018-10-02
Dataset catalog bugs

tadeork:


src/libs/ajax.js
```diff
@@ -342,7 +342,7 @@ export const DataSet = {
   deleteDataset: async (datasetObjectId, dacUserId) => {
     const url = `${await Config.getApiUrl()}/dataset/${datasetObjectId}/${dacUserId}`;
     const res = await fetchOk(url, _.mergeAll([Config.authOpts(), { method: 'DELETE' }]));
-    return await res.json();
+    return await res;
```
dags, line 345: this returns a whole response object and should be handled properly : resp.status may be any 200 , 401, 403, 500 . It seems it is not being handled properly where used.

tadeork: Added error messages.

dags (approved):

PR #84, opened 2018-10-02
General Fix

vvicario:


src/libs/ajax.js
```diff
@@ -193,7 +193,7 @@ export const DAR = {
   deletePartialDarRequest: async (darId) => {
     const url = `${await Config.getApiUrl()}/dar/partial/${darId}`;
     const res = await fetchOk(url, _.mergeAll([Config.authOpts(), { method: 'DELETE' }]));
-    return await res.json();
+    return await res;
```
dags, line 196: this returns a whole response object and should be handled properly : resp.status may be any 200 , 401, 403, 500 . It seems it is not being handled properly where used.

dags (approved):

PR #151, opened 2019-02-28
fixing configs

dags:
Intermediate step on fixing configuration handling.

### DIFF
--- a/config/prod.json
+++ b/config/prod.json
@@ -1,0 +1,8 @@
+{
+  "apiUrl": "https://consent.dsde-dev.broadinstitute.org/api",
+  "ontologyApiUrl": "https://consent-ontology.dsde-dev.broadinstitute.org/",
+  "clientId": "1087760099392-u18o6sqvb8uljmdltflohmd34h6hik9b.apps.googleusercontent.com",
+  "firecloudUrl": "https://firecloud-orchestration.dsde-dev.broadinstitute.org/",
+  "gwasUrl": "",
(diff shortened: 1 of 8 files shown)


config/prod.json
```diff
@@ -0,0 +1,8 @@
+{
+  "apiUrl": "https://consent.dsde-dev.broadinstitute.org/api",
```
rushtong, line 2: Should these be the prod urls? Replace `dsde-dev` with `dsde-prod`

config/prod.json
```diff
@@ -0,0 +1,8 @@
...
+  "ontologyApiUrl": "https://consent-ontology.dsde-dev.broadinstitute.org/",
+  "clientId": "1087760099392-u18o6sqvb8uljmdltflohmd34h6hik9b.apps.googleusercontent.com",
```
rushtong, line 4: This will need to change for prod - its in the dev project now. Fine for this PR, but we need to remember when setting up the prod env.

config/prod.json
```diff
@@ -0,0 +1,8 @@
...
+  "firecloudUrl": "https://firecloud-orchestration.dsde-dev.broadinstitute.org/",
+  "gwasUrl": "",
+  "nihUrl": "http://mock-nih.dev.test.firecloud.org/link-nih-account/index.html"
```
rushtong, line 7: This should point to the prod shiboleth instance too.

rushtong (approved):
Minor suggestions, but 👍

rushtong (approved):
Still 👍

PR #160, opened 2019-05-20
DUOS-338  MultiDAC: DUOS Add DAC Name column to Manage sDUL and Manage DAR pages

nlopezzalba:
As an Admin, I want to a see a column with the DAC Name on Manage DAR and Manage DUL table pages


rushtong (approved):
👍

PR #161, opened 2019-05-21
DUOS-339  MultiDAC: Admin Selects DAC when Creating an sDUL Election

nlopezzalba:
Now, when an Admin opens a DUL election they must select a DAC to vote on the sDULs from a dropdown box with the current DUOS DACs as options, from the pop-up box which appears to confirm the opening of the election.

### DIFF
--- a/src/components/ConfirmationDialog.js
+++ b/src/components/ConfirmationDialog.js
@@ -1,7 +1,6 @@
 import { Component } from 'react';
-import { button, div, h, h2, span, hh, label } from 'react-hyperscript-helpers';
+import { button, div, h, h2, span, hh } from 'react-hyperscript-helpers';
 import Modal from 'react-modal';
-import AsyncSelect from 'react-select/lib/Async';
 import { Alert } from '../components/Alert';
 import './ConfirmationDialog.css';
 
@@ -30,10 +29,6 @@ const customStyles = {
     borderRadius: '4px',
     outline: 'none',
     padding: '10px 20px 20px 20px',
-  },
-
-  labelMargin: {
-    margin: '10px 0 5px 0', textAlign: 'left', float: 'left'
   }
 };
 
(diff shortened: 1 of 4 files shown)


nlopezzalba:
@rushtong Not quite sure about this change... ConfirmationDialog is being used from a lot of different places and so far no form field was ever necessary (for those cases we use Modals instead). Even though this new select field will only be visible if certain prop is defined, IMO this may "dirty" the component a bit.

src/components/ConfirmationDialog.js
```diff
@@ -75,6 +81,28 @@ export const ConfirmationDialog = hh(class ConfirmationDialog extends Component
...
+                    // onChange: (option) => this.onDacsChange(option),
+                    // value: this.state.dacName,
+                    // noOptionsMessage: () => this.state.optionMessage,
+                    // loadingMessage: () => this.state.optionMessage,
+                    classNamePrefix: "select",
+                    placeholder: "Select DAC...",
+                    className: 'select-autocomplete',
+                    required: true
+                  })
+              ]),
```
rushtong: I see what you mean here. There is a bunch of extra content here that is conditionally rendered. Instead of defining that here, I think it would make sense if the component accepted any number of child divs. In all cases where we call the component like we did before, nothing would be rendered. In the case where we call this dialog from this new location, that is where we would specify the div as another thing to be rendered. Will look for some examples that might illustrate the point.

rushtong: I tested this locally and that entire div (minus the render) can go in AdminManageDul right before the `div({ className: "dialog-description" }` line which is 564 in this PR.

rushtong: Actually, there is a slight difference between having the div in one place vs the other. Not sure if my suggestion will meet the requirement, but it does allow for a cleaner confirmation dialog

nlopezzalba: yes, we can put the select in there, whatever we put below 564 will be rendered as "child" in the dialog. Didn't do that in the first place cause I wasn't sure how the field will be validated if empty from the dialog.

nlopezzalba: @rushtong please take a look to see if this makes more sense now

rushtong (approved):
LGTM!

PR #162, opened 2019-05-24
  DUOS-363 DUOS DAR Application UI - Remove character limits for 2.2 RUS and 2.3 NTS

nlopezzalba:
Description
There are currently character limits on the Research Use Statements and Non-Technical Summary (2.2. and 2.3) in the DAR Application Step 2 page in DUOS. We would like to remove these character limits.

expected behavior: researchers can enter an unlimited amount of text in the RUS and NTS. The DAC reviewing the DAR should also be able to see all text entered in the RUS and NTS and not have character limits either.

Note: character limit will be removed but in the placeholders we're still encouraging to try to limit to certain amount of characters.


rushtong (approved):
👍  Merging

PR #166, opened 2019-06-13
[DUOS-378][risk=no] Staging deploy

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-378

## Changes
* Update staging and prod configs
* Add staging build/deploy jobs/tasks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,16 +204,15 @@ jobs:
           env: 'staging'
           project_name: 'duos-ui'
           google_project: 'broad-duos-dev'
-# TODO: Uncomment when cluster is properly configured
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -25,6 +25,7 @@ commands:
         type: string
         default: 'broad-duos-dev'
     steps:
+      - run: cat public/config.json
```
dags: this is for debugging  ?

dags (approved):

.circleci/config.yml
rushtong: Yes, it is - will remove this.

PR #167, opened 2019-06-14
Moving configs to dotenv

dags:
Configuration moved to ENV variables, read by dotenv.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "bootstrap": "3.4.1",
     "chart.js": "1.1.1",
     "classnames": "2.2.6",
+    "dotenv": "^8.0.0",
     "jquery": "3.4.0",
     "lodash": "4.17.11",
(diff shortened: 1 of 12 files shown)


rushtong:
This changes how we would execute the deploys to kubernetes. We will need one of two things:
1. `.env` files and changes to the circleci yaml to move those around to the right places
2. A separate job/process (jenkins?) that generates a configmap deploy to the clusters so the `env` arguments are populated correctly.

I suggest we close this PR for now so we can write up a tech-doc (along with jira tickets) on how to convert the current deploys.

tadeork (approved):

rushtong (changes requested):
As mentioned in comments, this will break the current deployment. Can we close this, but keep the branch around, and work on a technical design that covers both this subject and the deployment?

rushtong:
Closing until we can revisit the overall process.

PR #169, opened 2019-06-20
[DUOS-380][risk=no] Show the right help links for the support link

rushtong:
## Addresses
Ticket https://broadinstitute.atlassian.net/browse/DUOS-380

## Changes
This was a little more complicated than I envisioned. We're doing something funky with how we're storing user login state, and that's not using state. This PR adds that to the app level as a state variable so that children components can re-render based on that.

Additionally, our Routes are not passing props correctly. The only props that are passed are the default Route ones and anything passed into `Routes` is ignored, except for the `Login` component, which this PR fixes. Will address the other routes in a different task. 

The end result is that the footer can now conditionally render the right `Support` link depending on the user's logged in state.

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -41,7 +41,7 @@ const Routes = ( props ) => (
   <Switch>
     <Route exact path='/' component={Home} props={props} />
     <Route exact path='/home' component={Home} props={props} />
-    <Route path='/login' component={Login} props={props} />
+    <Route path='/login' render={(routeProps) => <Login {...routeProps} {...props} />} />
     <Route path='/home_help' component={HomeHelp} />
     <Route path='/home_about' component={HomeAbout} />
     <Route path='/home_register' component={HomeRegister} props={props} />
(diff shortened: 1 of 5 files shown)


src/Routes.js
```diff
@@ -41,7 +41,7 @@ const Routes = ( props ) => (
   <Switch>
     <Route exact path='/' component={Home} props={props} />
     <Route exact path='/home' component={Home} props={props} />
-    <Route path='/login' component={Login} props={props} />
+    <Route path='/login' render={(routeProps) => <Login {...routeProps} {...props} />} />
```
rushtong, line 44: Will deal with all of the other routes in a separate PR.

src/components/DuosHeader.js
```diff
@@ -8,8 +8,6 @@ import './DuosHeader.css';
 
 class DuosHeader extends Component {
 
-  navBarCollapsed = true;
```
rushtong, line 11: Was unused

rushtong:
Have shopped this around to no avail. Merging and will deal with fall-out as needed.

PR #171, opened 2019-07-02
[DUOS-395][risk=no] Reference info for DAR & Dataset to DAC Review Pages

jaime-bernales-broad:
## Addresses
See https://broadinstitute.atlassian.net/browse/DUOS-395
See also: https://github.com/DataBiosphere/consent/pull/368

## Changes:
Updated the DAR summary at the top of the page for the following pages (some of the screenshots are older ... the dataset id and research use statements are fixed in all of them, will update soon):
- Access Preview
- Vote (Access Review)
- Collect Votes (Access Collect - DAC Chair only)
- Final Vote (Final Access - DAC Chair only)
- Review Record (Access Results)

Updated the underlying DAR ajax call to streamline it and make use of the new API updates.

### Access Preview
Screen Shot 2019-10-02 at 6 01 47 PM

### Vote (Access Review)
Screen Shot 2019-10-03 at 1 44 22 PM

### Collect Votes (Access Collect)
Screen Shot 2019-10-03 at 1 46 49 PM

### Final Vote (Final Access)
Screen Shot 2019-10-03 at 2 04 12 PM

### Review Record (Access Results)
Screen Shot 2019-10-02 at 6 02 50 PM


src/pages/AccessCollect.js
```diff
@@ -313,6 +314,7 @@ class AccessCollect extends Component {
     let dar = await DAR.getDarFields(this.props.match.params.referenceId, "rus");
     let request = await DAR.getDarFields(this.props.match.params.referenceId, "projectTitle");
     let darInfo = await DAR.describeDar(this.props.match.params.referenceId);
+    console.log('dar: ' + JSON.stringify(darInfo));
```
dags: Remove console.log

src/components/DarDetails.js
```diff
@@ -0,0 +1,39 @@
...
+        
+                    span({}, [
+                        this.props.consentName
+                    ])
+                ])
+            ])
+        )
+    }
+
+});
```
rushtong: nit: newline

src/components/DarDetails.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const DarDetails = hh(class DarDetails extends PureComponent {
+    
+    constructor(props) {
+        super(props);
+    }
+
+    render() {
+        return (
+            span({className: "consent-data-alt"}, [
```
rushtong: I know you're following a the current pattern, but the current pattern is one I really want to get away from - it's very much an artifact of the conversion from angular. Instead of defining css class names and css classes that are outside of the component, I strongly prefer to reference style objects in the component itself. Something like this:

Define a styles constant in your component somewhere:
```
const styles = {
  span: {
    'font-size': 19,
    'font-style': 'normal',
    'margin': '10px 0 30px 56px',
    'display': 'block',
    'color': '#333333',
    'font-weight': 'normal',
    'line-height': 2
  }
};
```
Then just reference it when you need it:
```
     span({style: styles.span}
```

Pros:
* The styles are closer to the code that uses them - much easier to tell if something is unused or not.
* It keeps the component isolated and free from visual side effects of multiple developers working on shared css files.
* Allows for IDEs to do refactoring since they are working with javascript objects which they are better at doing than css string values.

src/components/PageHeading.css
```diff
@@ -40,3 +40,13 @@
...
+
+.consent-data-alt {
+  font-size: 19px;
+  font-style: normal;
+  margin: 10px 0 30px 56px;
+  display: block;
+  color: #333333;
+  font-weight: normal;
+  line-height: 27px;
+}
```
rushtong: With my suggestion above, this could go away.

src/components/DarDetails.js
```diff
@@ -0,0 +1,39 @@
+import { PureComponent } from 'react';
+import { div, hh, label, h4, h, hr, span } from 'react-hyperscript-helpers';
+import * as Utils from '../libs/utils';
```
rushtong: Unused?

src/components/DarDetails.js
```diff
@@ -0,0 +1,39 @@
+import { PureComponent } from 'react';
+import { div, hh, label, h4, h, hr, span } from 'react-hyperscript-helpers';
```
rushtong: `label, h4, h, hr` look unused

rushtong (commented):
Looks good, some minor comments inline. Be sure to look at the eslint warnings ... there are a number of them in this PR.

Screen Shot 2019-07-02 at 6 01 54 PM

jaime-bernales-broad:
@rushtong branch updated

src/components/DarDetails.js
rushtong: Not sure why this was resolved - the file still needs a newline here.

src/components/DarDetails.js
```diff
@@ -0,0 +1,48 @@
...
+import { div, hh, span } from 'react-hyperscript-helpers';
+
+const styles = {
+    consentDataAlt: {
+        'font-size': 19,
+        'font-style': 'normal',
+        'margin': '10px 0 30px 56px',
+        'color': '#333333',
+        'font-weight': 'normal',
+        'line-height': '27px'
```
rushtong: All of these dashed style properties should be camel cased: 

Screen Shot 2019-07-11 at 1 39 33 PM

src/pages/AccessResultRecords.js
```diff
@@ -180,8 +181,14 @@ class AccessResultRecords extends Component {
...
               id: "recordAccess", imgSrc: "/images/icon_access.png", iconSize: "medium", color: "access",
-              title: "Data Access - Results Record", description: consentData
-            }),
+              title: "Data Access - Results Record"}),
+              DarDetails({
+                projectTitle: this.state.projectTitle,
+                darCode: this.state.darInfo.darCode,
+                datasetId: this.state.darInfo.datasetId,
+                datasetName: this.state.darInfo.datasetName,
+                consentName: this.state.consentName
```
rushtong: In testing, I'm seeing undefined values here.

src/pages/AccessResultRecords.js
```diff
@@ -180,8 +181,14 @@ class AccessResultRecords extends Component {
           div({ className: "col-lg-10 col-md-9 col-sm-9 col-xs-12 no-padding" }, [
             PageHeading({
               id: "recordAccess", imgSrc: "/images/icon_access.png", iconSize: "medium", color: "access",
-              title: "Data Access - Results Record", description: consentData
-            }),
+              title: "Data Access - Results Record"}),
+              DarDetails({
+                projectTitle: this.state.projectTitle,
+                darCode: this.state.darInfo.darCode,
+                datasetId: this.state.darInfo.datasetId,
```
rushtong: Also seeing strange behavior here, it's evaluating to `--` instead of the dataset id

src/pages/AccessResultRecords.js
rushtong: I see what's happening here. `consentName` is not a state variable on `AccessResultRecords`, but it is on `AccessCollect`.

rushtong (dismissed):
See comments inline. This PR is not fully tested - please retest and let me know when it's ready for review.

src/pages/AccessResultRecords.js
jaime-bernales-broad: I can't reproduce

rushtong: copy config/dev.json to public/config.json
npm install
npm start

Navigate to Admin Console
Navigate to Manage DAR page
Click the Approved button for the first DAR, DAR-12

Screen Shot 2019-07-15 at 7 52 09 AM

jaime-bernales-broad: @rushtong datasetID is not coming from back-end (image below), when I create from zero in my local this info appears, but doesn't appear on an existing DAR using the setting you provided. 

image

Note: I figured out what I mention above just because Vero's help. It surpasses my skills (designing and mocking HTML)

rushtong: If the back end doesn't support this, then it isn't ready for merging. Can you add a new story for implementing the back end requirement for this? We can leave this branch in place until that is complete.

jaime-bernales-broad: https://broadinstitute.atlassian.net/browse/DUOS-410?atlOrigin=eyJpIjoiOTFlZDA5MjBkN2Y2NDk5MmJiZmNjOTBkNjkyNWMwYmMiLCJwIjoiaiJ9

rushtong: I think I see the problem now that I've picked this up and am re-working it. We're making horribly inefficient API calls in the UI to get very narrow and specific pieces of DAR information. See `async showDarData(electionReview)` in `AccessResultsRecord.js` for an example. We're making at least 3 API calls to get info, then do a ton of calculations on that data. We could, instead, get the whole DAR in one request, but we've broken that up for some reason. I'm pretty sure we don't need any additional back end API calls to make this work.

src/components/DataAccessRequest.js
```diff
@@ -0,0 +1,57 @@
+import _ from 'lodash';
+import { Component } from 'react';
+import { div, hh, span } from 'react-hyperscript-helpers';
+
+
+const parseAlias = (alias) => {
```
dags, line 6: alternative function body: 

```
const prefix = 'DUOS-'; 
const pad = '000000'; 
return  prefix + (pad+n).slice(-pad.length);
```

src/pages/AccessReview.js
```diff
@@ -37,8 +41,9 @@ class AccessReview extends Component {
             return prev;
           });
         }).catch(error => {
-          this.setState({ showConfirmationDialogOK: true, alertMessage: "Sorry, something went wrong when trying to submit the vote. Please try again." });
-        });
+        this.setState(
```
dags: indentation ?

src/pages/AccessReview.js
```diff
@@ -49,8 +54,9 @@ class AccessReview extends Component {
...
```
dags: indentation here ?

src/pages/AccessReview.js
```diff
@@ -72,8 +78,9 @@ class AccessReview extends Component {
           });
           this.alertRPVoteRemember();
         }).catch(error => {
-          this.setState({ showConfirmationDialogOK: true, alertMessage: "Sorry, something went wrong when trying to submit the vote. Please try again." });
-        });
+        this.setState(
```
dags: indentation here ?

dags (approved):

src/pages/AccessReview.js
rushtong: Yes, should be indented. Intellij's auto-formatter seems to be moving these back after make the changes. Will look at that setting.

rushtong: Actually, no, the auto-formatter is correct here. It's confusing to the eye due to the previous line, `}).catch(error => {`. I'll update the blocks so it looks better and the formatter works.

src/components/DataAccessRequest.js
rushtong: Interesting approach - will investigate. I originally stole this from https://github.com/DataBiosphere/consent/blob/3d8cd299580ce729d1cd57652843ef9b8448047d/src/main/java/org/broadinstitute/consent/http/util/DatasetUtil.java#L7

rushtong: Done!

PR #186, opened 2019-07-31
[DUOS-419][risk=no] Show configuration error when trying to sign in

rushtong:
## Addresses
See https://broadinstitute.atlassian.net/browse/DUOS-419
We had a client id configuration error for staging, but the error was being swallowed and nothing was displaying in the UI leaving the user wondering what was happening. This PR adds:
* A new oauth client id configured for the staging cluster.
* Show error instead of letting the user try to sign in.

## Error case
Screen Shot 2019-07-31 at 11 46 17 AM

## Success case
Screen Shot 2019-07-31 at 11 47 46 AM

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -11,10 +11,6 @@ export const formatDate = (dateval) => {
   return datestr;
 };
 
-export const isObjectEmpty = (obj) => {
-  return Object.getOwnPropertyNames(obj).length === 0;
-};
-
 export const USER_ROLES = {
   admin: 'Admin',
   chairperson: 'Chairperson',
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -11,6 +11,10 @@ export const formatDate = (dateval) => {
   return datestr;
 };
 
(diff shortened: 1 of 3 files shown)


vvicario (approved):

LeoForconesi (approved):
Looking good! 
I'll grab [this]( https://github.com/DataBiosphere/duos-ui/pull/186/files#diff-292902966ff83b931071dc57fc60a179R15) and slowly walk away to my code. :smile:

src/libs/utils.js
```diff
@@ -11,6 +11,10 @@ export const formatDate = (dateval) => {
   return datestr;
 };
 
+export const isObjectEmpty = (obj) => {
```
dags: Could not use lodash's  _.isEmpty() ?

rushtong: Thank you - good idea. I am not that familiar with lodash ... need to spend some time with it.

dags: me too. I just saw it is on the project and have that function.

dags (approved):

PR #193, opened 2019-08-26
[DUOS-431][risk=no] Handle upload DAA error

rushtong:
## Addresses
Alongside https://github.com/DataBiosphere/consent/pull/360, this PR addresses https://broadinstitute.atlassian.net/browse/DUOS-431
Neither PR is dependent upon one another.

## Changes
There exists an edge case where the current profile's DAA is undefined, but the profile is complete. That case was sending invalid data to the API (`undefined` as the url for the existing DAA). This PR fixes this on the front end. 

Additionally, changed a button to a div to prevent unwanted page refreshes from happening underneath the file selection dialog box.


src/pages/ResearcherProfile.js
```diff
@@ -461,7 +461,8 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
 
         } else {
           if (this.state.file !== undefined && this.state.file.name !== '') {
-            DAR.postDAA(this.state.file.name, this.state.file, profile.urlDAA).then(response => {
+            const existentDAAUrl = (profile.urlDAA === undefined) ? '' : profile.urlDAA;
```
dags: checking only for "undefined" is enough, what about "null" ?

rushtong: Good idea ... lodash has a method for that now: https://lodash.com/docs/4.17.15#isNil

dags: nice !

dags (approved):

src/pages/ResearcherProfile.js
rushtong: There's something else going on here that happened to me earlier in development, but then went away for some reason. Clicking the upload button triggers a refresh of the page so the file upload dialog becomes disconnected to the original form element that triggered it. Will add a WIP to this and dig back in.

rushtong:
@dags Ready for re-review. I updated the surrounding box to a div and styled it so it looked like the button tag that it replaces.

dags (approved):

PR #204, opened 2019-09-17
[DUOS-452][risk=no] Fix download links for FF

rushtong:
## Addresses
Fixes the UI component of https://broadinstitute.atlassian.net/browse/DUOS-452
Tested on both FF and Chrome (original approach was not broken on Chrome).


dags (approved):

vvicario (approved):
I took it as reference for ORSP and it works perfect!

PR #206, opened 2019-09-19
[DUOS-416][risk=no] DAC member management

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-416
Requires https://github.com/DataBiosphere/consent/pull/354 for the API work.

## Changes
Additions to the DAC resource for handling DAC membership. This PR adds functionality missing from the UI mock-ups, most of which is found in the [dacManagement branch](https://github.com/DataBiosphere/duos-ui/blob/dacManagement).

### Manage DACs Changes
Added view and edit actions for a DAC
Screen Shot 2019-09-27 at 10 22 56 AM

### View Members
Screen Shot 2019-09-27 at 10 23 06 AM

### Edit DAC
Added section that displays the current users and enabled the autocomplete for chair persons and members. If a current chair or member is slated for removal, they are not filtered and allowed to be re-added in with any role.
Screen Shot 2019-09-27 at 10 23 20 AM

Autocomplete filters out users that are currently chairs/members.
Screen Shot 2019-09-27 at 10 32 20 AM

An example where we are swapping member and chair persons.
Screen Shot 2019-09-27 at 10 23 55 AM


src/components/modals/AddDacModal.js
```diff
@@ -1,27 +1,42 @@
...
     this.OKHandler = this.OKHandler.bind(this);
+    this.userSearch = this.userSearch.bind(this);
+    this.onChairSearchChange = this.onChairSearchChange.bind(this);
+    this.onMemberSearchChange = this.onMemberSearchChange.bind(this);
+    this.removeChairperson = this.removeChairperson.bind(this);
+    this.removeMember = this.removeMember.bind(this);
+    this.handleChange = this.handleChange.bind(this);
   }
 
   componentDidMount() {
```
dags: not used, should be removed ?

rushtong: Still very much a work in progress, not ready for review.

src/components/modals/AddDacModal.js
```diff
@@ -1,27 +1,42 @@
...
+
 
 export const AddDacModal = hh(class AddDacModal extends Component {
 
   constructor(props) {
     super(props);
     this.state = {
       isEditMode: this.props.isEditMode,
-      dac: this.props.isEditMode ? this.props.dac : {},
+      dac: this.props.isEditMode ? this.props.dac : { dac: {}, chairpersons: [], members: [] },
```
dags: this.props.dac have same { dac: {}, chairpersons: [], members: [] } structure ?  So, there is a state.dac.dac element ?

rushtong: See previous response.

src/components/DacUsers.js
```diff
@@ -0,0 +1,85 @@
...
+      removeButton: props.removeButton,
+      removeHandler: props.removeButton ? props.removeHandler : () => {},
+      removedIds: []
+    };
+    this.onRemove = this.onRemove.bind(this);
+    this.columnClass = this.columnClass.bind(this);
+    this.makeRow = this.makeRow.bind(this);
+  }
+
+  onRemove(dacId, dacUserId, type) {
```
dags, line 25: Just confirming: this works like a toggle ? if dacUserId exists in  removedIds, remove it from removedIds, otherwise add it. ?

rushtong: It does function like a toggle, yes. This allows the component to maintain who is selected for removal and who isn't. Those that are removed can be re-added (with same role or different, doesn't matter) and those who are not removed (i.e. they are unchanged), cannot be re-added with a different role.

dags (approved):

PR #209, opened 2019-10-07
[no jira][risk=no] UI Status Page

rushtong:
## Changes
* Add a status page that iterates over consent, ontology, and firecloud statuses.

Screen Shot 2019-10-08 at 10 38 11 AM


src/Routes.js
```diff
@@ -1,81 +1,97 @@
...
-    <AuthenticatedRoute path='/dul_review/:voteId/:consentId' component={DulReview} props={props} rolesAllowed={[USER_ROLES.member, USER_ROLES.chairperson]} />
-    <AuthenticatedRoute path='/dul_preview/:consentId' component={DulPreview} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson]} />
-    <AuthenticatedRoute path='/dul_collect/:consentId' component={DulCollect} props={props} rolesAllowed={[USER_ROLES.chairperson, USER_ROLES.admin]} />
-    <AuthenticatedRoute path='/final_access_review/:referenceId/:electionId' component={FinalAccessReview} props={props} rolesAllowed={[USER_ROLES.chairperson]} />
-    <AuthenticatedRoute path='/reviewed_cases' component={ReviewedCases} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.member, USER_ROLES.alumni]} />
-    <Route path='*' component={NotFound} />
+      <Route exact path="/" component={ Home } props={ props }/>
+      <Route exact path="/home" component={ Home } props={ props }/>
+      <Route exact path="/status" render={ (routeProps) => Status(_.mergeAll([routeProps, props])) }/>
```
rushtong, line 48: This is the only real change here - the rest is eslint reformatting

src/components/DuosFooter.js
```diff
@@ -13,6 +14,7 @@ class DuosFooter extends Component {
             li({ className: "footer-links__item" }, [a({ target: '_blank', href: "https://www.broadinstitute.org/privacy-policy" }, ["Privacy Policy"]),]),
             li({ className: "footer-links__item" }, [a({ target: '_blank', href: "https://www.broadinstitute.org/terms-conditions" }, ["Terms of Service"]),]),
             li({ className: "footer-links__item" }, [a({ href: supportLink }, ["Support"]),]),
+            li({ className: 'footer-links__item' }, [a({ href: 'status' }, ['Status'])])
```
nlopezzalba: just out of curiosity... why is status a string in href?

src/Routes.js
nlopezzalba: okie

src/components/DuosFooter.js
rushtong: I would have liked to make this a `Link`, but the footer is outside the routes structure. Actually, I think this might be a bug ... `status` will break on deep page links. Will make this root relative.

nlopezzalba (approved):

PR #213, opened 2019-11-15
[DUOS-471][risk=no] Enforce eRA Auth for DAR submissions

rushtong:
## Addresses
- https://broadinstitute.atlassian.net/browse/DUOS-471
- https://broadinstitute.atlassian.net/browse/DUOS-454

## Changes
- Lint formatting on the changed files results in lots of unrelated white space and quote changes.
- Extract all common code from the profile and data access request pages that both handle era commons authentication into a new component
  - Minor refactoring around the order of operations for auth
- Fix the remove era commons bug where removal didn't immediately display on-screen
- Add a callback handler for any parent components that might need to know the state of current authentication
- Enforce only eRA commons of the 4 different RI types. Other 3 are now optional
- Did *not* enforce in profile, but mark as required - currently only required for DAR submission.

## DAR Submission Enforcement
Screen Shot 2019-11-20 at 10 25 08 AM

## Profile Language Update
Screen Shot 2019-11-18 at 8 40 34 AM

### DIFF
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -15,8 +15,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
       isAuthorized: false,
       expirationCount: 0,
       nihUsername: '',
-      nihError: false,
-      validationError: _.isNil(props.validationError) ? false : props.validationError
+      nihError: false
     };
     this.authenticateAsNIHFCUser = this.authenticateAsNIHFCUser.bind(this);
     this.getResearcherProperties = this.getResearcherProperties.bind(this);
@@ -115,12 +114,12 @@ export const eRACommons = hh(class eRACommons extends React.Component {
   }
 
   render() {
-    const validationErrorStyle = this.state.validationError ? {
+    const validationErrorStyle = _.isNil(this.props.validationError) ? {} : {
       color: "#D13B07",
       border: "1px solid #D13B07",
       borderRadius: 5,
       padding: 6
-    } : {};
+    };
     const nihErrorMessage = 'Something went wrong. Please try again.';
     return (
       div({ className: this.props.className }, [
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -104,7 +104,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
   };
 
   render() {
-    const validationErrorStyle = (!_.isNil(this.props.validationError) && this.props.validationError) ? {
+    const validationErrorStyle = _.get(this.props, 'validationError', false) ? {
       color: "#D13B07",
       border: "1px solid #D13B07",
       borderRadius: 5,
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -40,10 +40,9 @@ export const eRACommons = hh(class eRACommons extends React.Component {
     let isFcUser = this.verifyUser();
     if (!isFcUser) {
       Researcher.getPropertiesByResearcherId(currentUserId).then(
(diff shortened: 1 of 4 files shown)


src/pages/DataAccessRequestApplication.js
```diff
@@ -247,19 +179,16 @@ class DataAccessRequestApplication extends Component {
       formData.zipcode = rpProperties.zipcode != null ? rpProperties.zipcode : '';
       formData.country = rpProperties.country != null ? rpProperties.country : '';
       formData.state = rpProperties.state != null ? rpProperties.state : '';
-      formData.piName = rpProperties.piName !== null ? rpProperties.piName : '' ;
+      formData.piName = rpProperties.piName !== null ? rpProperties.piName : '';
```
tadeork, line 182: I think this can be improved with `get(rpProperties, "piName", "");` from lodash.

rushtong: Good call. Most of this was copy/paste from the other components ... I can clean it up some more.

rushtong: Ah ... re-reading this ... this was a lint checking change. I prefer to leave as much of the old code in place if it isn't directly related to the era commons stuff.

src/components/eRACommons.js
```diff
@@ -0,0 +1,212 @@
...
+    if (isFcUser) {
+      const parsedToken = qs.parse(searchArg);
+      this.verifyToken(parsedToken).then(
+        (decodedNihAccount) => {
+          AuthenticateNIH.saveNihUsr(decodedNihAccount).then(
+            () => this.getResearcherProperties(),
+            () => this.setState({ nihError: true })
+          );
+        },
+        () => this.setState({ nihError: true })
```
tadeork, line 54: The second fn on the `then` is for the reject case of the promise?
Why then is needed the catch? Can you explain this please?

src/components/eRACommons.js
```diff
@@ -0,0 +1,212 @@
...
+      );
+    }
+  };
+
+  getResearcherProperties = async () => {
+    const currentUserId = Storage.getCurrentUser().dacUserId;
+    Researcher.getPropertiesByResearcherId(currentUserId).then(response => {
+      const isAuthorized = _.isNil(response.eraAuthorized) ? false : JSON.parse(response.eraAuthorized);
```
tadeork: I think here would be more readable with `get()` from lodash. Unless I missing something here.

src/components/eRACommons.js
rushtong: I don't see the catch in this method you're referring to. What line is that?

tadeork (approved):
Some questions inline. Looks good.

src/components/eRACommons.js
tadeork: Not the catch, sorry. The line 54 looks the same as 51.

src/pages/DataAccessRequestApplication.js
tadeork: It makes sense.

src/components/eRACommons.js
rushtong: They're handling different error cases.
- Line 51 handles errors in `AuthenticateNIH.saveNihUsr`
- Line 54 handles errors in `this.verifyToken`

So, `verifyToken` can error out even if `saveNihUsr` succeeds, so we need to handle both cases.

tadeork: Yes! I see it now. Thanks Greg. :+1:

PR #216, opened 2019-12-04
[DUOS-319][risk=no] Reorder Columns

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-319
Recommended to hide whitespace when viewing diffs.

## Changes
Reorder columns. New display order:
* Dataset ID
* Dataset Name
* dbGAP
* Structured Data Use Limitations
* Data Type
* Phenotype/Indication
* Principal Investigator
* \# of participants
* Description
* Species
* Data Depositor
* Consent ID
* SC-ID
* Approved Requestors

### Old Screen
Screen Shot 2019-12-04 at 8 19 44 AM

### New Screen
Left hand side of the table:
Screen Shot 2019-12-05 at 10 02 05 AM

Right hand side of the table:
Screen Shot 2019-12-05 at 10 02 14 AM

### DIFF
--- a/src/pages/DatasetCatalog.js
+++ b/src/pages/DatasetCatalog.js
@@ -29,7 +29,6 @@ class DatasetCatalog extends Component {
       allChecked: false,
       dataSetList: {
         catalog: [],
-        dictionary: [],
         showDialogDelete: false,
         showDialogEnable: false,
         showDialogDisable: false,
@@ -64,20 +63,14 @@ class DatasetCatalog extends Component {
         this.props.history.push('profile');
       }
     } else {
-      const dictionary = await DataSet.findDictionary();
       let catalog = await DataSet.findDataSets(this.USER_ID);
       catalog.forEach((row, index) => {
         row.checked = false;
         row.ix = index;
       });
 
-      const data = {
-        catalog: catalog,
-        dictionary: dictionary
-      };
-
       this.setState({
-        dataSetList: data,
+        dataSetList: { catalog: catalog },
         currentPage: 1
       });
     }
--- a/src/pages/DatasetCatalog.js
+++ b/src/pages/DatasetCatalog.js
@@ -48,13 +48,12 @@ class DatasetCatalog extends Component {
     this.openConnectDataset = this.openConnectDataset.bind(this);
     this.closeConnectDatasetModal = this.closeConnectDatasetModal.bind(this);
     this.okConnectDatasetModal = this.okConnectDatasetModal.bind(this);
-    this.openTranslatedDUL = this.openTranslatedDUL.bind(this);
     this.closeTranslatedDULModal = this.closeTranslatedDULModal.bind(this);
     this.okTranslatedDULModal = this.okTranslatedDULModal.bind(this);
 
     this.download = this.download.bind(this);
-    this.selectAll = this.selectAll.bind(this);
     this.exportToRequest = this.exportToRequest.bind(this);
+    this.defaultString = this.defaultString.bind(this);
   }
 
   async getDatasets() {
@@ -316,6 +315,11 @@ class DatasetCatalog extends Component {
     });
   };
 
+  defaultString(str, defaultStr) {
+    if (str.length === 0) { return defaultStr; }
+    return str;
+  };
+
   render() {
 
     const { searchDulText, currentPage, limit } = this.state;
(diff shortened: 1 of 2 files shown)


src/pages/DatasetCatalog.js
```diff
@@ -438,18 +438,18 @@ class DatasetCatalog extends Component {
...
+                                span({ id: trIndex + "_dataType", name: "dataType", isRendered: property.propertyName === 'Data Type' }),
+                                span({ id: trIndex + "_species", name: "species", isRendered: property.propertyName === 'Species' }),
                                 property.propertyValue
                               ]),
 
                               a({
                                 id: trIndex + "_linkdbGap",
                                 name: "link_dbGap",
-                                isRendered: property.propertyName === 'dbGAP',
+                                isRendered: property.propertyName === 'Repository',
```
rushtong: This looks like a bug. Without the right property name here, the dbgap link doesn't display.

src/pages/DatasetCatalog.js
```diff
@@ -438,18 +438,18 @@ class DatasetCatalog extends Component {
...
-                                span({ id: trIndex + "_dataType", name: "dataType", isRendered: property.propertyName === 'Data Type' }),
-                                span({ id: trIndex + "_species", name: "species", isRendered: property.propertyName === 'Species' }),
-                                span({ id: trIndex + "_phenotype", name: "phenotype", isRendered: property.propertyName === 'Phenotype/Indication' }),
                                 span({ id: trIndex + "_participants", name: "participants", isRendered: property.propertyName === '# of participants' }),
+                                span({ id: trIndex + "_phenotype", name: "phenotype", isRendered: property.propertyName === 'Phenotype/Indication' }),
                                 span({ id: trIndex + "_description", name: "description", isRendered: property.propertyName === 'Description' }),
+                                span({ id: trIndex + "_dataType", name: "dataType", isRendered: property.propertyName === 'Data Type' }),
+                                span({ id: trIndex + "_species", name: "species", isRendered: property.propertyName === 'Species' }),
```
rushtong: I don't think this is working as expected. Some of the columns in this table are hard-coded, so those are easy. Those are:
* Dataset ID
* Consent ID
* SC-ID
* Structured DUL
* Action buttons

Everything between Dataset ID and Consent ID is dynamically rendered on line 369. It depends on the display order of `this.state.dataSetList.dictionary` which comes from the server and is not part of the UI. We'll need to either disregard the display order and come up with our own, or change this on the server side. My preference (at this point) is to create a custom sort order that we can apply to the dictionary list and use that here instead of going back to the server.

rushtong: Turns out there are two things we need to provide custom sorting on. The `dictionary` controls a portion of the header columns and `catalog.properties` controls those same elements in the row data.

solideoglori:
Hi Greg - Can we have the columns in this order?

- Dataset ID
- Dataset Name
**- dbGAP**
**- Structured Data Use Limitations**
**- Data Type**
**- Phenotype/Indication**
**- Principal Investigator**
- # of participants
- Description
- Species
- Data Depositor
- Consent ID
- SC-ID
- Approved Requestors

rushtong:
@solideoglori  - Pretty sure that's reasonable. Will let you know if I run into any concerns.

src/pages/DatasetCatalog.js
```diff
@@ -365,111 +364,189 @@ class DatasetCatalog extends Component {
...
                   tr({}, [
                     th({}),
-                    th({ className: "table-titles dataset-color cell-size" }, ["Dataset Id"]),
-                    this.state.dataSetList.dictionary.map((dictionary, dIndex) => {
-                      return h(Fragment, { key: dIndex }, [
-                        th({ isRendered: dictionary.key !== 'Sample Collection ID', className: "table-titles dataset-color cell-size", id: dictionary.key }, [
-                          dictionary.key
-                        ])
-                      ]);
-                    }),
```
rushtong, line 375: @solideoglori ... this was the root of the complication. The dictionary is basically a list of columns, but it isn't the full list of columns, just some of them. So, if we want to rearrange some of the columns not defined in the dictionary, we'd have to do some complicated breaks to check the index, insert the undefined ones where we wanted them, and then continue on with looping over the rest of the dictionary elements. 

To get around this problem (and to solve a few other inefficiencies), I ditched the dictionary approach altogether. Instead, we just define our headers here as we want them. Then, when writing out the rows, we grab the property we want that matches the column. It should be trivial now to do any kind of re-arrangement/addition in the future.

solideoglori (approved):

PR #221, opened 2019-12-11
load dataset catalog on initial landing

gordonwade:
Address problem seen in [DUOS-477](https://broadinstitute.atlassian.net/browse/DUOS-477), where dataset catalog does not load on initial landing.

### DIFF
--- a/src/pages/DatasetCatalog.js
+++ b/src/pages/DatasetCatalog.js
@@ -56,25 +56,18 @@ class DatasetCatalog extends Component {
   }
 
   async getDatasets() {
-
-    if (this.state.isResearcher && '?reviewProfile' === this.props.location.search) {
-      const researcher = await Researcher.getPropertiesByResearcherId(Storage.getCurrentUser().dacUserId);
-      if (researcher.completed !== 'true') {
-        this.props.history.push('profile');
-      }
-    } else {
-      let catalog = await DataSet.findDataSets(this.USER_ID);
-      catalog.forEach((row, index) => {
-        row.checked = false;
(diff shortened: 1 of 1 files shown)


src/pages/DatasetCatalog.js
```diff
@@ -56,25 +56,18 @@ class DatasetCatalog extends Component {
...
-      });
-
-      this.setState({
-        dataSetList: { catalog: catalog },
-        currentPage: 1
-      });
-    }
-  }
+    // const currentUserId = Storage.getCurrentUser().dacUserId;
+    // let catalog = await DataSet.findDataSets(currentUserId);
```
rushtong: nit: remove these lines.

src/pages/DatasetCatalog.js
```diff
@@ -56,25 +56,18 @@ class DatasetCatalog extends Component {
...
+    let catalog = await DataSet.findDataSets(this.USER_ID);
```
rushtong: The use of `this.USER_ID` is an unusual pattern here, I would prefer that we access the user id from something that seems more reliable. In `componentDidMount`, which should execute before this method, we set `this.currentUser`. I think this would read more consistently with the rest of the app if we used `this.currentUser.dacUserId`

rushtong (approved):
Minor suggestions inline, but looks good 👍, thank you! No need to re-review if you want to address suggestions.

src/pages/DatasetCatalog.js
gordonwade: Okay, that makes sense!

PR #286, opened 2020-02-07
[no-jira][risk=no] Update zendesk address

rushtong:
`.org` -> `.com`


solideoglori (approved):
"duos-support@broadinstitute.zendesk.com" - copied from Customer Delivery's doc

PR #299, opened 2020-02-20
[no-jira][risk=no] Build Action

rushtong:
Testing out new github action for build/test


rushtong:
Interesting option as a replacement for circleci. Will investigate this alongside other deployment options Devops is reviewing.

PR #304, opened 2020-02-21
[no-jira][risk=no] Eslint and ajax library cleanup

rushtong:
## Addresses
Minor cleanup. Removes a bunch of unused API calls, at least one of which isn't in Consent.
Ran `eslint --fix` on the whole repo. 
Strongly suggested that you turn off whitespace changes when diffing the files.


rushtong:
Closing this ... will PR the eslint change and we can deal with formatting changes over time.

PR #305, opened 2020-02-23
Update to Home Page to improve user guidance

solideoglori:
Updating button text from "Sign In" to "Start here!" to improve user guidance, particularly for new users, in response to recent ZenDesk tickets.


rushtong (approved):
:+1:

PR #306, opened 2020-02-23
Updating DUOS home page graphic and bullet colors to match

solideoglori:
Updating DUOS home page graphic and bullet colors to match
- ignore branch name


rushtong (commented):
:+1:

PR #308, opened 2020-02-24
[DUOS-565][risk=no] Add check for eRA and collaborator

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-565

If `I am a collaborator of the PI/Data Custodian for the selected dataset(s)` is unchecked, and the user has not eRA authenticated, then prevent the DAR from being submitted.

### DIFF
--- a/public/config.json
+++ b/public/config.json
@@ -1,9 +1,9 @@
 {
-  "apiUrl": "https://consent.dsde-dev.broadinstitute.org/api",
-  "ontologyApiUrl": "https://consent-ontology.dsde-dev.broadinstitute.org/",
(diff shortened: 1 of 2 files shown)


solideoglori (approved):
Looks good.
Do we have an asterisk on the eRA Auth field in the UI when 'collaborator' is unchecked?

rushtong:
@solideoglori - I didn't change that at all  - it's already there. Adding some screenshots for clarity. The asterisk changes to Optional if collaborator is checked.

## Unchecked
Screen Shot 2020-02-24 at 12 25 46 PM

## Checked
Screen Shot 2020-02-24 at 12 25 54 PM

PR #316, opened 2020-02-27
DUOS-556 Shell page

mackenziekerwin:
Screen Shot 2020-02-27 at 10 48 01 AM
Includes some factoring out divs into logical components. There are 3 components on the Access Review page--header, vote, and application--and within the application summary there are "application sections," or components with a simple header and content.


public/index.html
```diff
@@ -58,4 +68,4 @@
     -->
 </body>
 
-</html>
+</html>
```
rushtong: nit: please add a newline. It looks like your IDE is auto-formatting these files and stripping out the trailing newline.

src/Routes.js
```diff
@@ -41,53 +42,56 @@ import { SummaryVotes } from './pages/SummaryVotes';
...
+    <AuthenticatedRoute path="/invalid_restrictions" component={InvalidRestrictions} props={props} rolesAllowed={[USER_ROLES.admin]} />
+    <AuthenticatedRoute path="/manage_ontologies" component={ManageOntologies} props={props} rolesAllowed={[USER_ROLES.admin]} />
+    <AuthenticatedRoute path="/researcher_review/:dacUserId" component={ResearcherReview} props={props} rolesAllowed={[USER_ROLES.admin]} />
+    <AuthenticatedRoute path="/access_result_records/:referenceId/:electionId" component={AccessResultRecords} props={props}
+      rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.member, USER_ROLES.alumni]} />
+    <AuthenticatedRoute path="/dul_results_record/:electionId" component={DulResultRecords} props={props}
+      rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.member, USER_ROLES.alumni]} />
+    <AuthenticatedRoute path="/access_review/:darId/:voteId/:rpVoteId?" component={AccessReview} props={props}
+      rolesAllowed={[USER_ROLES.member, USER_ROLES.chairperson]} />
+    <AuthenticatedRoute path="/mackenzie/:darId/:voteId/:rpVoteId?" component={MacKenzieAccessReview} props={props}
```
rushtong: Can we rename this component? Something like `AccessReviewV2` or anything along those lines.

src/components/AccessReviewHeader.jsx
```diff
@@ -0,0 +1,89 @@
...
+  fontSize: Theme.font.size.small,
+  lineHeight: Theme.font.leading.dense,
+}
+
+const BUTTON = {
+  ...SMALL,
+  fontWeight: Theme.font.weight.semibold,
+}
+
+export const AccessReviewHeader = hh(class AccessReviewHeader extends React.PureComponent {
```
rushtong: Can we make the `jsx` components `js` instead so we're consistent?

src/components/AccessReviewHeader.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+  back = () => {
+    const currentUser = Storage.getCurrentUser();
+    const page = currentUser.isChairPerson
+      ? "/chair_console"
+      : currentUser.isMember
+        ? "/member_console"
+        : "/";
+    this.props.history.push(page);
+  };
```
rushtong: This code looked familiar so I searched for it in the codebase. We have two other cases that look very similar to this and we should extract them all into a utility method. Here's the code that looks like the above:
```
  back = () => {
    const user = this.state.currentUser;
    const page = user.isChairPerson ? '/chair_console' :
      user.isMember ? '/member_console' :
        user.isAdmin ? '/admin_console' :
          user.isResearcher ? '/dataset_catalog?reviewProfile' :
            user.isDataOwner ? '/data_owner_console' :
              user.isAlumni ? '/summary_votes' : '/';
    this.props.history.push(page);
  };
```
You'll find this in `AccessReview.js` and `DulReview.js`. I think one opportune place for this method is in utils, perhaps in a constant like `Navigation` that can be called from anywhere.

src/theme.js
```diff
@@ -0,0 +1,27 @@
+export const Theme = {
```
rushtong: I think this would be better located in the `libs` directory. Also, this is fantastic ... love this idea 👍

src/index.css
```diff
@@ -1862,3 +1862,29 @@ input:-webkit-autofill, input:-webkit-autofill:hover, input:-webkit-autofill:foc
...
+
+.button-contained {
+  background-color: #00609f;
+  color: #ffffff;
+}
+
+.button-contained:hover {
+  background-color: rgba(0,96,159, 0.9);
+  color: #ffffff;
+}
```
rushtong: nit: please add a newline. Additionally, this is the case (css state such as hover) where style objects in js don't work. There are approaches to resolve this, but that's far beyond the scope of this PR and I don't want to sidetrack this effort. Will file a story so we can address this existing technical debt.

src/libs/ajax.js
```diff
@@ -219,6 +219,13 @@ export const DAC = {
 
 export const DAR = {
 
+  describeDarWithElectionInfo: async (darId, voteId) => {
+    const darInfo = await DAR.describeDar(darId);
+    const electionInfo = await Election.findElectionByVoteId(voteId);
+    darInfo.translatedUseRestriction = electionInfo.translatedUseRestriction;
+    return darInfo;
+  },
```
rushtong, line 229: Nice! We should file a tech-debt story to provide an API method that gives us all of the information we need here so we don't have to make 3 calls.

src/components/Application.jsx
```diff
@@ -0,0 +1,34 @@
...
+  fontSize: Theme.font.size.header,
+  lineHeight: Theme.font.leading.regular,
+};
+
+const HEADER_BOLD = {
+  ...HEADER,
+  fontWeight: Theme.font.weight.semibold,
+}
+
+export const Application = hh(class Application extends React.PureComponent {
```
rushtong: Naming is hard again. The word `Application` has too much context associated with it. Can we name this `DarApplication` or something similar?

src/components/VoteAsX.jsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const TAB_SELECTED = {
+  ...TAB_UNSELECTED,
+  backgroundColor: Theme.palette.background.secondary,
+  borderRadius: '9px 9px 0px 0px',
+  color: Theme.palette.primary,
+};
+
+export const VoteAsX = hh(class VoteAsX extends React.PureComponent {
```
rushtong: Naming: How about something like `DacVote`?

src/pages/MacKenzieAccessReview.jsx
```diff
@@ -0,0 +1,80 @@
...
+import { DAR } from '../libs/ajax';
+
+const SECTION = {
+  margin: '16px',
+};
+
+class MacKenzieAccessReview extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = this.initialState();
```
rushtong: I'd like to get away from this pattern of calling a method for initial state. I don't have a good handle on what I do want, so will look into that some more.

rushtong (changes requested):
I think this is coming along nicely. Feedback inline.

src/components/AccessReviewHeader.js
```diff
@@ -0,0 +1,82 @@
...
+
+  render() {
+    const currentUser = Storage.getCurrentUser();
+    const { history } = this.props;
+    return div(
+      {
+        style: {
+          justifyContent: "space-between",
+          alignItems: "center",
+          display: "flex",
```
rushtong: nit: try running eslint ... I think it will change these double quotes to single quotes.

src/components/AccessReviewHeader.js
```diff
@@ -0,0 +1,82 @@
...
+        div(
+          [
+            a(
+              {
+                id: "btn_back",
+                onClick: () => Navigation.back(currentUser, history),
+                className: "button-outlined secondary",
+                style: BUTTON,
+              },
+              "DAC Console"
```
rushtong, line 85: Can we add another button here to link back to the original version of the page? That would allow for us to vote on elections on dev while this feature is in development.

src/libs/utils.js
```diff
@@ -24,3 +24,16 @@ export const USER_ROLES = {
...
+    const page = user.isChairPerson ? '/chair_console'
+      : user.isMember ? '/member_console'
+        : user.isAdmin ? '/admin_console'
+          : user.isResearcher ? '/dataset_catalog?reviewProfile'
+            : user.isDataOwner ? '/data_owner_console'
+              : user.isAlumni ? '/summary_votes'
+                : '/';
+    history.push(page);
+  }
+}
```
rushtong: Thank you for this change!
nit: please add a newline.

rushtong (approved):
Looks great! One feature request and a couple minor nits. If you want to add that feature, I don't need to re-review.

PR #318, opened 2020-02-27
UI only - Updated fields for DUOS DAR step 1 & step 2

solideoglori:
Re-ordered existing DAR fields for usability and in order to improve parity between checkbox and narrative RUS.

### DIFF
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -930,24 +930,23 @@ class DataAccessRequestApplication extends Component {
                   div({ className: 'form-group' }, [
                     div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group' }, [
                       label({ className: 'control-label rp-title-question' }, [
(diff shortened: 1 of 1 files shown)


rushtong:
Something definitely broke here. The checkboxes in this screenshot are not selectable:
Screen Shot 2020-02-28 at 10 18 01 AM

rushtong:
Selecting the checkbox for 
```
2.4.3 Population structure or normal variation studies: The primary purpose of the research is to understand variation in the general population (e.g., genetic substructure of a population).
```
not only selects that one, it selects the one below it too, 
```
2.4.5 Commercial or For-Profit Purpose: The primary purpose of the research is exclusively or partially for a commercial purpose
```
Note that they are not sequential ... looks like 2.4.4 disappeared.

rushtong:
Addendum to the 2.1 checkbox issue, even without selecting any of the options, I was able to submit the application.

rushtong (changes requested):
See main comments. We'll need to resolve those bugs before continuing.

PR #326, opened 2020-03-03
DUOS-562 Research purpose

mackenziekerwin:
Right now, the four components of the application summary are `ApplicationSection` components that simply render a header and content. Since most of the sections require additional logic to format their content, they will eventually be broken out into more specific components and the `ApplicationSection` component will be renamed to `ResearchPurpose`. For now, I'm just adding the research use statement to the existing `ApplicationSection`. :)


rushtong (approved):
👍 Fantastic, I love short PRs.

PR #329, opened 2020-03-04
dependency: bump js-file-download from 0.4.9 to 0.4.10 (redoing undone update) (#294) + DAR Updates (Step 1&2) 

solideoglori:


rushtong:
@solideoglori It would be cleaner if you just included your changes for DUOS-482 and not worry about mixing in the dependabot one. Dependabot will re-file a new PR for the update as soon as it finds that it's out of date.

PR #333, opened 2020-03-05
DUOS-557 Member vote

mackenziekerwin:
This is a work in progress! I wrote all of my error/alert handling as `console.log`s, but those will of course be turned into dialogs once we make those decisions with the UX team. Let me know what you think of this approach--I had to make a few design decisions, and I included some comments to explain them, but I am open to suggestions.
One more thing--sorry about all the commits. I think I made this branch off of my un-merged shell branch.


src/components/DacVotePanel.js
```diff
@@ -30,31 +31,183 @@ const TAB_SELECTED = {
...
+   * this is called when VoteAsMember mounts
+   */
+  getVotesAsMember = async () => {
+    const { darId, voteId, rpVoteId } = this.props.ids;
+    try {
+      const vote = await Votes.getDarVote(darId, voteId);
+      this.setState({ vote });
+    }
+    catch (e) {
+      console.log('Something went wrong trying to get the data.')
```
rushtong, line 80: Yeah, good call on bringing this up. We need a more cohesive error reporting strategy.  See these links for what Terra is doing ... I'd like to try and use a similar approach if it makes sense.
* https://github.com/DataBiosphere/terra-ui/blob/dd4152c8280f5e6b19fff249a2a9490cfb868114/src/components/ErrorWrapper.js
* https://github.com/DataBiosphere/terra-ui/blob/5b6c7b45fb43931c9884ddc16fe50a8514852f29/src/libs/error.js

rushtong (approved):
👍 Looks good. Please make sure we have a ticket for the error handling.

PR #336, opened 2020-03-09
DUOS-558 Chair vote

mackenziekerwin:
Includes functionality to vote as a chair. Again, there's no real error handling, but it submits the final access vote and closes the election. Let me know if there's anything else that needs to be done when a chair clicks the vote button!

### DIFF
--- a/src/components/DacVotePanel.js
+++ b/src/components/DacVotePanel.js
@@ -67,6 +67,21 @@ export const DacVotePanel = hh(class DacVotePanel extends React.PureComponent {
   };
 
   /**
+   * gets data for final access vote
+   * this is called when VoteAsChair mounts
+   */
+  getVotesAsChair = async () => {
+    const { electionId } = this.state.vote;
+    try {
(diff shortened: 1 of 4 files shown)


myessail (approved):
👍 assuming the catch blocks will eventually go to actual error handling

src/components/DacVotePanel.js
```diff
@@ -139,28 +167,51 @@ export const DacVotePanel = hh(class DacVotePanel extends React.PureComponent {
...
   };
 
+  // closes the election for this DAR
+  async closeElection() {
+    const { election } = this.props;
+    const electionClone = _.cloneDeep(election);
+    electionClone.status = 'Closed';
+    try {
+      await Election.updateElection(election.electionId, electionClone);
+      console.log('Election closed');
```
rushtong: Not sure we need to capture this as something to show the user ... I would leave this out. Keep the error logging and we'll deal with that in the error handling story.

rushtong (approved):
👍 Looks great! One request inline.

PR #345, opened 2020-03-16
DUOS-560 Structured research purpose

mackenziekerwin:
Technically still a work in progress. Currently, the SRP and DUSL require different logic to format the content, but they will turn into an abstract component when we have the backend returning the JSON object.

### DIFF
--- a/src/components/DownloadLink.js
+++ b/src/components/DownloadLink.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import _ from 'lodash';
+import { div, a, i, hh } from "react-hyperscript-helpers";
+import { Theme } from '../libs/theme';
+
+const ROOT = {
+  margin: '8px 0px'
+};
+
+const ICON = {
+  color: Theme.palette.link,
+  marginRight: '6px',
+};
+
+export const DownloadLink = hh(class DownloadLink extends React.PureComponent {
+  render() {
+    const { label, onDownload } = this.props;
+    return div({ style: ROOT }, [
+      a({
+        id: _.kebabCase(label),
+        onClick: () => onDownload()
+      },
+        [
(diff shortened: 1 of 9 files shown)


src/components/DownloadLink.js
```diff
@@ -0,0 +1,29 @@
...
+        id: _.kebabCase(label),
+        onClick: () => onDownload()
+      },
+        [
+          i({ className: 'glyphicon glyphicon-download-alt', style: ICON }),
+          label,
+        ])
+    ]);
+  }
+});
```
rushtong: This looks good, but I think it might be more concise as a pure function (caveat, untested code follows):
```
export const DownloadLink = (props) => {
  return div({style: ROOT}, [
    a({
        id: _.kebabCase(props.label),
        onClick: () => props.onDownload(),
      },
      [
        i({className: 'glyphicon glyphicon-download-alt', style: ICON}),
        props.label,
      ]),
  ]);
};
```

rushtong:
@mackenziekerwin FYI, the ontology endpoint for getting a structured translation is here: https://consent-ontology.dsde-dev.broadinstitute.org/#/Translate/post_translate_summary

To generate a request to this endpoint, we'll need to format the relevant elements of the data access request into a properly formed [data use schema](https://consent-ontology.dsde-dev.broadinstitute.org/#/Data%20Use/get_schemas_data_use). We don't have any code to do that in the UI now so we'll need to make that happen to utilize the ontology endpoint.

src/components/DownloadLink.js
mackenziekerwin: Great suggestion. I haven't used functional components before, but I see how that cleans it up a lot!

rushtong (approved):
👍

PR #351, opened 2020-03-20
DUOS-594 Error feedback

mackenziekerwin:
Implements two types of user feedback: 1) a short message inline with the vote button, and 2) a boxed notification that slides in from the side. Please give any thoughts on how to clean it up or make it better!


mackenziekerwin:
Inline messages:

incomplete-vote
log-vote

mackenziekerwin:
Error notification:

show-error

package.json
```diff
@@ -13,7 +13,7 @@
     "js-file-download": "0.4.10",
     "lodash": "4.17.15",
     "mixin-deep": "2.0.1",
-    "query-string": "6.11.1",
```
rushtong: We use this in `eRACommons.js`

src/libs/utils.js
```diff
@@ -48,3 +51,16 @@ export const download = (fileName, text) => {
...
+export const Notifications = {
+  showError: error => {
+    return new Noty({
+      type: 'error',
+      layout: 'bottomRight',
+      theme: 'duos',
+      text: `Something went wrong. Please try again.\nError code: ${error.status}`,
+      timeout: '3500',
+      progressBar: false,
+    }).show();
```
rushtong, line 72: What I'd like to see here is an ability to customize the message. Absent arguments, this is a great default. To be more useful outside of the current context, I would introduce some props here.  Something like this (caveat ... completely untested):
```
showError: (props) => {
    return new Noty({
      type: 'error',
      layout: 'bottomRight',
      theme: 'duos',
      text: props.message,
      timeout: props.timeout,
      progressBar: false,
    }
```
It would be nice if props were an object with defined, default properties so a caller could pass in merged props with default props maybe like this (caveat, also untested):
```
    catch (e) {
      const mergedProps = {...Notifications.defaultProps, ...{message: 'my error message'}};
      Notifications.showError(mergedProps);
    }
```

rushtong (changes requested):
Overall, I like it. One required change, and one suggestion on improving the reusability of the notification.

src/libs/utils.js
rushtong: Addendum ... callers probably shouldn't do the merge, that should happen inside `Notifications`. Instead, callers should just pass in what they specifically care about, i.e. `Notifications.showError({message: 'my error message'})`

package.json
mackenziekerwin: Sorry, I'm not sure how it got deleted but it's back in there now.

src/libs/utils.js
mackenziekerwin: Hopefully this is a good solution... The order in which the objects are merged allows the caller to customize `text`, `timeout`, etc., but not properties like `type` and `theme`, which should be left to their defaults. I also kept the `defaultNotification` object pretty general in case we wanted to implement a `showSuccess` function or something similar in the future.

package.json
rushtong: No worries!

src/libs/utils.js
rushtong: I like this! 👍

rushtong (approved):

rushtong:
Just need to merge package.json before squashing.

PR #354, opened 2020-03-20
[no-jira][risk=no] Fix the commercial checkbox selection

rushtong:
* Re-implement https://github.com/DataBiosphere/duos-ui/pull/348 which ended up breaking the deploy.
* Included several library updates that were submitted by dependabot. I think the react-scripts update was the reason for the broken build.
* Added a build to circleci config so that step, in addition to test, is run for PRs. Previously, it wasn't, and that lead to the deploy failure.


mackenziekerwin (approved):

rushtong:
Thank you @mackenziekerwin!

PR #363, opened 2020-03-26
[DUOS-610][risk=no] Fix UI error when removing chairs

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-610

## Changes
We don't have json in the response body when removing a dac member/chair. Instead, return the status so the promises can complete correctly.


mackenziekerwin (approved):
Nice!

rushtong:
Thank you @mackenziekerwin!

PR #380, opened 2020-04-02
[no jira] DAR updates to reflect LC DAA for researchers

solideoglori:
- Add PI specific DAA for their acknowledgement
- Update DAR Step 4 text, linked button to new PDF, removed upload button, and linked text to SO guidance


src/pages/DataAccessRequestApplication.js
```diff
@@ -1612,44 +1612,24 @@ class DataAccessRequestApplication extends Component {
...
                       div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group' }, [
                         label({ className: 'control-label default-color' },
-                          ['1. Download the Data Access Agreement template and have your organization\'s Signing Official sign it'])
+                          ['By submitting this data access request, you agree to comply with all terms relevant to Authorized Users put forth in the agreement.'])
                       ]),
 
                       div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group' }, [
                         a({
-                          id: 'link_downloadAgreement', href: 'BroadDataAccessAgreement.pdf', target: '_blank',
+                          id: 'link_downloadAgreement', href: 'DUOS_LibraryCardAgreementTemplate_3.28.20_PIaacknowledgement.pdf', target: '_blank',
```
rushtong: Typo ... should be `DUOS_LibraryCardAgreementTemplate_3.28.20_PIacknowledgement.pdf`

src/pages/DataAccessRequestApplication.js
```diff
@@ -1612,44 +1612,24 @@ class DataAccessRequestApplication extends Component {
...
-                      p({ id: 'txt_uploadFile', className: 'fileName daa', isRendered: this.state.file.name !== '' || this.state.formData.nameDAA }, [
-                        'Your currently uploaded Data Access Agreement: ',
-                        span({ className: 'italic normal' }, [this.state.file.name !== '' ? this.state.file.name : this.state.formData.nameDAA])
-                      ]),
-                      span({
-                        className: 'col-lg-12 col-md-12 col-sm-6 col-xs-12 cancel-color required-field-error-span',
-                        isRendered: step4.uploadFile.invalid && showValidationMessages
-                      }, ['Required field'])
+                          ['For your request to be reviewed, your Signing Official must authorize you by sending a signed Library Card data access agreement']),
+                        a({ href: 'https://duos.broadinstitute.org/home_about', target: '_blank' }, '(instructions here)')
```
rushtong: We should not link to absolute urls like this. I think `/home_about` is good enough but please double check.

rushtong (changes requested):
Two fixes needed, but otherwise looks good.

src/pages/DataAccessRequestApplication.js
rushtong: See https://github.com/DataBiosphere/duos-ui/blob/develop/src/components/DuosHeader.js#L157 for an example of how we do this now.

src/pages/DataAccessRequestApplication.js
```diff
@@ -1612,44 +1612,24 @@ class DataAccessRequestApplication extends Component {
...
-                        'Your currently uploaded Data Access Agreement: ',
-                        span({ className: 'italic normal' }, [this.state.file.name !== '' ? this.state.file.name : this.state.formData.nameDAA])
-                      ]),
-                      span({
-                        className: 'col-lg-12 col-md-12 col-sm-6 col-xs-12 cancel-color required-field-error-span',
-                        isRendered: step4.uploadFile.invalid && showValidationMessages
-                      }, ['Required field'])
+                          ['For your request to be reviewed, your Signing Official must authorize you by sending a signed Library Card data access agreement']),
+                        a({ href: '/home_about', target: '_blank' }, '(instructions here)')
+                      ])
```
rushtong, line 1632: One of the side effects of this is that `handleFileChange` is no longer used since we're no longer uploading a `S.O. Signed Agreement`. There are likely state variables that are also no longer necessary. Would be fine with a cleanup PR if you want to get this out for tomorrow's release.

rushtong (approved):
👍 Looks good. One suggestion inline that can be part of this PR or a separate one, I'm fine either way.

PR #399, opened 2020-04-10
[no jira] Adding Machine Readable Consent info to about page

solideoglori:
Unsure of correctness of PR. I believe lines 63-72 and the machine-readable-consent-forms.png should be the only new lines compared to what is currently on dev/stag/prod. Not sure why others lines and files are appearing as new.


solideoglori:
Aiming for this: 

image

src/pages/HomeAbout.js
```diff
@@ -17,19 +18,58 @@ class HomeAbout extends Component {
...
+                ]),
+
+                div({ className: "home-content" }, [
+                  h3({}, ["Machine Readable Consent Forms & DUOS"]),
+                  p( { className: "home-sections-description" }, ["Standardizing data sharing language for expedited and compliant sharing"]),
+                  p( {style: { fontWeight: '400' }}, ["To date, many consent forms contain ambiguous or region-specific data sharing language, making widespread data sharing and aggregation of datasets difficult or near impossible."]),
+                  span({}, []),
+                  a({ href: 'https://blog.primr.org/is-your-data-sharing-consent-language-transparent-and-machine-readable/', target: '_blank' }, 'GA4GH\'s Machine Readable Consent Form Initiative'),
+                  p( {style: { fontWeight: '400' }}, ["This initiative provides standard data sharing language for IRBs, funders, and investigators to use in their consent forms, based on the Data Use Ontology (DUO)."]),
+                  img({ src: "/images/machine readable consent forms.png", className: "col-lg-12 col-md-12 col-sm-12 col-xs-12 home-content-graphic", alt: "About icon" }),
```
rushtong: Avoid spaces in file names:
```suggestion
                  img({ src: "/images/machine_readable_consent_forms.png", className: "col-lg-12 col-md-12 col-sm-12 col-xs-12 home-content-graphic", alt: "About icon" }),
```

rushtong (changes requested):
These files exist in your local install, but do not exist in github:
* public/DUOS_LibraryCardAgreementTemplate_3.28.20.pdf 
* public/images/DUOS_LC_Process.PNG 
* public/images/LC_only_process.PNG

Note that there are lower cased versions of the images here: https://github.com/DataBiosphere/duos-ui/tree/develop/public/images so I think the root issue has to do with different cased file names and possibly a windows platform issue. If you can try to stick to lower cased file names, it should help prevent this in the future. Until you remove them and check out a new copy of develop, they will be considered new files.

Also see inline comment about spaces in file names.

Outside of that, the content looks good.

solideoglori:
I updated the code to reference the existing lower case doc and images, and accepted the spacing fix. Is there a way to delete the doc and images I committed in GH before merging?

rushtong:
@solideoglori 
```
git rm public/DUOS_LibraryCardAgreementTemplate_3.28.20.pdf
git commit -m 'removing unused file' .
```
Lather, rinse, repeat for each file you don't want. Then ... 
```
git push origin HomeAbout_SO_LCDAA_Updates
```

rushtong:
Looks like there's a conflict between your version of `HomeAbout.js` and the latest version from develop. One way to fix that is to save your changes locally, checkout the version from develop, then re-apply your local changes.
Screen Shot 2020-04-14 at 1 54 23 PM

PR #410, opened 2020-04-15
Updating LCDAA per OSAP and relevant download links

solideoglori:
Deleted old PDFs (fix to #399 as well)
Added new PDF (1 to replace the 2) 
Updated links to ref to new PDF


rushtong (approved):
👍

PR #411, opened 2020-04-15
Expanding eRA Commons auth bypass ability to NIH intramural researchers

solideoglori:
NIH intramural researchers often do not have eRA Commons IDs. We are broadening the language of the 'Internal Collaborator' bypass checkbox to allow them to bypass eRA auth and still submit a DAR.

We will need to validate their library card existence based on their nih email as a uid rather than an era id.


rushtong (approved):
👍 Looks good. One thing to consider is the language on the Profile page:

> Please authenticate your eRA Commons account to submit Data Access Requests. Other profiles are optional:

PR #470, opened 2020-05-29
[DUOS-685][risk=no] Don't pass library cards when updating user

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-685

## Changes
The API is expecting a simple map of string -> string. If we pass in libraryCards as a list of strings, that breaks the API's json parsing.


rushtong:
👍  from slack

PR #477, opened 2020-06-03
[DUOS-689][risk=no] UI Stackdriver Logging

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-689

See helm/values PRs for how we'll get values into the config file
* ~https://github.com/broadinstitute/terra-helm/pull/197~ Merged for dev
* ~https://github.com/broadinstitute/terra-helmfile/pull/561~ Merged for dev
Until those are moved to production, we'll only have env values as proxies for the real ones.

## Changes
* Add js stackdriver logging as an infrastructure component. 
* Default log any non-successful ajax request that is expecting an OK response.
* Error API keys created added to configmap for each environment

## Example Reports
Screen Shot 2020-06-10 at 2 34 28 PM

Screen Shot 2020-06-10 at 2 34 43 PM


src/libs/config.js
```diff
@@ -21,11 +21,27 @@ export const Config = {
...
     return _.get(feature, featureName, false);
   },
 
+  getProject: async () => {
+    const env = await Config.getEnv();
+    switch (env) {
+      case 'prod':
+        return 'broad-duos-prod';
+      default:
+        return 'broad-duos-dev';
```
JVThomas, line 41: Is there no need to distinguish between staging and dev?

rushtong: Staging and dev clusters are both in the same `broad-duos-dev` google project. This will eventually change when we host the UI on the terra clusters, but until that happens, both are still in the dev one.

JVThomas (approved):

PR #491, opened 2020-06-22
[DUOS-683][risk=no]UI: Help Screen

Marianie-Alphonse:
### **Addresses**
https://broadinstitute.atlassian.net/browse/DUOS-683

### **Changes**

- Added CSS style for the select element in the BaseModal.css
- Added CSS style for the "contact us" button in the DuoHeader.css
- Added the SupportRequestModal, and it's helper functions to DuoHeader.js
- Created SupportRequestModal.js
- Created SupportRequestBaseModal.js
- Added called to Zendesk APIs for ticket creation and file upload in ajax.js
- Added a header for attachment body in Config.js

### **Rebase**
- 7/21/20 - 12:30pm


src/components/DuosHeader.css
```diff
@@ -104,6 +104,19 @@
...
+  font-size: 16px;
+  border: none;
+  background-color: transparent;
+  outline: none;
+  transition: background 0.3s ease !important;
+}
+
+.contact-us-button:hover {
+  color: #5BB0CA !important;
+}
```
rushtong: In general, I try to avoid as much css as possible in separate css files. Instead, I like to use inline styling. The hover case, however, is more complicated to implement inline. There are some small examples in the codebase of inlined hoverstate, but I'm fine with this.

src/components/DuosHeader.js
```diff
@@ -27,6 +29,13 @@ class DuosHeader extends Component {
       return prev;
     });
   };
+   
+  SupportRequestModal = (e) => {
```
rushtong: nit: lower case the `S` for consistent method naming.
```suggestion
  supportRequestModal = (e) => {
```

src/components/DuosHeader.js
```diff
@@ -149,7 +177,21 @@ class DuosHeader extends Component {
...
+                    onClick: this.SupportRequestModal,
+                    "data-tip": "Need help? Contact us for some assistance", "data-for": "tip_requestAccess"
+                  }, ["Contact Us"]),
+                  SupportRequestModal({
+                    showModal: this.state.showSupportRequestModal,
+                    onOKRequest: this.okSupportRequestModal,
+                    onCloseRequest: this.closeSupportRequestModal,
+                    onAfterOpen: this.afterSupportRequestModalOpen,
+                    url: this.props.location.pathname
+                  }),
```
rushtong: To avoid duplication, I think you can make this button a const at the top of the render method then just call the const here and below.

src/components/SupportRequestBaseModal.js
```diff
@@ -0,0 +1,78 @@
+import { Component } from 'react';
+import { button, div, h, span, hh } from 'react-hyperscript-helpers';
+import Modal from 'react-modal';
+import './BaseModal.css';
```
rushtong: I would leave this out and instead define any styles you need as js objects here in this component. It's a lot easier to refactor and find dead code this way. CSS in css files is very difficult to maintain over time.

src/components/SupportRequestBaseModal.js
```diff
@@ -0,0 +1,78 @@
...
+            ]),
+
+            div({ className: "modal-content" }, [
+              this.props.children
+            ]),
+
+            div({ className: "modal-footer" }, [
+              button({ id: "btn_action", className: "col-lg-5 col-md-3 col-sm-4 col-xs-6 btn " + this.props.color + "-background",
+                onClick: this.props.action.handler, disabled: disableOkBtn }, [this.props.action.label]),
+              button({ isRendered: this.props.type !== "informative", id: "btn_cancel", className: "col-lg-5 col-md-3 col-sm-4 col-xs-6 btn dismiss-background", onClick: this.props.onRequestClose }, ["Cancel"]),
```
rushtong: These class names look like bootstrap classes. It would be preferable to style them inline if possible.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+    };
+    this.closeHandler = this.closeHandler.bind(this);
+    this.afterOpenHandler = this.afterOpenHandler.bind(this);
+    this.OKHandler = this.OKHandler.bind(this);
+    this.nameChangeHandler = this.nameChangeHandler.bind(this);
+    this.typeChangeHandler = this.typeChangeHandler.bind(this);
+    this.subjectChangeHandler = this.subjectChangeHandler.bind(this);
+    this.descriptionChangeHandler = this.descriptionChangeHandler.bind(this);
+    this.attachmentChangeHandler = this.attachmentChangeHandler.bind(this);
+    this.emailChangeHandler = this.emailChangeHandler.bind(this);
```
rushtong: You can avoid all of these boilerplate bindings by using the arrow syntax, i.e.
```
functionName = () => {
  // do something
};
```
Or for async functions ... 
```
functionName = async () => {
  // do something
};
```
If they are plain functions, then you have to use these binding statements in the constructor, which I find annoying and error prone.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+              ]),  
+            ]),
+            div({ className: "form-group first-form-group" }, [
+              label({ id: "lbl_description", className: "common-color" }, ["How can we help you" + this.state.first_name + "? *"]),
+              input({ id: "txt_subject", placeholder:'Enter a subject', rows: "5", className: "form-control col-lg-12 vote-input", onChange: this.subjectChangeHandler, required: true }),
+              textarea({ id: "txt_description", placeholder:'Enter a description', rows: "5", className: "form-control col-lg-12 vote-input", onChange: this.descriptionChangeHandler, required: true }),
+            ]),
+            div({ className: "form-group first-form-group" }, [
+              label({ id: "lbl_attachment", className: "common-color" }, ["Attachment"]),
+              input({ type: "file", id: "txt_attachment", placeholder: 'Attach a file?', className: "form-control col-lg-12 vote-input common-color", onChange: this.attachmentChangeHandler, ref: 'fileUpload', required: false }),
```
rushtong: TOL (thinking out loud) ... I think this should be triggered on submit, not change. We could chain the file upload to happen before the submit so the order of operations is maintained.

More TOL, this only allows for a single file upload which is fine. I can see the need for multiples, but not for this story ... possibly a future enhancement.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+        imgSrc: "/images/icon_add_help.png",
+        color: "common",
+        title: "Contact Us",
+        height: this.state.height,
+        top: this.state.top,
+        action: { label: "Submit", handler: this.OKHandler },
+      },
+
+        [
+          form({ className: "form-horizontal css-form", name: "consentForm", noValidate: true, encType: "multipart/form-data" }, [
```
rushtong: nit: Should give this a more meaningful name

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+            !this.state.isLogged && div({ className: "form-group first-form-group" }, [
```
rushtong: This construction is throwing me for a loop. Does this mean the div is only rendered if `!this.state.isLogged` resolves to true?

rushtong (commented):
First round of feedback before I try to test this functionally. Fantastic work overall!!! See comments inline. Also, not sure if your editor lets you, but there are some eslint rules specified in the project that should help you get past the codacy warnings. Will dig back into this tomorrow.

rushtong:
The circle error is an OOM error ... very weird. We might need to adjust the circle config to account for this.
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x8fa0c0 node::Abort() [node]
 2: 0x8fa10c  [node]
 3: 0xb0026e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xb004a4 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xef49b2  [node]
 6: 0xef4ab8 v8::internal::Heap::CheckIneffectiveMarkCompact(unsigned long, double) [node]
 7: 0xf00b92 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [node]
 8: 0xf014c4 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 9: 0xf04131 v8::internal::Heap::AllocateRawWithRetryOrFail(int, v8::internal::AllocationSpace, v8::internal::AllocationAlignment) [node]
10: 0xecd5b4 v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationSpace) [node]
11: 0x116d73e v8::internal::Runtime_AllocateInNewSpace(int, v8::internal::Object**, v8::internal::Isolate*) [node]
12: 0x225f715dbe1d 
```

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
+import { Component } from 'react';
+import { div, form, input, label, textarea, hh, select, option } from 'react-hyperscript-helpers';
+import { SupportRequestBaseModal } from '../SupportRequestBaseModal';
+import { Alert } from '../Alert';
+import { Support, User } from '../../libs/ajax';
+import { Storage } from "../../libs/storage";
+import { convertCompilerOptionsFromJson } from 'typescript';
```
rushtong: unused

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
+import { Component } from 'react';
+import { div, form, input, label, textarea, hh, select, option } from 'react-hyperscript-helpers';
+import { SupportRequestBaseModal } from '../SupportRequestBaseModal';
+import { Alert } from '../Alert';
+import { Support, User } from '../../libs/ajax';
```
rushtong: `User` is unused here

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+    const attachmentText = await Support.uploadAttachment(file);
+    this.setState(prev => {
+      prev.attachment = attachmentText;
+      return prev;
+    });
+  };
+
+  emailChangeHandler = (e) => {
+    let emailText = e.target.value;
+    this.state.vaild = /.+@.+\.[A-Za-z]+$/.test(emailText) ? true: false;
```
rushtong: use `setState` instead of modifying a state value directly

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+    await this.setState(prev => {
+      prev.name = '';
+      prev.type = "question";
+      prev.subject = '';
+      prev.description = '';
+      prev.attachment = '';
+      prev.email = '';
+      return prev;
+    });
+    this.props.onOKRequest('support');;
```
rushtong: nit: only need one `;`

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+      custom_fields: [
+        { id: 360012744452, value: this.state.type},
+        { id: 360007369412, value: this.state.description},
+        { id: 360012744292, value: this.state.name},
+        { id: 360012782111, value: this.state.email },
+        { id: 360018545031, value: this.state.email }
+      ],
+      comment: {
+        body: this.state.description + '\n\n------------------\nSubmitted from: ' + this.props.url,
+        //${currUrl}
```
rushtong: nit: rm comment

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+    this.typeChangeHandler = this.typeChangeHandler.bind(this);
+    this.subjectChangeHandler = this.subjectChangeHandler.bind(this);
+    this.descriptionChangeHandler = this.descriptionChangeHandler.bind(this);
+    this.attachmentChangeHandler = this.attachmentChangeHandler.bind(this);
+    this.emailChangeHandler = this.emailChangeHandler.bind(this);
+  }
+
+  async OKHandler() {
+    const ticket = {};
+    ticket.request = {
```
rushtong: The ticket object construction seems a little out of place to me in the component. I think if it were a separate method in `Support`, like `Support.createTicket(type, name, email, subject, description);` (or something like that ... you could also pass a map of params to simplify the arg list), that would be a more context-appropriate location for this code.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+    });
+    this.props.onOKRequest('support');;
+  }
+
+  closeHandler() {
+    // this is the method to handle Cancel click
+    // could do some cleaning here
+    // or delegate it to the parent
+    // we need to use it to close the
+    // DO SOMETHING HERE ...
```
rushtong: You could add a little noty notification here, "Support request canceled" that goes away after a few seconds. Just a suggestion - I'm open to whatever you decide.

src/components/modals/SupportRequestModal.js
rushtong: Also, typo in `vaild`

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+      prev.email = emailText;
+      return prev;
+    });
+  };
+
+  render() {
+
+    return (
+      SupportRequestBaseModal({
+        id: "SupportRequestModal",
```
rushtong: Here and throughout, use single quotes where possible.
```suggestion
        id: 'SupportRequestModal',
```

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,194 @@
...
+
+export const SupportRequestModal = hh(class SupportRequestModal extends Component {
+  constructor(props) {
+    super(props);
+    const isLogged = Storage.userIsLogged();
+    let name = isLogged ? Storage.getCurrentUser().displayName : '';
+    let first_name = isLogged ? ", " + name.split(" ")[0] : '';
+    let email = isLogged ? Storage.getCurrentUser().email : '';
+    let height = isLogged ? "550px": '700px';
+    let top = isLogged ? '400px': '1';
```
rushtong: All of these `let`s can be `const`s. Also, you could avoid this altogether and just put the ternary expressions into the state object directly.

src/libs/ajax.js
```diff
@@ -719,6 +719,22 @@ export const Summary = {
...
+  //fetchany
+  //Noty
+  createSupportRequest: async (ticket) => {
+    const res = await fetchOk(`https://broadinstitute.zendesk.com/api/v2/requests.json`, fp.mergeAll([Config.jsonBody(ticket), { method: 'POST' }]));
+    return await res.json();
+  },
+
+  //Create File Name
+  uploadAttachment: async (file) => {
+    const res = await fetchOk(`https://broadinstitute.zendesk.com/api/v2/uploads?filename=${file.name}`, fp.mergeAll([Config.attachmentBody(file), { method: 'POST' }]));
```
rushtong: Have not tested this yet, but there is a potential issue if `file.name` is not url encoded safely or has strange characters. I can imagine some weird things happening with unicode characters. As discussed in review session, would probably be safer to generate a file name instead of using what was passed in.

rushtong (commented):
One more round of feedback before I look at this functionally. Please let me know if you need any help or explanations - would be happy to go into more depth.

src/components/modals/SupportRequestModal.js
Marianie-Alphonse: When I put them in the state objects I get this error "Unexpected use of 'name'  no-restricted-globals"

src/components/modals/SupportRequestModal.js
Marianie-Alphonse: I took this construction from the Terra code, I believed it checks if the first conditional is true/false and execute the second part if it's true

src/libs/ajax.js
Marianie-Alphonse: I made the name default to "Attachment" since the token will always be unique

src/components/modals/SupportRequestModal.js
rushtong: 👍 Makes sense, thanks!

src/components/modals/SupportRequestModal.js
rushtong: I think that's because you're using `name` to calculate `first_name`. I think this should be easy to resolve.

rushtong: Try this:
```
    this.state = {
      name: Storage.userIsLogged() ? Storage.getCurrentUser().displayName : '',
      isLogged: Storage.userIsLogged(),
      type: 'question',
      subject: '',
      description: '',
      attachment: '',
      email: Storage.userIsLogged() ? Storage.getCurrentUser().email : '',
      first_name: Storage.userIsLogged() ? Storage.getCurrentUser().displayName.split(' ')[0] : '',
      height: Storage.userIsLogged() ? '550px': '700px',
      top: Storage.userIsLogged() ? '400px': '1',
      valid: Storage.userIsLogged()
    };
```

src/components/SupportRequestBaseModal.js
Marianie-Alphonse: I got rid of most of them, the ones that are not in the basemodal.css like ```col-lg-5 col-md-3 col-sm-4 col-xs-6 btn ``` I was unable to remove

src/components/BaseModal.css
```diff
@@ -80,3 +80,28 @@ textarea {
 .modal-footer button.default-background {
     color: #ffffff;
 }
+
```
Marianie-Alphonse: I forgot about this file, I'm editing it now

rushtong:
Any chance of using the same file upload library that Terra uses for attachments? Here is a side by side comparison:

Firefox:
Screen Shot 2020-07-14 at 10 07 43 AM

Chrome:
Screen Shot 2020-07-14 at 10 08 03 AM

rushtong:
Same feedback applies to the select menu for Type. They are using [`react-select`](https://github.com/DataBiosphere/terra-ui/blob/535e52e3027bd51834cdb722200fe4283470dfcf/src/components/common.js#L6) which is something we should try to be consistent with.

src/components/DuosHeader.js
```diff
@@ -36,28 +52,47 @@ class DuosHeader extends Component {
     window.location = 'help_reports';
   };
 
+  okSupportRequestModal = () => {
+    this.setState(prev => {
+      prev.showSupportRequestModal = false;
+      prev.currentPage = 1;
```
rushtong: What do we use this for?

src/components/DuosHeader.js
```diff
@@ -36,28 +52,47 @@ class DuosHeader extends Component {
...
     this.setState(prev => {
       prev.showHelpModal = false;
       return prev;
     });
   };
 
+  closeSupportRequestModal = () => {
+    this.setState(prev => {
+      prev.showSupportRequestModal = false;
+      prev.currentPage = 1;
```
rushtong: Same question as above.

src/components/DuosHeader.js
```diff
@@ -36,28 +52,47 @@ class DuosHeader extends Component {
...
+
   afterModalOpen = () => {
     this.setState(prev => {
       prev.showHelpModal = false;
       return prev;
     });
   };
 
+  afterSupportRequestModalOpen = () => {
+    this.setState(prev => { prev.showSupportRequestModal = false; return prev; });
```
rushtong: nit: please split into separate lines for consistent formatting to other methods.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,250 @@
...
+        {
+          results.push(Support.uploadAttachment(this.state.attachment[i]));
+        }
+        const allToken = await Promise.all(results);
+        for (var t = 0; t < allToken.length; t++)
+        {
+          attachmentToken.push(allToken[t].token);
+        }
+      }
+      //const attachmentToken =  this.state.attachment !== '' ? await Support.uploadAttachment(this.state.attachment): '';
```
rushtong: nit: rm comment

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,250 @@
...
+export const SupportRequestModal = hh(class SupportRequestModal extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      name: Storage.userIsLogged() ? Storage.getCurrentUser().displayName : '',
+      isLogged: Storage.userIsLogged(),
+      type: 'question',
+      subject: '',
+      description: '',
+      attachment: '',
```
rushtong: The attachments are maintained in state even after the support request is canceled. We should clear those out if the user clicks `Cancel` or the close icon.

src/components/DuosHeader.js
```diff
@@ -71,6 +106,30 @@ class DuosHeader extends Component {
     }
 
     let helpLink = isAdmin ? '/help_reports' : '/home_help';
+    const contactUsIcon = isLogged ? '' : (this.state.hover ? img({src: '/images/navbar_icon_contact_us_hover.svg', style: {display: 'inline-block', margin: '0 8px 0 0', verticalAlign: 'baseline'}}) : img({src: '/images/navbar_icon_contact_us.svg', style: {display: 'inline-block', margin: '0 8px 0 0', verticalAlign: 'baseline'}}));
```
rushtong: There's a lot of duplicated code in the two conditions. I would suggest you create an image source based on logged in state. Something like this ...
```
const contactUsSource = isLogged ? '/images/navbar_icon_contact_us_hover.svg' : '/images/navbar_icon_contact_us.svg';
const contactUsIcon = img({src: contactUsSource, style: {display: 'inline-block', margin: '0 8px 0 0', verticalAlign: 'baseline'}})
```
That makes for shorter lines and less ternary duplication.

src/components/SupportRequestBaseModal.js
```diff
@@ -0,0 +1,117 @@
...
+              onClick: this.props.onRequestClose
+            }, ["Cancel"]),
+          ]),
+
+        ])
+
+      ])
+    );
+  }
+});
```
rushtong: nit: pages should have a newline

src/components/SupportRequestBaseModal.js
```diff
@@ -0,0 +1,117 @@
+import { Component } from 'react';
+import { button, div, h, span, hh } from 'react-hyperscript-helpers';
+import Modal from 'react-modal';
+import { PageSubHeading } from './PageSubHeading';
+
+
+Modal.setAppElement('#root');
+
+export const SupportRequestBaseModal = hh(class SupportRequestBaseModal extends Component {
```
rushtong: I can see where you got this pattern from. The original, `BaseModal`, is, I think, super inefficient. This could be a pure function with no need to be a component. We should take a stab at simplifying this class - will follow up with you on this.

src/components/SupportRequestBaseModal.js
```diff
@@ -0,0 +1,117 @@
...
+        background: 'rgb(255, 255, 255)',
+        outline: 'none',
+      }
+    };
+  };
+  render() {
+
+    const {
+      disableOkBtn = false
+    } = this.props;
```
rushtong: This is a good pattern, destructuring variables from props. We should do this up front so it's 💯 clear what the component needs. Scrolling down, we could also destructure `this.props.showModal`, `this.props.afterOpen`, etc., all in one place. When a developer needs to use this, they can go to there to see everything they need to provide to the component, sort of like a constructor. In typical react components, you would pull these out in the actual constructor instead of here in render. However, if we make this a pure function, we could do the destructuring right at the top.

src/libs/ajax.js
```diff
@@ -713,6 +713,43 @@ export const Summary = {
...
+
+  },
+  createSupportRequest: async (ticket) => {
+    const res = await fetchAny(`https://broadinstitute.zendesk.com/api/v2/requests.json`, fp.mergeAll([Config.jsonBody(ticket), { method: 'POST' }]));
+    return await res;
+  },
+
+  uploadAttachment: async (file) => {
+    const res = await fetchOk(`https://broadinstitute.zendesk.com/api/v2/uploads?filename=Attachment`, fp.mergeAll([Config.attachmentBody(file), { method: 'POST' }]));
+    return (await res.json()).upload
```
rushtong: semicolon

src/libs/ajax.js
```diff
@@ -713,6 +713,43 @@ export const Summary = {
...
+        { id: 360018545031, value: email }
+      ],
+      comment: {
+        body: description + '\n\n------------------\nSubmitted from: ' + url,
+        uploads: attachmentToken
+      },
+      ticket_form_id: 360000669472
+    };
+
+    return ticket
```
rushtong: semicolon

src/libs/ajax.js
rushtong: Since you're calling `fetchOk`, that can fail. If it does fail, `(await res.json()).upload` will be undefined.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,250 @@
...
+            id: 'modal',
+            type: 'danger',
+            title: alert.title,
+            description: alert.msg,
+          }),
+        ]),
+      ])
+    );
+  }
+});
```
rushtong: nit: newline

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,250 @@
...
+      name: Storage.userIsLogged() ? Storage.getCurrentUser().displayName : '',
+      isLogged: Storage.userIsLogged(),
+      type: 'question',
+      subject: '',
+      description: '',
+      attachment: '',
+      email: Storage.userIsLogged() ? Storage.getCurrentUser().email : '',
+      first_name: Storage.userIsLogged() ? Storage.getCurrentUser().displayName.split(' ')[0] : '',
+      height: Storage.userIsLogged() ? '550px': '700px',
+      top: Storage.userIsLogged() ? '400px': '1',
```
rushtong: These settings have an odd effect. If you change window sizes between showing and closing the modal, it remembers the original pixel values and shows up in odd positions along the vertical axis. We should look at how to get it to place itself relative to the screen in a flexible manner.

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,250 @@
...
+
+    this.afterOpenHandler = () => {
+      // call parent's after open handler
+      this.props.onAfterOpen('support');
+    };
+
+    this.OKHandler = async () => {
+      const attachmentToken = '';
+      if(this.state.attachment !== ''){
+        const attachmentToken = [];
```
rushtong: This is defined two lines up as a constant empty string.

rushtong (commented):
See comments inline. Let's set up a time to chat and I'd like to pull in Justin as well.

src/components/modals/SupportRequestModal.js
Marianie-Alphonse: I fixed the effect, but it looks a bit janky to me, we should discuss a cleaner option if possible

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,382 @@
...
+        first_name: Storage.userIsLogged() ?
+          Storage.getCurrentUser().displayName.split(' ')[0] :
+          '',
+        height: Storage.userIsLogged() ? (window.innerHeight < 550 ? window.innerHeight : '550px') : (window.innerHeight < 700 ? window.innerHeight : '700px'),
+        top: Storage.userIsLogged() ? '400px' : '1',
+        valid: Storage.userIsLogged(),
+        validAttachment: true
+      };
+
+      this.closeHandler = () => {
```
rushtong: Why are these methods in the constructor and not just component methods?

src/libs/ajax.js
```diff
@@ -713,6 +713,43 @@ export const Summary = {
...
+
+  },
+  createSupportRequest: async (ticket) => {
+    const res = await fetchAny(`https://broadinstitute.zendesk.com/api/v2/requests.json`, fp.mergeAll([Config.jsonBody(ticket), { method: 'POST' }]));
+    return await res;
+  },
+
+  uploadAttachment: async (file) => {
+    const res = await fetchAny(`https://broadinstitute.zendesk.com/api/v2/uploads?filename=Attachment`, fp.mergeAll([Config.attachmentBody(file), { method: 'POST' }]));
+    return (await res.json()).upload;
```
rushtong, line 749: upload might not exist on the response if the response is not successful. We should probably have a check to display an error if something goes wrong here.

rushtong (commented):
Couple minor comments, but haven't functionally tested yet. Will do that tomorrow. Looking good though!

src/components/modals/SupportRequestModal.js
```diff
@@ -0,0 +1,382 @@
...
+          height: this.state.height,
+          padding: '0px 450px 700px 0px',
+        },
+
+        content: {
+          position: 'fixed',
+          top: '1',
+          left: '1',
+          right: '0',
+          bottom: '0',
```
rushtong: Here, and in the overlay, I would suggest setting right and bottom at 20px. That sets the widget off the edges enough to look more like a widget.

rushtong (approved):
I like how the form moves with the page resizing. It works very similarly to how it does in Terra. 
Please see previous comments about functions in the constructor. If you want to resolve those, feel free to merge after that. Thank you!

src/components/modals/SupportRequestModal.js
Marianie-Alphonse: No real reason. I was following code from pages, I can move them out.

PR #493, opened 2020-06-25
[DUOS-488] Use new shibboleth service

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-488
See also: https://broadworkbench.atlassian.net/browse/DDO-462

## Changes
* Use the new Shibboleth server set up to relay era commons authentication attempts for Terra services
* Use the modified return url and token format
* Use the new shibboleth profile server to verify the token


config/staging.json
```diff
@@ -4,7 +4,9 @@
   "ontologyApiUrl": "https://consent-ontology.dsde-staging.broadinstitute.org/",
   "clientId": "1087760099392-t0nb61ehicbeuqkeftvql765ncu1707r.apps.googleusercontent.com",
   "firecloudUrl": "https://firecloud-orchestration.dsde-staging.broadinstitute.org/",
-  "nihUrl": "http://mock-nih.dev.test.firecloud.org/link-nih-account/index.html",
+  "gwasUrl": "",
+  "profileUrl": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
+  "nihUrl": "https://broad-shibboleth-prod.appspot.com/login",
```
dmohs: Prod login, dev save for token?

src/components/eRACommons.js
```diff
@@ -44,7 +44,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
     }
     if (isFcUser) {
       const parsedToken = qs.parse(searchArg);
-      this.verifyToken(parsedToken).then(
+      await this.verifyToken(parsedToken).then(
```
dmohs: This is a curious pattern. I'd love to discuss it with you offline.

dmohs (commented):
Mostly just wanted to make sure the new URLs looked right, which they do.

config/staging.json
rushtong: Oops, yes, need to fix that

src/components/eRACommons.js
rushtong: Don't think I need to await here, but we do need to verify

PR #505, opened 2020-07-08
Admin-only page adapting DAC Console to an SO Console

solideoglori:


src/pages/SigningOfficialConsole.js
```diff
@@ -0,0 +1,383 @@
...
+  };
+
+  openAccessCollect = (referenceId, electionId) => (e) => {
+    this.props.history.push(`access_collect/${electionId}/${referenceId}`);
+  };
+
+
+  handleOpenModal() {
+    this.setState({ showModal: true });
+  }
```
rushtong, line 144: Back on line 30, you have this: `this.handleOpenModal = this.handleOpenModal.bind(this);`
That's only necessary because of the construction of this method. You could eliminate lines 30-31 if you changed these two modal methods like this:

```
  handleOpenModal = () => {
    this.setState({ showModal: true });
  }
```
and 
```
  handleCloseModal = () => {
    this.setState({ showModal: false });
  }
```

rushtong (approved):
Looks good, see my suggestion inline to make it a little cleaner. One thing I want to get away from is the bootstrap styling (i.e. `const oneColumnClass = 'col-lg-1 col-md-1 col-sm-1 col-xs-1';`), but we're not there yet.

solideoglori:
Cleaned up 30/31 as suggested, merging now.

PR #518, opened 2020-07-20
[DUOS-705][risk=no] Fix for creating dars from catalog

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-705

Hotfix for creating DARs from dataset catalog selection.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -320,12 +320,6 @@ export const DAR = {
     return await res.json();
   },
 
-  partialDarFromCatalogPost: async (userId, datasetIds) => {
-    const url = `${await Config.getApiUrl()}/dar/partial/datasetCatalog?userId=${userId}`;
-    const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(datasetIds), { method: 'POST' }]));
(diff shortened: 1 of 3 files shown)


src/libs/ajax.js
```diff
@@ -320,12 +320,6 @@ export const DAR = {
     return await res.json();
   },
 
-  partialDarFromCatalogPost: async (userId, datasetIds) => {
-    const url = `${await Config.getApiUrl()}/dar/partial/datasetCatalog?userId=${userId}`;
```
rushtong, line 324: I broke this api call in https://github.com/DataBiosphere/consent/pull/663 
The [comments for the method](https://github.com/DataBiosphere/consent/commit/5770ac5f8b8bb4b3eac42a2dd8ef5056e06f7b53#diff-c9121f07792289da1889b8b24514dc8cL345) were incorrect, but I failed to follow up on whether this was used in the UI. It was, here. Instead of using this endpoint, I switched the UI code to use the same api call (undocumented, will file a new ticket for that) we use to create partial DARs via the save button on the DAR form. 

This also allows us to clean up the initial data population in the DAR form since we no longer need to handle that special condition.

PR #519, opened 2020-07-20
Help page text updates

solideoglori:


rushtong (commented):
I pushed a commit for codacy fixes.
Other feedback:
* Use single quotes instead of double quotes around string values throughout. 
* We should look at removing the `className: "home-content"` and using an inline style object.

PR #525, opened 2020-07-21
Console welcome. DS Cat guidance. Remove email on home. Chair buttons.

solideoglori:
Mainly small text changes _except_ I attempted to hide the legacy DAC Chair buttons which take the chair to Collect/Final votes pages in the old UI, and instead am trying to drive them to the new UI page.

### DIFF
--- a/src/pages/ChairConsole.js
+++ b/src/pages/ChairConsole.js
@@ -177,7 +177,7 @@ export const ChairConsole = hh(class ChairConsole extends Component {
       div({ className: 'container' }, [
         div({ className: 'col-lg-10 col-lg-offset-1 col-md-10 col-md-offset-1 col-sm-12 col-xs-12' }, [
           PageHeading({
-            id: 'chairConsole', color: 'common', title: 'Welcome ' + currentUser.displayName + '!',
+            id: 'chairConsole', color: 'common', title: 'Welcome to your DAC Chair Console, ' + currentUser.displayName + '!',
             description: 'These are your pending cases for review'
           }),
 
@@ -327,7 +327,7 @@ export const ChairConsole = hh(class ChairConsole extends Component {
                         className: 'cell-button ' + (pendingCase.alreadyVoted === true ? 'default-color' : 'cancel-color')
                       }, [
                         span({ isRendered: (pendingCase.alreadyVoted === false) && (pendingCase.electionStatus !== 'Final') }, ['Vote']),
-                        span({ isRendered: pendingCase.alreadyVoted === true }, ['Edit'])
+                        span({ isRendered: pendingCase.alreadyVoted === true }, ['Log Final Vote'])
                       ])
                     ]),
                     div({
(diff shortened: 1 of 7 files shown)


src/pages/ChairConsole.js
```diff
@@ -344,23 +344,11 @@ export const ChairConsole = hh(class ChairConsole extends Component {
...
-                      }, ['Collect Votes'])
                     ]),
                     div({
                       isRendered: (pendingCase.alreadyVoted === true) && (pendingCase.electionStatus === 'Final'),
                       className: twoColumnClass + ' cell-body f-center'
                     }, [
-                      button({
-                        id: pendingCase.frontEndId + '_btnFinal',
-                        name: 'btn_finalAccess',
-                        onClick: this.openFinalAccessReview(pendingCase.referenceId, pendingCase.electionId, pendingCase.rpElectionId),
```
rushtong, line 361: The method, `openFinalAccessReview`, is no longer used so we can delete that.

src/pages/ChairConsole.js
```diff
@@ -344,23 +344,11 @@ export const ChairConsole = hh(class ChairConsole extends Component {
                       isRendered: this.isAccessCollectEnabled(pendingCase),
                       className: twoColumnClass + ' cell-body f-center'
                     }, [
-                      button({
-                        id: pendingCase.frontEndId + '_btnCollect',
-                        name: 'btn_collectAccess',
-                        onClick: this.openAccessCollect(pendingCase.referenceId, pendingCase.electionId),
```
rushtong, line 350: The method, `openAccessCollect`, is no longer used so we can delete that.

src/pages/ChairConsole.js
rushtong: Looking into `openFinalAccessReview`, since that is no longer used, we also no longer need the page that it refers to, which is `FinalAccessReview`. That whole code path can be deleted.

rushtong (commented):
The small text changes look good. The button removals, however, could use some work - it might be worth either separating that out into another task or adding a cleanup story to deal with the fallout. Let me know how you'd like to proceed.

Specifically, there are a bunch of pages we should no longer be able to get to if we remove access to the original view. The pages that I think should go away include:
* Final Access Review
* Access Collect
* the old Access Review

Some of these are still referenced by admin pages, but Final Access Review would be completely dead code now.

solideoglori:
Right, it's likely time then to remove the former pages. 

However, if we do that will that cause an issue with the Access Result (i.e. the summary page accessed via 'Reviewed Cases Record')?

rushtong:
> Right, it's likely time then to remove the former pages.
> 
> However, if we do that will that cause an issue with the Access Result (i.e. the summary page accessed via 'Reviewed Cases Record')?

That's exactly why I suggest we need to look into it in a full story, I don't know all of the tendrils that are out there. It's a non-trivial amount of sleuthing/debugging to find all the unused code.

solideoglori:
@rushtong I removed the chair console changes, so I think this should be good to go.

rushtong (approved):

PR #527, opened 2020-07-23
[DUOS-591] React-GA/Google Analytics set-up

JVThomas:
PR aims to add GA integrations via the react-ga package. As of now the PR will only track pageviews however the react-ga package can be used to track even more metrics (to be defined later on). 

Some key notes on the PR that I'd be willing to discuss in a call...
- The App component was rewritten from a class component to a functional component. The reason behind it revolves around the app being an SPA. It doesn't have the normal page load/refresh behavior you would expect from a website so you have to actively listen for page views within the app. 
    - Normally this can be done by accessing the history object provided by react-router, however the app uses BrowserRouter which only allows access within components that have been wrapped with the ```withRouter``` function, meaning you would have to register a listener on every component.
    - Rather than doing that we can use a nifty React hook that has been introduced in react-router v5 called ```useHistory```, which will return the nearest history instance (basically from the closest Router parent) regardless of which type of Router you use. This allows me to set up a single listener on the app level rather than individual instances on all current and future components. But hooks can only be used within functional components, not classes, hence the refactor.
    - As a part of the refactor I noticed there were spinner methods that didn't seemed to be called or passed in as props anywhere, so I removed them. However that leaves the question of what to do with ```var loading```, which seems to change value only when the spinner functions are called.
- Currently the GA tracking id is a dummy id that I've set up to test the implementation. At some point we need to replace it with keys that are dynamically assigned based on the current environment (since you should have separate tracking ids for development, staging, and production)

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -7,7 +7,7 @@
   "gwasUrl": "",
   "profileUrl": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
   "nihUrl": "https://broad-shibboleth-prod.appspot.com/dev/login",
-  "gaId": "UA-173273990-1",
+  "gaId": "UA-173408794-1",
   "features": {
     "newDarUi": true
   }
--- a/config/dev.json
+++ b/config/dev.json
@@ -7,6 +7,7 @@
   "gwasUrl": "",
   "profileUrl": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
   "nihUrl": "https://broad-shibboleth-prod.appspot.com/dev/login",
+  "gaId": "UA-173273990-1",
   "features": {
     "newDarUi": true
   }
--- a/config/dev.json
+++ b/config/dev.json
@@ -7,7 +7,7 @@
   "gwasUrl": "",
   "profileUrl": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
   "nihUrl": "https://broad-shibboleth-prod.appspot.com/dev/login",
-  "gaId": "UA-173408794-1",
+  "gaId": "UA-159441671-1",
   "features": {
     "newDarUi": true
   }
--- a/src/App.js
+++ b/src/App.js
@@ -1,18 +1,18 @@
-import React, {useState, useEffect} from 'react';
+import {useState, useEffect} from 'react';
 import ReactGA from 'react-ga';
 import { div, h } from 'react-hyperscript-helpers';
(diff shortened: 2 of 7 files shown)


src/App.js
```diff
@@ -1,77 +1,71 @@
...
-      loading: false,
-      isLoggedIn: false
-    };
-  }
-
-  showSpinner = () => {
-    this.setState({
-      loading: true
+  const initializeReactGA = () => {
+    ReactGA.initialize('UA-173273990-1', {
```
rushtong: We should pull this from configs similarly to how we grab other keys like `Config.getGoogleClientId()` The complication here is that we have to wait for the app to load completely before doing so, similarly to how we wait for the config object to load before enabling the sign-in button.

src/App.js
```diff
@@ -1,77 +1,71 @@
...
 
-  signOut = () => {
+    Modal.setAppElement(document.getElementById('modal-root'));
+    initializeReactGA();
+    //call trackPageView to register initial page load
+    trackPageView(history.location);
+    //pass trackPageView as callback function for url change listener
+    history.listen(trackPageView);
+    setUserIsLogged();
+  }, []);
```
rushtong: This should fix the codacy warning
```suggestion
  }, [history]);
```

rushtong (commented):
See minor comments inline.
👍 to refactoring App - love the changes
👍 to getting rid of spinner here - I strongly dislike that approach and would love to see it be refactored throughout the app
👍 Love seeing `useEffect` and `useHistory`, let's keep those approaches in future work.

Side note: if the hyperscript helpers library becomes cumbersome, I'm not opposed to ditching it.

src/App.js
JVThomas: Right, I was worried about potential re-renders on dependency updates which is why I opted for the empty array (only executes on mount), however the history instance doesn't change (only its properties) so passing it in as a dependency is the right call.

config/dev.json
```diff
@@ -7,6 +7,7 @@
   "gwasUrl": "",
   "profileUrl": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
   "nihUrl": "https://broad-shibboleth-prod.appspot.com/dev/login",
+  "gaId": "UA-173408794-1",
```
rushtong: We will need to add real values to this field in the other config files before we push to develop. I'll get started on that today.

rushtong (approved):
👍 Please squash and merge when you're ready.

PR #528, opened 2020-07-23
[No Ticket] Install Trivy to scan docker image

zbedo:
This PR introduces:

- Trivy installation
- Docker image security scan, if there are CRITICAL findings, job will fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,10 +48,15 @@ commands:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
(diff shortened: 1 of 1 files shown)


rushtong (approved):
👍 Thanks for doing this! We also have dependabot scan for docker updates, but this is a nice extra layer.

rushtong:
@adrazhi Can you 👍 this when you're able to? I'd like to move this through, thank you!

adrazhi (approved):

PR #533, opened 2020-07-29
[DUOS-553] Footer style update

lindojong:
### Addresses
https://broadinstitute.atlassian.net/browse/DUOS-553


src/components/DuosFooter.js
```diff
@@ -1,32 +1,23 @@
...
-          img({ src: "/images/broad_logo.svg", className: "footer-logo", alt: "Broad Institute logo" })
+  return (
+    div({ className: "footer" }, [
+      footer({ className: "main-footer" }, [
+        img({ src: "/images/broad_logo_allwhite.png", className: "footer-logo", alt: "Broad Institute logo" }),
+        ul({ className: "footer-links" }, [
+          li({ className: "footer-links__item" }, ["\u00A9 Broad Institute"]),
+          li({ className: "footer-links__item" }, [a({ target: '_blank', href: "https://www.broadinstitute.org/privacy-policy" }, ["Privacy Policy"]),]),
+          li({ className: "footer-links__item" }, [a({ target: '_blank', href: "https://www.broadinstitute.org/terms-conditions" }, ["Terms of Service"]),]),
+          li({ className: "footer-links__item" }, [a({ href: supportLink }, ["Support"]),])
```
rushtong: Pretty sure the test is failing because `li({ className: 'footer-links__item' }, [a({ href: '/status' }, ['Status'])])` is missing.

src/components/DuosFooter.js
```diff
@@ -1,32 +1,23 @@
-import { Component } from 'react';
 import { a, div, footer, img, li, ul } from 'react-hyperscript-helpers';
 import { Storage } from '../libs/storage';
 
+const DuosFooter = (props) => {
+  const isAdmin = props.isLogged ? Storage.getCurrentUser().isAdmin : false;
```
rushtong: You could clean this up a bit. For example, `Storage.getCurrentUser()` returns null if it isn't set. So, instead of relying on props, you could just call that and check if null, then look for the `isAdmin` property. I like lodash/functional lodash for this kind of thing.
* Lodash equivalent: https://lodash.com/docs/#isEmpty
* https://github.com/lodash/lodash/wiki/FP-Guide

src/components/DuosFooter.js
rushtong: To test the very few integration tests we do have, do this.
* Terminal window 1: `npm start`
* Terminal window 2: `npm run-script cypress:run`

src/index.css
```diff
@@ -1444,16 +1444,16 @@ input:-webkit-autofill {
...
 
 .footer-logo {
-  float: right;
+  float: left;
   height: 32px;
-  width: 127px;
-  margin-top: 22px;
+  margin-top: 15px;
+  margin-right: 35px;
 }
```
rushtong, line 1457: When possible, I prefer to use inline style objects instead of relying on css class names. Here are some docs that describe this approach: https://reactjs.org/docs/dom-elements.html#style

For this PR, don't worry, it's not a requirement. For future work, we'll try to move towards using style objects instead of css classes. Some of my favorite reasons for this:

1. Style objects can be refactored easily with modern IDEs, but css is more difficult.
2. Style objects can be easily merged/manipulated with event handlers.
3. Style objects can live close to the rendered objects and are less likely to become orphaned. 
4. CSS classes, on the other hand, are a maintenance headache.

rushtong (approved):
Looks fantastic! 👍

rushtong:
@lindojong Please squash and merge when you're ready!

PR #536, opened 2020-07-30
DUOS-674 UI: DAR Application Step 1 Component

JVThomas:
Step 1 has been pulled out into its own functional component, going to layout questions comments below...

- There's a lot of props being passed in to the component that mainly consist of parent state manipulators and read-only variables necessary for conditional rendering. Maybe Step 1 can be broken down further, maybe the data flow from parent to child can be optimized. It's hard to say for sure until all the steps have been converted to components.
- Questions about prompts and my code have been added as comments for the purpose of review. I'll remove them once the code is good to go.
- Researcher profile data is now being pulled in from the profile as read-only variables and is no longer being saved as part of the form data. I remember you mentioning that there were aspects of the saved data that just repeated information from other tables unnecessarily and it seemed like the profile data was one example of this.


src/pages/dar_application/ResearcherInfo.jsx
```diff
@@ -0,0 +1,183 @@
...
+    const name = event.target.name;
+    const value = event.target[attr];
+    //NOTE: come up with a better name for this function, sounds too similar to formStateChange
+    props.formFieldChange(name, value);
+    stateVarSetter(value);
+  };
+
+  return (
+    div({ className: 'col-lg-10 col-lg-offset-1 col-md-12 col-sm-12 col-xs-12' }, [
+      fieldset({ disabled: props.darCode !== null }, [
```
rushtong: I like https://lodash.com/docs/#isNil for these kinds of checks - it's mostly a matter of style.

rushtong (approved):
One minor comment inline - great work and ready for merging with the cleanup mentioned.

Question:
* Do you prefer jsx over js? I'm not opposed to going down either path, but want to make a conscious decision that yes, we're doing "X" for reasons "Y", "Z".

Other thoughts:
* I'm trying to rethink the approach of billions of little fields, field handlers, field updaters, etc. being passed around like we've discussed in standup.
* One idea might be to move field handlers into their respective components and instead of passing in field values, pass in an uber-object. The parent then can look at that object when the component updates it and decide what to do with it, post, show validation errors, etc.
* Inside each component, it deals with the part of the object that it cares about so handlers are local to the component.
* That object could also be an object specific to that component, i.e. `section2Form` and the same idea would apply. The parent then needs to only know about the 4 different section forms and deal with them as each component modifies their state. 
* This approach would mean moving some validation methods into the components too and tracking validation status across all of them in the parent.

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -145,7 +141,6 @@ export default function ResearcherInfo(props) {
           div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group' }, [
             label({ className: 'control-label rp-title-question' }, [
               '1.3 Principal Investigator* ',
-              //NOTE: user doesn't actually type in the name, it's prefilled. Should I adjust phrasing?
```
rushtong: I did miss this question the first time around, sorry. Yes, if you want to adjust the phrasing, that would be a good idea. Perhaps something like 
```
"I certify that the principal investigator is aware of this study."
```
@solideoglori Do you have a better suggestion?

rushtong (approved):
Approved pending @solideoglori

src/pages/dar_application/ResearcherInfo.js
solideoglori: "I certify that the principal investigator listed below is aware of this study."

Thanks @JVThomas and @rushtong

rushtong (approved):
👍 Ready for merging!

PR #538, opened 2020-08-03
DUOS-675 DAR Application Step 2 Component

JVThomas:
## Addresses 
DUOS-675: https://broadinstitute.atlassian.net/browse/DUOS-675

## Notes
- Added a nextPage and a prevPage helper method to replace to individual goToStep functions currently used
- Passed in TypesOfResearch component as a prop for step 2 since it's prop values are used for it alone
- Switched out onChange for onBlur to limit re-rendering on text inputs for non Async search text inputs to limit re-renders
- Moved searchDatasets function into step 2 component (it's only used there)
- Misc comments added to code for later plans on component conversions


src/pages/DataAccessRequestApplication.js
```diff
@@ -264,6 +264,23 @@ class DataAccessRequestApplication extends Component {
     }, () => this.checkValidations());
   };
 
+//NOTE: use nextPage and previous page instead of having indiividual go to pages for each step
```
rushtong: nit: typo
```suggestion
//NOTE: use nextPage and previous page instead of having individual go to pages for each step
```

src/pages/DataAccessRequestApplication.js
```diff
@@ -264,6 +264,23 @@ class DataAccessRequestApplication extends Component {
...
+  window.scrollTo(0,0);
+}
+
+prevPage = (e) => {
+  this.setState(prev => {
+    prev.step = prev.step - 1;
+    return prev;
+  });
+  window.scrollTo(0,0);
+}
```
rushtong, line 282: Nice logic abstraction 👍

src/pages/DataAccessRequestApplication.js
```diff
@@ -616,6 +620,25 @@ class DataAccessRequestApplication extends Component {
     const isTypeOfResearchInvalid = this.isTypeOfResearchInvalid();
     const genderLabels = ['Female', 'Male'];
     const genderValues = ['F', 'M'];
+
+    //NOTE: component is only here temporarily until component conversion has been complete
+    //ideally this, along with the other variable initializtion should be done with a useEffect hook
```
rushtong: nit: typo
```suggestion
    //ideally this, along with the other variable initialization should be done with a useEffect hook
```

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -67,7 +59,7 @@ export default function ResearcherInfo(props) {
               className: 'checkbox-inline rp-checkbox',
               disabled: !isNil(props.darCode),
               checked: checkCollaborator,
-              onChange: (e) => formStateChange(setCheckCollaborator, 'checked', e)
+              onChange: (e) => props.formStateChange(setCheckCollaborator, 'checked', e)
```
rushtong: One bit of code structure that I struggle with in our existing code is that referring to props directly, spread out through the html, is confusing. I like having all of the required props destructured into constants in one place, usually at the top of the render method, so as a dev, I can see all of the component requirements in one place. In a typical component, that's usually in the constructor. In a functional world, we should choose someplace that makes sense for us going forward.

rushtong (approved):
Looks good. One bit of structural feedback - feel free to address if you think it makes sense.

rushtong (approved):
Thank you for the changes, they look great 👍

PR #539, opened 2020-08-03
DUOS-552 Homepage Redesign

lindojong:
Work to redesign the old homepage. https://broadinstitute.atlassian.net/browse/DUOS-552

Still WIP to clean up CSS styling + classes but have most of the general look finished:

duos1
duos2


src/components/HomeReadMore.js
```diff
@@ -0,0 +1,56 @@
+import get from 'lodash/get';
```
lindojong: Created a new file so I wouldn't touch the old ReadMore which is still used on another page. The original ReadMore expects pure text, and outputs `...Read More` in-line. I could spend more time to combine the two (which is probably ideal, design-wise).

src/components/SignIn.js
```diff
@@ -140,8 +140,6 @@ export const SignIn = hh(class SignIn extends Component {
           clientId: this.state.clientId,
           onSuccess: this.responseGoogle,
           onFailure: this.forbidden,
-          icon: false,
-          buttonText: 'Start here!'
```
lindojong, line 144: Changes the look from a plain 'Start here!' button to the Google-branded button.

src/pages/Home.js
```diff
@@ -1,72 +1,149 @@
...
-                  a({
-                    href: 'https://support.google.com/accounts/answer/176347',
-                    target: '_blank'
-                  }, ['using these instructions.'])
-                ])
-              ])
+      div({ className: 'row' }, [
+        div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12' }, [
+          div({ className: 'row no-margin', style: { backgroundColor: 'white', height: '350px', position: 'relative' }}, [
+            img({ style: { height: 'inherit', minWidth: '100%', transform: 'scale(1.1,1)', zIndex: '-99' }, src: '/images/home_header_background.png'}),
```
lindojong: The banner image had a transparent border (might be from a shadow effect?) so I scaled it up as a hacky fix to have the actual image span the width of the page. Might want to modify the image itself to have it cropped properly.

solideoglori:
Can we format the 'About DUOS' paragraphs as 'justified'?

solideoglori:
Can we also remove the 'by leveraging' from the end of the 'What is DUOS and How does it work subtitle description'?

solideoglori:
For the 'Are you a researcher?' subtitle description: "Click here to register and start your data access request!"

For the 'Are you a DAC Member?' subtitle description: "Click here to learn how DUOS is helping DACs efficiently manage data access and use compliance"

Can we also invert the positioning so that the 'researcher' text is under the 'data requestor' side of the diagram?

src/pages/Home.js
```diff
@@ -1,72 +1,160 @@
...
```
JVThomas: So on Firefox the banner image has a zIndex of -99, which puts it behind every element on the page. That's fine except for the fact that its parent div has ```backgroundColor:white```, which will cover up the banner image. Use ```backgroundColor: transparent``` on the parent instead to let the banner show.

src/components/ReadMore.js
```diff
@@ -44,15 +70,23 @@ export const ReadMore = hh(class ReadMore extends Component {
     });
   };
 
+  getReadLink = (fun, text, classes) => {
+    if (this.state.inline) {
+      return a({ onClick: () => fun() }, [text])
```
rushtong: Codacy suggestion:
```suggestion
      return a({ onClick: () => fun() }, [text]);
```

src/components/ReadMore.js
```diff
@@ -44,15 +70,23 @@ export const ReadMore = hh(class ReadMore extends Component {
...
+    } else {
+      return a({ onClick: () => fun(), style: this.state.readStyle }, [
+        text,
+        span({className: classes, style: {padding: '0 1rem'}, 'aria-hidden': 'true'})
```
rushtong: Codacy suggestion:
```suggestion
        span({className: classes, style: {padding: '0 1rem'}, 'aria-hidden': 'true'});
```

src/pages/Home.js
```diff
@@ -1,72 +1,172 @@
...
+            div({ className: 'col-lg-4 col-lg-offset-2 col-md-4 col-md-offset-2' }, [
+              p({ style: header }, ['Are you a researcher?']),
+              p({ style: description }, [
+                'Click here to register and start your data access request!']),
+              div({className:'row', style: { display: 'flex', justifyContent: 'center' }}, [
+                button({ className: 'btn-primary', style: buttonStyle }, [
+                  a({
+                    href: 'https://accounts.google.com/SignUp?continue:https%3A%2F%2Faccounts.google.com%2Fo%2Foauth2%2Fauth%3Fopenid.realm%26scope%3Demail%2Bprofile%2Bopenid%26response_type%3Dpermission%26redirect_uri%3Dstoragerelay%3A%2F%2Fhttp%2Flocalhost%3A8000%3Fid%253Dauth721210%26ss_domain%3Dhttp%3A%2F%2Flocalhost%3A8000%26client_id%3D832251491634-smgc3b2pogqer1mmdrd3hrqic3leof3p.apps.googleusercontent.com%26fetch_basic_profile%3Dtrue%26hl%3Des-419%26from_login%3D1%26as%3D43c5de35a7316d00&ltmpl:popup',
+                    style: { color: '#fff' }}, [
+                      'REGISTER'
```
rushtong: Codacy suggestion:
```suggestion
                    'REGISTER'
```

rushtong:
One bit of visual feedback. Can we remove the white strip between the large header image and the top nav?

Screen Shot 2020-08-05 at 9 43 45 AM

cypress/integration/home_spec.js
```diff
@@ -5,8 +5,11 @@ describe('Home', function() {
   it('Home page loads correctly', function() {
     cy.visit('');
     cy.contains('DUOS');
-    cy.contains('Start here!');
-    cy.contains('What is DUOS?');
+    cy.contains('Sign in with Google');
+    cy.contains('What is DUOS and how does it work?');
+    cy.contains('Are you a DAC member?');
+    cy.contains('Are you a researcher?');
+    cy.contains('About DUOS');
```
rushtong, line 12: Nice test additions!

src/components/ReadMore.js
```diff
@@ -23,10 +26,33 @@ export const ReadMore = hh(class ReadMore extends Component {
   }
 
   getContent = () => {
+    if (this.state.inline) {
+      return this.getInlineContent();
+    }
+    else {
+      return this.getFormattedContent();
+    }
+  };
```
rushtong, line 30: This could be condensed to a single ternary, which is mostly a matter of style:
```
    return this.state.inline ? this.getInlineContent() : this.getFormattedContent();
```

src/components/ReadMore.js
```diff
@@ -23,10 +26,33 @@ export const ReadMore = hh(class ReadMore extends Component {
...
+        style: this.state.style,
+      }, this.state.content);
     } else {
-      return this.state.content.slice(0, this.state.charLimit) + ' ...';
+      return span({
+        className: this.state.className,
+        style: this.state.style,
+      }, this.state.content.slice(0, this.state.charLimit) + ' ...');
+    }
+  };
```
rushtong, line 38: Here's another example where the code can be shortened a lot with a ternary that describes the only thing different in the two conditions, the content:
```
    const content = this.state.expanded ? this.state.content : this.state.content.slice(0, this.state.charLimit) + ' ...';
    return span({
      className: this.state.className,
      style: this.state.style,
    }, content);
```

src/components/ReadMore.js
```diff
@@ -44,15 +70,23 @@ export const ReadMore = hh(class ReadMore extends Component {
...
+  getReadLink = (fun, text, classes) => {
+    if (this.state.inline) {
+      return a({ onClick: () => fun() }, [text])
+    } else {
+      return a({ onClick: () => fun(), style: this.state.readStyle }, [
+        text,
+        span({className: classes, style: {padding: '0 1rem'}, 'aria-hidden': 'true'})
+      ])
+    }
+  }
```
rushtong, line 71: I think this is another example where ternaries can help reduce code duplication:
```
    const linkElements = this.state.inline ? [text] :
      [text, span({className: classes, style: {padding: '0 1rem'}, 'aria-hidden': 'true'})];
    const linkElementsStyle = this.state.inline ? {} : this.state.readStyle;
    return a({ onClick: () => fun(), style: linkElementsStyle}, linkElements);
```

src/pages/Home.js
```diff
@@ -1,72 +1,172 @@
...
+              img({ style: duosLogo , alt: 'DUOS logo', src: '/images/duos_logo.svg' }),
+              h1({ style: homeTitle }, ['Data Use Oversight System']),
+              div({ style: homeBannerDescription }, [
+                'Expediting data access for researchers, by facilitating and \nenhancing data access committee\'s workflows'])
             ]),
-            p({ style: { fontSize: 'larger', fontWeight: '500' } }, ['Are you a DAC member who wants to learn more about DUOS?']),
-            p({ style: { fontSize: 'large', fontWeight: '350' } }, ['Email us! ']),
-            p({ style: { fontSize: 'large', fontWeight: '500' } }, [
-              h(Mailto, { email: 'DUOS-support@broadinstitute.zendesk.com' })
+            div({ isRendered: !this.props.isLogged, style: { padding: '2em 2em 0 0', display: 'flex', alignItems: 'center', position: 'absolute', top: '1rem', right: '1rem'}}, [
```
rushtong: One of the suggestions I made for Justin's PR is that we try to destructure as many props at the top of the render method into constants. That way, anyone who works on this code will know all of the props that are required in one place. If they're spread out through the html like this, some can be easily overlooked.

rushtong (changes requested):
Visually, looks great. See feedback in line as well as conversation in slack about the register button.

lindojong:
@rushtong About the white space, I think we could get rid of it if we could get a version of the image with the transparent pixels cropped out. That white space comes from the photo and the header.
Capture

src/components/HomeReadMore.js
rushtong: If you want to take a stab at merging those files, that would be ideal, but I'm fine either way.

src/components/ReadMore.js
rushtong: For bonus points, here is a way to destructure two constants from a single ternary expression:
```
    const {linkElements, linkElementsStyle} = this.state.inline ?
      {linkElements: [text], linkElementsStyle: {}} :
      {
        linkElements: [
          text,
          span({
            className: classes,
            style: {padding: '0 1rem'},
            'aria-hidden': 'true',
          })], linkElementsStyle: this.state.readStyle,
      };
    return a({ onClick: () => fun(), style: linkElementsStyle}, linkElements);
```

src/pages/Home.js
lindojong: I believe I've addressed this now (wasn't sure if I needed to handle `props: this.props`, but destructured the rest), if you could check it again.

rushtong (approved):
Thank you for the changes, they look great! 👍

PR #540, opened 2020-08-03
Remove link to NHGRI info page in header, maintain page

solideoglori:


JVThomas:
The link is removed from the header, but the "'/NHGRIpilotinfo" link can still be visited by manually adding that to the url. Is that by design?

rushtong (commented):
See also these references:
* https://github.com/DataBiosphere/duos-ui/blob/04c0d005c0e9fc54996ca7ca54e322616bb9ebdb/src/Routes.js#L42
* https://github.com/DataBiosphere/duos-ui/blob/4f78aebde1a0c910a59b11c92f4b2947de8510d1/src/pages/NHGRIpilotinfo.js#L4

Do we need to keep this page around or can we remove it altogether?

solideoglori:
I'd like to hide the link but leave the page accessible since NIH folks will want to view 
It's telling me there is now a conflict. How do I resolve this?

rushtong:
From email:
> I'd like to hide the link but leave the page accessible since NIH folks will want to view
> It's telling me there is now a conflict. How do I resolve this?

@solideoglori If that's the case, then this is intentional and we can leave the page and routing in place. 👍

rushtong (approved):

PR #541, opened 2020-08-03
Remove links to legacy DAA and PDF

solideoglori:


src/pages/ResearcherProfile.js
```diff
@@ -944,49 +944,6 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
-                    ['2. Upload your template of the Data Access Agreement signed by your organization\'s Signing Official'])
-                ]),
-
-                div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12' }, [
-                  div({
-                    style: { paddingTop: 7 }, className: 'fileUpload col-lg-4 col-md-4 col-sm-6 col-xs-12 btn-secondary btn-download-pdf hover-color'
-                  }, [
-                    span({ className: 'glyphicon glyphicon-upload' }),
-                    'Upload S.O. Signed Agreement',
-                    input({ id: 'uploadFile', type: 'file', onChange: this.handleFileChange, className: 'upload' })
```
rushtong, line 981: The `handleFileChange` method is no longer necessary and can be removed. Lines 30 and 164

Removing that method also means we don't need `this.state.file` which is used for several calculations in the submit handler. It looks like we can remove this along with the calls to `DAR.postDAA`.

`DAR.postDAA` is only used by this page, `DataAccessRequestRenewal`, and `DatasetRegistration`. I believe those pages could also be cleaned up a lot.

rushtong (changes requested):
I approve of this change, but there is a good amount of technical debt left behind with this removal. We should also look into `DataAccessRequestRenewal` and `DatasetRegistration` as they have similar logic around posting DAA files.

solideoglori:
Great. I will merge this and work on cleanup in the others. Those pages are less priority to resolve such issues as they are hidden pages to non-Admin users.

PR #552, opened 2020-08-11
DUOS-717 Expanding Alert Notification on Inputs

JVThomas:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-717

## Notes
- Code now indicates (with links) to the specific steps that have failed verification on step 4 component mount. Input fields now have labels(step 2) or background color highlights (step 3) to indicate missing data for required fields
  - Opted to use background colors instead of labels for step 3 because having "This is a required field" on a list of tightly grouped questions (in terms of layout) would look messy
- One lingering question regarding step 3:
  - The front-end initializes the radio questions as empty strings for the sake of having them registered as controlled inputs. However when a partial application is saved and the page is refreshed, the radio questions are set to false as default values, leaving me to believe that the back-end initializes the empty values as false
  - The cause for concern is that under normal entry the user has the assumption that blank values will be caught and highlighted upon reaching step 4. However that assumption is false when resuming a partial application, creating a potential scenario where a user can accidentally submit a request with incorrect values for step 3.
  - It also leaves me with the awkward position of having to do <variableName>.toString().length === 0 to account for both empty string (which is not evaluated properly by isNil) and false ( which is not evaluated properly with isEmpty)


src/pages/dar_application/ResearchPurposeStatement.js
```diff
@@ -31,168 +30,315 @@ export default function ResearchPurposeStatement(props) {
...
+                ['3.1 In order to ensure appropriate review, please answer the questions below:'])
+            ]),
           ]),
-          div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group' }, [
-            YesNoRadioGroup({
-              value: forProfit, onChange: handleRadioChange, id: 'forProfit', name: 'forProfit',
-              required: true
-            })
+          div({className: 'radio-question-container row no-margin', style: {
+            backgroundColor: showValidationMessages && forProfit.toString().length === 0 ? alertBackgroundColor : 'inherit'
```
rushtong, line 54: I hit an error here:

1. Create DAR from selected datasets
2. Navigate to step 4
3. Follow validation error link to step 2.
4. Enter data in step 2 until it is valid
5. Navigate to step 3, get error:
```
TypeError: Cannot read property 'toString' of undefined
ResearchPurposeStatement
src/pages/dar_application/ResearchPurposeStatement.js:54
  51 |   ]),
  52 | ]),
  53 | div({className: 'radio-question-container row no-margin', style: {
> 54 |   backgroundColor: showValidationMessages && forProfit.toString().length === 0 ? alertBackgroundColor : 'inherit'
     | ^  55 | }}, [
  56 |   div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group ' }, [
  57 |     label({ className: 'control-label rp-choice-questions' },
```

rushtong: I hit the same error if having completed step 2, then navigating to step 4, and follow the validation error link to step 3.

rushtong: Another thought on this, I think `forProfit` is unique in that it is in step 2 and in step 3.

rushtong (changes requested):
One minor functionality bug, but otherwise, this looks great!

src/pages/dar_application/ResearchPurposeStatement.js
JVThomas: I can't seem to reproduce it on my end, but the toString check is prone to error if the props are not initialized on controller mount. I can add a variable check (ex: ```forProfit !== undefined && forProfit.toString()...```) on the inputs to ensure that if it's undefined the toString check won't go through

src/pages/DataAccessRequestApplication.js
```diff
@@ -192,7 +192,7 @@ class DataAccessRequestApplication extends Component {
     }
     this.setState(prev => {
       prev.completed = completed;
-      prev.formData = formData;
+      prev.formData = fp.merge(prev.formData, formData);
```
JVThomas, line 195: When creating a DAR from Dataset Catalog it generates json that doesn't have all of the keys initialized (something the DARApplication does in its constructor). So when this init function was called, it would completely overwrite the initial blank values on the state values with undefined due to missing keys in the payload. Merging the payload into the state variable will ensure proper updates while maintaining the initialized blank values for the missing values.

rushtong (approved):
👍 Especially like the error styling on Step 3 radio buttons, nice work!

PR #562, opened 2020-08-14
[DUOS-660][risk=no] Rename Changed DAR Fields

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-660
Required for https://github.com/DataBiosphere/consent/pull/698

## Changes
Field renaming for all DAR properties that are being modified on the back end. 
As a result of that change, the structure of what's needed for a DAR is minimized. For example, we no longer need to send a user id or a fully populated datasets field. The user is determined from the auth token and the datasets are determined from the datasetIds. Also, no user/researcher information is stored on the DAR any longer.


JVThomas (commented):
The code looks good, but I can't test the code with npm start until the back-end code is deployed. There are some parts (like non_tech_rus) where the data doesn't save properly, but I'm assuming that's because I'm connecting to dev which has the previous API endpoints.

rushtong:
@JVThomas - Yes, that would be the issue. To test locally, you have to spin up consent locally, then point your UI to the local.broadinstitute url.

JVThomas (approved):
Pointed the config file to the local consent api, code saves and populates as expected.

PR #573, opened 2020-09-01
DUOS 673 Remove user details from darInfo & describeDar refactor

JVThomas:
Addresses [DUOS-673](https://broadinstitute.atlassian.net/browse/DUOS-673)

Basic premise is to extract user details out of darInfo and rely on the data returned from ResearcherProfile. Additional changes also include removing dependency on the modal summary endpoint and instead using the v2 DAR GET and User GET endpoints as well as adding additional helper functions and details to the dataUseTranslation file to define purpose statements and researchTypes so that the back-end does not have to (nor should it, data presentation is a front-role responsibility) 

In all cases:

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1,5 +1,4 @@
 import fileDownload from 'js-file-download';
-import * as ld from 'lodash';
 import * as fp from 'lodash/fp';
 import { Config } from './config';
 import { Models } from './models';
@@ -289,8 +288,7 @@ export const DAR = {
     // const piName = ld.get(ld.head(ld.filter(darInfo.researcherProperties, { 'propertyKey': 'piName' })), 'propertyValue', "");
     darInfo.pi = rawDar.investigator;
     darInfo.havePI = rawDar.havePI || rawDar.isThePI;
-    darInfo.profileName = researcher.displayName
-    
+    darInfo.profileName = researcher.displayName;
     // dataUse from Models.dar has properties denoting what research the data will be used for.
     // Get these properties directly from the DAR.
     const dataUseModel = fp.keys(darInfo.dataUse);
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -263,7 +263,6 @@ export const DAR = {
     darInfo.country = rawDar.country;
     darInfo.status = rawDar.status;
 
-    //NOTE:does rationale need to come from researcher or current user (guessing researcher)?
     darInfo.hasAdminComment = researcher.rationale != null;
     darInfo.adminComment = researcher.rationale;
     const purposeStatements = DataUseTranslation.generatePurposeStatement(darInfo);
@@ -283,7 +282,6 @@ export const DAR = {
     darInfo.datasets = rawDar.datasets;
     darInfo.pi = rawDar.investigator;
     darInfo.havePI = rawDar.havePI || rawDar.isThePI;
-    //NOTE:Is this meant to display the researcher name?
     darInfo.profileName = researcher.displayName;
     // dataUse from Models.dar has properties denoting what research the data will be used for.
     // Get these properties directly from the DAR.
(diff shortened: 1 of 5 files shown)


src/libs/ajax.js
```diff
@@ -251,39 +251,40 @@ export const DAR = {
...
+    darInfo.darCode = rawDar.darCode;
+    darInfo.projectTitle = rawDar.projectTitle;
+    darInfo.institution = rawDar.institutionName;
+    darInfo.department = rawDar.department;
+    darInfo.city = rawDar.city;
+    darInfo.country = rawDar.country;
+    darInfo.status = rawDar.status;
+
+    //NOTE:does rationale need to come from researcher or current user (guessing researcher)?
+    darInfo.hasAdminComment = researcher.rationale != null;
```
lindojong, line 266: I'm inclined to believe that rationale should come from the current user? Rationale appears to be the comment made for the reasoning behind a vote approving/denying a request (see all the finalRationale fields across elections and votes). This particular field only seems to be accessed from admin level roles though.

src/libs/ajax.js
```diff
@@ -251,39 +251,40 @@ export const DAR = {
...
-    const profileName = ld.get(ld.head(ld.filter(darInfo.researcherProperties, { 'propertyKey': 'profileName' })), 'propertyValue', "");
-    const piName = ld.get(ld.head(ld.filter(darInfo.researcherProperties, { 'propertyKey': 'piName' })), 'propertyValue', "");
-    darInfo.pi = isThePI ? profileName : piName;
-    darInfo.havePI = havePI || isThePI;
-    darInfo.profileName = profileName;
+
+    darInfo.datasets = rawDar.datasets;
+    darInfo.pi = rawDar.investigator;
+    darInfo.havePI = rawDar.havePI || rawDar.isThePI;
+    //NOTE:Is this meant to display the researcher name?
```
lindojong: Yes this is the researcher's name

src/libs/ajax.js
rushtong: This is correct, rationale is on the user. It has a bit of a complicated history, but in essence, it is what the admin enters for reasoning when approving or denying a user as a researcher.

src/libs/dataUseTranslation.js
```diff
@@ -10,6 +104,86 @@ export const DataUseTranslation = {
    * @param darInfo
    * @returns {{primary: [{code: '', description: ''}], secondary: [{code: '', description: ''}]}}
    */
+
+  //NOTE: backend categorization of purposeStatement/researchType differs from front-end primary/secondary designations
+  //Should the two categories be unified or are they distinct for a reason?
```
rushtong: The back end categorization of `purposeStatement/researchType` is out of date which is why we want to move them to the front end and make them consistent with how we're describing primary/secondary designations.

src/libs/dataUseTranslation.js
```diff
@@ -1,5 +1,99 @@
 import * as fp from 'lodash/fp';
 
+const translations = {
```
rushtong, line 3: FYI, love this refactor.

src/libs/dataUseTranslation.js
```diff
@@ -1,5 +1,99 @@
...
+    manualReview: false
+  },
+  poa: {
+    code: 'POA',
+    desciption: 'The dataset will be used for the study of Population Origins/Migration patterns.',
+    manualReview: true
+  },
+  diseases: (diseases) => {
+    const diseaseArray = diseases.sort();
+    const diseaseString = diseaseArray.length > 1 ? fp.join('; ')(diseaseArray) : diseaseArray[0];
```
rushtong, line 16: I don't think you need to check for length here, pretty sure it does the right thing with arrays of any length, even 0.

src/libs/ajax.js
rushtong: Just realized my comment was confusing. The rationale here is attached to the researcher that made the data access request, not the currently logged in user.

rushtong (approved):
Looks good. See one minor comment inline, otherwise 👍

JVThomas:
I recently added commits to correct some errors I made when processing the researcherType statements (The statements under "Types of Research" in ApplicationSummary), so I'd appreciate a re-review of those changes. Thanks!

rushtong (approved):
LGTM 👍

PR #579, opened 2020-09-08
[DUOS-746][risk=no] Fix user profile update.

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-746

## Changes
Consolidate calls to clone the user's profile for update. Library cards and entries are not valid properties for update.


JVThomas (approved):
Looks good, user profiles now save and redirect

PR #599, opened 2020-09-18
DUOS-768 Implement User query via ID [risk=no]

JVThomas:
Addresses [DUOS-768](https://broadinstitute.atlassian.net/browse/DUOS-768)

PR supplies bugfix regarding User querying. There are certain instances where the email tied to a DAR differs from the actual email on the researcher's user profile. The PR aims to utilize the newly implemented User query via id (back-end) for more predictable results (no ambiguity, user is either registered or not registered).

Potential followups to this PR is finding ways to optimize calls that get Researcher Properties. Currently the only difference between the User query and Researcher Properties query is that the User return lacks Library Card info, so further development of the API can help reduce redundant/overlapping calls on the front-end.

In all cases:

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -931,6 +931,9 @@ export const StatFiles = {
 
 export const User = {
 
+  //DEPRECATION NOTE: consider this method deprecated, a user's email can change with their employment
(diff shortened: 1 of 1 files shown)


rushtong (approved):
Looks good. Minor suggestion, we should start adding deprecation notes to APIs that we don't want to be using. The `getByEmail` seems like a good candidate.

PR #608, opened 2020-09-23
[DUOS-724][risk=no] Implement Create Dataset Page

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-724
Screen Shot 2020-09-23 at 2 09 26 PM

Takes the /dataset_registration mock up and hooks it up to createDataset v2, creating only a dataset from a successfully filled out form that is NOT associated with a consent. Removes most DAR logic + fields, the ability for a save (datasets do not have that capability). Adds some validation that the required fields are actually filled out and auto-populates the 1.1 data custodian field.

In all cases:


src/pages/DatasetRegistration.js
```diff
@@ -16,34 +16,25 @@ import * as fp from 'lodash/fp';
...
       file: {
         name: ''
       },
+      properties: {},
       showModal: false,
-      completed: '',
       showDialogSubmit: false,
-      showDialogSave: false,
-      step: 1,
       formData: {
```
lindojong, line 28: I'm wondering if I should further split formData into 2 groups for datasetData and consentData to keep it organized?

src/pages/DatasetRegistration.js
```diff
@@ -119,65 +96,16 @@ class DatasetRegistration extends Component {
...
-      if (this.props.location.props.formData.darCode !== undefined) {
-        formData = this.props.location.props.formData;
-        formData.ontologies = this.getOntologies(formData);
-      } else if (this.props.location.props.formData.datasetIds !== undefined) {
-        // set datasets sent by data set catalog
-        formData.datasets = this.processDataSet(this.props.location.props.formData.datasetIds);
-      }
-    }
+    let formData = Storage.getData('dataset_registration') === null ? this.state.formData : Storage.getData('dataset_registration');
+    Storage.removeData('dataset_registration');
```
lindojong: Not really sure if I need to use `Storage.getData` for this page

src/pages/DatasetRegistration.js
```diff
@@ -217,52 +145,45 @@ class DatasetRegistration extends Component {
...
-    window.scrollTo(0, 0);
   };
 
-  step3 = (e) => {
-    this.setState(prev => {
-      prev.step = 3;
-      return prev;
-    });
-    window.scrollTo(0, 0);
+  validateRequiredFields(formData) {
```
lindojong, line 118: Manually validates the required fields since the html attribute `required` didn't seem to do anything

src/pages/DatasetRegistration.js
```diff
@@ -551,89 +317,118 @@ class DatasetRegistration extends Component {
     this.props.history.goBack();
   };
 
+  createProperties = () => {
```
lindojong, line 274: this can create a partial set of properties if it doesn't receive them all (just like how the backend only actually requires datasetName). However, we have other code that blocks form submission in case the required fields on the form aren't filled out

src/pages/DatasetRegistration.js
```diff
@@ -16,34 +16,25 @@ import * as fp from 'lodash/fp';
 import './DataAccessRequestApplication.css';
 
 
-const noOptionMessage = 'Start typing a Dataset Name, Sample Collection ID, or PI';
+// const noOptionMessage = 'Start typing a Dataset Name, Sample Collection ID, or PI';
```
lindojong: Think this is part of the broken TypeOfResearch component, I'm still troubleshooting atm but that's not necessary for dataset submission

rushtong:
Merge the latest from develop to get past the test-docker failure.

src/pages/DatasetRegistration.js
```diff
@@ -16,34 +16,25 @@ import * as fp from 'lodash/fp';
...
 
 class DatasetRegistration extends Component {
 
   constructor(props) {
     super(props);
-    this.dialogHandlerSave = this.dialogHandlerSave.bind(this);
-    this.setShowDialogSave = this.setShowDialogSave.bind(this);
```
rushtong, line 26: nice cleanup

src/pages/DatasetRegistration.js
rushtong: Agreed, that makes a lot of sense

src/pages/DatasetRegistration.js
rushtong: We use storage to stash content into a cookie in case we go back and need to re-read it for some reason. I don't think we should be using it very much, and in fact, I'd like to remove it if at all possible. I would try and not use it here.

src/pages/DatasetRegistration.js
rushtong: I could be wrong, but I don't think we're submitting a form here which is why required doesn't work. We're really just manipulating state and calling API actions based on that state. 

https://www.w3schools.com/tags/att_input_required.asp

rushtong (commented):
Still reviewing, but this is looking good. Some comments in the meantime.

rushtong (approved):
Looks good, but we have more work to do so let's add some follow-on tickets to make sure we cover the missing pieces: 

* We already know the names of all datasets, so we can pre-validate the user entry value here and prevent an invalid submission
* Work with Jonathan on to determine if we need controlled vocabulary validation on Data Type, Species, and Phenotype/Indication
* Add list of DACs for selection in the DAC question
* Validation on the Repository URL
* Validation on the # of participants, we should not be able to submit a -100 count study.
* Fully implement Section 2 questions once we have consent creation handled on the back end

Aside from the other cleanup tasks you want to do, I think this is merge-able.

PR #630, opened 2020-10-06
[DUOS-789][risk=no] Disable voting on non-open elections

rushtong:
## Addresses
Partially addresses https://broadinstitute.atlassian.net/browse/DUOS-789
See also: https://github.com/DataBiosphere/consent/pull/768

## Changes
This PR:
1. Filters closed data access elections from the chair/member manage DAR pages
2. Disables vote submission on non-open elections, effectively making a read-only view of the voting page
    - This does not provide any navigation to a read only-view
3. Minor language changes to the manage dar console to more accurately reflect the button actions
    - `Log Final Vote` is now `Update Vote` which is what it really does.

## TODO:
In separate PRs:
~1. Update pending case API so it does not return closed elections~ See https://github.com/DataBiosphere/consent/pull/768

----

In all cases:

### DIFF
--- a/src/libs/theme.js
+++ b/src/libs/theme.js
@@ -6,6 +6,7 @@ export const Theme = {
     link: '#216fb4',
     error: '#DB3214',
     success: '#00928A',
+    disabled: '#cccccc',
     background: {
       secondary: 'rgba(0, 96, 159, 0.1)',
       highlighted: 'rgba(193,107,12, 0.1)',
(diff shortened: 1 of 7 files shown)


JVThomas:
If the election is not open, would it be easier to not render the VotePanel as opposed to disabling the panel/hiding panel elements from the Chair/Member tabs?

rushtong:
I chose this path so that we can still show the votes, but show them as disabled. As it currently stands, this PR is a bit of a workaround in that we're preventing the user from submitting invalid input. However, it also serves as a potential read-only view regardless of election status.

rushtong:
Also marking this as draft as there are some new wrinkles around what to do for elections with status of `Final`.

JVThomas (approved):

PR #631, opened 2020-10-06
[DUOS-783][risk=no] Better validation on dataset required fields

lindojong:
Addresses follow up comments on initial PR for DUOS-724 as described here: https://broadinstitute.atlassian.net/browse/DUOS-783

Added validation for dataset name (can't be a duplicate of preexisting name) and # of participants (must be a positive integer). Also replaced DAC text input field with a Select picker. Spoke with Jonathan and confirmed that at this time we still don't need to validate specific urls or vocabulary.

Screen Shot 2020-10-06 at 4 40 18 PM
Screen Shot 2020-10-06 at 4 41 04 PM

In all cases:


src/pages/DatasetRegistration.js
```diff
@@ -105,6 +112,19 @@ class DatasetRegistration extends Component {
     });
   };
 
+  handlePositiveIntegerOnly = (e) => {
+    const field = e.target.name;
+    const value = e.target.value.replace(/[^\d]/,'');
+
+    if (value === '' || parseInt(value) > -1) {
```
rushtong: Codacy suggestion:
```suggestion
    if (value === '' || parseInt(value, 10) > -1) {
```

src/pages/DatasetRegistration.js
```diff
@@ -126,6 +147,24 @@ class DatasetRegistration extends Component {
       this.isValid(formData.description);
   };
 
+  validateDatasetName(name) {
+    let datasets = this.state.allDatasets;
+    let val = fp.filter(function(ds) {
+      let properties = ds.properties;
+      return fp.find(function(p) { return p.propertyName === "Dataset Name" && p.propertyValue === name}, properties);
```
rushtong: Codacy suggestion:
```suggestion
      return fp.find(function(p) { return p.propertyName === "Dataset Name" && p.propertyValue === name; }, properties);
```

src/pages/DatasetRegistration.js
```diff
@@ -427,16 +489,15 @@ class DatasetRegistration extends Component {
...
                           required: true,
                         }),
                         span({
                           className: 'cancel-color required-field-error-span',
-                          isRendered: fp.isEmpty(this.state.datasetData.datasetName) && showValidationMessages,
+                          isRendered: (fp.isEmpty(this.state.datasetData.datasetName) && showValidationMessages)
+                            || (!fp.isEmpty(this.state.datasetData.datasetName) && !this.validateDatasetName(this.state.datasetData.datasetName)),
                         },
-                        ['Required field']),
+                        [this.validateDatasetName(this.state.datasetData.datasetName) ? 'Required field' : 'Dataset Name already in use']),
```
rushtong, line 522: I'm not seeing this error pop up. If I create a dataset, then immediately try to create a new one with the same name, I get a 404 back from consent and a generic error message. We might need a bug fix on the POST endpoint ... have not looked closely at that yet.

Screen Shot 2020-10-07 at 2 48 12 PM

rushtong: Ancillary problem (and unrelated to this PR) with the above screen shot. Both buttons are disabled and the only way to close out of the modal is to click the little "X" icon.

src/pages/DatasetRegistration.js
```diff
@@ -637,18 +700,25 @@ class DatasetRegistration extends Component {
...
+                          isClearable: true,
+                          isMulti: false,
+                          isSearchable: true,
+                          options: this.dacOptions(),
+                          placeholder: 'Select a DAC...',
                           className: (fp.isEmpty(this.state.datasetData.dac) && showValidationMessages) ?
-                            'form-control required-field-error' :
-                            'form-control',
+                            'required-field-error' :
+                            '',
```
rushtong, line 741: The required field error label doesn't clear state after being invalid, then selecting a valid one.

Screen Shot 2020-10-07 at 2 52 08 PM

rushtong (changes requested):
Looking good! Still some functional issues here though. See comments inline.

src/pages/DatasetRegistration.js
rushtong: More info on this. On the back end, we throw a 404 in this case. I think that should really be a 409 Conflict instead. We should add a PR for that separate from this story.

lindojong: Made a ticket for the error messaging here: https://broadinstitute.atlassian.net/browse/DUOS-797

lindojong: Ticket for modal: https://broadinstitute.atlassian.net/browse/DUOS-800

rushtong (approved):
Thank you for the changes! 👍

PR #684, opened 2020-10-28
DUOS-681 [risk="low"] Add question 2.7 to DAR Application

JVThomas:
Addresses [DUOS-681](https://broadinstitute.atlassian.net/browse/DUOS-681)

PR adds dynamic DUL questions to the end of Step 2 of the DAR Application. As such, additional keys are required to process the formData on the backend.

1) gsoAcknowledgement: Boolean
2) pubAcknowledgement: Boolean
3) dsAcknowledgement: Boolean
4) collaborationDocument: File
5) irbDocument: File
6) activeDULDocument: Object (used for front-end processing, retaining previous inputs, possibly queries)
a) ethicsApprovalRequired: Boolean
b) collaboratorRequired: Boolean
c) publicationResults: Boolean
d) diseaseRestrictions: Boolean
e) geneticStudiesOnly: Boolean

In all cases:


src/pages/DataAccessRequestApplication.js
```diff
@@ -81,7 +84,11 @@ class DataAccessRequestApplication extends Component {
         cloudUse: '',
         cloudProvider: '',
         cloudProviderType: '',
-        cloudProviderDescription: ''
+        cloudProviderDescription: '',
+        gsoAcknowledgement: false,
+        pubAcknowledgement: false,
+        dsAcknowledgement: false,
+        activeDULQuestions: {}
```
rushtong, line 94: Can you add a story for adding these to the back end? I'd love it if you could take that on too. Here are a couple PRs that do very similar things for additional string/boolean fields as well as an object based on for collaborators:
* https://github.com/DataBiosphere/consent/pull/704
* https://github.com/DataBiosphere/consent/pull/778

src/pages/dar_application/DataAccessRequest.js
```diff
@@ -1,22 +1,204 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { a, br, div, fieldset, h, h3, input, label, span, textarea } from 'react-hyperscript-helpers';
 import isNil from 'lodash/fp/isNil';
 import isEmpty from 'lodash/fp/isEmpty';
+import forEach from 'lodash/fp/forEach';
+import includes from 'lodash/fp/includes';
+import cloneDeep from 'lodash/fp/cloneDeep';
+import isEqual from 'lodash/fp/isEqual';
+import every from 'lodash/fp/every';
```
rushtong, line 9: Side note, I've come to the realization that this is a better pattern than * imports, similar to what we do on the java side.

src/pages/dar_application/DataAccessRequest.js
```diff
@@ -1,22 +1,204 @@
...
+        verticalAlign: 'middle'
+      }
+    }, [
+      input({
+        id,
+        type: 'file',
+        style: uploadFileInput,
+        ref: fileRef,
+        onChange: (e) => updateFile(formAttribute, e.target.files[0])
+        // onChange: (e) => formFieldChange({name: formAttribute, value: e.target.files[0]})
```
rushtong: nit: rm

src/pages/dar_application/DataAccessRequest.js
```diff
@@ -1,22 +1,204 @@
...
+      })
+    ])
+  );
+};
 
 export default function DataAccessRequest(props) {
   const {
     darCode,
-    datasets,
+    // datasets,
```
rushtong: nit: rm

src/pages/dar_application/DataAccessRequest.js
```diff
@@ -339,9 +594,132 @@ export default function DataAccessRequest(props) {
...
+          style: uploadFileDiv(showValidationMessages, irbDocument)
+        }, [
+          div({className: 'row no-margin'}, [
+            div({className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12'}, [
+              span({className: 'rp-choice-questions', style: uploadFileDescription}, [
+                //NOTE: ask for question to be rephrased, grammar seems odd and I don't know how the conditions tie to each other
+                //second statement can either be a condition for the first or last statements (or maybe both?)
+                `One or more of the datasets you selected requires local IRB approval for use. Please upload your local IRB apporval(s) here as a single document. 
+                When IRB approval is required and Expedited or Full Review is required and must be completed annually.
+                Determinations of Not Human Subjects Research (NHSR) by IRBs will not be accepted as IRB approval.`
```
rushtong, line 563: @solideoglori Can you comment here?

rushtong (approved):
Looks good 👍 

### Notes:
We are already processing these fields on the back end:
* String irbDocumentLocation;
* String irbDocumentName;
* String collaborationLetterLocation;
* String collaborationLetterName;

We don't want to store real file objects, just string values of names and locations. These should be accessed individually via the file upload APIs which will return the full DAR with these fields populated correctly and file content stored safely.

src/components/UploadLabelButton.js
```diff
@@ -0,0 +1,209 @@
...
+          if(!isNil(targetFile)) {
+            setTargetDownload(URL.createObjectURL(targetFile));
+          }
+        }catch(error) {
+          Notifications.showError({text: 'Error: Failed to retreive saved file'});
+        }
+      }
+    };
+    downloadFile();
+  }, [currentFileLocation, currentFileName, formAttribute, newDULFile, referenceId]);
```
rushtong, line 102: I still struggle with this structure, but from what I've read on `useEffect`, this is the accepted syntax. This: `anonymous function () => { blah }, [args available to function]` is super non-intuitive to me. It would be much cleaner if it followed the standard function syntax, `(args) => { blah }`.

src/libs/ajax.js
```diff
@@ -372,6 +398,17 @@ export const DAR = {
     return await res.json();
   },
 
+  getIRBDocument: async (referenceId, fileType) => {
```
rushtong: naming nit. This could also be used for Collaboration document. Maybe something more generic like `getDARDocument`?

src/libs/ajax.js
```diff
@@ -347,6 +363,16 @@ export const DAR = {
     return await res.json();
   },
 
+  //v2 endpoint for DAR POST
+  postDar: async (dar) => {
+    const filteredDar = fp.omit(['createDate', 'sortDate', 'data_access_request_id'])(dar);
+    const url = `${await Config.getApiUrl()}/dar/v2`;
+    // const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(filteredDar), { method: 'POST' }]));
```
rushtong: nit: rm

src/libs/ajax.js
```diff
@@ -426,6 +463,19 @@ export const DAR = {
     return manualReview;
   },
 
+  //NOTE: endpoints requires a dar id
+  uploadDULDocument: async(file, darId, fileType) => {
```
rushtong: naming nit: similar to above, this is really focused on DARs, not DULs.

src/libs/utils.js
```diff
@@ -16,6 +17,13 @@ export const formatDate = (dateval) => {
   return datestr;
 };
 
+//Custom empty check needed on File
+//lodash's isEmpty checks for enumerated keys, something a File does not have (ends up being an empty array)
+//leads to incorrect evaluation of File
+export const isFileEmpty = (file) => {
+  return isNil(file) || file.size < 1 || file.length < 1;
+};
```
rushtong, line 25: Nice

src/pages/DataAccessRequestApplication.js
```diff
@@ -115,6 +131,29 @@ class DataAccessRequestApplication extends Component {
     }, () => this.checkValidations());
   };
 
+  changeDULDocument = (dataset) => {
```
rushtong: naming nit: same as above.

src/pages/DataAccessRequestApplication.js
```diff
@@ -481,43 +556,87 @@ class DataAccessRequestApplication extends Component {
     this.setState({ showDialogSave: true });
   };
 
-  dialogHandlerSubmit = (answer) => (e) => {
+  //Can't do uploads in parallel since endpoints are post and they both alter attributes in json column
+  //If done in parallel, updated attribute of one document will be overwritten by the outdated value on the other
```
rushtong, line 563: I had to reckon with this as well with DAC membership updates. We have a utility function, `utils/PromiseSerial` that will force a list of async functions to execute serially. Probably overkill for what you're doing here, but might be helpful in the future.

src/pages/DataAccessRequestApplication.js
```diff
@@ -481,43 +556,87 @@ class DataAccessRequestApplication extends Component {
...
+  saveDULDocuments = async(uploadedIrbDocument = null, uploadedCollaborationLetter = null, referenceId) => {
```
rushtong: Similar naming nit. We use `DUL` to refer to data use letter documents which are usually a hard-copy (pdf, word, etc.) of the data use restrictions on the dataset. These are provided by the study's PI or PI's institution directly to our Office of Research Subject Protection.

src/pages/DataAccessRequestApplication.js
```diff
@@ -428,12 +467,48 @@ class DataAccessRequestApplication extends Component {
...
+      const dulQuestionMap = {
+        'geneticStudiesOnly': 'gsoAcknowledgement',
+        'publicationResults': 'pubAcknowledgement',
+        'diseaseRestrictions': 'dsAcknowledgement',
+        'ethicsApprovalRequired': 'irbDocument',
+        'collaboratorRequired': 'collaborationLetter'
+      };
+      let result = false;
+
+      //NOTE: verify validation runs correctly here
```
rushtong: This could use some extra documentation. I'm not following all of the validation rules here.

src/pages/DataAccessRequestApplication.js
```diff
@@ -481,43 +556,87 @@ class DataAccessRequestApplication extends Component {
...
+      }
+      formattedFormData.datasetIds = fp.map('value')(formattedFormData.datasets);
+      formattedFormData.userId = userId;
+
+      try {
+        //NOTE: the pre-processing saves are adding time to record generation (makes front-end seem slow)
+        //pre-prcessing saves are needed since you can't save documents without a reference id
+        //saves ensure record has a reference id
+        //actual fix would involve generating a blank draft record that is saved on console button click
+        //however that would fall outside the scope of this pr, which is already large enough due to refactored code
```
rushtong, line 608: Great point. Let's file a story for that and flesh out how we'd tackle it technically.

src/pages/DataAccessRequestApplication.js
```diff
@@ -481,43 +556,87 @@ class DataAccessRequestApplication extends Component {
...
-        formData.userId = Storage.getCurrentUser().dacUserId;
-        this.setState(prev => {
-          prev.disableOkBtn = true;
-          return prev;
+      }
+      formattedFormData.datasetIds = fp.map('value')(formattedFormData.datasets);
+      formattedFormData.userId = userId;
+
+      try {
+        //NOTE: the pre-processing saves are adding time to record generation (makes front-end seem slow)
```
rushtong, line 604: We should look at improving the endpoint performance if this is affecting the UI. It really shouldn't, we don't do a lot of work when saving a draft DAR. Will file a story for that.

src/pages/DataAccessRequestApplication.js
```diff
@@ -545,38 +664,81 @@ class DataAccessRequestApplication extends Component {
...
+      }, () => this.saveDarDraft());
     } else {
       this.setShowDialogSave(false);
     }
   };
 
-  savePartial = () => {
-    let formData = this.state.formData;
+
+  //NOTE: work on this, expect similar issue with submit?
```
rushtong: nit: Can you expand on this? What more work do we need to do?

rushtong (commented):
Have not manually tested yet, will do that today. Code generally looks good, just a few naming nits, see comments inline.

src/libs/ajax.js
JVThomas: Yeah, its already being used for both, I have no idea why I named it just for IRB

src/pages/DataAccessRequestApplication.js
JVThomas: It was more of a note for me during development, but a small explanation for this block is still a good idea

src/pages/DataAccessRequestApplication.js
JVThomas: Another old comment, will remove

src/components/UploadLabelButton.js
JVThomas: The arguments are essentially telling the hook what variables to watch for updates. This way, instead of the hook firing off on every re-render (which can happen on any prop variable or state variable update), it'll only fire off on the values that it cares about. 

You can have an empty array as an argument if you want the hook to run only on mount, but if the process and/or result of the hook is dependent on the arguments then there's no reason to leave it empty at all.

rushtong:
Probably not for this PR, but the disease selection is not being displayed correctly. This is after selecting sleep apnea, then reloading, the selection in the UI is not ~populating~ saving correctly.

UPDATE: It is the post. It looks like the call to PUT does not contain the ontologies or the diseases list. So, when it comes back from the server, it doesn't have the right information populated.

UPDATE: Note that the back end is expecting a particular format. See https://github.com/DataBiosphere/consent/blob/34ee6caad7d66466314287491e6fb2f13cb524b7/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java#L816. We're expecting:
```
ontologies: [{"id": "blah", "label": "blah", definition: "blah", "synonyms": ["blah", "blah"]}]
```

Screen Shot 2020-11-10 at 7 16 46 AM

rushtong (approved):
Functionality works as expected, aside from ontologies not being saved (we can work on this separately, but should be addressed soon). Approving pending feedback from @solideoglori

src/pages/DataAccessRequestApplication.js
rushtong: Created https://broadinstitute.atlassian.net/browse/DUOS-835 to look at this.

JVThomas:
> Probably not for this PR, but the disease selection is not being displayed correctly. This is after selecting sleep apnea, then reloading, the selection in the UI is not ~populating~ saving correctly.
> 
> UPDATE: It is the post. It looks like the call to PUT does not contain the ontologies or the diseases list. So, when it comes back from the server, it doesn't have the right information populated.
> 
> UPDATE: Note that the back end is expecting a particular format. See https://github.com/DataBiosphere/consent/blob/34ee6caad7d66466314287491e6fb2f13cb524b7/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java#L816. We're expecting:
> 
> ```
> ontologies: [{"id": "blah", "label": "blah", definition: "blah", "synonyms": ["blah", "blah"]}]
> ```
> 
> Screen Shot 2020-11-10 at 7 16 46 AM

So I've fixed the issue of it not being in the request payload by swapping out Object.assign with lodash's assign method. However the saving issue is still there. For this test I simply added 'asthma' as a disease (the only disease chosen) and saved the draft.
Screen Shot 2020-11-10 at 10 15 04 AM
Screen Shot 2020-11-10 at 10 13 28 AM

rushtong:
@JVThomas There are 3 required fields on each ontology entry: {"id": "blah", "label": "blah", definition: "blah"}

JVThomas:
@rushtong Then that means dev, staging, and production can't save the diseases either. The ontology processing was left untouched, and that only sent id and label. I'll make a separate PR for this issue for immediate review and incorporate it into this PR later.

Screenshot of develop branch version of the ontology processing

Screen Shot 2020-11-10 at 10 43 36 AM

rushtong:
@JVThomas 
> Then that means dev, staging, and production can't save the diseases either.

I believe that to be true, but haven't tested. Thanks for jumping on this one, it's a fairly important feature.

JVThomas:
Added additional commits to incorporate ontology picker bug as well as adding rendering logic for question 2.7 on submitted dar views

rushtong (approved):
Looks good! 👍

PR #701, opened 2020-11-12
[DUOS-493][risk=no] Config Map Updates

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-493
Requires 
* https://github.com/DataBiosphere/duos-ui/pull/715
* https://github.com/broadinstitute/duos-deploy/pull/24

## Changes
* Use a blank config when building image
* Consolidate docker images
* Update readme to reflect current tooling
* Update deployment yaml to use configmap

##  Notes
We don't have an automated process for updating configmaps on the clusters so these are currently done manually. They do not change frequently so this is a manageable process. In the near future, we will be deploying using helm charts and that process will include a configmap update to the target cluster. This PR is intended to prep the repo for the helm chart conversion.

~Ensure that a configmap is deployed to each cluster in the correct namespace before merging.~ Configmaps are live on dev, staging, and prod clusters.

----

In all cases:


Dockerfile
```diff
@@ -1,6 +1,25 @@
...
+WORKDIR /usr/src/app
+
+# add `/usr/src/app/node_modules/.bin` to $PATH
+ENV PATH /usr/src/app/node_modules/.bin:$PATH
+
+# install and cache app dependencies
+COPY src /usr/src/app/src
+COPY public /usr/src/app/public
+COPY package.json /usr/src/app/package.json
+COPY config/base_config.json /usr/src/app/public/config.json
```
JVThomas, line 16: Asking for a reminder, what determines the ENV config source for Docker?

rushtong: In our current deployment project, this is what we would do:
1. In this code, we're replacing what we used to use for the env specific config with a blank one in the docker image that we store. 
2. Now, look at the k8s.yml file and you can see that we're mounting a config json file from a configmap that exists on the cluster and effectively replacing `base_config.json` in the docker image with a real version from the cluster. 
3. If you go to each cluster (we have one for dev, one for staging, and one for prod) you will be able to see what's in that configmap. I created env specific configmap entries manually.

That sets us up for moving to the Terra clusters where:
1. When we eventually move to a helm/helmfile process, I won't need to manually update those values on each cluster, that will happen via the helm chart deployment process. See https://github.com/broadinstitute/terra-helm/pull/197 for that chart. Look at the [configmap template file](https://github.com/broadinstitute/terra-helm/pull/197/files#diff-a04dd5e9be58e94835694c7b69ee256d72b08c9a9b2d5651101909f2ff670868) and you'll see that each value is replaced by a value that comes from an environment specific values file. 
2. That values file is defined in this PR: https://github.com/broadinstitute/terra-helmfile/pull/561

JVThomas (approved):

PR #703, opened 2020-11-16
DUOS-718 [risk=no] Bugfix for disease picker

JVThomas:
Addresses [DUOS-718](https://broadinstitute.atlassian.net/browse/DUOS-718)

PR provides fix to ontology picker on DAR by setting the key, id, and item values on each ontology item within the TypeOfResearch component on prop update to ensure that the React-Select component (AsyncSelect) will not break. It also adds the definition (now being checked on the back-end) as well as change the labelling so that it displays the name of the disease rather than the ontology definition link. Branch may end up with merge conflicts as a result of DUOS-681

There may be one issue with regards to previously saved ontologies in that their labels will show the ontology links since...
1) they lack a saved reference to item
2) their label value is set to equal the ontology link.

Otherwise these elements are still saved and can be removed as usual.

In all cases:


rushtong (approved):
Really nice cleanup 👍

PR #735, opened 2020-12-07
[DUOS-266][risk=no] Added page component for BackgroundSignIn to be used for testing

rfricke-asymmetrik:
https://broadinstitute.atlassian.net/browse/DUOS-266 

Added a react page component to be used to perform the login to the DUOS web app using a google cloud access token.

The component accepts an access token either via property or query string provided as ?token= to the /backgroundsignin route and will use this token to call the user/me api and redirect to the home page of DUOS.

In all cases:

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -51,7 +51,14 @@ const Routes = (props) => (
     <Route exact path="/" render={(routeProps) => <Home {...routeProps} {...props} />} />
     <Route exact path="/home" render={(routeProps) => <Home {...routeProps} {...props} />} />
     <Route exact path="/status" render={(routeProps) => Status(_.mergeAll([routeProps, props]))} />
-    <Route exact path="/backgroundsignin" component={BackgroundSignIn} />
+    <Route exact path="/backgroundsignin" render={
+      (routeProps) =>
+        props.env
+          ? props.env !== 'prod'
+            ? <BackgroundSignIn {...routeProps} />
+            : <NotFound />
+          : <div />
+    } />
     <Route path="/FAQs" component={FAQs} />
     <Route path="/home_about" component={HomeAbout} />
     <Route path="/home_signing_official" component={HomeSigningOfficial} />
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -43,6 +43,7 @@ import NHGRIpilotinfo from './pages/NHGRIpilotinfo';
 import { Status } from './pages/Status';
 import { SummaryVotes } from './pages/SummaryVotes';
 import HomeResearcherInfo from "./pages/HomeResearcherInfo";
+import BackgroundSignIn from './pages/BackgroundSignIn';
 
 
 const Routes = (props) => (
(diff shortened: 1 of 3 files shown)


src/Routes.js
```diff
@@ -43,13 +43,15 @@ import NHGRIpilotinfo from './pages/NHGRIpilotinfo';
...
 import HomeResearcherInfo from "./pages/HomeResearcherInfo";
+import BackgroundSignIn from './pages/BackgroundSignIn';
 
 
 const Routes = (props) => (
   <Switch>
     <Route exact path="/" render={(routeProps) => <Home {...routeProps} {...props} />} />
     <Route exact path="/home" render={(routeProps) => <Home {...routeProps} {...props} />} />
     <Route exact path="/status" render={(routeProps) => Status(_.mergeAll([routeProps, props]))} />
+    <Route exact path="/backgroundsignin" component={BackgroundSignIn} />
```
JVThomas: This is more of a comment for discussion, feel free to reply back @rfricke-asymmetrik @rushtong. I don't have any issues with the test exclusive component, but I'd feel a lot better if the Route was only rendered in select environments (say if env="test" or env="local"). 

This way you could pass in the env variable as a prop into Routes and have a line like ```{props.env === 'test' && <Route exact path="/backgroundsignin" component={BackgroundSignIn} />}``` to ensure that the Route only shows on a test env.

In terms of getting the env variable, we have a file called Config that is already being imported within App.js that contains a function called getEnv. You could use that function as part of the Application initialization process and pass it down into props from there.

src/pages/BackgroundSignIn.js
```diff
@@ -0,0 +1,141 @@
...
+              default:
+                setInvalidToken(true);
+                break;
+            }
+          });
+      }
+    };
+
+    if (accessToken)
+      performLogin();
```
JVThomas, line 99: You also have an if statement for accessToken within performLogin itself, so you can remove the if block either there or from the highlighted context above. I lean towards removing it from the function since it simplifies it's logic (albeit slightly), but it's a matter of preference.

src/Routes.js
rushtong: I think I'd prefer that we have this option available on any env **_except_** prod. In the (hopefully near) future, we'll have on-demand k8s clusters that might point to a perf, alpha, or staging env and I'd like to be able to have comparison runs across different envs with different builds.

src/pages/BackgroundSignIn.js
rfricke-asymmetrik: That is a good call. I think removing it from the function makes sense.

src/Routes.js
rfricke-asymmetrik: Agreed. This should, at the very least, not be visible on prod.

rushtong (commented):
Looks good, pending suggested changes 👍🏽 Please ping me for a re-review when that happens.

src/Routes.js
```diff
@@ -51,7 +51,14 @@ const Routes = (props) => (
...
     <Route exact path="/status" render={(routeProps) => Status(_.mergeAll([routeProps, props]))} />
-    <Route exact path="/backgroundsignin" component={BackgroundSignIn} />
+    <Route exact path="/backgroundsignin" render={
+      (routeProps) =>
+        props.env
+          ? props.env !== 'prod'
+            ? <BackgroundSignIn {...routeProps} />
+            : <NotFound />
+          : <div />
+    } />
```
rushtong, line 61: Nice!

rushtong (approved):
Works like a charm and looks great too 👍🏽

PR #748, opened 2020-12-10
NIH IC Webform (public) creation

solideoglori:
**ADDRESSES**
https://broadinstitute.atlassian.net/browse/DUOS-884

In all cases:


solideoglori:
updating per Codacy recommendations

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2795 @@
...
+          return prev;
+        });
+
+        if (this.state.showValidationMessages) {
+          this.setState({showDialogSubmit: false});
+        }
+        else {
+          let ds = this.formatFormData(formData);
+          if (fp.isEmpty(this.state.updateDataset)) {
+            DataSet.postDatasetForm(ds).then(resp => {
```
rushtong: We need to remove any functionality that does modification. This whole if block should go away.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2795 @@
...
+                prev.problemSavingRequest = true;
+                prev.submissionSuccess = false;
+                prev.errorMessage = errorMessage;
+                return prev;
+              });
+            });
+          }
+          else {
+            const { datasetId } = this.props.match.params;
+            DataSet.updateDataset(datasetId, ds).then(resp => {
```
rushtong: here too, this should be removed.

rushtong (changes requested):
The changes recommended here will prevent any attempts at updating content on the back end. This version is fine as it stands, but there is a tremendous amount of duplicated code. We need to add a story in the backlog to update the real version of the form so the html content can be shared with a demo page like this one. That way, both pages can render the same component, one with real functionality and one with no functionality.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2795 @@
...
+    }
+  };
+
+
+  async init() {
+    const { datasetId } = this.props.match.params;
+    let updateDataset = {};
+    // update dataset case
+    if (!fp.isNil(datasetId)) {
+      DataSet.getDataSetsByDatasetId(datasetId).then(data => {
```
rushtong: If you're not logged in (and this is an open demo page) then this call will fail

src/Routes.js
```diff
@@ -57,6 +58,7 @@ const Routes = (props) => (
     <Route path="/home_researcher_info" component={HomeResearcherInfo} />
     <Route path="/election404" component={Election404} />
     <Route path="/NHGRIpilotinfo" component={NHGRIpilotinfo} />
+    <Route path="/NIHICWebform" component={NIHICWebform} />
```
rushtong: It's preferable to keep urls all lower case. It's less of a concern in our world as we have everything deployed to linux based hosts, but casing mistakes are easy to make especially in one-word-paths.
```suggestion
    <Route path="/nih_ic_webform" component={NIHICWebform} />
```

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+  };
+
+  handleChange = (e) => {
+    const field = e.target.name;
+    const value = e.target.value;
+    this.setState(prev => {
+      prev.datasetData[field] = value;
+      prev.disableOkBtn = false;
+      prev.problemSavingRequest = false;
+      prev.submissionSuccess = false;
```
rushtong: Since this is a demo page, we should remove all traces of trying to save anything.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+  };
+
+  handleCheckboxChange = (e) => {
+    const field = e.target.name;
+    const value = e.target.checked;
+    this.setState(prev => {
+      prev.formData[field] = value;
+      prev.disableOkBtn = false;
+      prev.problemSavingRequest = false;
+      prev.submissionSuccess = false;
```
rushtong: Same thing here

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+    if (fp.isEmpty(oldDataset)) {
+      return !fp.contains(name, datasets);
+    }
+    // update dataset case (include old dataset name as valid)
+    else {
+      let updateDatasetName = fp.find(p => p.propertyName === "Dataset Name", this.state.updateDataset.properties).propertyValue;
+
+      return (name === updateDatasetName || !fp.contains(name, datasets));
+    }
+  };
```
rushtong: This method should probably be removed ... we don't need to do any dataset validation, do we?

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+      // if the old dataset name has been edited
+      else {
+        return this.validateDatasetName(name) ? 'form-control' : 'form-control required-field-error';
+      }
+    }
+    // if a new dataset name is being edited
+    else {
+      return this.validateDatasetName(name) ? 'form-control' : 'form-control required-field-error';
+    }
+  };
```
rushtong: Same thing here.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+    this.setState(prev => {
+      if (fp.isNil(option)) {
+        prev.selectedDac = {};
+        prev.datasetData['dac'] = '';
+      } else {
+        prev.selectedDac = option.item;
+        prev.datasetData['dac'] = option.label;
+        prev.disableOkBtn = false;
+        prev.problemSavingRequest = false;
+        prev.problemLoadingUpdateDataset = false;
```
rushtong: These should likely be removed

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+      ethics = false,
+      geographic = false,
+      moratorium = false,
+      methods = false,
+      generalUse = false,
+    } = this.state.formData;
+    const ontologies = this.formatOntologyItems(this.state.formData.ontologies);
+    const { publicAccess = false } = this.state.datasetData;
+    const { npoa } = !poa;
+    const { problemSavingRequest, problemLoadingUpdateDataset, showValidationMessages, submissionSuccess } = this.state;
```
rushtong: All of these constants suggest data uploads ... we should remove them

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+    const isTypeOfResearchInvalid = false;
+    const isUpdateDataset = (!fp.isEmpty(this.state.updateDataset));
```
rushtong: Same here.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2502 @@
...
+    const profileUnsubmitted = span({}, [
+      'Please make sure ',
+      h(Link, { to: '/profile', className: 'hover-color' }, ['Your Profile']),
+      ' is updated, as it will be linked to your dataset for future correspondence'
+    ]);
```
rushtong: Since this is a demo page, there won't be a logged in user. This logic and usage should be removed.

rushtong:
We should get you into the habit of adding a ticket to the backlog for these changes and adding that to the PR title and description. They are non-trivial and we should be tracking them alongside other functional changes.

solideoglori:
Definitely, will add tickets for this and the other work.

solideoglori:
@rushtong I tried cleaning up the code per your comments and was able to resolve a few, but some I couldn't seem to remove and still have it render locally on the rest of them. Thoughts/suggestions on what I might be doing wrong?

rushtong:
Will dig in this afternoon!

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2704 @@
...
+                          className: this.showDatasetNameErrors(this.state.datasetData.datasetName, showValidationMessages),
+                          required: true,
+                        }),
+                        span({
+                          className: 'cancel-color required-field-error-span',
+                          isRendered: fp.includes('required-field-error', this.showDatasetNameErrors(this.state.datasetData.datasetName, showValidationMessages))
+                        },
+                        [this.validateDatasetName(this.state.datasetData.datasetName) ? 'Required field' : 'Dataset Name already in use']),
+                      ])
+                  ]),
```
rushtong: Here's an example of something that is confusing because of copy/paste. We have a form field asking for the PI's name.
Then, as input for that, we are calling it `datasetName`. That input has an error handler attached to it, which we don't need because were not validating a dataset name. I'll file a PR to this PR that does as much cleanup as I can.

solideoglori: Thanks @rushtong!

solideoglori:
Hi @rushtong! I merged your suggested changes into this branch and resolved a routes conflict. I _think_ this is ready to go, but GH is saying there's a change I haven't addressed. I'm not sure if this is due to not closing the comments above or if it's getting that from the code itself. Could you advise?

rushtong:
@solideoglori FYI, see https://github.com/DataBiosphere/duos-ui/pull/759 for another attempt at cleaning up unused code and clarifying form elements.

src/Routes.js
rushtong: Can you make this change?

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+                          'NIH Grant or Contract Number ',
+                        ]),
+                      ]),
+                    div(
+                      {className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group rp-last-group'},
+                      [
+                        input({
+                          type: 'text',
+                          name: 'description',
+                          id: 'inputDescription',
```
rushtong: name and id should reflect the form field that they are identifying.
ID also needs to be unique on the page.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+                          maxLength: '256',
+                          value: this.state.datasetData.description,
```
rushtong: I think this should be
```suggestion
                          value: '',
```

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+                          onChange: () => {},
+                          className: (fp.isEmpty(this.state.datasetData.description) && showValidationMessages) ?
+                            'form-control required-field-error' :
+                            'form-control',
```
rushtong: This should just be: 
```suggestion
                          className: 'form-control',
```

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+                          required: true,
+                        }),
+                        span({
+                          className: 'cancel-color required-field-error-span',
+                          isRendered: fp.isEmpty(this.state.datasetData.description) && showValidationMessages,
+                        },
+                        ['Required field']),
```
rushtong: We aren't doing any validation on these fields so this span is unnecessary. These should be removed.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+class NIHICWebform extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      dacList: [],
+      selectedDac: {},
+      nihValid: false,
+      showValidationMessages: false,
+      formData: {
```
rushtong: Once you remove all references to `formData`, this state object can go away

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+        forProfit: '',
+        hmb: false,
+        poa: false,
+        diseases: false,
+        ontologies: [],
+        other: false,
+        otherText: '',
+        generalUse: false
+      },
+      datasetData: {
```
rushtong: Once you remove all references to `datasetData`, this state variable can go away.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+                      RadioButton({
+                        style: {
+                          margin: '2rem',
+                          color: ' #010101',
+                        },
+                        id: 'checkPublicAccess_yes',
+                        name: 'checkPublicAccess',
+                        value: 'yes',
+                        defaultChecked: publicAccess,
+                        onClick: () => this.setPublicAccess(true),
```
rushtong: The original code where this came from was for updating a dataset. This radio button really doesn't have anything to do with that. There probably shouldn't be any handler on this, I think it might work with just
```suggestion
                        onClick: () => {},
```

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+            value: item.id,
+            label: item.label,
+            item: item,
+          };
+        });
+        callback(options);
+      });
+  };
+
+  setPublicAccess = (value) => {
```
rushtong: All references to this method should go away - it was dataset specific.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+      publication = false,
+      collaboration = false,
+      ethics = false,
+      geographic = false,
+      moratorium = false,
+      methods = false,
+      generalUse = false,
+    } = this.state.formData;
+    const ontologies = this.formatOntologyItems(this.state.formData.ontologies);
+    const publicAccess = false;
```
rushtong: Once other refactoring happens, this should also go away.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+    const { npoa } = !poa;
+    const { showValidationMessages } = this.state;
```
rushtong: I don't think we're using this any more ... it should be removed from here and state.

src/pages/NIHicWebform.js
```diff
@@ -0,0 +1,2266 @@
...
+    // NOTE: set this to always false for now to submit dataset without consent info
```
rushtong: unnecessary comment.

rushtong (changes requested):
See comments inline ... these are the kinds of things that need to happen throughout the file. I didn't point out every case, but there are quite a few.

solideoglori:
@rushtong I think I closed out everything with this one except for the formData, which I couldn't seem to tease out and still have it render locally.

rushtong (approved):
👍🏽

PR #758, opened 2020-12-15
[DUOS-890][risk=no] Can't edit Data Custodian/PI fields on dataset registration form

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-890
Added missing onChange handler to the affected fields so that they are editable.

In all cases:

### DIFF
--- a/src/pages/DatasetRegistration.js
+++ b/src/pages/DatasetRegistration.js
@@ -769,6 +769,7 @@ class DatasetRegistration extends Component {
                         name: 'researcher',
                         id: 'inputResearcher',
                         value: this.state.datasetData.researcher,
+                        onChange: this.handleChange,
                         disabled: !isUpdateDataset,
                         className: (fp.isEmpty(this.state.datasetData.researcher) && showValidationMessages) ? 'form-control required-field-error' : 'form-control',
                         required: true
@@ -790,6 +791,7 @@ class DatasetRegistration extends Component {
                         name: 'principalInvestigator',
                         id: 'inputPrincipalInvestigator',
                         value: this.state.datasetData.principalInvestigator,
+                        onChange: this.handleChange,
                         className: (fp.isEmpty(this.state.datasetData.principalInvestigator) && showValidationMessages) ? 'form-control required-field-error' : 'form-control',
                         required: true
                       }),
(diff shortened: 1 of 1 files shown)


src/pages/DatasetRegistration.js
```diff
@@ -769,6 +769,7 @@ class DatasetRegistration extends Component {
                         name: 'researcher',
                         id: 'inputResearcher',
                         value: this.state.datasetData.researcher,
+                        onChange: this.handleChange,
```
JVThomas: If you're using text inputs and they're not tied to a real-time update feature it's better to use `onBlur`. This way instead of firing the `handleChange` function on every keystroke you'll fire it once the text input is out of focus.

JVThomas (changes requested):

src/pages/DatasetRegistration.js
rushtong: Great point ... I would have missed this.

rushtong (approved):
Pending Justin's feedback, 👍🏽

src/pages/DatasetRegistration.js
```diff
@@ -1143,7 +1143,7 @@ class DatasetRegistration extends Component {
                                 isDisabled: isUpdateDataset || !diseases,
                                 isMulti: true,
                                 loadOptions: (query, callback) => this.searchOntologies(query, callback),
-                                onChange: (option) => this.onOntologiesChange(option),
+                                onBlur: (option) => this.onOntologiesChange(option),
```
JVThomas: This is a real-time update so this **should** be onChange, not onBlur. With each keystroke you provide a new string to filter out the preset list of ontologies obtained via API (should've provided an example earlier, my mistake). The code also breaks if you select a disease with onBlur, which means the AsyncSelect component probably does not create an initial list, so you'll end up trying to filter out an undefined array (which leads to errors).

JVThomas (changes requested):

src/pages/DatasetRegistration.js
lindojong: Ah whoops, nice catch!

JVThomas (approved):

PR #788, opened 2021-01-08
DUOS-912[risk=no]DAR Application Question 2.3.3 - Hide text box unless "Other" button is selected

raejohanek:
## Address
https://broadworkbench.atlassian.net/browse/DUOS-912?

## Scope of Changes
Check to make sure props.other is true before rendering the textbox
Set the textbox's enabled value to always be true

In all cases:

### DIFF
--- a/src/pages/dar_application/TypeOfResearch.js
+++ b/src/pages/dar_application/TypeOfResearch.js
@@ -170,19 +170,19 @@ export const TypeOfResearch = hh(class TypeOfResearch extends Component {
             disabled: props.disabled,
           }),
 
-          props.other ?
-            textarea({
-              style: otherTextStyle,
-              value: props.otherText,
-              onChange: props.otherTextHandler,
-              name: 'otherText',
-              id: 'otherText',
-              maxLength: '512',
-              rows: '2',
-              required: props.other,
-              placeholder: 'Please specify if selected (max. 512 characters)',
-              enabled: true
-            }) : undefined
+          textarea({
+            isRendered: props.other,
+            style: otherTextStyle,
+            value: props.otherText,
+            onChange: props.otherTextHandler,
+            name: 'otherText',
+            id: 'otherText',
+            maxLength: '512',
+            rows: '2',
+            required: props.other,
+            placeholder: 'Please specify if selected (max. 512 characters)',
+            enabled: true
+          })
         ])
     );
   }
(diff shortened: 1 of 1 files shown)


src/pages/dar_application/TypeOfResearch.js
```diff
@@ -170,19 +170,19 @@ export const TypeOfResearch = hh(class TypeOfResearch extends Component {
...
+              style: otherTextStyle,
+              value: props.otherText,
+              onChange: props.otherTextHandler,
+              name: 'otherText',
+              id: 'otherText',
+              maxLength: '512',
+              rows: '2',
+              required: props.other,
+              placeholder: 'Please specify if selected (max. 512 characters)',
+              enabled: true
```
JVThomas: Going to add a brief overview of the html syntax we use before adding a code suggestion. The syntax is known as hyperscript, which swaps out JSX or traditional html templating with a more JS functional approach. 

For the example above, you can see that the textarea is rendered with a call to `textarea()`, which is provided by the `react-hyperscript-helpers` package. The argument is a Javascript object that can have HTML/CSS properties like `placeholder`, but there are also target keys that the function will evaluate internally. One such key is `isRendered` whose value will be evaluated as a Boolean. If the value is true, the element will render, otherwise it won't show at all.

Now onto the style suggestion. Instead of using a ternary operator, you can simply add `isRendered: props.other` to the JS object to get the same functionality, but with less typing.

raejohanek: Thanks Justin! I will make this change

src/pages/dar_application/TypeOfResearch.js
```diff
@@ -170,19 +170,19 @@ export const TypeOfResearch = hh(class TypeOfResearch extends Component {
...
+            style: otherTextStyle,
+            value: props.otherText,
+            onChange: props.otherTextHandler,
+            name: 'otherText',
+            id: 'otherText',
+            maxLength: '512',
+            rows: '2',
+            required: props.other,
+            placeholder: 'Please specify if selected (max. 512 characters)',
+            enabled: true
```
JVThomas: One last comment, and I apologize for not catching it earlier. This template is a small part of an application form that we show before and after submission. Before submission it makes sense to edit (its a form after all), but afterwards the information should be read-only, so we need to make the textarea is disabled.

HTML inputs generally have a `disabled` property that determines if an input should be editable or not. On top of that this file has the value `props.disabled` which carries the needed boolean value. So simply swap out `enabled: true` with `disabled: props.disabled` and you should be good to go.

One more note: The styling may not look like a disabled element when you do this and that's expected. Textareas require some custom styling on our part to achieve that look, but that's outside the scope of this PR so don't worry about tackling that here!

JVThomas (changes requested):

src/pages/dar_application/TypeOfResearch.js
raejohanek: that makes sense! I will fix that

JVThomas (approved):

PR #792, opened 2021-01-11
DUOS-858[risk=no]Add two buttons to Home page midsection

raejohanek:
## Address
https://broadworkbench.atlassian.net/browse/DUOS-858?atlOrigin=eyJpIjoiZGZhYjU2YWI4ODA4NDFjOTg5NDM1YjJjMzBkYjdiMzYiLCJwIjoiaiJ9

## Scope of Change
- Added constant buttonStyleBig
- Added "register a dataset in DUOS" button
- Added "submit a data access request" button
- Modified margins and padding
- Pass 'customStyle' prop to SignIn from Home
- render a button in the custom style (buttonStyleBig) if there is a customStyle prop
- otherwise render the regular google sign in button

In all cases:

### DIFF
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -143,7 +143,7 @@ export const SignIn = hh(class SignIn extends Component {
       h(GoogleLogin,
         _.isNil(this.props.customStyle) ? defaultStyle : {
           render: (props) => button({className: 'btn-primary', onClick: props.onClick, style: this.props.customStyle}, 'Submit a Data Access Request'),
-            defaultStyle
+          defaultStyle
         });
   };
 
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -129,7 +129,7 @@ export const SignIn = hh(class SignIn extends Component {
     return disabled ?
       div({ style: { textAlign: 'center', height: '44px', width: '180px' } }, [
         img({ src: '/images/loading-indicator.svg', alt: 'spinner' })
-      ]) : //_.isNil(this.props.customStyle) ?
+      ]) :
       h(GoogleLogin,
         _.isNil(this.props.customStyle) ? {
           scope: 'openid email profile',
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -2,7 +2,7 @@ import _ from 'lodash/fp';
 import { Component } from 'react';
 import GoogleLogin from 'react-google-login';
 import { div, h, hh, img, span, button } from 'react-hyperscript-helpers';
-import { Alert } from '../components/Alert';
+import { Alert } from './Alert';
 import { User } from '../libs/ajax';
 import { Config } from '../libs/config';
 import { Storage } from '../libs/storage';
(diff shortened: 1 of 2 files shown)


src/pages/Home.js
```diff
@@ -86,6 +86,19 @@ class Home extends Component {
       margin: '2rem auto'
     };
 
+    const buttonStyleBig = {
+      borderRadius: '5px',
```
lindojong: nit: for the stuff that's duplicated from `buttonStyle`, you can use it here as `...buttonStyle, fontSize: '22px', ...` and override the stuff to be changed directly

src/pages/Home.js
```diff
@@ -141,7 +154,19 @@ class Home extends Component {
...
+          div({ className: 'row', style: { background: '#eff0f2', margin: '50px 0', padding: '48px 0 60px 0' } }, [
+            div({ className: 'row', style: { background: '#eff0f2', margin: '0 0 48px 0'} }, [
+              div({ className:'col-lg-6 col-md-6', style: { display: 'flex', justifyContent: 'center' }}, [
+                button({ className: 'btn-primary', style: buttonStyleBig }, [
+                  a({href: 'https://databiosphere.github.io/duos-ui/DatasetRegistrationInfo', style: {color: '#fff'}}, ['Register a dataset in DUOS'])
+                ])
+              ]),
+              div({ className:'col-lg-6 col-md-6', style: { display: 'flex', justifyContent: 'center' }}, [
+                button({ className: 'btn-primary', style: buttonStyleBig }, [
+                  a({href: '/dataset_catalog', style: {color: '#fff'}}, ['Submit a Data Access Request'])
```
lindojong: Might be more of a question for Jonathan, but should we show this button if a user is not logged in?

src/pages/Home.js
```diff
@@ -86,6 +86,19 @@ class Home extends Component {
...
+    const buttonStyleBig = {
+      borderRadius: '5px',
+      backgroundColor: '#00689F',
+      fontFamily: 'Montserrat',
+      fontSize: '22px',
+      fontWeight: 500,
+      color: '#FFFFFF',
+      padding: '0 5rem',
+      margin: '2rem auto',
+      height: '50px',
```
lindojong, line 92: nit: (optional) the buttons seems a bit wide (a lot of extra empty space beyond the text horizontally), but if you like how it looks you don't need to change it!

src/pages/Home.js
```diff
@@ -141,7 +154,19 @@ class Home extends Component {
               ])
             ])
           ]),
-          div({ className: 'row', style: { background: '#eff0f2', margin: '50px 0', padding: '60px 0 72px 0' } }, [
+          div({ className: 'row', style: { background: '#eff0f2', margin: '50px 0', padding: '48px 0 60px 0' } }, [
+            div({ className: 'row', style: { background: '#eff0f2', margin: '0 0 48px 0'} }, [
+              div({ className:'col-lg-6 col-md-6', style: { display: 'flex', justifyContent: 'center' }}, [
+                button({ className: 'btn-primary', style: buttonStyleBig }, [
+                  a({href: 'https://databiosphere.github.io/duos-ui/DatasetRegistrationInfo', style: {color: '#fff'}}, ['Register a dataset in DUOS'])
```
lindojong: Are we going to be linking to the github pages instead of creating new pages inside duos-ui for future FAQ docs?

src/pages/Home.js
raejohanek: great idea! thank you!

src/pages/Home.js
raejohanek: That's a good point, this button won't work for someone not logged in, so I'll check with him.

src/pages/Home.js
raejohanek: I do kind of like the wider buttons

src/pages/Home.js
raejohanek: I'm not sure about future documents, but as of right now I do believe we are linking the github pages.

src/pages/Home.js
```diff
@@ -141,7 +148,21 @@ class Home extends Component {
...
```
rushtong: When we link outside of the app itself, it should be in a new tab.

src/components/SignIn.js
```diff
@@ -130,14 +130,27 @@ export const SignIn = hh(class SignIn extends Component {
...
+          render: (props) => button({className: 'btn-primary', onClick: props.onClick, style: this.props.customStyle}, 'Submit a Data Access Request'),
+          scope: 'openid email profile',
+          height: '44px',
+          width: '180px',
+          theme: 'dark',
+          clientId: this.state.clientId,
+          onSuccess: this.responseGoogle,
+          onFailure: this.forbidden,
+          disabledStyle: { 'opacity': '25%', 'cursor': 'not-allowed' }
+        });
```
rushtong, line 147: This is a good example where we can extract out the common set of properties between these two things and then merge in something additional for the changes. For example, scope, height, etc. are all the exact same values. To clean this up a bit, add this before the return statement on line 129:
```
    const defaultStyle = {
      scope: 'openid email profile',
      height: '44px',
      width: '180px',
      theme: 'dark',
      clientId: this.state.clientId,
      onSuccess: this.responseGoogle,
      onFailure: this.forbidden,
      disabledStyle: { 'opacity': '25%', 'cursor': 'not-allowed' }
    };
```
Then, you can replace the google login component like this:
```
      h(GoogleLogin,
        _.isNil(this.props.customStyle) ? defaultStyle : {
          ...{render: (props) => button({className: 'btn-primary', onClick: props.onClick, style: this.props.customStyle}, 'Submit a Data Access Request')},
          ...defaultStyle
        });
```
And the end result is that you have defined one set of component properties but used it in two places.

rushtong (commented):
One bit of feedback to help reduce the copy/paste, but otherwise, looks great!

src/components/SignIn.js
rushtong: https://www.javascripttutorial.net/object/javascript-merge-objects/

rushtong (approved):
👍🏽  Thanks for the changes!

rushtong:
👍🏽  Pending checks.

PR #795, opened 2021-01-12
Add PDF to public folder(part of: DUOS-862[risk=no]Update DS Registration Section 3 text and PDF)

raejohanek:
## Address
https://broadworkbench.atlassian.net/browse/DUOS-862?

## Scope of Changes
Adding Data Provider Agreement PDF to the public folder so that I can easily link it in the update for DUOS-862

In all cases:


rushtong:
Some initial comments that we can deal with in stand-up:
* There is a required text change to the DAR form that's not reflected in this PR.
* The story requests a change to the underlying library card PDF file in [this directory](https://github.com/DataBiosphere/duos-ui/tree/develop/public). There are some options here.
  * One is to rename the file to something more generic, and then changing the content of the binary itself to be the binary of the new pdf. This would also require refactoring any links to the old file name - there should be several, IIRC.
  * Another is to just change the content of the underlying pdf file without changing the file name, but since it is date specific, the file name is no longer an accurate representation of the content.

raejohanek:
> Some initial comments that we can deal with in stand-up:
> 
> * There is a required text change to the DAR form that's not reflected in this PR.
> * The story requests a change to the underlying library card PDF file in [this directory](https://github.com/DataBiosphere/duos-ui/tree/develop/public). There are some options here.
>   
>   * One is to rename the file to something more generic, and then changing the content of the binary itself to be the binary of the new pdf. This would also require refactoring any links to the old file name - there should be several, IIRC.
>   * Another is to just change the content of the underlying pdf file without changing the file name, but since it is date specific, the file name is no longer an accurate representation of the content.
 
Maybe I misunderstood the story, it says "update the PDF to the one attached", so I was simply going to replace the current PDF file with this one, instead of making any changes to the current file. Should I be taking a different approach, like the ones you described, instead?

rushtong:
@raejohanek From the story:

> Change all text in "3." section from 'Dataset Registration Agreement(s)' to 'Data Provider Agreement'

This is asking to make a small text change in the Dataset Registration page.

> Update PDF to the one attached.

This is saying to replace [this file](https://github.com/DataBiosphere/duos-ui/blob/develop/public/DUOSLibraryCardAgreement_10.14.2020.pdf) with the PDF that's linked in the story. This PR only adds the new PDF file to that same public directory, but it doesn't actually replace `DUOSLibraryCardAgreement_10.14.2020.pdf` with anything. My comments above were meant to illustrate a couple of options in how to do that.

raejohanek:
> @raejohanek From the story:
> 
> > Change all text in "3." section from 'Dataset Registration Agreement(s)' to 'Data Provider Agreement'
> 
> This is asking to make a small text change in the Dataset Registration page.
> 
> > Update PDF to the one attached.
> 
> This is saying to replace [this file](https://github.com/DataBiosphere/duos-ui/blob/develop/public/DUOSLibraryCardAgreement_10.14.2020.pdf) with the PDF that's linked in the story. This PR only adds the new PDF file to that same public directory, but it doesn't actually replace `DUOSLibraryCardAgreement_10.14.2020.pdf` with anything. My comments above were meant to illustrate a couple of options in how to do that.

Hi Greg! I have those changes made on the DUOS-862 Branch, but before I make that PR, I believe I need to make this PR (putting the pdf in the public folder) so then I can use the link to this PDF in the PR I make for DUOS-862. I can include a link to this PDF from google drive at the moment, but I want to be consistent with the linked PDFs other places on the website, they all use links from the public folder on github. As far as I understand, I cannot use this link until the PDF is in the folder. So, this is an intermediary request, thats why I included "part of" in the title, but I think this was confusing. If there is a way for these two to happen in the same PR, that would be great to learn, if not, once this PDF is added I will submit the PR for DUOS-862.

rushtong:
Ah ... I see where you're coming from. All of this can be done in one PR, no need for an intermediary one. Will set up a session so I can walk you through the process.

PR #801, opened 2021-01-14
DUOS-916[risk=no]UI Updates to Dataset Catalog

raejohanek:
## Address
https://broadworkbench.atlassian.net/browse/DUOS-916?

## Scope of Changes
-add new table style to CSS index (white, gray outline, gray shadow, thicker margins)
-add new text style
-replace current table style and text color variables with the new ones

In all cases:

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -322,6 +322,7 @@ small {
 
 .search-wrapper {
   margin-top: 35px !important;
+  margin-bottom: 20px !important;
 }
 
 @media (max-width: 800px) {
@@ -463,15 +464,6 @@ small {
   height: auto;
   background: #eeeeee;
   margin: 15px 0 20px 0;
-}
-
-.table-catalog {
-  padding: 15px 25px !important;
-  box-shadow: 3px 3px 0 #cccccc;
-  height: auto;
-  border: 1px solid #cccccc;
-  background: #ffffff;
-  margin: 25px 0 10px 0;
 }
 
 .box-vote, .box-vote-results, .box-vote-singleresults, .box-vote-stats {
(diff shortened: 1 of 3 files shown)


src/index.css
```diff
@@ -1622,6 +1631,13 @@ td:first-child .checkbox label {
   margin: 0!important;
   color: #777777;
 }
+.catalog-table-items {
+  font-weight: normal !important;
+  font-size: 14px !important;
+  padding: 10px 20px 5px 5px !important;
+  margin: 0!important;
+  color: #000000;
+}
```
rushtong: Please look at `theme.js` which is where I think this should exist. Make these two classes style objects and then import the object directly into the catalog page directly. Have a look [here](https://github.com/DataBiosphere/duos-ui/blob/develop/src/pages/access_review/ApplicantInfo.js) where we create constants that contain style information, then set the `style` property directly on the html elements where we want them. This way, if these are pulled into a common theme, other classes can then start using them. Rename the styles to better reflect what they mean, i.e. `lightTable` or something that makes sense to you.

rushtong (changes requested):
Looks fantastic on screen. See one suggestion inline for how to make it more useful across components.

src/index.css
```diff
@@ -322,6 +322,7 @@ small {
 
 .search-wrapper {
   margin-top: 35px !important;
+  margin-bottom: 20px !important;
```
rushtong, line 325: You don't have to make any changes for this comment, the downstream impact is pretty trivial, but it's a great learning example. Open up a tab with your branch, and then another side by side on dev and view the admin manage DAC page. This css change slightly pushes down the table below the search field. This is the primary reason I push for moving our styles to inline styles. If you make a change to the inline style on a component, you won't impact any other pages/components in any way, either positive or negative. If you make a change to a css file, you need to visually assess the impact in all pages where that style class is used. In this case, the search wrapper is used quite a bit.

rushtong (approved):
This looks great, thank you for the changes!

src/index.css
rushtong: The second big reason I push for inline styles are that they are tremendously flexible. I can refactor them easily, find all usages easily, merge a base style with new properties easily, filter out/replace existing properties, etc. just using standard javascript. The only real limitation to this approach is css state. We need to manage things like `hover` state via javascript instead of css.

PR #803, opened 2021-01-14
DUOS-887[risk=no]Researcher Console - Visual Ease/Accessibility Updates

raejohanek:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-887

## Scope of Change
-changed CSS styling of Pageheader descriptions to be non italic and black
-changed CSS styling for black cell body text
-changed CSS styling of the table to be white with an outline
-changed CSS styling of table headers so they are nonitalic and purple on the research console

\<your comments for this PR go here \>

In all cases:

### DIFF
--- a/src/components/PageSubHeading.css
+++ b/src/components/PageSubHeading.css
@@ -33,8 +33,7 @@
 }
 
 .page-sub-heading-description {
-  color: #777777;
-  font-style: italic;
+  color: #000000;
   font-weight: 500;
   font-size: 15px;
 }
--- a/src/components/PageHeading.css
+++ b/src/components/PageHeading.css
@@ -25,8 +25,7 @@
 }
 
 .page-heading-description {
-  color: #777777;
-  font-style: italic;
+  color: #000000;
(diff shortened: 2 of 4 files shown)


rushtong:
This is still a draft PR, sorry if this feedback might be premature. 

The text in the cells should be better aligned. Right now, it is pushed up to the top of the table row which throws off the alignment with the icons on the left and the buttons on the right.

Screen Shot 2021-01-15 at 7 16 24 AM

src/components/PageSubHeading.css
```diff
@@ -33,8 +33,7 @@
 }
 
 .page-sub-heading-description {
-  color: #777777;
-  font-style: italic;
+  color: #000000;
```
rushtong, line 36: Instead of making changes to this class (same feedback in `PageHeading.css`), I would update remove the use of this class from the component, and then apply whatever style you need to it through a style object. It could go into `theme.js` as `PageHeading` and `PageSubheading` objects to make the transition a little more legible.

PR #807, opened 2021-01-15
DUOS-887[risk=no]Researcher Console - Visual Ease/Accessibility Updates

raejohanek:
## Address
https://broadworkbench.atlassian.net/browse/DUOS-887

## Scope of Changes
-remove change made to index.css
- remove use of jumbotron and table-box css classes, replace with lightTable java style object, turning table white, font black, removing italics
- refactor heading description from css class to js object

In all cases:


src/components/PageHeading.css
```diff
@@ -19,19 +19,6 @@
...
-
-.page-heading-description {
-  color: #777777;
-  font-style: italic;
-  font-weight: normal;
-  font-size: 19px;
-  line-height: 25px;
-}
-
 .page-heading-description .consent-data {
```
lindojong: You might want to double check if there are any `PageHeading`s in the code that still have a `consent-data` class as well, and verify if there's a change. This particular CSS block would never get executed now that `PageHeading` no longer uses the `page-heading-description` class.

src/pages/ResearcherConsole.js
```diff
@@ -223,33 +224,35 @@ class ResearcherConsole extends Component {
...
+              div({ style: Theme.lightTable }, [
                 div({ className: "row no-margin" }, [
-                  div({ className: "col-lg-2 col-lg-offset-1 col-md-2 col-md-offset-1 col-sm-2 col-sm-offset-1 col-xs-2 col-xs-offset-1 cell-header default-color" }, ["Temporary id"]),
-                  div({ className: "col-lg-5 col-md-5 col-sm-5 col-xs-5 cell-header default-color" }, ["Project title"]),
-                  div({ className: "col-lg-2 col-md-2 col-sm-2 col-xs-2 cell-header default-color" }, ["Date"]),
-                  div({ className: "col-lg-2 col-md-2 col-sm-2 col-xs-2 cell-header f-center default-color" }, ["Resume"]),
+                  div({ style: Theme.textTableHead, className: "col-xs-2 access-color" }, ["Temporary ID"]),
+                  div({ style: Theme.textTableHead, className: "col-xs-5 col-xs-offset-1 access-color" }, ["Project Title"]),
+                  div({ style: Theme.textTableHead, className: "col-xs-2 access-color" }, ["Date"]),
+                  div({ style: Theme.textTableHead, className: "col-xs-2 f-center access-color" }, ["Resume"]),
```
lindojong: Screen Shot 2021-01-15 at 2 20 08 PM
Looks like the heading for this table got misaligned here! The 1 col worth of space for the trash icon was provided by the `col-xx-offset-1`, but as an alternative you could add a new empty column placeholder, if you don't want to depend on the classes.

lindojong (commented):
(optional) I think this could be a good opportunity to move out all the CSS from both `PageHeading` and `PageSubHeading` into the components instead, since you've already taken some action to move some of it.

src/components/PageHeading.js
```diff
@@ -1,22 +1,63 @@
...
+        style: ICON
       }),
-      div({ className: "page-heading-text " + this.props.iconSize}, [
-        h2({ id: this.props.id + "_title", className: "page-heading-title  " + this.props.color + "-color", style: this.props.titleStyle }, [this.props.title]),
-        span({ id: this.props.id + "_description", className: "page-heading-description", style: this.props.descriptionStyle }, [this.props.description]),
+      div(this.props.iconSize === 'none' ?
+        { style: NONE }
+        : this.props.iconSize === 'large' ?
+          { style: LARGE }
+          : { style: MEDIUM }, [
```
lindojong: One last suggestion I have is to move checking what kind of iconSize out to a separate method, and then just applying `{ style: newMethod(this.props.iconSize) }` so it looks cleaner, especially since each style for now is just a different number for the margin size. Otherwise, the changes look great!

src/components/PageHeading.js
```diff
@@ -1,22 +1,75 @@
...
+    };
+
+    const NONE = {
+      marginLeft: '0'
+    };
+
+    if (iconSize === 'none') {
+      return NONE;
+      }
+      else {
```
lindojong: nit: can simplify this to multiple `if (case1) {} if (case2) {}` before the final `else {}`

lindojong (commented):
Thank you for the changes! To resolve the codacy error, be sure to try out the built in formatting in intellij from `Code > Reformat Code.` Intellij should also be able to flag some of these linting errors inside the IDE with a yellow highlight.
Screen Shot 2021-01-19 at 11 34 45 AM

src/components/PageHeading.js
```diff
@@ -1,22 +1,76 @@
-import { Component } from 'react';
-import { div, hh, img, h2, span } from 'react-hyperscript-helpers';
-import './PageHeading.css';
+import {Component} from 'react';
+import {div, hh, img, h2, span} from 'react-hyperscript-helpers';
 
 export const PageHeading = hh(class PageHeading extends Component {
```
rushtong, line 4: Outside the scope of this PR, but these components can easily be pure functions instead. This shouldn't hold up the PR, but if you want an example, see the work done [in the DAR pages](https://github.com/DataBiosphere/duos-ui/blob/develop/src/pages/dar_application/DataAccessRequest.js#L38)

rushtong (approved):
Console looks fantastic! 👍🏽

PR #808, opened 2021-01-15
[DUOS-900][risk=no] Update the UI to validate dataset name more efficiently

lindojong:
Addresses: https://broadworkbench.atlassian.net/browse/DUOS-900
Implements the validate dataset name endpoint on the dataset registration field, verifying if the given name is available or not, right after it's typed.

### Scope of Changes
---
- Validate dataset name availability on blur (after exiting the dataset name input field), IF the input value is different than what it was last
- Renamed `showDatasetNameErrors` to the slightly more descriptive `showDatasetNameErrorHighlight` and added a note for what it does

In all cases:

### DIFF
--- a/src/pages/DatasetRegistration.js
+++ b/src/pages/DatasetRegistration.js
@@ -300,7 +300,6 @@ class DatasetRegistration extends Component {
       (!fp.isEmpty(this.state.updateDataset) || !this.isTypeOfResearchInvalid());
   };
 
-  // call the new ajax function and save it to the state?
   async validateDatasetName(name) {
     return DataSet.validateDatasetName(name).then(datasetId => {
       const isValid = (datasetId < 0);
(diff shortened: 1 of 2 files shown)


src/pages/DatasetRegistration.js
```diff
@@ -244,6 +244,21 @@ class DatasetRegistration extends Component {
     });
   };
 
+  // same as handleChange, but adds call to validate dataset name and only affects state if a change has been made
+  handleDatasetNameChange = (e) => {
+    const value = e.target.value;
+    if (this.state.datasetData.datasetName !== value) {
+      this.validateDatasetName(value);
```
JVThomas: This function is asynchronous, so the code (as is) will execute before ```validateDatasetName()``` resolves. This causes a side-effect where the error message shows briefly while the dataset name is being validated (on first input). Easy way to fix this is to make ```handleDatasetNameChange``` an async function and add an ```await``` keyword in front of ```validateDatasetName```. This way the body of the function will wait on the validation resolve before moving forward.

JVThomas (changes requested):

JVThomas (approved):

PR #832, opened 2021-01-26
DUOS-760 [risk=no] Initial Page layout and styling of Chair Console

JVThomas:
Addresses [DUOS-760](https://broadworkbench.atlassian.net/browse/DUOS-760)

Initial styling and table layout of the new Chair Console, as well as modal depicting DAR details on click. Some scaffolding added for pagination and search, but they serve as as an initial stepping stone and are not indicative of the end result of that option. As such the table right now will show all DAR entries up front since view control (since view control is tied to pagination and search).

In all cases:


raejohanek (commented):
this looks great! I am wondering if we want the chair console tables to match the new styling of the dataset catalog and researcher console, white background/black body text/no italics ?

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
-import { Component } from 'react';
-import { a, div } from 'react-hyperscript-helpers';
-import {PageHeading} from "../components/PageHeading";
+import {filter, isEmpty, isNil, find, cloneDeep} from 'lodash/fp';
+import Modal from 'react-modal';
+import { useState, useEffect, useRef} from 'react';
+import { div, h, img } from 'react-hyperscript-helpers';
+// import { SearchBox } from '../components/SearchBox';
```
rushtong: nit: rm

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
...
+import { DAR, DataSet, PendingCases, User } from '../libs/ajax';
+import { Storage } from '../libs/storage';
+// import { NavigationUtils, USER_ROLES } from '../libs/utils';
```
rushtong: nit: rm

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
...
+  RIGHT_HEADER_SECTION: {
+    display: 'flex',
+    alignItems: 'flex-end'
+  },
+  LEFT_HEADER_SECTION: {
+    display: 'flex',
+    flexDirection: 'row',
+    paddingTop: "3rem"
+  },
+  TABLE: {
```
rushtong: I'd love for these to be re-usable by other components. Can the table styles be extracted to themes.js?

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
...
+      justifyContent: "left",
+      alignItems: "center",
+    },
+    ELECTION_ACTIONS_CELL: {
+      width: "23%",
+      margin: "0 2%",
+      display: "flex",
+      justifyContent: "left",
+      alignItems: "center",
+    }
```
rushtong: There's a lot of duplicate code in the cells that could be in a single `CELL` style, then specific cells can merge in their own customized properties.

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
...
+      researcherInfo.researcherProperties.forEach((property) => {
+        researcherProperties[property.propertyKey] = property.propertyValue;
+      });
+      return Object.assign({}, researcherInfo, researcherProperties);
+    }catch(error){
+      Notifications.showError({text: 'Error: Failed to initialize researcher information'});
+      return {};
+    }
+  }
+};
```
rushtong: This, along with `getDatasetNames` are modal specific functions that I think would be better off if we pulled them out into a separate chair console summary modal component. It would also provide an opportunity to extract out all of the modal style info which would clean up this class a lot.

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
...
+  //     setCurrentPage(targetPage);
+  //     setVisibleList(filteredList.slice(targetStartIndex, targetEndIndex));
+  //   }
+  // }, [setCurrentPage, setVisibleList]);
+
+  useEffect(() => {
+    const init = async() => {
+      try {
+        const currentUser = Storage.getCurrentUser();
+        const caseList = await PendingCases.findDataRequestPendingCasesByUser(currentUser.dacUserId);
```
rushtong: This is using the old API, not the v2 api.

src/pages/NewChairConsole.js
```diff
@@ -1,40 +1,426 @@
...
+        const pendingList = filter((e) => { return e.electionStatus !== 'Closed'; }, caseList.access);
+
+        //NOTE: commented out code below are thoughts/pseudocode for pagination/search initialization
+        // setTotalPendingVotes(pendingList.length);
+        setElectionList(pendingList);
+        // setFilteredList(pendingList);
+        // if(pendingList.length > 0) {
+        //   const endIndex = darLimit > pendingList.length ? darLimit : pendingList.length;
+        //   setVisibleList(filteredList.slice(0, endIndex));
+        // }
```
rushtong: Can we extract these comments into the ticket for pagination/search?

rushtong (changes requested):
See comments inline ... the page starts to load fine, but fails when looking at an election dac. We need to know if this is a UI problem or an API problem, but since this is not using the correct API, it's hard to tell at the moment. 

This is my screen as a DAC Chairperson:
Screen Shot 2021-01-27 at 10 16 04 AM

src/pages/NewChairConsole.js
JVThomas: Merge in itself is a relatively expensive operation (lodash effectively iterates over keys and applies them to the next argument in the method, depth first search). So when you're dealing with multiple cells across multiple rows it could affect performance. Assign is only one level deep, so there's some issues with nested attributes.

JVThomas:
> See comments inline ... the page starts to load fine, but fails when looking at an election dac. We need to know if this is a UI problem or an API problem, but since this is not using the correct API, it's hard to tell at the moment.
> 
> This is my screen as a DAC Chairperson:
> Screen Shot 2021-01-27 at 10 16 04 AM

Screen Shot 2021-01-27 at 10 45 29 AM

Code loads fine on my end. Error is indicating that there's no name attribute on null, which would mean there's no DAC assigned to an election? Is that possible? Or maybe it's a timing issue?

Here's a screenshot hooked up to dev
Screen Shot 2021-01-27 at 10 49 57 AM

src/pages/NewChairConsole.js
JVThomas: Sure, but it's pseudocode/thoughts that are supposed to act as a guide for the next ticket.

rushtong:
@JVThomas Yes, it could be true (especially in dev) that there are elections missing a DAC. We also need to be looking at the `/api/dar/manage/v2` endpoint for data.

src/pages/NewChairConsole.js
JVThomas: What's the v2 variant for pendingCases? Is it just using the standard DAR get and then filtering it out manually?

src/pages/NewChairConsole.js
JVThomas: Sure, makes sense to have them as exportable constants there

JVThomas:
> @JVThomas Yes, it could be true (especially in dev) that there are elections missing a DAC. We also need to be looking at the `/api/dar/manage/v2` endpoint for data.

For the elections missing a DAC, is that a possible format for an election or is it a bugged record that needs to be filtered out from the view?

rushtong:
@JVThomas 

> For the elections missing a DAC, is that a possible format for an election or is it a bugged record that needs to be filtered out from the view?

The point is moot - the API this PR is using is buggy & being deprecated. The v2 dar manage API should be doing the right things with respect to getting the DACs for the datasets of the DARs instead of trying to get it from the election.

src/libs/ajax.js
```diff
@@ -411,6 +412,19 @@ export const DAR = {
...
+    const url = `${await Config.getApiUrl()}/dar/manage/v2`;
+    const res = await axios.get(url, Config.authOpts());
+    return res.data;
+  },
+
+  updateDar: async (dar, id) => {
+    const url = `${await Config.getApiUrl()}/dar/${id}`;
+    const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(dar), { method: 'PUT' }]));
+    return await res;
+  },
```
rushtong: This was removed in https://github.com/DataBiosphere/duos-ui/pull/799. I don't see this used in your code, so it looks like you might need to rebase.

rushtong (approved):
Looks and works great. The modal will need more iteration on the style, but this is a great place to start from so we can pull in the team for more design feedback. One suggestion for rebasing/merging the latest from develop, but otherwise, 👍🏽

PR #851, opened 2021-01-29
[DUOS-983] PDF Exchange: DPA for LCA on SO and DAR Step 4 pages

solideoglori:
Replaced DPA PDF with LCA PDF on SO and DAR Step 4 pages.

In all cases:


src/pages/HomeSigningOfficial.js
```diff
@@ -80,7 +80,7 @@ class HomeSigningOfficial extends Component {
               div({style: {marginTop: '2rem', marginBottom: '2rem'}},
                 [a({
                   id: 'link_downloadAgreement',
-                  href: '/Data_Provider_Agreement.pdf',
+                  href: '/DUOS_LCA_01.2021.pdf',
```
rushtong: The old file is referenced in 4 places. Instead of adding a new one in two of the 4, can we instead just replace the contents of `Data_Provider_Agreement.pdf` with the new one?

rushtong (changes requested):
See comment inline.

rushtong (changes requested):
Same feedback applies - can we instead replace the current PDF contents with the new file instead of linking this new pdf to 2 of the 4 places it exists now?

solideoglori:
@rushtong - possibly, but so we're clear, there are 2 distinct PDFs we need to have available. 

The LCA needs to be replaced the SO page and DAR Step 4 (this is a prod error disrupting DAR submissions). 

The DPA needs to be on Dataset Registration and it's accompanying info page, which is current state.

I think this means we need two versions of a PDF reference, since we need to consistently present 2 different PDFs in different places.

rushtong:
@solideoglori - I can understand the need for 2 separate PDFs, but that's not what we have now - all 4 locations reference the same PDF. They are:

* Data Access Renewal
* Dataset Registration
* Home Signing Official (updated in this PR)
* Data Use Agreements (updated in this PR)

Let's clarify the exact requirements and implementation in the Jira ticket before proceeding.

solideoglori:
@rushtong updated ticket with the following text:

The LCA PDF needs to be referenced in the downloadable link on the following pages:

https://duos.broadinstitute.org/home_signing_official

https://duos.broadinstitute.org/dar_application

https://duos.broadinstitute.org/dar_renewal

The DPA PDF needs to be referenced in the downloadable link on the following pages:

https://duos.broadinstitute.org/dataset_registration

rushtong:
Looking at the ticket and this PR, I think we need one more change to the DAR Renewal page.
Also, it would be good if we could name the PDF document more generically so devs can understand what it is. I think `Library_Card_Agreement.pdf` is clear enough.

solideoglori:
@rushtong I updated the name of the agreement and added it to the DAR Renewal page as well.

rushtong (approved):
👍🏽  Thank you for the changes!

PR #860, opened 2021-02-04
DO NOT MERGE - Create codeql-analysis.yml

davidbernick:
Hi all -- this is, what should be, an innocent enough PR that enables CodeQL in our repo. it should be non-gating.


rushtong (approved):
I'm fine with this, the PR title (`DO NOT MERGE`) is misleading though. Will be happy to merge when we're happy with the final result.

rushtong:
Closing until we need this.

PR #861, opened 2021-02-05
DUOS-765 Pagination/Search added to New Chair Console

JVThomas:
Addresses [DUOS-765](https://broadworkbench.atlassian.net/browse/DUOS-765)

PR adds Pagination and search to New Chair Console. New Pagination component was made to remove dependency on `react-pagination` library. Search has been slightly modified to allow for a drill-down search on space or comma separated terms (search will split the list of terms and reduce the list iteratively).

NOTE: Noticed an issue on dev where multiple DARs with the same content (referenceID was identical) were being returned from the endpoint, which is why certain DARs will look to be repeated. From the UI's point of view these DARs are unique, so it'll render all of them. Choosing a specific DAR from the list of duplicates was not implemented since that would have to be a standard that must be applied to all UI elements (since you're effectively designating a specific election as the de-facto choice). Will hold off on implementing such a filter until further details are hashed out.

In all cases:

### DIFF
--- a/src/components/PaginationBar.js
+++ b/src/components/PaginationBar.js
@@ -1,6 +1,7 @@
 import { div, input, span } from 'react-hyperscript-helpers';
 import { useRef, useEffect } from 'react';
 import { toNumber } from 'lodash';
+import { Theme } from '../libs/theme';
 
 //NOTE: This is a new pagination made to fit with the updated table look
 //Component does not use third-party libraries for pagination, whereas the old version relies on 'react-pagination'
@@ -25,7 +26,7 @@ export default function PaginationBar(props) {
           span({
             onClick: (e) => goToPage(toNumber(currentPage.current.value) - 1),
             onMouseEnter: applyTextHover,
-            onMouseLeave: (e) => removeTextHover(e, '#1f3b50'),
+            onMouseLeave: (e) => removeTextHover(e, Theme.palette.primary),
           },['Prev']),
         ]),
         div({style: Styles.TABLE.CURRENT_PAGE}, [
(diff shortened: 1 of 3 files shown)


src/components/PaginationBar.js
```diff
@@ -0,0 +1,64 @@
...
+
+  useEffect(() => {
+    currentPage.current.value = props.currentPage;
+    tableSize.current.value = props.tableSize;
+  }, [props.currentPage, props.tableSize]);
+
+  return (
+    div({style: Styles.TABLE.FOOTER}, [
+      div({style: Styles.TABLE.PAGINATION_SECTION_OFFSET}),
+      div({style: Styles.TABLE.PAGINATION_BUTTON_SECTION}, [
```
rushtong, line 22: Nice style abstraction here!

src/components/PaginationBar.js
```diff
@@ -0,0 +1,64 @@
...
+        div({
+          style: Styles.TABLE.PAGINATION_BUTTON
+        }, [
+          span({
+            onClick: (e) => goToPage(toNumber(currentPage.current.value) - 1),
+            onMouseEnter: applyTextHover,
+            onMouseLeave: (e) => removeTextHover(e, '#1f3b50'),
```
rushtong: It feels like these should be named color constants somewhere. Same comment in the other usage on line 48.

rushtong (approved):
This looks great. Tested with a variety of filters and pagination options and found no issues. One minor comment inline that shouldn't hold up the PR if you choose not to address it.

PR #865, opened 2021-02-08
DUOS-761[risk=no]UI: New Chair Console Election Action Buttons

raejohanek:
## Scope
- Vote button: route user to access_reveiw/darid/voteid (for all open elections)
- Open/re-open button: ask user to confirm, refresh the page with now open status for that election (for all closed, canceled, final, or unreviewed elections)
- create and use ConfirmationModal to ask the user to confirm instead of window.confirm() or ConfirmationDialog
- display confirmation text notification after election is opened or re-opened

## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-761

In all cases:


src/pages/NewChairConsole.js
```diff
@@ -151,6 +190,30 @@ const NewChairConsole = () => {
     }
   };
 
+  const createElection = (darId, index) => {
+    if (darId !== null) {
+      let copy;
+      Election.createDARElection(darId)
+        .then(
```
JVThomas: .then is structured the same way as .catch in that it expects a callback function to execute the code.

```
.then((optionalResponseDataVar) => {
   //lines of code here
})
  ```

File: src/pages/NewChairConsole.js [hunk 2]
```diff
@@ -151,6 +190,30 @@ const NewChairConsole = () => {
...
+          Notifications.showSuccess({text: "Election successfully opened"}),
+          copy = cloneDeep(filteredList),
+          copy[index].election = { status: "Open", finalAccessVote: false },
```
JVThomas: Two key notes here, bit long but it's important to read through it:

1. We don't want to overwrite keys on objects directly unless we know we're not removing key data. If copy[index].election was something simple like a string or an int then we can overwrite it with no hesitation. But if it's something like an object we need to think about the other keys that can exist within it. So if the election looked like this...
```
{
  name: 'Election Name',
  id:273,
  status: 'Closed',
  finalAccessVote: false
}
```
... then we're going to lose the name and the id on the election, since we are now setting election to only ```{status: "Open", finalAccessVote: false}```. So instead of overwriting the election completely, you can do something like...
```
  copy[index].election.status = "Open";
  copy[index].finalAccessVote = false;
```
...to update the values we want.


2. Even if we do the above, nothing will update on the screen!

Kapture 2021-02-11 at 09 49 50

That's because React updates the screen by comparing the references of variables! It does this to optimize re-renders since it's faster to compare references than it is to compare every single key on an object (which can be very broad or, if it has a bunch of nested keys, very deep). With the above method, even though we've changed ```election.status```, election itself is still the same. To make a copy and update the attributes, we can use Object.assign(). Here's an example...
```
copy[index].election = Object.assign({}, copy[index].election, {status: "Open", finalAccessVote: false}
```

The way it works is that, iterating from right to left, apply the keys of the current object to the object on the left. If both objects have the same keys, then the value of the current object will be used to "update" the value of the object to the left. (I say "update" because the only object that has its value changed is the left most object). We use the empty object on the left because it's a fresh empty object, which means that it'll have a new reference (the second condition we're aiming for). Once we do the above, the ```setFilteredList(copy)``` should work as expected!

NOTE: lodash has it's own assign function that you can use that has the same result, but performs the operation differently (applies from left to right, always returns a copy so you don't need to add an empty object, etc.) Feel free to use whichever one you want, just note that it's important to be aware of implementation differences when using third-party libraries.

Lodash assign docs -> https://lodash.com/docs/4.17.15#assign

JVThomas (changes requested):
Some early feedback/explanations for now, still going through the PR

src/pages/NewChairConsole.js
raejohanek: Okay, thank you for this explanation!!! That makes sense about tracking variable references instead! The reason I was overwriting is because the election may not exist yet in the case that the status is unreviewed, so we can't always assign direct field values, also this is the same overwriting that happens in createDARElection so I was thinking we would want the elections to match in both places. But, looking at the Object.assign() line you have written above, that would work even if copy[index].election is null, right? the new object would just have those two properties, and if it isn't null the new object will have the additional data as well

JVThomas: Yep, exactly!

src/pages/NewChairConsole.js
JVThomas: I just realized from our Object.assign conversation that you don't have to worry about null objects when dealing with a successful create because you should be getting the data back from the response. The issue however is that the Election.createDARElection doesn't work the way to expect since it's part of legacy code. The current function returns a raw response that you have to parse with .json(). However .json() in of itself is asynchronous, so you'll have to do some promise chaining

```
Election.createDARElection(darId)
.then(electionResponse => electionResponse.json())
.then(newElection => {
  //earlier code block//
  copy[index].election = newElection
}
```

The newElection data should be updated with the expected status and finalAccessVote values already, and it's already a different object so you can just assign it as the new election.

NOTE: @rushtong I'm not a fan of this approach, but the current implementation of createDARElection leaves us with little choice (await/async approach will have similar issues). createDARElection is used in AdminManageAccess, so I don't want Rae to deal with the function refactor in this PR (it can be handled in a separate PR). Is this reasonable?

raejohanek: I see what you're saying, that makes sense to me!

src/pages/NewChairConsole.js
```diff
@@ -151,6 +190,30 @@ const NewChairConsole = () => {
...
 
+  const createElection = (darId, index) => {
+    if (darId !== null) {
+      let copy;
+      Election.createDARElection(darId)
+        .then(() => {
+          Notifications.showSuccess({text: "Election successfully opened"});
+          copy = cloneDeep(filteredList);
+          copy[index].election = Object.assign( {}, copy[index].election, {status: "Open", finalAccessVote: false});
+          setFilteredList(copy);
```
JVThomas, line 203: Another thing to note: filteredList is a subset of the total election list, where the subset can be, at most, as large as the electionList itself. This is important because whenever a user updates the search terms on the search bar, it uses the electionList as the initial master list and then cuts it down for each term. Once that's done, the trimmed list is assigned as the filteredList.

When you update the filteredList, you're updating a list that exists only with the current search terms. If the user changes the search terms, the filteredList is reset, and you lose the updates you've made.

Updating the filteredList is the right move to show immediate changes on screen, but you'll also need to make updates to the electionList for future search updates.

JVThomas (changes requested):
I've noticed another issue where the button/status doesn't update once you open an election, even with the code updates. I think this is a slightly more complex issue and will spend some time looking into it.

To reproduce it, just open an election on any non-open election with nothing in the search bar. The reason why this is interesting is because if you have search terms present, the row updates properly. My guess this problem has more to do with my templates and not the button itself, but I'll need some time to look into it. In the meantime, I left one more comment on the code that does need to be addressed (I left an explanation as to why).

src/pages/NewChairConsole.js
raejohanek: good thinking! I'll add an update to electionList as well. 

> I've noticed another issue where the button/status doesn't update once you open an election, even with the code updates. I think this is a slightly more complex issue and will spend some time looking into it.
> 
> To reproduce it, just open an election on any non-open election with nothing in the search bar. The reason why this is interesting is because if you have search terms present, the row updates properly. My guess this problem has more to do with my templates and not the button itself, but I'll need some time to look into it. In the meantime, I left one more comment on the code that does need to be addressed (I left an explanation as to why).

I'm not seeing this error, is this what you mean
https://user-images.githubusercontent.com/77015255/107695604-d9c24a00-6c7e-11eb-9f93-9137d4b70f56.mov

src/pages/NewChairConsole.js
rushtong: @JVThomas Good catch. The admin page will eventually be updated. We should create a new createElection method that works the way we want and deprecate the current function. Having the new function return the result of `await rawData.json()` (assuming a successful response) is the better way to go.

src/pages/NewChairConsole.js
JVThomas: Ok, I've figured it out (can't believe I didn't realize it sooner). The issue occurs on any page that's not page 1, and it happens due to index. Index is derived from the list of visible rows on the table, which means that the value will occur within the range [0, tableSize). If we're on page 2 or greater, the index needs to be processed based on the current page. 
```(ex: currentPage = 2, tableSize = 10, index = 0 ===> index on filteredList would be 10, not 0)```

Comment (raejohanek) [reply src/pages/NewChairConsole.js:203]: Oh wow! Good thinking! So, I will add a check for the current page before we get the index

File: src/pages/NewChairConsole.js [hunk 4]
```diff
@@ -151,6 +190,32 @@ const NewChairConsole = () => {
...
+    if (darId !== null) {
+      let copy;
+      const i = index + ((currentPage - 1) * tableSize);
+      Election.createDARElection(darId)
+        .then((electionResponse) => electionResponse.json())
+        .then((newElection) => {
+          Notifications.showSuccess({text: "Election successfully opened"});
+          copy = cloneDeep(filteredList);
+          copy[parseInt(i)].election = newElection;
+          setFilteredList(copy);
```
JVThomas, line 203: Alright now you'll need to update the electionList right after the filterList update. Remember that the filteredList is derived from applying filters on the electionList, so filteredList is more of a temporary state (changes with filters) while the electionList can be considered as a source of truth. However, because filteredList is a subset of electionList, you can't use the index you've calculated to access electionList. Instead you'll need to search electionList for the proper record.

One thing to note is that you can't search based on elections since you might be creating one for the first time. However the dar the election is associated is a viable search target.

JVThomas (changes requested):

src/pages/NewChairConsole.js
raejohanek: right, thank you!!

src/pages/NewChairConsole.js
```diff
@@ -38,13 +39,39 @@ const getElectionDate = (election) => {
...
   const endIndex = currentPage * tableSize; //NOTE: endIndex is exclusive, not inclusive
   const visibleWindow = props.filteredList.slice(startIndex, endIndex);
   const dataIDTextStyle = Styles.TABLE.DATA_REQUEST_TEXT;
   const recordTextStyle = Styles.TABLE.RECORD_TEXT;
 
+  const createActionButton = (electionInfo, index) => {
+    const name = "cell-button hover-color";
+    const e = electionInfo.election;
+    const dar = electionInfo.dar;
+    if (e !== null) {
```
JVThomas: I think this should be the last bug left to be ironed out. After this the PR should be good to go.

In javascript variables can be ```null``` or ```undefined``` however these values are not equal to each other. So if ```e = undefined``` then the if statement will return true since ```undefined !== null```. A good way to deal with this is to use lodash's isNil method (which is imported at the top of the file) which will treat undefined and null as equivalent values (since both in a sense represent an absence of value).

JVThomas (changes requested):

src/pages/NewChairConsole.js
raejohanek: awesome! I will make that change

src/pages/NewChairConsole.js
```diff
@@ -151,6 +190,34 @@ const NewChairConsole = () => {
     }
   };
 
+  const createElection = (darId, index) => {
+    if (darId !== null) {
```
JVThomas: Sorry, forgot to mark this one as well. It's pretty much the same issue as the e !== null check

JVThomas (changes requested):

JVThomas (approved):
Looks good, great work!

PR #866, opened 2021-02-08
[DUOS-1019][risk=no] Update DAR Modal Design

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1019

This PR updates the style a little bit and re-arranges the elements in the modal to more closely align with the previous version. In addition, this pulls user information from the DAR create user instead of using the unreliable data in `darInfo`.

### New Modal UI:

Screen Shot 2021-02-08 at 3 48 54 PM

### Old Modal UI:

Screen Shot 2021-02-08 at 3 47 11 PM

----

In all cases:

### DIFF
--- a/src/components/modals/DarModal.js
+++ b/src/components/modals/DarModal.js
@@ -28,8 +28,8 @@ const returnPIName = (researcher) => {
 };
 
 const returnInstitution = (researcher) => {
-  const piNameProp = find({propertyKey: "institution"})(researcher.researcherProperties);
-  return isNil(piNameProp) ? '- -' : piNameProp.propertyValue;
+  const institutionProp = find({propertyKey: "institution"})(researcher.researcherProperties);
+  return isNil(institutionProp) ? '- -' : institutionProp.propertyValue;
 };
 
 const processResearchTypes = (researchTypes) => {
--- a/src/components/modals/DarModal.js
+++ b/src/components/modals/DarModal.js
@@ -1,8 +1,11 @@
-import { div, h } from 'react-hyperscript-helpers';
+import {div, h} from 'react-hyperscript-helpers';
+import { Alert } from '../Alert';
+import { User } from '../../libs/ajax';
 import { Styles } from '../../libs/theme';
+import { Notifications } from '../../libs/utils';
 import Modal from 'react-modal';
-import {isEmpty} from 'lodash/fp';
-import {Alert} from '../Alert';
+import {find, isEmpty, isNil} from 'lodash/fp';
+import {useEffect, useState} from 'react';
 
 const ModalDetailRow = (props) => {
   return (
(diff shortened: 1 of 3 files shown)


src/components/modals/DarModal.js
```diff
@@ -32,37 +42,52 @@ const processResearchTypes = (researchTypes) => {
...
 
   return h(Modal, {
     isOpen: showModal,
     onRequestClose: closeModal,
     shouldCloseOnOverlayClick: true,
     style: {
-      content: Styles.MODAL.CONTENT
+      content: {
+        ...Styles.MODAL.CONTENT,
+        ...{boxShadow: '3px 3px 0 #cccccc', borderRadius: "10px"}
```
JVThomas: Instead of using a box-shadow, change the overlay to make the foreground a bit darker.

```
style: {
      content: {
        ...Styles.MODAL.CONTENT,
      },
      overlay: {
        backgroundColor: 'rgba(0, 0, 0, 0.5)'
      }
    }
 ```

File: src/components/modals/DarModal.js [hunk 2]
```diff
@@ -71,7 +96,16 @@ const DarModal = (props) => {
       h(ModalDetailRow, {
         label: 'Type of Research',
         detail: processResearchTypes(darDetails.researchType)
-      })
+      }),
+      div({
+        isRendered: requiresManualReview(darDetails),
+        style: Styles.ALERT
+      }, [
+        Alert({
```
JVThomas, line 104: The alert is great! Just wanted to add that with this addition you can remove the ternary string processing in processResearchTypes and just return type.description.

Current code for reference...
```
const processResearchTypes = (researchTypes) => {
  let researchStatements = '';
  if(!isEmpty(researchTypes)) {
    researchStatements = researchTypes.map(type => {
      return `${type.description} ${type.manualReview ? 'Requires manual review.' : ''}`;
    });
  }
  return researchStatements;
};
```

src/pages/NewChairConsole.js
```diff
@@ -98,13 +99,15 @@ const NewChairConsole = () => {
...
     let darData = darInfo.data;
-    if(!isNil(darData)) {
+    if (!isNil(darData)) {
       setShowModal(true);
       darData.researchType = DataUseTranslation.generateResearchTypes(darData);
       if(!darData.datasetNames) {
         darData.datasetNames = getDatasetNames(darData.datasets);
       }
       setDarDetails(darData);
+      const researcher = await User.getById(darData.userId);
```
JVThomas, line 109: Just for future reference, the investigator and pi keys I see on some of the DARs are holdovers from the old API?

JVThomas (changes requested):

src/components/modals/DarModal.js
rushtong: Thank you! 👍🏽

src/components/modals/DarModal.js
rushtong: Nice, thank you!

src/pages/NewChairConsole.js
rushtong: Yes. All fields that are here: https://github.com/DataBiosphere/consent/blob/e063adc816fe0afc4b861fc3c8a2072029a8243a/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java#L22 need to eventually go away. These are legacy fields that were stuffed into the dar mongo document simply because we didn't want to go the extra mile and look up user information from the actual user. In all cases in the UI, we should be moving away from using the composite objects that obscure where the underlying data is stored ... these DAR fields are great examples.

JVThomas (approved):

PR #888, opened 2021-02-18
DUOS-1037[risk=no]UI: Refactor eRA Commons component

raejohanek:
Scope:
-change API call from getResearcherPropertiesById to getMe
-eliminate getCurrentUser id calls in Storage
-rename method to better reflect underlying function

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1037

In all cases:

### DIFF
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -1,4 +1,4 @@
-import {_, isNil} from 'lodash';
+import {get, isNil, isUndefined, merge } from 'lodash';
 import { find, getOr } from 'lodash/fp';
 import * as qs from 'query-string';
 import React from 'react';
@@ -90,8 +90,8 @@ export const eRACommons = hh(class eRACommons extends React.Component {
         this.setState({ nihError: true });
         return false;
       });
-    if (_.isUndefined(isFcUser)) { return false; }
-    return _.get(isFcUser, 'enabled.google', false);
+    if (isUndefined(isFcUser)) { return false; }
+    return get(isFcUser, 'enabled.google', false);
   };
 
   registerUserToFC = async (properties) => {
@@ -145,11 +145,11 @@ export const eRACommons = hh(class eRACommons extends React.Component {
       fontSize: '.9em',
       transition: 'all .2s ease'
     };
-    const validationErrorState = _.get(this.props, 'validationError', false) ? {
+    const validationErrorState = get(this.props, 'validationError', false) ? {
       color: '#D13B07'
     } : {};
     const buttonStyle =
-      _.merge(this.state.isHovered ? _.merge(buttonNormalState, buttonHoverState) : buttonNormalState, validationErrorState);
+      merge(this.state.isHovered ? merge(buttonNormalState, buttonHoverState) : buttonNormalState, validationErrorState);
     const nihErrorMessage = 'Something went wrong. Please try again.';
 
     return (
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -55,7 +55,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
   };
 
   getUserInfo = async () => {
-    User.getMe().then(response => {
+    User.getMe().then((response) => {
       const isAuthorized = JSON.parse(_.get(response, 'eraAuthorized', "false"));
       const expirationCount = AuthenticateNIH.expirationCount(_.get(response, 'eraExpiration', 0));
       const nihValid = isAuthorized && expirationCount > 0;
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -2,9 +2,8 @@ import _ from 'lodash';
 import * as qs from 'query-string';
 import React from 'react';
 import { a, button, div, hh, label, span } from 'react-hyperscript-helpers';
-import { AuthenticateNIH, Researcher } from '../libs/ajax';
+import { AuthenticateNIH, User } from '../libs/ajax';
 import { Config } from '../libs/config';
-import { Storage } from '../libs/storage';
 
 
 export const eRACommons = hh(class eRACommons extends React.Component {
@@ -21,7 +20,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
     if (this.props.location !== undefined && this.props.location.search !== '') {
       this.authenticateAsNIHFCUser(this.props.location.search);
     } else {
-      this.getResearcherProperties();
+      this.getUserInfo();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/components/eRACommons.js
```diff
@@ -56,9 +54,8 @@ export const eRACommons = hh(class eRACommons extends React.Component {
     }
   };
 
-  getResearcherProperties = async () => {
-    const currentUserId = Storage.getCurrentUser().dacUserId;
-    Researcher.getPropertiesByResearcherId(currentUserId).then(response => {
+  getUserInfo = async () => {
+    User.getMe().then(response => {
```
rushtong, line 58: Codacy has a fix for the issue: [Expected parentheses around arrow function argument.](https://app.codacy.com/gh/DataBiosphere/duos-ui/pullRequest?prid=6975326)
```suggestion
    User.getMe().then((response) => {
```

src/components/eRACommons.js
```diff
@@ -34,10 +33,9 @@ export const eRACommons = hh(class eRACommons extends React.Component {
...
 
   authenticateAsNIHFCUser = async (searchArg) => {
-    const currentUserId = Storage.getCurrentUser().dacUserId;
     let isFcUser = await this.verifyUser();
     if (!isFcUser) {
-      await Researcher.getPropertiesByResearcherId(currentUserId).then(
+      await User.getMe().then(
         (response) => isFcUser = this.registerUserToFC(response),
         () => this.setState({ nihError: true })
       );
```
rushtong, line 42: The responses are different between these two API calls. You should also be making sure that the `response` sent to `registerUserToFC` doesn't break that method, which I think it will.

src/components/eRACommons.js
```diff
@@ -56,9 +54,8 @@ export const eRACommons = hh(class eRACommons extends React.Component {
...
   };
 
-  getResearcherProperties = async () => {
-    const currentUserId = Storage.getCurrentUser().dacUserId;
-    Researcher.getPropertiesByResearcherId(currentUserId).then(response => {
+  getUserInfo = async () => {
+    User.getMe().then((response) => {
       const isAuthorized = JSON.parse(_.get(response, 'eraAuthorized', "false"));
       const expirationCount = AuthenticateNIH.expirationCount(_.get(response, 'eraExpiration', 0));
       const nihValid = isAuthorized && expirationCount > 0;
```
rushtong: Here, again, the responses are different between the two API calls. I'm pretty sure that `_.get(response, 'eraAuthorized', "false"))` will always be false since we don't store `eraAuthorized` directly on the user.

rushtong (changes requested):
See comments inline

src/components/eRACommons.js
```diff
@@ -21,7 +20,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
     if (this.props.location !== undefined && this.props.location.search !== '') {
       this.authenticateAsNIHFCUser(this.props.location.search);
     } else {
-      this.getResearcherProperties();
+      this.getUserInfo();
```
rushtong: Could use an `await` here since `getUserInfo` is async

src/components/eRACommons.js
```diff
@@ -56,17 +54,16 @@ export const eRACommons = hh(class eRACommons extends React.Component {
...
 
-  getResearcherProperties = async () => {
-    const currentUserId = Storage.getCurrentUser().dacUserId;
-    Researcher.getPropertiesByResearcherId(currentUserId).then(response => {
-      const isAuthorized = JSON.parse(_.get(response, 'eraAuthorized', "false"));
-      const expirationCount = AuthenticateNIH.expirationCount(_.get(response, 'eraExpiration', 0));
+  getUserInfo = async () => {
+    User.getMe().then((response) => {
+      const isAuthorized = JSON.parse(_.get(response.researcherProperties, 'eraAuthorized', "false"));
+      const expirationCount = AuthenticateNIH.expirationCount(_.get(response.researcherProperties, 'eraExpiration', 0));
```
rushtong: Some points here:
* Here and elsewhere, we should move away from using `_` to refer to lodash functions and instead import them directly. 
* I also prefer the fp version as it is more legible and composable. 
* I don't think we need JSON.parse here, the object should already be a json structure. 
* We're not parsing the user's properties correctly here. There are no properties with the names as described, each property has the form of `{"propertyId":2579,"userId":3351,"propertyKey":"eraAuthorized","propertyValue":"true"}` so you have to look for the `propertyKey` key in each property.

Something like this should help clean this up. You might want to add some null checks as well.
```suggestion
import { find, getOr } from 'lodash/fp';
...
      const props = response.researcherProperties;
      const authProp = find({'propertyKey':'eraAuthorized'})(props);
      const expProp = find({'propertyKey':'eraExpiration'})(props);
      const isAuthorized = getOr(false,'propertyValue')(authProp);
      const expirationCount = AuthenticateNIH.expirationCount(getOr(0,'propertyValue')(expProp));
```

src/components/eRACommons.js
```diff
@@ -56,17 +54,16 @@ export const eRACommons = hh(class eRACommons extends React.Component {
...
       const nihValid = isAuthorized && expirationCount > 0;
       this.props.onNihStatusUpdate(nihValid);
       this.setState(prev => {
         prev.isAuthorized = isAuthorized;
         prev.expirationCount = expirationCount;
-        prev.nihUsername = _.get(response, 'nihUsername', '');
+        prev.nihUsername = _.get(response.researcherProperties, 'nihUsername', '');
```
rushtong: Same feedback here, we're not parsing the properties correctly.

src/libs/ajax.js
```diff
@@ -1032,17 +1032,18 @@ export const AuthenticateNIH = {
...
-    fireCloudProfileObj.institute = (profile.institution !== undefined && profile.institution !== "") ? profile.institution : "n/a";
+    fireCloudProfileObj.institute = (properties.institution !== undefined && properties.institution !== "") ? properties.institution : "n/a";
     fireCloudProfileObj.institutionalProgram = "n/a";
     fireCloudProfileObj.programLocationCity = "n/a";
     fireCloudProfileObj.programLocationState = "n/a";
     fireCloudProfileObj.programLocationCountry = "n/a";
-    fireCloudProfileObj.pi = (profile.havePi !== undefined && profile.havePi === true)
-      ? profile.piName : profile.isThePI === true
+    fireCloudProfileObj.pi = (properties.havePi !== undefined && properties.havePi === true)
+      ? properties.piName : properties.isThePI === true
```
rushtong: Same feedback here, we aren't processing properties correctly.

rushtong (changes requested):

src/components/eRACommons.js
```diff
@@ -1,10 +1,10 @@
-import _ from 'lodash';
+import {_, isNil} from 'lodash';
```
rushtong: This is causing some compilation problems in the page. I changed this to 
```
import {get, isNil, isUndefined, merge } from 'lodash';
```
and removed all of the `_.` references which allows the page to render.

rushtong (changes requested):
Looks good aside from the lodash issue. See comments inline.

rushtong (approved):
Thank you for the changes!

PR #886, opened 2021-02-18
DUOS-1036[risk=no]UI: Refactor DAC Vote page

raejohanek:
Scope:
- AccessReview: get accessElection from darId instead of voteId, remove voteId
- Routes: update path to no longer include voteId
- NewChairConsole, ChairConsole, MemberConsole, AccessReviewHeader: update references to this page so they no longer include voteId

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1036

Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] PR is labeled with a Jira ticket number and includes a link to the ticket
- [ ] PR is labeled with a security risk modifier [no, low, medium, high] 
- [ ] PR describes scope of changes

In all cases:

- [ ] Get a minimum of one thumbs worth of review, preferably 2 if enough team members are available
- [ ] Get PO sign-off for all non-trivial UI or workflow changes
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #895, opened 2021-02-19
DUOS-641[risk=no]UI: Add support for a Global Banner

raejohanek:
Scope:
Add getActiveBanners to notificationService
Add notificationData state to DuosHeader 
Pull data from API call in componentDidMount
Render notifications using makeNotifications helper 

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-641

In all cases:


src/libs/notificationService.js
```diff
@@ -19,6 +19,18 @@ export const NotificationService = {
...
   },
 
+  /**
+   * Get the raw banner content from GCS
+   * @returns {Promise<JSON>}
+   */
+  getActiveBanners: async () => {
+    const banners = await NotificationService.getBanners();
+    if (!fp.isEmpty(banners)) {
+      return fp.find({active: true})(banners);
```
JVThomas: find() will only return the first element that is true, which can be an issue if there are multiple messages that are active. Use something like map() to create a new array of elements that should be included.

JVThomas (changes requested):

src/libs/notificationService.js
raejohanek: good catch!! thank you!

src/Routes.js
```diff
@@ -92,25 +92,25 @@ const Routes = (props) => (
...
     <AuthenticatedRoute path="/researcher_review/:dacUserId" component={ResearcherReview} props={props} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/access_result_records/:referenceId/:electionId" component={AccessResultRecords} props={props}
-      rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
+                        rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
     <AuthenticatedRoute path="/dul_results_record/:electionId" component={DulResultRecords} props={props}
-      rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.member, USER_ROLES.alumni]} />
-    <AuthenticatedRoute path="/access_review/:darId/:voteId" component={AccessReview} props={props}
-      rolesAllowed={[USER_ROLES.member, USER_ROLES.chairperson]}/>
+                        rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.member, USER_ROLES.alumni]} />
+    <AuthenticatedRoute path="/access_review/:darId" component={AccessReview} props={props}
```
rushtong, line 98: This is from an earlier PR, https://github.com/DataBiosphere/duos-ui/pull/886
I think you need to update your dev branch and merge from that.

rushtong: Never mind ... this is a closed PR.

PR #898, opened 2021-02-22
DUOS-641[risk=no]UI: Add support for a Global Banner

raejohanek:
## Scope:
Add getActiveBanners to notificationService
Add notificationData state to DuosHeader
Pull data from API call in componentDidMount
Render notifications using makeNotifications helper

## Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-641

In all cases:

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -67,7 +67,7 @@ class DuosHeader extends Component {
   makeNotifications = () => {
     const dataList = [];
     const data = this.state.notificationData;
-    //JSON Object, indicating 1 notification only
+    //JSON Object, indicating 1 active notification only
     if (!data.hasOwnProperty('length')) {
       return Notification({notificationData: data});
     } else {
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -24,7 +24,7 @@ class DuosHeader extends Component {
   async componentDidMount() {
     let dacChairPath = await NavigationUtils.dacChairConsolePath();
     this.setState({dacChairPath: dacChairPath});
-    const notificationData =  await NotificationService.getBanners();
+    const notificationData =  await NotificationService.getActiveBanners();
     this.setState(prev => {
       prev.notificationData = notificationData;
       return prev;
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -1,4 +1,4 @@
-import { map } from 'lodash';
+
 import {Component, Fragment, useState} from 'react';
 import {a, button, div, h, hr, img, li, nav, small, span, ul} from 'react-hyperscript-helpers';
 import Drawer from '@material-ui/core/Drawer';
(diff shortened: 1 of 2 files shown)


src/components/DuosHeader.js
```diff
@@ -9,7 +9,9 @@ import {Link, withRouter} from 'react-router-dom';
 import {Storage} from '../libs/storage';
 import {SupportRequestModal} from './modals/SupportRequestModal';
 import './DuosHeader.css';
-import {NavigationUtils} from '../libs/utils';
+import {NavigationUtils} from "../libs/utils";
```
rushtong: nit: revert this change

src/components/DuosHeader.js
```diff
@@ -161,6 +170,17 @@ class DuosHeader extends Component {
...
+  makeNotifications = () => {
+    const dataList = [];
+    const data = this.state.notificationData;
+    for (let i in data) {
+      if (data[parseInt(i)].hasOwnProperty("id")) {
+        dataList.push(data[parseInt(i)]);
+      }
+    }
+    return dataList.map((d) => Notification({notificationData: d}));
+  };
```
rushtong: Whenever you find yourself iterating over lists and using temporary holding variables, there is almost always a cleaner, functional way to do it with filters and maps. For example, I started with this:
```
...
import { get, filter, flow, map } from 'lodash/fp';
...
    const data = this.state.notificationData;
    const dataList = filter(get('id'))(data);
    return map((d) => Notification({notificationData: d}))(dataList);
```
which breaks out the three main concepts in this method,
* get the data
* filter on those with an id
* map those to a new notification

Once we have that, we can simplify even further:
```
    const dataList = filter(get('id'))(this.state.notificationData);
    return map((d) => Notification({notificationData: d}))(dataList);
```
And then reducing again:
```
    return map((d) => Notification({notificationData: d}))(filter(get('id'))(this.state.notificationData));
```

That gets a little hard to read, but it's a one-liner with no temporary variables. Another way to clean this up is to use lodash flow which is a way to chain functions together, one at a time, passing the result of the first to the next, etc., which is what we're doing in that one-liner:
```
    return flow(
      filter(get('id')),
      map((d) => Notification({notificationData: d}))
    )(this.state.notificationData);
```

TBH, I don't think we need the filter line, but I think it's OK for safety.

src/libs/notificationService.js
```diff
@@ -19,6 +19,15 @@ export const NotificationService = {
     });
   },
 
+  /**
+   * Get the raw banner content from GCS
+   * @returns {Promise<JSON>}
+   */
+  getActiveBanners: async () => {
+    const banners = await NotificationService.getBanners();
+    return banners.filter((note) => note.active === true);
```
rushtong: I find lodash's `filter` a lot cleaner to use in these cases.

rushtong (changes requested):
The global banner doesn't display in the new navbar when it is condensed, only in the expanded state. See also comments inline.

rushtong (approved):
Thank you for the changes! The styling alignment looks a little off, but that's outside the scope of this PR. 

Please file a ticket to address the alignment issue:
Screen Shot 2021-02-24 at 7 01 45 AM

Nice work!

PR #901, opened 2021-02-23
DUOS-1045[risk=no]New DAC Chair Navigation Bug

raejohanek:
## Scope:
-route user to correct page depending on config flag
-- in DuosHeader set empty path, assign path in componentDidMount using NavigationUtils to check the flag (I didn't use Navigation.console(user, history) bc it requires the current user which may or may not exist to check the user roles which is not needed as it is done in the header already)

## Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1045

In all cases:

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -116,7 +116,7 @@ class DuosHeader extends Component {
     this.state = {
       showSupportRequestModal: false,
       hover: false,
-      dacChairPath: '/chair_console',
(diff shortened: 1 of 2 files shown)


rushtong (changes requested):
You'll also find a hard-coded reference to the old console page in `DuosHeader.js` that needs to be updated to use the `console` method in this class.

rushtong (approved):
👍🏽

PR #915, opened 2021-03-01
DUOS-1060[risk=no] Upgrade to Babel 7, React 17, and react-scripts 4

JVThomas:
Addresses [DUOS-1060](https://broadworkbench.atlassian.net/browse/DUOS-1060)

PR aims to upgrade react-scripts to 4.0 as a result of a recurring false positive on its ```immer``` dependency that has been solved in version 4.0.4. However this upgrade would, as a result of upgrading from 3.6.x, allow us to upgrade to Babel 7 and React 17 as well. Both upgrades are warranted: the current project structure does not have images within the src folder, meaning that none of the images used in the app are bundled and chunked with webpack. At the same time the current linter, ```babel-eslint``` is not longer supported (repo is now read-only) in favor of the new version ```@babel/eslint-parser```. The linter upgrade is solely needed as ```babel-eslint``` is unable to parse template strings without erroring out (which causes the linter to stop working for that file). 

This PR includes the configuration updates needed to accomplish above goals. This will include moving images out of the public folder and into src, which is most of the PR. As such, css files reference these files via url will need to have that attribute updates, whereas JS files need to utilize imports when referencing images due to their (images) inclusion in the webpack bundle.

The Babel upgrade now has a linter working properly, which is good and bad. It's good in that it works, it's bad because there are a lot of errors (especially in the legacy code), that are now being caught. I've corrected some of the flagrant ones here and will mark them with comments explaining why the code change is necessary. The rest are warnings, however because of the breadth of the image updates (in terms of files updates) Codacy will most likely flag these issues. To keep this PR from becoming too large I have opted not to address these issues here. Otherwise the vast majority of code modifications will be changes to ```src``` and ```url```.

In all cases:


src/pages/ReviewedCases.js
```diff
@@ -52,7 +55,7 @@ class ReviewedCases extends Component {
     let data = this.state.electionsList.dul;
     let sortedData = data.sort(function(a, b) {
 
-      if (!a.hasOwnProperty(sortKey) || !b.hasOwnProperty(sortKey)) {
+      if ((isNil(a) || isNil(a[sortKey])) || (isNil(b) || isNil(b[sortKey]))) {
```
JVThomas, line 58: hasOwnProperty is a function that's used to determine whether or not an object has a property defined on itself. You usually use this in extended classes to determine whether a property was inherited from it's parent class. I have no idea why it's being used here and hasOwnProperty is typically seen as an anti-pattern (now highlighted with the new Babel configuration), hence the adjustment here. Not familiar with ReviewedCases, simply corrected with an equivalent statement, so extra eyes would be appreciated.

.eslintrc
```diff
@@ -1,41 +1,47 @@
 {
-  "extends": "react-app",
-  "parser": "babel-eslint",
+  "extends": ["eslint:recommended", "plugin:react/recommended"],
```
JVThomas, line 2: A lot of the configurations had to be updated due to babel switching over to a monorepo format. I've switched some of the rulesets to warnings (default was error) since the configuration has caught a lot of linting issues in the whole codebase. Erroring out will just prevent the app from running. These rules should be adjusted over time to fit our code standards.

src/App.js
```diff
@@ -7,6 +7,7 @@ import { Config } from './libs/config';
 import DuosFooter from './components/DuosFooter';
 import DuosHeader from './components/DuosHeader';
 import {useHistory} from 'react-router-dom';
+import loadingImage from './images/loading-indicator.svg';
```
JVThomas, line 10: Switching over to imports will allow Webpack to include images in its bundle and adjust internal references as needed.

src/components/AdminConsoleBox.css
```diff
@@ -21,7 +21,7 @@
     display: inline-block;
     width: 35px;
     height: 40px;
-    background: url('/images/admin_icons.png') no-repeat 0 0;
+    background: url('../images/admin_icons.png') no-repeat 0 0;
```
JVThomas, line 24: CSS url ruleset is handled by a webpack package called css-loader, which is managed and configured by react-scripts. As such, relative links still need to be used within css files.

src/components/DuosFooter.js
```diff
@@ -27,7 +28,7 @@ function DuosFooter() {
   return (
     div({ style: footerStyle }, [
       footer({ style: mainFooterStyle }, [
-        img({ src: "/images/broad_logo_allwhite.png", style: footerLogoStyle, alt: "Broad Institute logo" }),
+        img({ src: footerLogo, style: footerLogoStyle, alt: "Broad Institute logo" }),
```
JVThomas, line 31: Within js files we can use import references in place of relative links

src/components/SpinnerComponent.js
```diff
@@ -15,14 +15,10 @@ export class SpinnerComponent extends React.Component {
     }
 
     this.state = {
-      show: this.props.hasOwnProperty('show') ? this.props.show : false
+      show: this.props.show || false
```
JVThomas, line 18: Again, not sure why hasOwnProperty was used on props. There's no reason to do it (unless I'm missing something)

src/components/eRACommons.js
```diff
@@ -120,7 +120,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
       width: 38,
       backgroundRepeat: 'no-repeat',
       backgroundSize: 'contain',
-      backgroundImage: 'url("/images/era-commons-logo.png")',
+      backgroundImage: `url(${eraIcon})`,
```
JVThomas, line 125: Even though we're defining a css rule here, because we're in a js file we'll have to use import references and template strings to get the desired result.

src/components/modals/AddUserModal.js
```diff
@@ -91,16 +92,18 @@ export const AddUserModal = hh(class AddUserModal extends Component {
       roles: this.state.updatedRoles
     };
     switch (this.state.mode) {
-      case 'Add':
+      case 'Add': {
```
JVThomas, line 95: New linting rule in place that enforces block scoping of variables within switch cases. Without them variables defined in each case may be hoisted and made visible to the entire switch block, allowing potential errors due to things such as multiple declarations of variables of the same name across different cases, or accessing variable values before they're even defined due to hoisting.

src/components/modals/SupportRequestModal.js
```diff
@@ -310,48 +311,47 @@ export const SupportRequestModal = hh(
...
-                          alignItems: 'center',
-                          border: this.state.attachment.length === 0 ?
-                            '1px dashed' :
-                            'none',
-                        },
-                      }, [
-                        <div {...getRootProps()}>
-                          <input {...getInputProps()} />
-                          <p>{this.state.attachment.length === 0 ?
+                  h(Dropzone, {
```
JVThomas, line 314: Linter ended up being confused with the mix of jsx syntax and js syntax within this html return (didn't match expected format of array). While the error message wasn't entirely accurate, it still highlighted a syntax concern correctly.

src/libs/ajax.js
```diff
@@ -690,13 +690,11 @@ export const Match = {
   findMatch: async (consentId, purposeId) => {
     const url = `${await Config.getApiUrl()}/match/${consentId}/${purposeId}`;
     const res = await fetchOk(url, Config.authOpts());
-    let answer = {};
     try {
-      answer = await res.json();
-    } catch (error) {
-      answer = {};
-    } finally {
+      const answer = await res.json();
```
JVThomas, line 689: Not sure why the finally block was used here. Try and catch is sufficient to handle the response. Having a finally block will suspend try/catch operations until finally resolves, which can lead to unusual order of operations for async events if not planned properly. Syntax will now be flagged by linter.

src/pages/DataAccessRequestRenewal.js
```diff
@@ -959,35 +961,35 @@ class DataAccessRequestRenewal extends Component {
...
-                      required: true,
-                      disabled: this.state.formData.darCode !== null,
-                    }),
-                    span({
-                      className: 'cancel-color required-field-error-span',
-                      isRendered: fp.isEmpty(this.state.formData.projectTitle) && showValidationMessages,
-                    },
-                    ['Required field']),
-                  ]),
+                  div(
```
JVThomas, line 964: Nothing wrong with this block other than indentation.

src/components/modals/SupportRequestModal.js
JVThomas: Scratch that, there's some styling issues cropping up from this conversion

JVThomas: Found the issue, forgot to add the parenthesis at the end of the getInputProps call, styling is working as expected

JVThomas:
If you're going to test this, you'll need to delete your node-modules and package-lock.json files before running ```npm install```. Odds are you'll need to run ```sudo xcode-select --reset``` as well if ```node-gyp``` fails to install.

rushtong (approved):
👍🏽  Manually tested both local and docker based builds and it works like a charm. Did not test every code path, but did validate most of the admin and researcher ones.

rushtong (approved):
Still 👍🏽

PR #916, opened 2021-03-01
[DUOS-1071][risk=no] Fix DS loading bug

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1071

Possible fix for infinite DS query bug

----

In all cases:

### DIFF
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -735,7 +735,7 @@ class DataAccessRequestApplication extends Component {
 
     let dataUseArray = await Promise.all(datasetPromises);
     dataUseArray.forEach((datasetRecord, index) => {
-      currentDatasets[index].dataUse = datasetRecord.dataUse;
+      currentDatasets[index].dataUse = datasetRecord.dataUse || {};
     });
     return currentDatasets;
   }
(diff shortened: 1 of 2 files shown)


JVThomas:
Found the issue in the code. Certain datasets do not have a dataUse object attached to them (null value), so the code manually fetches the data use object for a dataset via API call and attaches it to the dataset. However, certain datasets will return null, so null ends up being assigned. Dataset gets updated with null, component refreshes, sees that dataset has a null dataUse, cue endless cycle of fetches.

It's been a while since I wrote this code, but I guess I assumed that datasets would always return a dataUse object no matter what since the restrictions are added to the object if it's true (no false representation). However if no such value exists, a null value is returned. So, the simple fix is just default to an empty object if a null value is given.

I'm also questioning if I should even have the API fallback be there at all since it probably just gets the data from the same source anyway. I'll refactor this part of the code as another PR.

JVThomas:
Closing this and making a new DUOS-1071 for team review.

rushtong:
@JVThomas The data use that comes back for a dataset **_should_** be the data use on the consent that backs the data use. There can be the case where a dataset doesn't have a backing consent, which would result in a null value.

PR #917, opened 2021-03-01
DUOS-1071[risk=no] Fix DS loading bug

JVThomas:
Addresses [DUOS-1071](https://broadworkbench.atlassian.net/browse/DUOS-1071)

PR addresses infinite spinner and fetches caused by null assignment on dataUse attribute for specific datasets. The assumption was that a dataUse object was guaranteed, with two calls, one through AsyncSelect, and another with prop update should the AsyncSelect component fail. However, datasets with no restrictions return a null value rather than an empty object. Since the code executes the fetch under the empty object assumption, the code enters an endless cycle of reading a null, fetching, assigning returned value (which is null), repeat. Solution is to simply assign an empty object as a default value if a null is returned.

In all cases:

### DIFF
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -735,7 +735,7 @@ class DataAccessRequestApplication extends Component {
 
     let dataUseArray = await Promise.all(datasetPromises);
     dataUseArray.forEach((datasetRecord, index) => {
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍🏽

JVThomas:
Leaving a note that the npm audit is a false positive (immer dependency for react-scripts).

PR #921, opened 2021-03-02
[DUOS-1075][risk=no] Update docker run instructions

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1075

----

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,5 +53,5 @@ Update your local environment configuration file and mount that to `/usr/share/n
 
 ```
 docker build . -t duos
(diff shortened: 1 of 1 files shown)


raejohanek (commented):
what does this change do ? config/dev -> ${PWD}/path/to/config

rushtong:
@raejohanek 
> what does this change do ? config/dev -> ${PWD}/path/to/config

There is no functional change here, just an example to show that the user would need to indicate the correct path to the correct configuration file. We removed many of the config files in /config so the old content doesn't reflect that any more.

raejohanek (approved):

PR #922, opened 2021-03-02
DUOS-1072[risk=no]Dataset Update Validation Bug

raejohanek:
Scope:
This fixes the "complete all required fields" bug for all datasets, and the "Some errors occurred, the Dataset was not updated." bug that was appearing for a few datasets (447, 454)

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1072

In all cases:

### DIFF
--- a/src/pages/DatasetRegistration.js
+++ b/src/pages/DatasetRegistration.js
@@ -62,7 +62,7 @@ class DatasetRegistration extends Component {
         dac: '',
         consentId: '',
         needsApproval: false,
-        validName: false
+        isValidName: false
       },
       problemSavingRequest: false,
       problemLoadingUpdateDataset: false,
@@ -166,7 +166,7 @@ class DatasetRegistration extends Component {
       prev.datasetData.dac = dac;
       prev.selectedDac = dac;
       let validName = this.validateDatasetName(prev.datasetData.datasetName);
-      prev.datasetData.validName = validName;
+      prev.datasetData.isValidName = validName;
 
       return prev;
     });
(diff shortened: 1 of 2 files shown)


rushtong (changes requested):
I'm still seeing a failure to update, but the error message looks incorrect as the POST operation succeeded. This still needs some debugging.

Screen Shot 2021-03-03 at 8 15 44 AM

src/pages/DatasetRegistration.js
```diff
@@ -62,6 +62,7 @@ class DatasetRegistration extends Component {
         dac: '',
         consentId: '',
         needsApproval: false,
+        validName: false
```
rushtong: nit: rename to `isValidName`

rushtong:
The `onNihStatusUpdate` and `nihValid` state variable look unused to me, we should clean that up in this PR to minimize whatever validation bugs there are.

rushtong:
I think there is an API bug here as well. If we post a dataset with a valid dac id, it doesn't look like that is being honored. When retrieving the dataset again, the dac id is null.

rushtong:
There is at least one API bug. We need to update this code: https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java#L148

by adding this block to update the DAC on the consent
```
        if (Objects.nonNull(dataset.getDacId())) {
            Consent consent = consentDAO.findConsentFromDatasetID(datasetId);
            if (Objects.nonNull(consent)) {
                consentDAO.updateConsentDac(consent.getConsentId(), dataset.getDacId());
            }
        }
```

rushtong:
Have confirmed locally that with the above fix ☝🏽  I can successfully update datasets and have them return with their correct DAC. This also fixes the initial load of the dataset in the case that it does have a DAC. Will file a separate PR for that.

rushtong:
See https://github.com/DataBiosphere/consent/pull/956 for the API fix

raejohanek:
> The `onNihStatusUpdate` and `nihValid` state variable look unused to me, we should clean that up in this PR to minimize whatever validation bugs there are.

the onNihStatusUpdate method is used by eraCommons

rushtong:
@raejohanek 
> the onNihStatusUpdate method is used by eraCommons

`eraCommons` is not on the dataset registration page - I think this might have been a copy/paste issue.

rushtong (approved):
👍🏽

PR #927, opened 2021-03-04
DUOS-1084 Added disabled button functionality to Vote Button in DAC Chair Console

JVThomas:
Addresses [DUOS-1084](https://broadworkbench.atlassian.net/browse/DUOS-1084)

PR adds disabled button functionality to vote button should a vote not exist for the chair. Changes include conditional opacity assignment to styling, disabled button checks on event functions, and helper method/variable/styling refactors. Changes also include associated label changes to Final status and the Vote Button

Will merge after [DUOS-1086](https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=95&projectKey=DUOS&modal=detail&selectedIssue=DUOS-1086) is addressed/merged

In all cases:


rushtong (approved):
Looks great! For future thoughts, we might consider some info text somewhere to indicate why a specific vote button is disabled. We'll need to review the cases and come up with appropriate explanations.

JVThomas:
Requesting re-review. Added small update to correct the labels regarding 'Final' and 'Declined'. UI updates/adjustments regarding election status will be done on another PR.

src/pages/NewChairConsole.js
```diff
@@ -107,12 +107,12 @@ const Records = (props) => {
...
         case 'Open' : {
           const votes = filter({type: 'DAC', dacUserId: currentUserId})(electionInfo.votes);
-          const isFinal = !isEmpty(votes.find((voteData) => !isNil(voteData.vote)));
           return [
             h(TableTextButton, {
               key: `vote-button-${e.referenceId}`,
               onClick: () => props.history.push(`access_review/${dar.referenceId}`),
-              label: isFinal ? 'Final' : 'Vote'
+              label: 'Vote',
+              disabled: isEmpty(votes)
```
raejohanek, line 119: So, when there are no votes, but the election is open we want the Vote button to be disabled? Is that right? Then, how would anyone cast the first vote? And currently, it looks like none of the buttons are disabled even the ones that say 0/10 votes... this looks good to me and it works well, but is this the intended behavior?

JVThomas: The votes should only be disabled if a vote for the user doesn't exist. When elections are made, Vote instances are made for every member, they just have ```vote: null```. When a user submits their vote, the ```vote``` is assigned. The votes array represents the vote instances assigned to the user. If there are no vote instances, you shouldn't be able to use the button to visit the vote page.

Note this situation shouldn't happen, however there are buggy elections where a Chair was never given a vote (maybe they were recently added, maybe it's old holdovers from older elections, not sure on the exact causes), so this is a preventative measure to avoid needless API requests while providing visual feedback (though there'll need to be an explanation on why the button is disabled, haven't figured that out yet).

However I did notice that the filter I made is not correct. I should be checking on two different types: ```DAC``` and ```Chairperson```, so I'll have to make an adjustment on that.

raejohanek: Ahh, that makes a lot of sense thanks!

raejohanek (approved):

rushtong (approved):
With the latest filter change, 👍🏽

PR #928, opened 2021-03-04
DUOS-1086[risk=no]Redo reverted changes: Refactor the vote page on New Chair Console

raejohanek:
Scope:
- remove vote from access review link, leave votes and isFinal as is for the label

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1086

In all cases:

### DIFF
--- a/src/pages/NewChairConsole.js
+++ b/src/pages/NewChairConsole.js
@@ -107,11 +107,10 @@ const Records = (props) => {
         case 'Open' :
           const votes = filter({type: 'DAC', dacUserId: currentUserId})(electionInfo.votes);
           const isFinal = !isEmpty(votes.find((voteData) => !isNil(voteData.vote)));
-          const vote = head(votes);
           return [
             h(TableTextButton, {
               key: `vote-button-${e.referenceId}`,
-              onClick: () => props.history.push(`access_review/${dar.referenceId}/${vote.voteId}`),
+              onClick: () => props.history.push(`access_review/${dar.referenceId}`),
               label: isFinal ? 'Final' : 'Vote'
             }),
             h(TableIconButton, {


src/pages/NewChairConsole.js
```diff
@@ -107,11 +107,10 @@ const Records = (props) => {
         case 'Open' :
           const votes = filter({type: 'DAC', dacUserId: currentUserId})(electionInfo.votes);
           const isFinal = !isEmpty(votes.find((voteData) => !isNil(voteData.vote)));
-          const vote = head(votes);
```
JVThomas, line 110: You can keep this line here, I'm going to need to know if a vote exists for the disabled button styling

JVThomas (changes requested):

src/pages/NewChairConsole.js
raejohanek: Okay, I wonder if we will still need the disabled buttons? Maybe it was only erroring when there were no votes bc it was trying to put the voteid on the end of the link

JVThomas: Actually scratch that, I can just check if the array is empty

rushtong: @JVThomas - I don't see the usage of `vote` in this method, are you planning on it being there for another PR?

JVThomas: The disabled buttons still need to be there because we need to let the user know that the button won't work for them. We shouldn't have to let the figure it out.

JVThomas (approved):
Ignore previous change request, the updates are good. Votes array already filters by userId and by DAC type, so checking for an empty array should be enough to determine disabled status.

src/pages/NewChairConsole.js
JVThomas: The vote isn't needed, just the array, hence the "scratch that" comment. The array is needed for my PR, which is going to add disabled styling to the vote button if the user doesn't have a vote

PR #932, opened 2021-03-05
DUOS-892[risk=no] Unregister serviceWorkers from browsers

JVThomas:
Addresses [DUOS-892](https://broadworkbench.atlassian.net/browse/DUOS-892)

PR aims to address cache issues introduced by service workers used in previous app builds. Service workers are primarily used within Progressive Web Apps (PWAs) to cache assets/files for offline functionality. If not configured properly on initial use, the service worker can prove to be detrimental on application redeploys since it will attempt to restore assets that may or may not exist instead of fetching the files from the server. This is apparent with the recent move of images from the public folder to src (to be included in the webpack bundle). 

The PR proposes this: The application itself will unregister the service worker via function call, causing the browser to stop and remove the worker already registered in the user's browser. However by this point the user has already loaded the page and will still see a cached version. The user will have to hard reload one last time to see the current version of the application. Future visits to the site will not return a cached version via service worker.

Note the above scenario is only an issue for users who have previously visited the DUOS website. New users will never experience the issue since the code registering the service worker will no longer exist.

In all cases:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,12 +10,13 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
 import App from './App';
-import registerServiceWorker from './registerServiceWorker';
+import { unregister } from './registerServiceWorker';
 import { BrowserRouter } from 'react-router-dom';
+
+unregister();
 
 ReactDOM.render(
   <BrowserRouter>
     <App />
   </BrowserRouter>
   , document.getElementById('root'));
-registerServiceWorker();


rushtong:
I'm OK with this approach, but sure wish there were another option. I think the user impact can be managed with some help from customer support. Another thing we could do is to add a temporary banner with a message, "Images not loading? Try command-refresh" and maybe a link to a help page describing what to do. @solideoglori ... your thoughts?

solideoglori:
If this will effect all users, then we should use a banner - ideally one with hard refresh hot keys in it.

If it will only affect a select group, then we should rely on CS help requests.

rushtong (approved):
👍🏽 Let's see how this works with a banner as well.

PR #939, opened 2021-03-10
DUOS-818[risk=no]UI: New DAR results view

raejohanek:
## Scope:
This PR creates the ResultsReview page which replaces the AccessCollect, AccessPreview, and AccessResultRecords pages. This page provides information on DARs to Admin users by re-using the components used on the AccessReview page which provides the same info to Chair users. Those components are AccessReviewHeader, DarApplication, VoteSummary, ApplicantInfo, and AppSummary. So those files all contain minor style changes for compatibility. Additionally, the ReviewedCases and AdminManageDAR pages are slightly modified because they now call the ResultsReview page.

## Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-818

In all cases:


src/pages/access_review/AccessReviewHeader.js
```diff
@@ -16,10 +17,14 @@ const SMALL = {
 };
 
 export const AccessReviewHeader = hh(class AccessReviewHeader extends React.PureComponent {
-
-  openAccessReview = (referenceId) => {
```
raejohanek, line 20: unused

src/pages/access_review/AppSummary.js
```diff
@@ -74,7 +64,7 @@ export const AppSummary = hh(class AppSummary extends React.Component {
   render() {
     const { darInfo, accessElection, consent, researcherProfile } = this.props;
     const { piName, profileName, institution, department, city, country } = researcherProfile;
-    const mrDAR = JSON.stringify(accessElection.useRestriction, null, 2);
+    const mrDAR = !isNil(accessElection) ? JSON.stringify(accessElection.useRestriction, null, 2) : null;
```
raejohanek, line 67: address error: DARs with unreviewed or closed election statuses return a 404 when findElectionById is called, in AppSummary the accessElection is only used for the  mrDAR download link, so jus remove that if it is null

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,108 @@
...
+    const getElectionInformation = async (darId) => {
+      let accessElection;
+      let accessElectionReview;
+      let rpElectionReview;
+
+      try {
+        accessElection = await Election.findElectionByDarId(darId);
+        console.log(accessElection);
+      } catch (error) {
+        //access election is null
```
raejohanek: this is addressing the case for DARs with unreviewed, canceled, or closed elections. catch the error and send an information notification letting the user know why there is no vote data available

src/pages/access_review/AppSummary.js
```diff
@@ -108,7 +97,7 @@ export const AppSummary = hh(class AppSummary extends React.Component {
             },
             [
               StructuredDarRp({darInfo: darInfo}),
-              div({},[
+              div({isRendered: !isNil(mrDAR)},[
```
raejohanek, line 100: DARs with unreviewed, canceled, closed election statuses return a 404 when findElectionById is called (in ReviewResults line 63). So the access Election is null, and in AppSummary the accessElection is only used for the  mrDAR download link, so just remove that link if it is null. Similarly, in DarApplication VoteSummaries are not rendered for the DARs with null accessElection.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,107 @@
...
+import {DAR, DataSet, Election, Researcher} from '../libs/ajax';
+import {Notifications} from '../libs/utils';
+import * as fp from 'lodash/fp';
+import {DarApplication} from "./access_review/DarApplication";
+
+const SECTION = {
+  margin: '20px',
+};
+
+class ReviewResults extends React.PureComponent {
```
rushtong: In terms of construction, we should try to use default functions for all new work for consistency purposes. This will help us isolate older code that extends components so it's clear where we need to do re-writes. See `DataAccessRequest` for an example.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,107 @@
...
+  constructor(props) {
+    super(props);
+    this.props = props;
+  }
+
+  componentDidMount()  {
+    const darId = this.props.match.params.referenceId;
+    this.getData(darId);
+  }
```
rushtong: We should be using hooks to set this data. See https://reactjs.org/docs/hooks-state.html and `DataAccessRequest` for examples. All of the content in `getData` should be collected similarly using hooks.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,107 @@
...
+        [consent, datasets, researcherProfile] = await Promise.all([
+          consentPromise,
+          Promise.all(datasetsPromise),
+          researcherPromise
+        ]);
+
+      } catch (error) {
+        Notifications.showError({text: 'Error initializing DAR Data'});
+        return Promise.reject(error);
+      }
```
rushtong: I've seen similar code in a couple places and I think we need to extract this to a utility function. @JVThomas - please add your thoughts on this.

src/pages/ReviewedCases.js
```diff
@@ -45,7 +45,7 @@ class ReviewedCases extends Component {
       }
     };
     this.getReviewedConsents = this.getReviewedConsents.bind(this);
-    this.openAccessResultRecords = this.openAccessResultRecords.bind(this);
+    this.openAccessReviewResults = this.openAccessReviewResults.bind(this);
     this.openDulResultsRecord = this.openDulResultsRecord.bind(this);
```
rushtong: These lines are legacy and unnecessary code. Both `openAccessReviewResults` and `openDulResultsRecord` are anonymous functions so they don't need to be bound to `this`. `getReviewedConsents` is NOT an anonymous function, so it DOES need to be bound to `this`. Wherever I see non-anonymous functions that are bound in the constructor like this, I generally make them anonymous and remove the bind statements - it's less boilerplate and easier to read, IMHO.

src/pages/access_review/AccessReviewHeader.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { div, img, hh } from 'react-hyperscript-helpers';
 import { Theme } from '../../libs/theme';
 import lockIcon from '../../images/lock-icon.png';
+import * as fp from 'lodash/fp';
```
rushtong: We haven't officially decided as a team what to do here, but I think the consensus is moving towards only importing the specific methods you want to use. In this case, just `isNil`

src/pages/access_review/AccessReviewHeader.js
```diff
@@ -16,10 +17,14 @@ const SMALL = {
...
-    this.props.history.push(`/access_review/${referenceId}`);
-  };
+  constructor(props) {
+    super(props);
+    if (!fp.isNil(props.dacChairMessage)) {
+      this.state = { dacChairMessage: props.dacChairMessage };
+    } else {
+      this.state = { dacChairMessage: "" };
+    }
+  }
```
rushtong: Instead of making this a state variable, you can choose not to render the message at all (which is preferable) based o the existence of the prop. A general principle in React is to avoid setting state when something can be easily calculated from existing props.

rushtong: Secondarily, the naming of this, now that it is extracted into a prop, could be more generic. Something like, `message`

src/pages/access_review/AccessReviewHeader.js
```diff
@@ -55,7 +60,7 @@ export const AccessReviewHeader = hh(class AccessReviewHeader extends React.Pure
               div({ style: SMALL },
                 "Review the Application Summary and Data Use Limitations to determine if the researcher should be granted access to the data."
               ),
-              div({ style: SMALL }, "DAC Chairs can optionally vote as a member.")
+              div({ style: SMALL }, this.state.dacChairMessage )
```
rushtong: This should probably not be rendered at all if the prop isn't there:
```suggestion
              div({ isRendered: !fp.isNil(props.dacChairMessage), style: SMALL }, props.dacChairMessage)
```

src/pages/access_review/AppSummary.js
```diff
@@ -8,17 +8,7 @@ import { StructuredDarRp } from '../../components/StructuredDarRp';
...
-  margin: '32px 0px',
-  color: Theme.palette.primary,
-  opacity: '70%',
-  textTransform: 'uppercase',
-  fontSize: Theme.font.size.small,
-  lineHeight: Theme.font.leading.dense,
-  fontWeight: Theme.font.weight.semibold,
-  fontFamily: 'Montserrat',
-};
+import isNil from "lodash/fp";
```
rushtong, line 11: Nice

rushtong (changes requested):
This isn't a full review, will need to dig in a little more. See comments inline.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,107 @@
...
+    const getElectionInformation = async (darId) => {
+      let accessElection;
+      let accessElectionReview;
+      let rpElectionReview;
+
+      try {
+        accessElection = await Election.findElectionByDarId(darId);
+      } catch (error) {
+        //access election is null
+        Notifications.showInformation({text: 'There is not an active data access election for this DAR.'});
```
rushtong: This alert shows up in the Reviewed Cases link which doesn't make sense - we already expect that there's no active election for the DAR.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,107 @@
...
+
+    render() {
+     if (!fp.isNil(this.state)) {
+      const voteAsChair = true;
+      const {history, match} = this.props;
+      const {darId, darInfo, datasets, consent, researcherProfile, accessElection, accessElectionReview, rpElectionReview } = this.state;
+
+      return div({isRendered: !fp.isNil(this.state), id: 'container', style: {margin: '2rem'}}, [
+        div({id: 'header', style: {...SECTION, padding: '1rem 0'}}, [
+          AccessReviewHeader({history, match})
```
rushtong: `AccessReviewHeader` doesn't seem to use the history prop - that may have been something we deleted at some point and was never removed from existing calls.

rushtong: Ah ... I see it now, you removed it when moving to the constructor approach. Will comment there.

src/pages/access_review/AccessReviewHeader.js
rushtong: The functionality that this added was to ensure that the history was correctly populated. If not necessary anymore, we should remove `history` from all calls.

src/pages/ReviewResults.js
rushtong: For non-open elections (the reviewed cases records), we're not passing in any vote information, so the call to get that data from the open election is not coming back with any data and we don't see anything. We do need a way to get the votes from the most recent election for this DAR so that they do show up correctly, that will be valuable information for any closed election.

rushtong: I think we need a new story in consent to address this. See [describeDataRequestElection](https://github.com/DataBiosphere/consent/blob/7192e7f94308b986dea454d5dda47da5687f3703/src/main/java/org/broadinstitute/consent/http/service/DatabaseElectionAPI.java#L218) where we explicitly request an open election. This method should not be so restrictive and instead return the most recent DAR election if there are any. The status doesn't have to be 'OPEN' for this to work correctly.

raejohanek: Okay, I will make a ticket!

src/pages/ReviewResults.js
raejohanek: I just took it out of ReviewResults and AccessReview and put it in utils

src/libs/utils.js
```diff
@@ -150,6 +151,34 @@ export const NavigationUtils = {
...
+    });
+    const consentPromise = await DAR.getDarConsent(darId);
+    [consent, datasets, researcherProfile] = await Promise.all([
+      consentPromise,
+      Promise.all(datasetsPromise),
+      researcherPromise
+    ]);
+
+  } catch (error) {
+    Notifications.showError({text: 'Error initializing DAR Data'});
```
rushtong: I know this is not your code originally, but the error message seems unhelpful, especially with the acronym. I think this might be more helpful to a user:
```suggestion
    Notifications.showError({text: 'Error retrieving Data Access Request information, please contact support.'});
```

src/pages/access_review/DarApplication.js
```diff
@@ -27,29 +27,32 @@ export const DarApplication = hh(class DarApplication extends React.PureComponen
...
+        }),
+        VoteSummary({
+          isRendered: voteAsChair && !isNil(rpVotes),
+          question: 'Was the research purpose accurately converted to a structured format?',
+          questionNumber: '2',
+          votes: rpVotes,
+        }),
+        AppSummary({darInfo, accessElection, consent, researcherProfile})
+      ])
+      : null;
```
rushtong: I think the null should instead be an empty div.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,95 @@
...
+  const getElectionInformation = async (darId) => {
+    let accessElection;
+    let accessElectionReview;
+    let rpElectionReview;
+
+    try {
+      accessElection = await Election.findElectionByDarId(darId);
+    } catch (error) {
+      //access election is null, this is expected in the case of a closed, unreviewed, or canceled election status
+      //so for these cases there is currently no way to display the vote information
```
rushtong, line 57: Just a note that this comment will be incorrect once we fix the API to return an access election for these election statuses. Might be worth mentioning in the API ticket (when that's created) to come back to this code and show an error message here.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,95 @@
...
+import {isNil} from 'lodash/fp';
+import {DarApplication} from "./access_review/DarApplication";
+
+const SECTION = {
+  margin: '20px',
+};
+
+export default function ReviewResults(props) {
+  const [match] = useState(props.match);
+  const [darId] = useState(props.match.params.referenceId);
```
rushtong: These two things don't have to be in state - they can be plain constants derived directly from the props.

src/pages/ReviewResults.js
```diff
@@ -0,0 +1,95 @@
...
+    setAccessElection(accessElection);
+    setAccessElectionReview(accessElectionReview);
+    setRpElectionReview(rpElectionReview);
+  };
+
+    return (
+      (!isNil(darInfo)) ?
+        div({id: 'container', style: {margin: '2rem'}}, [
+          div({id: 'header', style: {...SECTION, padding: '1rem 0'}}, [
+            AccessReviewHeader({match})
```
rushtong: I'm confused by the `match` argument, should it be a message argument instead?

* tracing this through, `match` comes from `props.match`
* in `AccessReviewHeader`, we only have one usage of `props`: `this.props.message`

rushtong (changes requested):
Thank you for the changes! Please see latest feedback/questions inline.

rushtong (approved):

PR #940, opened 2021-03-11
DUOS-1096[risk=no]UI: DAC Admin Manage DAC Membership

raejohanek:
## Scope
In DuosHeader, make DAC Chair Console a dropdown, with the following choices:
Manage DARs -> New Chair Console Page
Manage DACs -> Manage DAC Page

Chair page 
image

Admin page
image

Compatibility changes:
Rename: AdminManageDac -> ManageDAC
Expand allowed user roles to include chairs
Add optional path param (role: "chair") to determine if the rendering of  the "Add DAC" button is needed

## Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1096

In all cases:


src/Routes.js
```diff
@@ -68,7 +68,7 @@ const Routes = (props) => (
     <Route path="/nih_pilot_info" component={NIHPilotInfo} />
     <AuthenticatedRoute path="/admin_console" component={AdminConsole} props={props} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/admin_manage_users" component={AdminManageUsers} props={props} rolesAllowed={[USER_ROLES.admin]} />
-    <AuthenticatedRoute path="/admin_manage_dac" component={AdminManageDac} props={props} rolesAllowed={[USER_ROLES.admin]} />
+    <AuthenticatedRoute path="/manage_dac/:role?" component={ManageDac} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson]} />
```
rushtong: I don't think you need to pass a role query param here. We know the user's roles from the stored user in session.

src/components/DuosHeader.js
```diff
@@ -216,7 +216,15 @@ class DuosHeader extends Component {
...
         }
+      },
+      chair: {
+      'Manage DARs': {
+        link: this.state.dacChairPath
+      },
+      'Manage DACs': {
+        link: '/manage_dac/chair'
       }
+    }
```
rushtong: nit: indentation

rushtong: Thinking out loud here: we're storing `dacChairPath` in state and it really should be a prop as it should never change. State is meant to hold things that are variable, and the path really is not. This should instead be a constant that is set in a `useEffect` block, but that's beyond the scope of this PR.

src/pages/manage_dac/ManageDac.js
```diff
@@ -189,86 +190,106 @@ class AdminManageDac extends Component {
...
+            div({className: 'col-2 cell-header common-color'}, ['DAC Datasets']),
+            div({className: 'col-2 cell-header f-center common-color'}, ['Actions'])
           ]),
 
-          hr({ className: 'table-head-separator' }),
+          hr({className: 'table-head-separator'}),
 
           this.state.dacs.filter(this.searchTable(searchDacText))
             .slice((currentPage - 1) * limit, currentPage * this.state.limit)
             .map(dac => {
```
rushtong, line 267: In the case of a chairperson on a single dac, this map shows me ALL DACs, not the ones that I'm a chair in. That's because all DACs are visible to all user roles. We need to filter the DACs on this page to the ones the user is a chairperson in, or all of them if they're an admin.

rushtong (changes requested):
The UI looks great. Couple of functional issues here.
1. If I'm not a chair of a DAC, then that DAC should not be visible to me in the manage dac screen. As a side effect of this, I can update users in DACs that I'm not a member of.
2. If I'm only a member in a DAC, I should not be able edit users in my DAC, but should be able to see my DAC and view the members of the DAC. This is possible now in this PR.

src/pages/manage_dac/ManageDac.js
```diff
@@ -58,7 +64,29 @@ class ManageDac extends Component {
...
       }
     );
-  }
+  };
+
+  getUserRole = async () => {
+    let roles;
+    let dacIDs = [];
+    let res;
+    await User.getMe().then(
```
rushtong: You don't need to do this here. When a user logs in, we store their roles as part of their stored user object. See https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/storage.js for where we do this.

rushtong (changes requested):
One suggestion inline. Also, the functional issues I mentioned last time are still an issue:

1. If I'm not a chair of a DAC, then that DAC should not be visible to me in the manage dac screen. As a side effect of this, I can update users in DACs that I'm not a member of.
2. If I'm only a member in a DAC, I should not be able edit users in my DAC, but should be able to see my DAC and view the members of the DAC. This is possible now in this PR.

raejohanek:
> One suggestion inline. Also, the functional issues I mentioned last time are still an issue:
> 
> 1. If I'm not a chair of a DAC, then that DAC should not be visible to me in the manage dac screen. As a side effect of this, I can update users in DACs that I'm not a member of.
> 2. If I'm only a member in a DAC, I should not be able edit users in my DAC, but should be able to see my DAC and view the members of the DAC. This is possible now in this PR.

 Okay so for 1, when I log in as a chair, I am a part of the Dac with dacID=3 and I only see that DAC. When I log in as an admin, I see all DACs, See the pics I just put in the PR description. I wrote this so that if you have both Admin and Chair roles assigned to your account, then the page will render with all DACs, maybe that is why you are seeing all?

Update: as I keep refreshing this page, I see that if I refresh it quick enough it does show all DACs, not just the one I am a chair on, I think maybe this is happening bc it is showing up before the user roles are assigned so I'll change render to wait for the user role value 

For 2, if you are only a member in a DAC you cannot navigate to this page (unless you are also an admin or chair). So this page is only accessible through the DAC Chair console, and the Admin Console. So currently, no members can see their dacs and view their members using this page. Also, on the manage dac page I filter to make sure the user role is chair and then add that dacid to the list of dacs to include in the list, so hopefully no dacs that the user is just a member of would be included.

src/pages/manage_dac/ManageDac.js
raejohanek: so, maybe you will know why this is happening, but I tried to use the Storage.getCurrentUserRole() method originally, but combined with the .then() I get this error: 
image
So I didn't know how to fix that, that is why I switched over to the User.getMe() call, and then just use result.roles instead of result

rushtong:
@raejohanek - Disregard the functionality issues I mentioned - I had not done a `git pull` to get your latest changes, my bad. I am seeing other issues, but those are not them.

rushtong:
I am seeing a 403 when trying to update DAC membership as a chairperson. It is likely the API for `PUT /api/dac` has not been updated with a chairperson role yet.

src/pages/manage_dac/ManageDac.js
rushtong: This code will do what you're looking to do without the /me call and doesn't trigger the error you see:
```
    const currentUser = Storage.getCurrentUser();
    const roles = currentUser.roles;
    let dacIDs = [];
    roles.forEach((role) => {
      if (role.name === CHAIR) {
        dacIDs.push(role.dacId);
      }
    });
```

rushtong (changes requested):
Looks and functions great! Just one suggestion to clean up an unnecessary api call.

src/pages/manage_dac/ManageDac.js
rushtong: That being said, the loop can be cleaned up by filtering the roles on name, then mapping to return a list of dacIds.

raejohanek: good idea!!!

src/pages/manage_dac/ManageDac.js
```diff
@@ -57,7 +70,18 @@ class AdminManageDac extends Component {
...
       }
     );
-  }
+  };
+
+  getUserRole = async () => {
+    const currentUser = Storage.getCurrentUser();
+    const roles = currentUser.roles.map(r => r.name);
+    const role = contains(ADMIN)(roles) ? ADMIN : CHAIR;
+    const dacIDs = currentUser.roles.filter(r => r.name === CHAIR).map(role => role.dacId);
```
rushtong: I find lodash/fp's filter much more legible ... `filter(name: CHAIR)(currentUser.roles)`
Same with lodash/fp map: `map('dacId')(currentUser.roles)`

rushtong (approved):
Looks good! Feel free to merge when ready 👍🏽

PR #947, opened 2021-03-16
DUOS-1055[risk=no]UI: Admin Manage Access & Member Console - link DAR ID to modal summary

raejohanek:
Addresses: https://broadworkbench.atlassian.net/browse/DUOS-1055

Scope: 
Admin manage access: remove info column, link modal to dar ID, update styling
member console: link modal to darId, updateStyling
utils: refactor reused code
darmodal: accomodate the differerent dataset structures, some called label some called name

In all cases:


src/pages/ReviewResults.js
```diff
@@ -22,11 +22,8 @@ export default function ReviewResults(props) {
   const [voteAsChair] = useState(true);
 
   useEffect(() => {
-    let darData;
-    let electionData;
-
     const getData = async (darId) => {
-      [darData, electionData] = await Promise.all([
```
raejohanek, line 29: unused

src/pages/NewChairConsole.js
```diff
@@ -49,23 +49,6 @@ const processElectionStatus = (election, votes) => {
...
-};
-
-const applyTextHover = (e) => {
-  e.target.style.color = Styles.TABLE.DAR_TEXT_HOVER.color;
-  e.target.style.cursor = Styles.TABLE.DAR_TEXT_HOVER.cursor;
-};
-
-const removeTextHover = (e, color) => {
-  e.target.style.color = color;
-};
```
raejohanek, line 67: these are all reused so I put them in utils

src/components/modals/DarModal.js
```diff
@@ -43,71 +43,77 @@ const processResearchTypes = (researchTypes) => {
 };
 
 const requiresManualReview = (darDetails) => {
-  return darDetails.illegalBehavior ||
```
raejohanek, line 46: dars sent from member and admin have a requiresManuelReiew property and none of the properties listed here, add a check so it works for bth dars from chairs and dars from admins and members

src/components/modals/DarModal.js
```diff
@@ -43,71 +43,77 @@ const processResearchTypes = (researchTypes) => {
...
-    darDetails.vulnerablePopulation;
+    darDetails.vulnerablePopulation)
+    : darDetails.requiresManualReview;
 };
 
 const DarModal = (props) => {
   //NOTE: Modal should be simple (raw information should be passed in as props) in order to ensure plug and play use
   const {showModal, closeModal, darDetails, researcher} = props;
 
-  return h(Modal, {
```
raejohanek, line 58: the only change in this return is the addition of the if statement

src/pages/AdminManageAccess.js
```diff
@@ -29,17 +30,13 @@ class AdminManageAccess extends Component {
       showDialogCreate: false,
       dacList: []
     };
-    this.getElectionDarList = this.getElectionDarList.bind(this);
```
raejohanek, line 32: made anonymous

src/pages/AdminManageAccess.js
```diff
@@ -216,98 +215,91 @@ class AdminManageAccess extends Component {
...
+                          className: "glyphicon glyphicon-hand-right hover-color",
+                          isRendered: dar.status === 'pending',
+                          "data-tip": "Researcher review pending",
+                          "data-for": "tip_pendingReview"
+                        }),
+                        span({
+                          className: "glyphicon glyphicon-hand-right dismiss-color",
+                          isRendered: dar.status === null
+                        }),
+                      ])
```
raejohanek: lines 226-296 feature no functional changes just formatting to address incorrect indentation warnings

src/libs/utils.js
```diff
@@ -46,6 +47,23 @@ export const sleep = (milliseconds) => {
   return new Promise(resolve => setTimeout(resolve, milliseconds));
 };
 
+export const getDatasetNames = (datasets) => {
+  if(!datasets){return '';}
+  const datasetNames = datasets.map((dataset) => {
+    return ((dataset.label) ?  dataset.label : dataset.name);
```
raejohanek, line 54: for some reason the datasets from chair console datasets have names stored as 'label' while datasets from admin and member have names stored as 'name'

rushtong: Could you dig into this a little more? In other words, there had to be a case where `name` was correct. What's different about the `datasets` argument from the other calls to this method? The source of the bug might be in the caller, not here.

raejohanek: so when name is present, name is always correct, when label is present, label is always correct. The datasets returned from the api call DAR.getDataAccessManageV2() have labels while the datasets returned by DAR.getDarModalSummary(dar.dataRequestId) have names, I'm not sure this is so much a bug as it is a discrepancy, my first attempt was to get all the pages using the same api calls since they all display similar info, but because the pages are all structured so differently this required to changes to every single component used on the pages and I ended up scrapping that approach.

raejohanek: so when name is present, name is always correct, when label is present, label is always correct. The datasets returned from the api call DAR.getDataAccessManageV2() have labels while the datasets returned by DAR.getDarModalSummary(dar.dataRequestId) have names, I'm not sure this is so much a bug as it is a discrepancy, my first attempt was to get all the pages using the same api calls since they all display similar info, but because the pages are all structured so differently this required to changes to every single component used on the pages and I ended up scrapping that approach.

src/components/modals/DarModal.js
```diff
@@ -43,71 +43,77 @@ const processResearchTypes = (researchTypes) => {
...
-    isOpen: showModal,
-    onRequestClose: closeModal,
-    shouldCloseOnOverlayClick: true,
-    style: {
-      content: { ...Styles.MODAL.CONTENT },
-      overlay: {
-        backgroundColor: 'rgba(0, 0, 0, 0.5)'
+  if (!isNil(darDetails)) {
```
JVThomas: Still going through the PR, here's a comment in the meantime...

Instead of using an if block here, use the ```isRendered``` conditional on ```DarModal```. This way the component won't render if the conditional evaluates to false.

Example: ```h(DarModal, {isRendered: !isNil(darDetails), ...```

Functional components in React are essentially render functions, so while you're not outputting anything, the component still initializes when it didn't have to.

That said, it could be confusing for a user to click the DAR name and not see a modal pop up when they were expecting it. Another possible approach is to output a simple modal that has a error statement like "DAR details were not provided" or something along those lines. Feel free to pick the option that best fits the situation.

src/libs/utils.js
rushtong: Ah, thank you - that makes sense.
1. `DAR.getDarModalSummary(dar.dataRequestId)` should be avoided and deprecated as soon as possible. The API is not reliable as it uses DTO style construction pulling from a variety of objects.
2. `DAR.getDataAccessManageV2()` is preferred as that gets the full objects instead of manually constructed value objects.

src/pages/AdminManageAccess.js
```diff
@@ -177,8 +162,23 @@ class AdminManageAccess extends Component {
...
 
   render() {
-    const { searchDarText, currentPage, limit } = this.state;
+    const { showModal, searchDarText, currentPage, limit, darDetails, researcher } = this.state;
     const pageCount = Math.ceil((this.state.darElectionList.filter(this.searchTable(searchDarText)).filter(row => !row.isCanceled).length).toFixed(1) / limit);
+    const handleCloseModal = () => {
+      this.setState({ showModal: false });
+    };
+    const openSummaryModal = async (dar) => {
+      let darDetails = await DAR.getDarModalSummary(dar.dataRequestId);
```
rushtong: Should use `DAR.describeDar()` instead - that will give you all of data necessary. The modal summary API should be avoided.

src/pages/MemberConsole.js
```diff
@@ -104,73 +104,91 @@ class MemberConsole extends Component {
 
   render() {
 
-    const { searchDarText } = this.state;
+    const { searchDarText, showModal, darDetails, researcher } = this.state;
     const pageCount = Math.ceil((this.state.electionsList.access.filter(this.searchTable(searchDarText)).length).toFixed(1) / (this.state.accessLimit));
-
+    const closeSummaryModal = () => this.setState({ showModal: false });
+    const openSummaryModal = async (dar) => {
+      let darDetails = await DAR.getDarModalSummary(dar.referenceId);
```
rushtong: Same comment as above.

rushtong (changes requested):
See comments inline, we should be getting the DAR info from the actual DAR instead of the value object returned in modal summary.

src/pages/AdminManageAccess.js
raejohanek: so DAR.describeDar() has a list of dataset ids, but it does not have a list of datasets or a list of dataset names, like dar.getDarModalSummary has. We need the dataset names for the modal, so if we do want to switch describeDar then we would need an extra step to get dataset by ID... I am thinking this would work  well if we move the getDatasetNames function there to the DarModal

raejohanek: or if we want to change the whole page over to the V2 API call then we wouldn't even need an extra API call in openSummary cause we'd already have all the info, but that would be a big change, not sure what you think?

rushtong: You can get the datasets from the datasets API.

rushtong (approved):
Thank you for the changes!

PR #948, opened 2021-03-17
DUOS-1097[risk=no]User Session Timeout is not redirecting to home

raejohanek:
Scope: 
- Wrap all axios methods (get, put, and post) if the call returns a 401 logging out the user and returning to the home page
- in fetchAny, check for 401 status and log out the user/return to the home page

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1097

In all cases:

### DIFF
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -33,6 +33,7 @@ export const SignIn = hh(class SignIn extends Component {
 
   responseGoogle = async (response) => {
     Storage.setGoogleData(response);
+    console.log(response);
     try{
       const userRes = await this.getUser();
       const user = Object.assign(userRes, setUserRoleStatuses(userRes, Storage));
@@ -45,7 +46,7 @@ export const SignIn = hh(class SignIn extends Component {
         this.props.history.push('/profile');
         setTimeout(() => {
           Storage.clearStorage();
-        }, 3000);
+        }, 3599);
       } catch(error) {
         try{
           const status = error.status;
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -43,6 +43,9 @@ export const SignIn = hh(class SignIn extends Component {
         this.setUserRoleStatuses(registeredUser, Storage);
         this.props.onSignIn(); //is this async?
         this.props.history.push('/profile');
+        setTimeout(() => {
+          Storage.clearStorage();
+        }, 3000);
       } catch(error) {
         try{
           const status = error.status;
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -25,15 +25,14 @@ export const SignIn = hh(class SignIn extends Component {
       prev.clientId = clientKey;
       return prev;
     });
-  }
+  };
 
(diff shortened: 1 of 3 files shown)


src/components/SignIn.js
```diff
@@ -25,11 +25,11 @@ export const SignIn = hh(class SignIn extends Component {
       prev.clientId = clientKey;
       return prev;
     });
-  }
+  };
```
raejohanek, line 28: address compilation warnings

JVThomas (changes requested):
You shouldn't have to change any of the axios methods at all for this fix. The only component you need to change is ```AuthenticatedRoute.js```, especially now that the backend has a ```/me``` endpoint that you can use as a safe way to check is a session is active on the backend.

The idea is to redirect before the component even loads. The changes here would only happen after the fact.

JVThomas:
As a starting point, look at ```Routes.js``` and ```AuthenticatedRoutes.js``` to get a high-level view of what the latter is trying to accomplish.

rushtong:
One of the effects we're seeing is that sessions are timing out after the component loads, and when the user follows a rendered link. Those cases are causing 401's that aren't getting caught.

JVThomas:
> One of the effects we're seeing is that sessions are timing out after the component loads, and when the user follows a rendered link. Those cases are causing 401's that aren't getting caught.

Oh, in that case look into axios interceptors
https://github.com/axios/axios#interceptors

Either way, you don't need a wrapper on every axios call. More likely than not third-party packages will have some sort of generic function or attribute that will allow you to define some sort of pre-processing step.

src/libs/ajax.js
```diff
@@ -1101,6 +1114,10 @@ const fetchAny = async (...args) => {
   if (res.status >= 500) {
     await reportError(args[0], res.status);
   }
+  if (!res.ok && res.status === 401) {
```
raejohanek, line 1117: Update fetchAny to include the same 401 check fetchOk includes

src/libs/ajax.js
```diff
@@ -11,6 +11,19 @@ import axios from 'axios';
 import {DataUseTranslation} from './dataUseTranslation';
 import {isFileEmpty} from './utils';
 
+//define axios interceptor
+//to log out user and redirect to home when response has 401 status
+//return responses with statuses in the 200s and reject the rest
+axios.interceptors.response.use(function (response) {
+  return response;
+}, function (error) {
+  if (error.status === 401) {
```
JVThomas: To get the status on an error, you'll need to do ```error.response.status```

JVThomas (changes requested):

src/libs/ajax.js
JVThomas: It's specific to axios, which is probably why you handle errors differently here compared to fetchOk and fetchAny.

https://github.com/axios/axios#handling-errors

JVThomas (approved):
Code looks good. This should cover instances where consent requests happen after a component has loaded and after a user session has timed out.

PR #950, opened 2021-03-18
DUOS-1118[risk=no]Data Sharing Language Tool

raejohanek:
SCOPE:
-Create data sharing language tool (DataSharingLanguageTool.js) which uses DataUseTranslation.translateDataUsefor its functionality
-Take reused method out of DatasetRegistration and put into utils
-Add page to routes
-Add page to overview section of Homepage
-Edit RadioButton to be flexible to a different style

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1118

In all cases:


src/pages/DatasetRegistration.js
```diff
@@ -72,7 +74,7 @@ class DatasetRegistration extends Component {
 
     this.handleOpenModal = this.handleOpenModal.bind(this);
     this.handleCloseModal = this.handleCloseModal.bind(this);
-  };
```
raejohanek: address compilation warnings

src/Routes.js
```diff
@@ -85,6 +86,7 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/signing_official_console" component={SigningOfficialConsole} props={props} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/dataset_registration/:datasetId" component={DatasetRegistration} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson]} />
     <AuthenticatedRoute path="/dataset_registration" component={DatasetRegistration} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson]} />
+    <AuthenticatedRoute path="/data_sharing_language_tool" component={DataSharingLanguageTool} props={props} rolesAllowed={[USER_ROLES.all]} />
```
rushtong: I didn't think this needed to be authenticated. Since it is linked from the overview, and that is also unauthed, I think we should move this to just a plain `Route`

src/components/RadioButton.js
```diff
@@ -7,7 +7,7 @@ export const RadioButton = (props) => {
     fontSize: 15,
     lineHeight: '2rem',
     color: 333,
-    fontFamily: '\'Roboto\', sans-serif',
+    fontFamily: props.sdsl ? 'Montserrat' : '\'Roboto\', sans-serif',
```
rushtong: This pattern presupposes that there will ever only be a single conditional for the font family, which is a limiting pattern. Instead, what we should do is extract this style block into a separate object. The component would then be able to accept additional props that would be enhancements to that style object and merge them. By default, there'd be no merge, and in your case, you'd pass in a style object with the font family you want so that gets merged over the default style. This approach makes the component flexible enough to be modified by any other parent component without having to add even more ternary logic to individual font settings.

src/components/RadioButton.js
```diff
@@ -36,7 +36,7 @@ export const RadioButton = (props) => {
 
   const labelStyle = {
     cursor: props.disabled ? 'not-allowed' : 'pointer',
-    color: '#603B9B',
+    color: props.sdsl ? '#1f3b50' : '#603B9B',
```
rushtong: Same pattern comment here.

src/Routes.js
rushtong: In testing, I can confirm that the link is visible when not authenticated and following it fails when in that state.

src/pages/DataSharingLanguageTool.js
```diff
@@ -0,0 +1,232 @@
...
+        ]),
+      ]),
+
+      div({className: 'form-group', style: {marginTop: '2rem'}}, [
+        label({style: Styles.MEDIUM}, [
+          '2. Choose any additional constraints you need to put on future uses of your data', br(),
+          span({style: Styles.MEDIUM_DESCRIPTION}, ['Then if necessary, you may choose additional terms on your study\'s data to govern it\'s use by adding requirements or limitations.']),
+        ]),
+        div({}, [
+          RadioButton({
```
rushtong: I believe all section 2 questions should be checkboxes. Using radio buttons to simulate checkboxes is an anti-pattern: https://www.nngroup.com/articles/checkboxes-vs-radio-buttons/

src/pages/Home.js
```diff
@@ -215,7 +215,12 @@ class Home extends Component {
                       a({
                         href: 'https://www.ga4gh.org/genomic-data-toolkit/regulatory-ethics-toolkit/#:~:text=Machine%20Readable%20Consent%20Guidance&text=Machine%20readable%20consent%20language%20is,to%20for%20their%20research%20purposes',
                         target: '_blank'
-                      }, ['Machine Readable Consent Guidance.'])
+                      }, ['Machine Readable Consent Guidance.']),
+                      " For help determining your data's permitted uses, try our ",
+                      a({
+                        href: '/data_sharing_language_tool',
+                      }, ['Data Sharing Language Tool']),
```
rushtong: This should be a `Link` component.

src/pages/DataSharingLanguageTool.js
```diff
@@ -0,0 +1,232 @@
...
+          span({style: Styles.MEDIUM_DESCRIPTION}, ['First, you must determine what type of secondary use is permitted for you study\'s data.' +
+          ' You do this by selecting one of the options in the following section:']),
+        ]),
+        div({}, [
+          RadioButton({
+            value: 'general',
+            defaultChecked: general,
+            onClick: () => {
+              setGeneral(!general), setHmb(false), setDiseases(false), setOther(false);
+            },
```
rushtong: The `setDiseases` method doesn't clear out existing selections (it only modifies state) so the ontology text box doesn't clear if you enter something and then select something else.

rushtong: Same feedback for `setOther`

src/pages/DataSharingLanguageTool.js
```diff
@@ -0,0 +1,232 @@
...
+        "This tool is made publicly available by the DUOS team for anyone" +
+        " interested in leveraging standardized data sharing language in their" +
+        " consent forms. The tool leverages the Global Alliance for Genomics " +
+        "and Health’s (GA4GH) companion standards of the Data Use Ontology (DUO) " +
+        "and Machine Readable Consent Guidance (MRCG). The DUO is a structured " +
+        "vocabulary describing permitted data uses and the MRCG is a suggested " +
+        "representation of those uses in consent form language. This tool enables" +
+        " users to easily define what types of data use they would like permitted" +
+        " in their consent forms and then suggests corresponding text for the " +
+        "consent form below, based on the MRCG."
```
rushtong, line 76: Might be outside the scope of the story, but it would be nice to have links here to the standards that we're describing. Let's follow up with @solideoglori about adding some link content here.

rushtong (changes requested):
Page looks good! See functional and stylistic feedback inline.

src/pages/DataSharingLanguageTool.js
raejohanek: I believe that info in in the MRCG link!

src/pages/Home.js
raejohanek: So I just changed it to a Link component and I can't seem to get the link to work, suggestions?

src/pages/DataSharingLanguageTool.js
```diff
@@ -0,0 +1,298 @@
...
+      : Notifications.showError({text: "Please complete question 1"});
+  };
+
+  const generateHelper = () => {
+    const darInfo = {
+      general: general, diseases: diseases, ontologies: ontologies, other: other, otherText: otherText,
+      hmb: hmb, methods: !nmds, forProfit: !npu
+    };
+    const summaries = [];
+    const dataUse = (DataUseTranslation.translateDarInfo(darInfo));
```
raejohanek: @rushtong DataUseTranslation.translateDarInfo(darInfo) is not a good fit for the functionality of this page. It returns inaccurate and often conflicting language. it does not include any checks/language for most of the information checkboxes, as you can see in the darInfo constant it only includes methods which is the opposite of NMDS and forProfit which is the opposite of NPU. Meaning it is pointless to have all of the other checkboxes currently. These opposite properties cause issues, for example if someone doesn't check NPU they get a statement that might not be true: "The dataset will be used in a study related to a commercial purpose" similarly if they don't check NMDS then they get this statement: "The primary purpose of the research is to develop and/or validate new methods for analyzing or interpreting data." which also might not be true and it conflicts with the first sentence returned that establishes the primary purpose to be either general, hmb, or disease related. Instead we should only use this method to generate the one sentence on primary purpose, and a different method to generate the sentences for all of the checkbox information. There already exists the constant consentTranslations in dataUsetranslation that has a good sentence description of all of the information from the checkboxes, so I think we should use that or something similar to generate the secondary use restrictions

raejohanek: I outlined the approach I am proposing in the code, what do you think?

rushtong: > DataUseTranslation.translateDarInfo(darInfo) is not a good fit for the functionality of this page. It returns inaccurate and often conflicting language.

If the above is true, then `translateDarInfo` has some serious problems. I'll look at this tomorrow. We should be using the exact same method for generating the language as that is the idea behind this entire page. If this doesn't generate the same language with the same questions, then it doesn't meet the requirements. The current functionality was approved by Jonathan, so we might need to loop him into this conversation.

src/components/RadioButton.js
```diff
@@ -7,13 +7,13 @@ export const RadioButton = (props) => {
     fontSize: 15,
     lineHeight: '2rem',
     color: 333,
-    fontFamily: '\'Roboto\', sans-serif',
+    fontFamily: props.font ? props.font : '\'Roboto\', sans-serif',
```
rushtong: This approach is very similar to the previous approach. To make RadioButton more flexible, allow it to take in a style object that can be merged into its own style. That way it is generic. With the current approach, we'd need to add new props for each style prop that might need to change. With the object approach, it is flexible enough to handle any new style without the need for modification.

src/pages/DataSharingLanguageTool.js
rushtong: What we should be using is `DataUseTranslation.translateDataUseRestrictions` since this should be passing in a `dataUse` object that describes the constructed `dataset`. The reason we're getting confusing results is that using `translateDarInfo` is meant for Data Access Requests, not datasets.

src/libs/dataUseTranslation.js
```diff
@@ -167,6 +167,32 @@ const consentTranslations = {
...
+    result.push(consentTranslations.collaboratorRequired.description);
+  }
+  if (restrictions.includes("IRB")) {
+    result.push(consentTranslations.ethicsApprovalRequired.description);
+  }
+  if (restrictions.includes("GS-")) {
+    result.push(consentTranslations.geographicalRestrictions.description);
+  }
+  return result;
+};
```
rushtong: I don't think we need this method. What we need to do is construct a valid `dataUse` and pass that into `translateDataUseRestrictions`. I should have caught this earlier, but didn't.

rushtong (changes requested):
See comments inline

src/libs/dataUseTranslation.js
raejohanek: okay, thank you!

src/pages/DataSharingLanguageTool.js
raejohanek: Just changed it to translate data use restrictions and the generated language is much better, more comprehensive and clear! I had to make a minor change to the method to accommodate the ontologies given, but it will not affect the functionality of the method elsewhere

src/pages/DataSharingLanguageTool.js
```diff
@@ -0,0 +1,299 @@
...
+              onChange: (e) => setMor(e.target.checked),
+              id: 'checkMoratorium',
+              type: 'checkbox',
+            }),
+            label({
+              style: labelStyle,
+              className: 'regular-checkbox',
+              htmlFor: 'checkMoratorium',
+            }, ['Publication Moratorium (MOR)']),
+          ]),
```
raejohanek: this checkbox is pointless currently because the translateDataRestriction method will never generate any language for this property, so we might as well remove it. Additionally, translateDataRestriction checks for and generates language for POA property, POA is intentionally not included in the DSLT but this may cause some discrepancies between the language generated here and the language generated in actual practice

rushtong: We have not decided on any language for this particular dataset restriction, so yes, there is currently no effect to selecting it. I think the right course of action should be to remove it for this PR, and file a new ticket to add it back in alongside the required language for this option. Let's follow up with @solideoglori for his 👍🏽

src/pages/Home.js
```diff
@@ -215,8 +216,11 @@ class Home extends Component {
                       a({
                         href: 'https://www.ga4gh.org/genomic-data-toolkit/regulatory-ethics-toolkit/#:~:text=Machine%20Readable%20Consent%20Guidance&text=Machine%20readable%20consent%20language%20is,to%20for%20their%20research%20purposes',
                         target: '_blank'
-                      }, ['Machine Readable Consent Guidance.'])
-                    ]),
+                      }, ['Machine Readable Consent Guidance.']),
+                      " For help determining your data's permitted uses, try our ",
+                      (Link, { to: '/data_sharing_language_tool', className: 'hover-color'}, ['Data Sharing Language Tool']),
```
rushtong: Link needs to be wrapped to use this construction syntax (see https://www.npmjs.com/package/react-hyperscript-helpers) 
```
h(Link, { to: '/data_sharing_language_tool', className: 'hover-color'}, ['Data Sharing Language Tool'])
```

rushtong (approved):
Once the link is fixed, 👍🏽

PR #955, opened 2021-03-19
DUOS-1088[risk=no] Update style on new DAC Chair, Member, Admin tables

raejohanek:
Scope:
Update title on NewChairConsole
Update title on MemberConsole
Add border to table container (surrounding header and pagination bar)
Add rounded edges

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1088
Update style on new DAC Chair, Member, Admin tables

In all cases:

### DIFF
--- a/src/libs/theme.js
+++ b/src/libs/theme.js
@@ -121,7 +121,9 @@ export const Styles = {
       backgroundColor: "#f3f6f7",
       display: "flex",
       justifyContent: "left",
-      height: "51px"
+      height: "51px",
+      borderTopLeftRadius: '10px',
+      borderTopRightRadius: '10px'
     },
     RECORD_ROW: {
       fontFamily: 'Montserrat',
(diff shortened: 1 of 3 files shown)


JVThomas (changes requested):
Screen Shot 2021-03-22 at 7 53 59 AM

You've added the border, but the background color of the table header and footer is poking through the corners. This is because the background colors are being defined in ```TABLE.HEADER_ROW``` and ```TABLE.FOOTER```. 

However you can't use ```borderRadius``` since it would change all four corners of the header and footer. Instead you'll have to manually define the radius on the necessary corners (dependent on which section you're working on). You'll need to use attributes like ```borderTopLeftRadius``` to handle the styling.

Here's a reference article on MDN
https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-left-radius

src/libs/theme.js
```diff
@@ -119,7 +121,9 @@ export const Styles = {
       backgroundColor: "#f3f6f7",
       display: "flex",
       justifyContent: "left",
-      height: "51px"
+      height: "51px",
+      borderTopLeftRadius: '10px',
```
JVThomas: Instead of defining the pixel value, you can use the term ```inherit```. This way the radius value will be whatever is defined on the parent, which in this case should be the Container div (where the border is drawn).

JVThomas (changes requested):

JVThomas (approved):

PR #964, opened 2021-03-25
DUOS-1132[risk=no]Update question 1.3 on Dar Application page

raejohanek:
SCOPE:
When the profile is not filed out at all:
This box is disabled while simultaneously displaying the error that the DAR application cannot be submitted because it is a required field. Update the error message to direct the user to their profile. 

When the profile is filled out and the user is not a PI and does not have a PI: 
the box is pre-populated with ‘--’ allowing it to pass validation checks and the DAR is able to be submitted when it shouldn’t be. Remove the placeholder so the question errors and does not allow submission.

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1132

In all cases:

### DIFF
--- a/src/pages/dar_application/ResearcherInfo.js
+++ b/src/pages/dar_application/ResearcherInfo.js
@@ -228,7 +228,7 @@ export default function ResearcherInfo(props) {
               }),
               span({
                 className: 'cancel-color required-field-error-span', isRendered: (invalidInvestigator) && (showValidationMessages)
-              }, ['Required field, ensure your Profile has a PI listed or marks you as the PI'])
+              }, ['Required field, ensure your Profile has a PI listed, or that you are selected as the PI'])
             ])
           ])
         ]),
--- a/src/pages/dar_application/ResearcherInfo.js
+++ b/src/pages/dar_application/ResearcherInfo.js
@@ -228,7 +228,7 @@ export default function ResearcherInfo(props) {
               }),
               span({
                 className: 'cancel-color required-field-error-span', isRendered: (invalidInvestigator) && (showValidationMessages)
-              }, ['Required field'])
+              }, ['Required field, your Profile must have a PI'])
             ])
           ])
         ]),
--- a/src/pages/dar_application/ResearcherInfo.js
+++ b/src/pages/dar_application/ResearcherInfo.js
@@ -82,7 +82,7 @@ export default function ResearcherInfo(props) {
           name: 'anvil-use-approval-status',
           checked: option.value === anvilUse,
           value: option.value,
-          onChange: (e) => (formFieldChange({name: 'anvilUse', value: option.value}))
+          onChange: () => (formFieldChange({name: 'anvilUse', value: option.value}))
         }),
         span({ className: "radio-check"}),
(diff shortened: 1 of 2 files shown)


src/pages/dar_application/ResearcherInfo.js
```diff
@@ -228,7 +228,7 @@ export default function ResearcherInfo(props) {
               }),
               span({
                 className: 'cancel-color required-field-error-span', isRendered: (invalidInvestigator) && (showValidationMessages)
-              }, ['Required field'])
+              }, ['Required field, your Profile must have a PI'])
```
rushtong: This, I think, is not technically true. A researcher can also BE the PI and in that case, not need to have a separate PI listed. In testing, it looks like this validation isn't working. I can turn off my PI status and have not PI listed:

----
Screen Shot 2021-03-26 at 9 40 06 AM

----
HOWEVER, I can still submit a DAR this way, so I don't think this message shows up. The PI field isn't selectable based on the user profile:

----
Screen Shot 2021-03-26 at 9 43 29 AM

----
I think this should be invalid from the start if the user's profile isn't the PI and doesn't have a PI. We need some form of messaging on the DAR form that explains this when the PI field isn't selectable like it is in the screenshot.

rushtong (approved):
This is fine as it is, but I think there is a bigger validation problem here. Feel free to merge this and we can tackle the validation issue in a new ticket.

src/pages/dar_application/ResearcherInfo.js
raejohanek: Good catch! I updated the comment to be more clear and removed the '--' placeholder that allows the DAR to be submitted when it shouldn't. I also updated the ticket to reflect this

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -150,7 +150,7 @@ export default function ResearcherInfo(props) {
               div({ isRendered: checkCollaborator !== true, className: 'display-inline' }, ['*']),
               div({ isRendered: checkCollaborator === true, className: 'display-inline italic' }, [' (optional)']),
               span({ className: 'default-color' },
-                ['Please autenticate with ',
+                ['Please authenticate with ',
```
rushtong, line 153: Nice catch!

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -228,7 +228,7 @@ export default function ResearcherInfo(props) {
               }),
               span({
                 className: 'cancel-color required-field-error-span', isRendered: (invalidInvestigator) && (showValidationMessages)
-              }, ['Required field'])
+              }, ['Required field, ensure your Profile has a PI listed or marks you as the PI'])
```
rushtong: Minor language nit, the researcher would have to mark themselves as a PI. This reads as if a PI must do something for the researcher. 

```suggestion
              }, ['Required field, ensure your Profile has a PI listed, or that you are selected as the PI'])
```

rushtong (approved):
Love the cleanup, thank you! One minor language suggestion inline. Feel free to merge if you want to make that change.

PR #970, opened 2021-03-30
DUOS-1138[risk=no] SDSLT bug fixes

raejohanek:
SCOPE:
Fix bug by checking for empty list
Fix bug on other textbox by removing value field
Capitalize r in readable

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1138

In all cases:

### DIFF
--- a/src/libs/dataUseTranslation.js
+++ b/src/libs/dataUseTranslation.js
@@ -5,6 +5,7 @@ import join from 'lodash/fp/join';
 import concat from 'lodash/fp/concat';
 import clone from 'lodash/fp/clone';
 import uniq from 'lodash/fp/uniq';
+import head from 'lodash/fp/head';
 import { searchOntology } from '../libs/ontologyService';
 import { Notifications } from '../libs/utils';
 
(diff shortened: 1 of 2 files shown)


JVThomas (dismissed):
One last change. When the textarea is disabled, the background color shifts to a greyish hue, causing the font to sort of blend in with the background. 
Screen Shot 2021-03-30 at 1 17 28 PM

A white background color should do the trick, and you should be able to do that by assigning ```style:{backgroundColor: '#fff'}``` as an attribute to the textarea

src/libs/dataUseTranslation.js
```diff
@@ -372,7 +372,7 @@ export const DataUseTranslation = {
       if (!isNil(value) && value) {
         if (key === 'diseaseRestrictions') {
           let resolvedLabels = [];
-          if (!value.isEmpty && !isNil(value[0].label)) {
+          if (!isNil(value[0]) && !isNil(value[0].label)) {
```
rushtong: nit: I prefer to use lodash `head` method to get the first element of an array - it's cleaner (and more functional) than referencing `[0]` directly.

rushtong (approved):
One nit (see inline), otherwise 👍🏽

PR #977, opened 2021-04-01
DUOS-1127[risk=low]Converted DAR Table to sub-components

JVThomas:
Addresses [DUOS-1127](https://broadworkbench.atlassian.net/browse/DUOS-1127)

- PR converts the Dar Table used in the NewChairConsole into smaller components that can be re-used in the AdminManageAccess console. As such a large part of this PR is the refactoring of both NewChairConsole and AdminManageAccess to use the components (and in AdminManageAccess' case the V2 endpoints). 
- Code also introduces new validation on Open and Close buttons to ensure that only Chairpeople of the DAC of target election can execute those functions (Chair Console).
- A new route was made called ```new_admin_manage_access``` to compare the new functionality against its older counterpart. I can either replace the current route with this new component or keep this route for post-deployment testing. (Feedback on this is appreciated).
- This PR does not apply any of the components to the Member console. Due to the large amount of refactoring I decided to limit this PR to the aforementioned pages.
- NewAdminManageAccess currently has a UI issue where the ResearcherReview link doesn't immediately load (it's waiting on a Researcher fetch response before rendering). I have plans to incorporate some sort of loading placeholder as well a memo function to avoid repeat fetches on election search updates
- Final/Closed status rendering issue is not solved with this PR (will be solved in a subsequent PR).
- Added ```exhaustive-dependencies-react-hooks``` warning for local development

In all cases:


src/components/SearchBar.js
```diff
@@ -0,0 +1,30 @@
...
+    width: '100%',
+    border: '1px solid #cecece',
+    backgroundColor: '#f3f6f7',
+    borderRadius: '5px',
+    height: '4rem',
+    paddingLeft: '2%',
+    fontFamily: 'Montserrat'
+  };
+
+  return div({className: "right-header-section", style: Styles.RIGHT_HEADER_SECTION}, [
```
rushtong, line 19: I can't find this css class name anywhere. Nor can I find its `left` complement.

src/components/SearchBar.js
```diff
@@ -0,0 +1,30 @@
...
+    input({
+      type: 'text',
+      placeholder: 'Enter search terms',
+      //Styling seems to only work when defined here, variable reference doesn't work
+      //Odds are there's a competing style, need to figure out where it's coming from
```
rushtong, line 24: The only external styling I can find here is in `index.css`:
```
a, input, button {
  outline: 0 !important;
  text-decoration: none !important;
}
```
Additionally, we have some `.less` files contributing to the style, but I'm not familiar with how those are being applied.

rushtong:
Functional issue. When updating the rows per page by hand (i.e. changing 10 to 100), the page freezes while all the data loads. It takes a while for all of the content to load and it looks like we're querying for user by id many, many times. Eventually the page unfreezes after everything is complete.

Screen Shot 2021-04-02 at 7 47 55 AM

rushtong:
Functional issue. With 42 records visible, the search bar is very, very slow. It looks like we're re-querying for all of the data, in addition to re-fetching users. I think to make this page work, we'll need to add the researcher information to the DAR record to prevent the explosion of client side queries.

src/components/dar_table/DarTable.js
```diff
@@ -0,0 +1,141 @@
...
+
+const calcPageCount = (tableSize, filteredList) => {
+  if(isEmpty(filteredList)) {
+    return 1;
+  }
+  return Math.ceil(filteredList.length / tableSize);
+};
+
+//////////////////
+//MAIN COMOPNENT//
```
rushtong: nit typo

src/components/dar_table/DarTableActions.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if(!isNil(dar) && !isNil(dar.data)) {
+      const userId = dar.userId;
+      const init = async(userId) => {
+        const user = await User.getById(userId);
+        setResearcher(user);
+      };
```
rushtong: I think this is where we are pre-loading all the researcher info before we need to show it in the modal window.

rushtong:
Functional issue. As a member of DAC 0001, I can vote on this DAR. However, I am not a Chair in this DAC, so I should not have access to the cancel icon. 

Screen Shot 2021-04-02 at 9 15 56 AM

src/components/dar_table/DarTableActions.js
```diff
@@ -0,0 +1,177 @@
...
+        h(DarTableOpenButton, {
+          dar,
+          index,
+          openConfirmation,
+          label: 'Open',
+          isRendered: !isElectionOpen(election),
+          disabled: isNil(chairVote) && consoleType === 'chair'
+        }),
+        createResearcherButtons(dar, visibilityOptions.showResearcher, history),
+        h(ReactTooltip, {
```
rushtong: It took me some time to understand what's happening with these tooltips. They are only necessary when `createResearcherButtons` is visible since they are tips on the different hand icons. I'm sure this will be confusing to any developer as the code is removed from where it is actually used. Can we integrate these into the researcher buttons?

rushtong (changes requested):
I love the extraction to smaller components, but there are some functional issues still to tackle (see comments). This PR goes pretty far in scope by trying to tackle the admin manage page and I would prefer to handle that separately after we're confident that the DAC admin works well.

src/components/dar_table/DarTableActions.js
JVThomas: I'm thinking of getting rid of ReactTooltop altogether. The big performance hit seems to come from ```ReactTooltip.rebuild()```, which is needed to link new DOM elements to its associated tooltip. Without this function call, the table performs significantly faster. I'll have to figure out a different way to show what the researcher icons represent.

src/components/SearchBar.js
JVThomas: It's for html markup purposes. Makes it easier to follow along when inspecting the template in the browser.

src/components/SearchBar.js
JVThomas: I'll dig into it. It's a pretty old comment, so it might not even apply now.

src/components/dar_table/DarTableActions.js
JVThomas: Yep, buttons assign the User locally before analyzing status. While it turns out it's not a major factor in the slowdown, it still causes a minor 1 second-ish pause before updating. Also, I'd prefer not making n calls to the backend just so an icon can show up on each row.

JVThomas:
Update: I added ReactTooltip back into the DarTable (it's now implemented in a more efficient manner) and removed the User queries (so its now dependent on the [updated manage response](https://github.com/DataBiosphere/consent/pull/999).

src/components/SearchBar.js
JVThomas: The left component is the title/subtitle portion on the page, both are components in the same flexbox row

src/components/dar_table/DarTableActions.js
JVThomas: Update: Re-implemented React Tooltip, see comment below for more information.

src/components/dar_table/DarTableActions.js
```diff
@@ -0,0 +1,132 @@
...
+    },
+    manageAccess: {
+      showVote: false,
+      showCancelIcon: false,
+      showResearcher: true
+    }
+  };
+
+  //applied extraOptions to template for final visibilityOptions reference
+  let visibilityOptions = templates[consoleType];
```
rushtong, line 35: To get past the codacy warning, would lodash's `get(consoleType)(templates)` work just as well here? Also, not entirely sure this would bypass codacy as we are passing in a variable as the path param here.

rushtong (approved):
As far as this PR goes wrt table refactoring, I think it looks good and works smoothly. I am seeing incorrect permissions issues, but those can be tackled in another PR. One comment inline that might help get past the codacy warning.

raejohanek (commented):
I did a bug fix on the New Chair Console DUOS-1149, https://github.com/DataBiosphere/duos-ui/pull/979/files I'm not sure if those changes will still be in place after this PR, can you make sure those changes are included in this PR
UPDATE: you can ignore this comment I see now that it is included thanks!

src/components/dar_table/DarTableActions.js
JVThomas: Not sure, the function fundamentally does the same thing (lodash goes a step further and executes recursively for nested objects). Either way it's a false positive.

raejohanek (changes requested):
The cancel and open election buttons on the new admin access page do not seem to be working for me additionally the manage access button on the admin console routes to the old admin access page, and the new one is only reached by adding new to the url

rushtong:
@raejohanek 
> additionally the manage access button on the admin console routes to the old admin access page, and the new one is only reached by adding new to the url

This was intentional and is not really necessary for approval of this PR. The final cutover to the new admin manage will be a separate event.

raejohanek:
> @raejohanek
> 
> > additionally the manage access button on the admin console routes to the old admin access page, and the new one is only reached by adding new to the url
> 
> This was intentional and is not really necessary for approval of this PR. The final cutover to the new admin manage will be a separate event.

Okay makes sense, just the buttons then!

rushtong:
@raejohanek, @JVThomas 
> The cancel and open election buttons on the new admin access page do not seem to be working for me

I also see some permissions wonkiness, but again, that's not part of this PR and could be from other things. I think the permissions issues can be resolved in a separate PR. 

For example (and I don't have root causes here, just UI visibility issues):

`grushton.dev2@gmail.com` should not have an enabled "Cancel" icon here, there really should be no icon at all as the member user should not be able to close this election:
Screen Shot 2021-04-07 at 12 11 34 PM

In addition, the user SHOULD be able to vote on this election as a member user.

Root causes can be back end or front end, but we need to iron them out.

JVThomas:
> > @raejohanek
> > > additionally the manage access button on the admin console routes to the old admin access page, and the new one is only reached by adding new to the url
> > 
> > 
> > This was intentional and is not really necessary for approval of this PR. The final cutover to the new admin manage will be a separate event.
> 
> Okay makes sense, just the buttons then!

Yep, that's a bug! Sorry for the slow reply, was figuring out what happened there. It's caused by a missing else statement for ```setDivAttributes``` in ```utils.js```. Going to test locally to make sure and then I'll push the fix to the branch.

JVThomas:
> @raejohanek, @JVThomas
> 
> > The cancel and open election buttons on the new admin access page do not seem to be working for me
> 
> I also see some permissions wonkiness, but again, that's not part of this PR and could be from other things. I think the permissions issues can be resolved in a separate PR.
> 
> For example (and I don't have root causes here, just UI visibility issues):
> 
> `grushton.dev2@gmail.com` should not have an enabled "Cancel" icon here, there really should be no icon at all as the member user should not be able to close this election:
> Screen Shot 2021-04-07 at 12 11 34 PM
> 
> In addition, the user SHOULD be able to vote on this election as a member user.
> 
> Root causes can be back end or front end, but we need to iron them out.

This is a front-end issue. I'm checking on Chair instead of Chair OR Member for the Vote button. I'll make the correction

**Update**: Scratch that, was looking on the wrong conditional. The button is active if the user has a vote for that election. If they don't, the button is disabled.

Here's the relevant code, can be found on ```DarTableActions```, around line 94.
```
const targetVotes = filter((v) => {
      const belongsToUser = (currentUserId === v.dacUserId);
      const targetTypes = (v.type === 'Chairperson' || v.type === 'DAC');
      return belongsToUser && targetTypes;
})(votes);
```

JVThomas:
> > @raejohanek, @JVThomas
> > > The cancel and open election buttons on the new admin access page do not seem to be working for me
> > 
> > 
> > I also see some permissions wonkiness, but again, that's not part of this PR and could be from other things. I think the permissions issues can be resolved in a separate PR.
> > For example (and I don't have root causes here, just UI visibility issues):
> > `grushton.dev2@gmail.com` should not have an enabled "Cancel" icon here, there really should be no icon at all as the member user should not be able to close this election:
> > Screen Shot 2021-04-07 at 12 11 34 PM
> > In addition, the user SHOULD be able to vote on this election as a member user.
> > Root causes can be back end or front end, but we need to iron them out.
> 
> This is a front-end issue. I'm checking on Chair instead of Chair OR Member for the Vote button. I'll make the correction
> 
> **Update**: Scratch that, was looking on the wrong conditional. The button is active if the user has a vote for that election. If they don't, the button is disabled.
> 
> Here's the relevant code, can be found on `DarTableActions`, around line 94.
> 
> ```
> const targetVotes = filter((v) => {
>       const belongsToUser = (currentUserId === v.dacUserId);
>       const targetTypes = (v.type === 'Chairperson' || v.type === 'DAC');
>       return belongsToUser && targetTypes;
> })(votes);
> ```

**Update 2:** Found the issue, it wasn't permissions, it was the userId memoization. Replaced it with a standard assignment. That said, Vote buttons should be disabled if the user doesn't have a Vote for an election. I can think of a couple possible scenarios (member added after election created?), but that falls outside the scope of this PR.

JVThomas:
> @raejohanek, @JVThomas
> 
> > The cancel and open election buttons on the new admin access page do not seem to be working for me
> 
> I also see some permissions wonkiness, but again, that's not part of this PR and could be from other things. I think the permissions issues can be resolved in a separate PR.
> 
> For example (and I don't have root causes here, just UI visibility issues):
> 
> `grushton.dev2@gmail.com` should not have an enabled "Cancel" icon here, there really should be no icon at all as the member user should not be able to close this election:
> Screen Shot 2021-04-07 at 12 11 34 PM
> 
> In addition, the user SHOULD be able to vote on this election as a member user.
> 
> Root causes can be back end or front end, but we need to iron them out.

In this instance that account is a Chairperson in Demo Dac New. Should chairs only have cancel rights if they have a vote?

rushtong:
> In this instance that account is a Chairperson in Demo Dac New. Should chairs only have cancel rights if they have a vote?

This must have been a mistake on my part. Chairs can cancel an election as long as they are in the election's DAC.

rushtong:
> Update 2: Found the issue, it wasn't permissions, it was the userId memoization. Replaced it with a standard assignment. That said, Vote buttons should be disabled if the user doesn't have a Vote for an election. I can think of a couple possible scenarios (member added after election created?), but that falls outside the scope of this PR.

Agreed, there are likely other issues here that we need to get to root causes for. I think one is that there are times when votes do not exist for an election when they should.

rushtong (approved):
There is a merge conflict that needs to be fixed, but after that, 👍🏽

raejohanek (changes requested):
the open and cancel buttons on the new chair console page do nothing when I click on them and the vote button is disabled (for me) for every single election even when I am a member and chair on that DAC, additionally for some of the rows where the vote button is disabled the cancel button is not, it seems maybe there is something wrong with how we are checking to see whether or not the vote button should be disabled, but now the buttons on the new admin access page work great!
Screen Shot 2021-04-07 at 1 31 45 PM

JVThomas:
> the open and cancel buttons on the new chair console page do nothing when I click on them and the vote button is disabled (for me) for every single election even when I am a member and chair on that DAC, additionally for some of the rows where the vote button is disabled the cancel button is not, it seems maybe there is something wrong with how we are checking to see whether or not the vote button should be disabled, but now the buttons on the new admin access page work great!
> Screen Shot 2021-04-07 at 1 31 45 PM

Need to make sure, but which account are you using? You have a chair and a member account, but the chair is only a chair for DAC 0001 and is a member for DAC 0002. The member account is a Member for DAC 0001 and nothing else.

raejohanek (approved):
Both pages look great and work well!

JVThomas:
I know the PR has been approved, but I'm not sure of an ideal time to merge this in. I know Jonathan briefly mentioned showcasing the DAC Chair page to others, so I want to make sure the merge doesn't get in the way of that. This is a big PR (again, sorry about this) and I want some extra time to test it once it's up.

rushtong:
The merge to dev should be fine, I won't be doing another staging release
unless Jonathan asks.

> I know the PR has been approved, but I'm not sure of an ideal time to
> merge this in. I know Jonathan briefly mentioned showcasing the DAC Chair
> page to others, so I want to make sure the merge doesn't get in the way of
> that. This is a big PR (again, sorry about this) and I want some extra time
> to test it once it's up.
>

JVThomas:
NOTE: Codacy flag (Injection Sink) is a false positive. Value used is a constant defined within the code itself, it does not come from outside input.

PR #979, opened 2021-04-02
DUOS-1149[risk=low]DAC Member (who is also a chair) can close/reopen election not in their DAC in the chair console

raejohanek:
SCOPE
- gather a list of dacIds that the current user is a chairperson in
- check to see if the dacId for the current row/dar is in that list
- if the user is not a chairperson on the dac for that row, then disable the open and re-open buttons and don't show the cancel election button

ADDRESSES
https://broadworkbench.atlassian.net/browse/DUOS-1149

In all cases:

### DIFF
--- a/src/pages/NewChairConsole.js
+++ b/src/pages/NewChairConsole.js
@@ -97,24 +97,20 @@ const Records = (props) => {
             const targetTypes = (v.type === 'Chairperson' || v.type === 'DAC');
             return belongsToUser && targetTypes;
           })(electionInfo.votes);
-          const voteButton = h(TableTextButton, {
-            key: `vote-button-${e.referenceId}`,
-            onClick: () => props.history.push(`access_review/${dar.referenceId}`),
-            label: 'Vote',
-            disabled: isEmpty(votes)
-          });
-          if (chairDacIds.includes(dacId)) {
-            return [
-              voteButton,
-              h(TableIconButton, {
-                icon: Block,
-                key: `cancel-button-${e.referenceId}`,
-                onClick: () => cancelElectionHandler(electionInfo, dar.referenceId, index)
-              })
-            ];
-          } else {
-            return [voteButton];
-          }
+          return [
+            h(TableTextButton, {
+              key: `vote-button-${e.referenceId}`,
+              onClick: () => props.history.push(`access_review/${dar.referenceId}`),
+              label: 'Vote',
+              disabled: isEmpty(votes)
+            }),
+            h(TableIconButton, {
+              isRendered: chairDacIds.includes(dacId),
+              icon: Block,
+              key: `cancel-button-${e.referenceId}`,
+              onClick: () => cancelElectionHandler(electionInfo, dar.referenceId, index)
+            })
+          ];
         }
         default :
           return h(TableTextButton, {
--- a/src/pages/NewChairConsole.js
+++ b/src/pages/NewChairConsole.js
@@ -1,4 +1,4 @@
-import { isEmpty, isNil, includes, toLower, filter, cloneDeep, find } from 'lodash/fp';
+import { isEmpty, isNil, includes, map, toLower, filter, cloneDeep, find } from 'lodash/fp';
 import { useState, useEffect, useRef } from 'react';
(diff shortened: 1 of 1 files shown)


src/pages/NewChairConsole.js
```diff
@@ -94,32 +97,39 @@ const Records = (props) => {
...
+          const voteButton = h(TableTextButton, {
+            key: `vote-button-${e.referenceId}`,
+            onClick: () => props.history.push(`access_review/${dar.referenceId}`),
+            label: 'Vote',
+            disabled: isEmpty(votes)
+          });
+          if (chairDacIds.includes(dacId)) {
+            return [
+              voteButton,
+              h(TableIconButton, {
```
rushtong: Instead of wrapping the whole thing in an if, you could set the is rendered property to the result of `chairDacIds.includes(dacId)` and save yourself an if/else block.

src/pages/NewChairConsole.js
```diff
@@ -94,32 +97,39 @@ const Records = (props) => {
...
+            return [voteButton];
+          }
         }
         default :
           return h(TableTextButton, {
             key: `reopen-button-${e.referenceId}`,
             onClick: () => openConfirmation(dar, index),
-            label: 'Re-Open'
+            label: 'Re-Open',
+            disabled: !chairDacIds.includes(dacId)
```
rushtong: Since we're using this same logic in multiple places, it makes sense to pull this calculation out into a constant and just re-use the constant for disabled state and/or rendered state.

rushtong (changes requested):
Looks good, but can be simplified. See comments inline.

src/pages/NewChairConsole.js
raejohanek: great idea!!!

src/pages/NewChairConsole.js
```diff
@@ -85,7 +85,11 @@ const Records = (props) => {
   const createActionButtons = (electionInfo, index) => {
     const e = electionInfo.election;
     const dar = electionInfo.dar;
-    const currentUserId = Storage.getCurrentUser().dacUserId;
+    const dacId = electionInfo.dac.dacId;
+    const currentUser = Storage.getCurrentUser();
+    const chairDacIds = map((role) => {return role.dacId;})(filter({roleId: 2})(currentUser.roles));
+    const currentUserId = currentUser.dacUserId;
+    const isChair = chairDacIds.includes(dacId);
```
rushtong: nit, naming. We know the user is a chair, so something like `isChairForDac` would be more intuitive.

src/pages/NewChairConsole.js
```diff
@@ -112,14 +117,16 @@ const Records = (props) => {
...
           });
       }
     }
     return h(TableTextButton, {
       onClick: () => openConfirmation(dar, index),
       key: `open-election-dar-${dar.referenceId}`,
-      label: 'Open Election'
+      label: 'Open Election',
+      disabled: !isChair
     });
```
rushtong: Should this even be rendered if the user isn't a chair for the dac?

src/pages/NewChairConsole.js
```diff
@@ -112,14 +117,16 @@ const Records = (props) => {
           return h(TableTextButton, {
             key: `reopen-button-${e.referenceId}`,
             onClick: () => openConfirmation(dar, index),
-            label: 'Re-Open'
+            label: 'Re-Open',
+            disabled: !isChair
           });
```
rushtong: Same idea here, should we even render it if the user can't perform the action?

raejohanek: I was trying to follow the logic of the vote buttons since these buttons look the same, and the vote button just get disabled when there is no vote for the user, so that's why I did this but what do you think?

rushtong: I get that. My opinion is that if something isn't useful to the user, then it shouldn't be there. Having a disabled button that will always be disabled (because they're not a chair in the dac) is not useful. It's just noise on the page.

raejohanek: makes sense to me!

rushtong (approved):
👍🏽

PR #996, opened 2021-04-07
DUOS-1151[risk=no]Enable Admin-only Removal/Delete of a Data Access Committee

raejohanek:
SCOPE:
- Add delete Icon to actions section for admin only
- Add "Delete DAC?" confirmation modal
- Check to see if the dac has any datasets to see whether or not it is safe to delete
- Call to DAC.delete, once confirmed there are no datasets assigned to the dac
- Display notifications for success and inability to delete

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1151

In all cases:


src/pages/manage_dac/ManageDac.js
```diff
@@ -309,7 +346,13 @@ class ManageDac extends Component {
...
                       style: actionButtonStyle,
                       onClick: () => this.editDac(dac)
-                    }, ['Edit'])
+                    }, ['Edit']),
+                    h(TableIconButton, {
+                      style: {paddingTop: '.5rem', color: '#00609f'},
+                      isRendered: userRole === ADMIN,
+                      onClick: () => this.deleteDac(dac),
+                      icon: Delete
+                    })
```
rushtong: This needs a cursor style so it is clear that it's clickable.

src/pages/manage_dac/ManageDac.js
```diff
@@ -114,6 +119,37 @@ class ManageDac extends Component {
...
+      isEmpty(response) ?
+        this.setState(prev => {
+          prev.showConfirmationModal = true;
+          prev.isEditMode = false;
+          prev.selectedDac = selectedDac;
+          return prev;
+        })
+        : Notifications.showError({text: "All datasets assigned to this DAC must be reassigned before this can be deleted."});
+    });
+  };
```
rushtong: We should be able to tell beforehand if the DAC has datasets and therefore the icon shouldn't be enabled and instead show a tooltip that has the error message in it. There are probably a few ways to do this, but I think the simplest way is to make a new API call that gets all DACs along with their Datasets so you don't have to make this extra call per DAC.

rushtong (changes requested):
Works well, but I think there is a more UX-friendly way to accomplish this feature. See comments inline.

src/pages/manage_dac/ManageDac.js
rushtong: Even better, we already have a datasetids property on the DAC, but it isn't populated in the query we're making for all DACs. We should update the underlying query so this value is always populated and you would not need to make the extra call for datasets by dacid.

rushtong: I think this will do it:
```
    @SqlQuery("SELECT distinct d.*, a.datasetid " +
            " FROM dac d " +
            " LEFT JOIN consents c ON d.dac_id = c.dac_id " +
            " LEFT JOIN consentassociations a ON a.consentid = c.consentid ")
    List<Dac> findAll();
```

raejohanek: good idea!

raejohanek: so there is only one datasetid being returned here, because theres only one on a consent association, but a dac could have any number of datasets, so we are missing some information I think but I'm not really sure how to solve this problem

rushtong: We have to be able to query for this, so that's where the work needs to happen. It could be the mapper as well - we might need to change how we're mapping DACs from the result set.

src/components/dar_table/DarTableCancelButton.js
```diff
@@ -28,7 +28,7 @@ export default function DarTableCancelButton(props) {
   const attributes = {
     key: `cancel-button-${election.referenceId}`,
     disabled,
-    dataTip: disabled ? 'You do not have permision to cancel this election' : 'Cancel Election',
```
raejohanek, line 31: typo

rushtong:
Functionally, I get an error onscreen when deleting a DAC. That error message goes away, the confirmation modal screen stays up, and the DAC is successfully deleted. I'm not sure, but `handleDeleteDac` might have a race condition happening. I would try moving the success/failure logic outside of the `then`. Await the response from the delete call and assign that to a constant, then check the value of the constant to determine what to do with the success/error case.

Screen Shot 2021-05-13 at 1 56 30 PM

src/libs/ajax.js
```diff
@@ -120,7 +120,7 @@ export const DAC = {
   delete: async (dacId) => {
     const url = `${await Config.getApiUrl()}/dac/${dacId}`;
     const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), { method: 'DELETE' }]));
-    return res.json();
+    return res;
```
raejohanek, line 124: by not parsing the response to a json, we can just check the response status in handleDeleteDAC. the reason the handleDeleteDAC method wasn't working was because the response was returning successfully but then the res.json() line was throwing an unexpected end of input error (I think because there is nothing in the body) and that was triggering the catch phrase in handleDeleteDAC causing it to display an error even though the delete was successful

raejohanek: also, the only place this is used is in handleDeleteDac on the ManageDac page

rushtong: We should add a comment here about why this is needed since it deviates from our standard pattern. Very much makes sense for API calls that don't have any response content.

rushtong (approved):
One minor comment suggestion, then 👍🏽

PR #1014, opened 2021-04-13
DUOS-1172[risk=no]UI: State Unselectable with no message explaining why

raejohanek:
SCOPE:
- updated state validation check 
- leave state box enabled when no country is selected, only disable when a non US country is selected
- fix tooltip for isPi
- add tooltip for state

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1172

In all cases:

### DIFF
--- a/src/pages/ResearcherProfile.js
+++ b/src/pages/ResearcherProfile.js
@@ -721,7 +721,7 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
                 div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 no-padding' }, [
                   div({ className: 'row fsi-row-lg-level fsi-row-md-level no-margin' }, [
                     div({ className: 'col-lg-6 col-md-6 col-sm-6 col-xs-6' }, [
-                      label({ id: 'lbl_profileCity', className: 'control-label' }, ['City* ']),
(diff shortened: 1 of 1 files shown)


src/pages/ResearcherProfile.js
```diff
@@ -140,6 +141,20 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
         this.researcherFieldsValidation();
       }
     });
+    //clear out state field if the user selects
```
raejohanek, line 144: before this code block was here when you would chose another country the state box would be disabled but still show the state value that was in it before

src/pages/ResearcherProfile.js
```diff
@@ -254,8 +269,8 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
   }
 
   isValidState(value) {
-    const stateSelected = (!isNil(value) || !isEmpty(value)) && (value !== "N/A");
-    const inUS = (this.state.profile.country === "United States of America");
+    const stateSelected = (!isNil(value) || !isEmpty(value));
```
rushtong: This is reporting valid even if I unselect the state option so validation is failing here.

rushtong (changes requested):
See comment inline. Ignore the failing cloud build check, we'll get that removed at some point.

rushtong (approved):
Thank you for the changes, works well 👍🏽

PR #1016, opened 2021-04-13
DUOS-1189/DUOS-1194[risk=no]Add 'Dataset Name' column to DAC Chair,  and Admin Consoles

raejohanek:
SCOPE:
-NOTE: this PR does not implement the additional column in the member console, like the ticket requests. This is because the member console is getting its data from the v1 endpoint that does not contain datasets or dataset id. We could implement this column without changing the api call if we query for the dar off of the access election and then query again for the dataset based off of the datasetid included in the dar. This would be two extra api calls for every row, which would slow things down quite a bit. I think it makes more sense to convert the member console over to the v2 api call and reuse the components used for the new chair and admin access pages. But that is out of the scope of this PR. Additionally, in the meantime, the datasets are already accessible from this page by clicking on the dar modal. Thoughts?
-implement Dataset column on new chair and new admin manage access

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1189

In all cases:

### DIFF
--- a/src/components/dar_table/DarTable.js
+++ b/src/components/dar_table/DarTable.js
@@ -17,6 +17,7 @@ import ReactTooltip from 'react-tooltip';
 export const tableHeaderTemplate = [
   div({style: Styles.TABLE.DATA_ID_CELL}, ["Data Request ID"]),
   div({style: Styles.TABLE.TITLE_CELL}, ["Project title"]),
+  div({style: Styles.TABLE.TITLE_CELL}, ["Dataset"]),
   div({style: Styles.TABLE.SUBMISSION_DATE_CELL}, ["Last Updated"]),
   div({style: Styles.TABLE.DAC_CELL}, ["DAC"]),
   div({style: Styles.TABLE.ELECTION_STATUS_CELL}, ["Election status"]),
@@ -26,6 +27,7 @@ const loadingMarginOverwrite = {margin: '1rem 2%'};
 
 export const tableRowLoadingTemplate = [
   div({style: assign(Styles.TABLE.DATA_ID_CELL, loadingMarginOverwrite), className: 'text-placeholder'}),
+  div({style: assign(Styles.TABLE.TITLE_CELL, loadingMarginOverwrite), className: 'text-placeholder'}),
   div({style: assign(Styles.TABLE.TITLE_CELL, loadingMarginOverwrite), className: 'text-placeholder'}),
   div({style: assign(Styles.TABLE.SUBMISSION_DATE_CELL, loadingMarginOverwrite), className: 'text-placeholder'}),
   div({style: assign(Styles.TABLE.DAC_CELL, loadingMarginOverwrite), className: 'text-placeholder'}),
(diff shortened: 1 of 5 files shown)


src/components/dar_table/DarTable.js
```diff
@@ -16,17 +16,19 @@ import ReactTooltip from 'react-tooltip';
 ////////////////////
```
raejohanek, line 16: consistent capitalization

src/pages/MemberConsole.js
```diff
@@ -81,7 +81,7 @@ class MemberConsole extends Component {
 
   }
 
-  openAccessReview = (referenceId) => async (e) => {
```
raejohanek, line 84: eslint unused constant

rushtong:
@raejohanek 

> We could implement this column without changing the api call if we query for the dar off of the access election and then query again for the dataset based off of the datasetid included in the dar. This would be two extra api calls for every row, which would slow things down quite a bit. I think it makes more sense to convert the member console over to the v2 api call and reuse the components used for the new chair and admin access pages. But that is out of the scope of this PR. Additionally, in the meantime, the datasets are already accessible from this page by clicking on the dar modal. Thoughts?
-implement Dataset column on new chair and new admin manage access

This conversation should be in the ticket, so let's move it there.

JVThomas (changes requested):
Since the dataset name is now visible on the table, you'll need to update the search function so that users can search on it. The function, ```darSearchHandler```, can be found on ```utils.js```

src/libs/utils.js
```diff
@@ -13,6 +13,14 @@ export const applyHoverEffects = (e, style) => {
   });
 };
 
+//currently, dars contain a list of datasets (any length) and a list of length 1 of a datasetId
+//go through the list of datasets and get the name of the dataset whose id is in the datasetId list
+export const getNameOfDatasetForThisDAR = (datasets, datasetId) => {
+  let id = datasetId[0].toString();
+  let data = find({"value" : id})(datasets);
```
rushtong: minor nits.
* Both can be `const`s
* Can be simplified a bit. Prefer lodash `head` or `first` over direct array access
  * this also works: `const data = find({"value" : first(datasetId).toString()})(datasets);`
* Could also use a nil check to ensure `datasetId` isn't an empty array. It should never be, otherwise we'd have an invalid DAR, but better to be safe than throw an error.

rushtong (changes requested):
Works well, but could be cleaned up in addition to Justin's feedback.

JVThomas (dismissed):
Screen Shot 2021-04-14 at 1 33 50 PM

Ok the search function is up and running, which means now we can tackle the column aesthetics

The table headers are being-forced into multi-line formatting due to the new columns, which means you'll have to play with the widths on some of the column styles to make sure they fall under one line. You can use the templates on ```DarTable``` to see which style corresponds to what cell, with each style having a width value represented as a percent. Because we're defining width in a flex display, width will act as an initial value: the browser will try to aim for this, but will make adjustments if the content does not fit. 

This actually helps us since we care about the relative look and not a pixel-perfect representation. That said, the rendered output is a lot easier to predict when the total width falls within 100%. Feel free to make your own cell styling (or simplify what's already there, your choice).

**NOTE:** There's nothing wrong with multi-line headers, but I think we're in a position where we can flatten them out. Plus, extra width on the dataset name column can help reduce the height of the rows and help reduce scrolling.

Screen Shot 2021-04-14 at 1 46 55 PM

rushtong (approved):
Looks good 👍🏽

PR #1020, opened 2021-04-15
[DUOS-1203][risk=no] Use audit-ci instead of npm audit

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1203

This PR uses an alternative approach to the standard `npm audit` checks to allow for bypassing known false positive alerts.

* Updated audit action to use [audit-ci](https://github.com/IBM/audit-ci)
* Added allowlist entry for advisory number 565 (`react-scripts>webpack>terser-webpack-plugin>cacache>ssri`) which is a false positive alert in the way that we use it. 
* Moved react-scripts to dev dependency. Both `npm audit` and `npx audit-ci` pick this up as a vulnerability regardless of where it is, but dev dependencies is a more accurate home for it.

---

In all cases:


config/audit-ci.json
```diff
@@ -0,0 +1,5 @@
+{
+  "allowlist": [
+    565
```
JVThomas: So this would put an exception on advisory 565 (regex vulnerability on ssri) and not advisory 565 as it applies to react-scripts. There is always the possibility that a new package can take on ssri as a dependency and use it in a way that makes 566 a real threat, but it would be over looked due to the blanket exception.

audit-ci has an alternative output where you can specify an advisory alongside a path.
https://github.com/IBM/audit-ci#testnpm-config-fileaudit-cijson

JVThomas (changes requested):

JVThomas (approved):
Looks good, audit-ci whitelist is working as expected. Just leaving a note to check if the build steps work as expected on staging/production now that react-scripts is a devDependency. It should work, there's no flags being passed to npm and NODE_ENV isn't set to production (I think), but I'm not to familiar with the deployment setup on k8s.

rushtong:
@JVThomas 
> Just leaving a note to check if the build steps work as expected on staging/production now that react-scripts is a devDependency. It should work, there's no flags being passed to npm and NODE_ENV isn't set to production (I think), but I'm not to familiar with the deployment setup on k8s.

Our prod deploy just does a build with the docker file same as other environments (https://github.com/DataBiosphere/duos-ui/blob/develop/Dockerfile) so there is no difference between env image builds. All env deploys are interchangeable with env-specific config files.

PR #1023, opened 2021-04-16
DUOS-1097[risk=no] User Session Timeout is not redirecting to home

JVThomas:
Addresses [DUOS-1097](https://broadworkbench.atlassian.net/browse/DUOS-1097)

There's still uncertainty on why Stackdriver cannot set a user when handling 401 requests. However this PR deals with this issue by checking for 401s first on fetchAny or fetchOK (axios requests already performs the 401 check). If the res.status if a 401, perform a redirect home, otherwise go through the Stackdriver flow as normal. Outcome is having the initial component template load (component mount) but then redirects back to the home page once a 401 status had been read.

In all cases:


rushtong (approved):
Works as expected. I think we can add some additional tweaks to add a notification to the user, but as it stands, this PR does what we need it to do.

PR #1026, opened 2021-04-16
DUOS-1195/1193[risk=no]Update Member Console to use v2 endpoint and reuse components

raejohanek:
SCOPE:
- create NewMemberConsole, Use v2 endpoint to get data
- reuse DarTable, SearchBar, and DarTableSkeletonLoader
- Design Choice: the old member console used the cases/pending api endpoint to get only the open cases so after switching to the v2 endpoint I filter the list so that it only shows open cases. This is because a member cannot do anything with canceled or accepted elections, they are there to vote on open elections.
-  I altered the cancel buttons to not just be disabled on the Member Console but to not render altogether. This is not needed for the open button bc the open button only appears for not-open elections which will never be present on the member console due to the filtering

To view the page visit /new_member_console I have not replaced the routing to the old member console

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1195
Also addresses: https://broadworkbench.atlassian.net/browse/DUOS-1193

In all cases:


src/pages/NewMemberConsole.js
```diff
@@ -0,0 +1,79 @@
...
+  const [currentPage, setCurrentPage] = useState(1);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const init = async () => {
+      try {
+        setIsLoading(true);
+        const pendingList = await DAR.getDataAccessManageV2();
+        const openElectionList = (pendingList).filter((i) => {
+          return i.election.status === "Open";
```
JVThomas: So the V2 endpoint returns all DARs that have been submitted in general, including DAR's that do not have an election (they've just been submitted, but the Chair or Admin has yet to open an election). ```i.election.status === "Open"``` will error out if an election does not exist, which breaks the init function

src/pages/NewMemberConsole.js
```diff
@@ -0,0 +1,79 @@
...
+            })
+          ]),
+          div({style: Styles.HEADER_CONTAINER}, [
+            div({style: Styles.TITLE}, ["DAC Member Console"]),
+            div({style: Styles.SMALL}, ["Should data access be granted to this applicant?"])
+          ]),
+        ]),
+        h(SearchBar, {handleSearchChange, currentPage})
+      ]),
+      h(DarTable, {
```
JVThomas, line 65: There's nothing wrong here, just a UI improvement that I just thought of. When I made the table components I never made the Vote label adapt to whether or not the user had already submitted a vote. It never bothered me with the Chair console since their vote is essentially the final verdict, making the election ```Accepted``` or ```Closed```. This in turns changes the row's output and hides the vote button. However with the Member console the Vote button stays up, so it's hard for members to know whether or not they already voted on an election.

I don't think you'll have to do too much work, just some minor adjustments to ```DarTableActions``` and ```DarTableVoteButton```. The votes are already filtered out by user on line 98 in ```DarTableActions```, and the Vote Button is rendered out in that same function, so you should have everything you need to make this work.

The label update should be specific to the Member Console, since the Chairperson can't update their Chair vote once submitted. (Admin Console doesn't even render the Vote Button, so you have nothing to worry about there).

JVThomas (changes requested):

src/pages/NewMemberConsole.js
rushtong: @raejohanek @JVThomas 
> The label update should be specific to the Member Console, since the Chairperson can't update their Chair vote once submitted. (Admin Console doesn't even render the Vote Button, so you have nothing to worry about there).

A chair can vote as a member and then navigate back to their console, never having submitted a final vote. I think we could use the previously-voted treatment on both console pages.

raejohanek: So do we want to change the vote button to a 'Voted' button? That could be confusing as a chair if they have voted as a member but not yet as a chair. We could do a 'voted' button on the member console and like a 'voted as member' button on the chair page? @rushtong @JVThomas

JVThomas: That will work, you'll just have to reference ```consoleType``` in ```DarTableActions``` to make sure the label is added correctly. 

Also maybe change the phrase to "Edit Member Vote" or "Update Member Vote" (or anything along those lines) to show that that button performs an action.

raejohanek: Good idea. So any member can update their vote? What if just across both the member and chair console pages if the member vote has been cast then the button says 'update vote'? Because I think on the chair console a button like 'update member vote' doesn't really capture the entire function of the button since it can also be used to vote as a chair.

JVThomas: Yeah, good point. Let's go with "Update Vote".

src/pages/NewMemberConsole.js
```diff
@@ -0,0 +1,80 @@
...
+  const [currentPage, setCurrentPage] = useState(1);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const init = async () => {
+      try {
+        setIsLoading(true);
+        const pendingList = await DAR.getDataAccessManageV2();
+        const openElectionList = (pendingList).filter((i) => {
+          return isNil(i.election.status) ? false : i.election.status === "Open";
```
JVThomas: You need to do a nil Check on ```i.election```, not ```i.election.status```. If the election doesn't exists, ```i.election.status``` will throw a "Cannot access ```status``` on ```undefined```" error message

JVThomas (changes requested):

src/pages/NewMemberConsole.js
rushtong: FYI, see https://broadworkbench.atlassian.net/browse/DUOS-1218 for more information on button text. That ticket contains some NIH feedback on labeling that's relevant to this discussion. It might be worth not addressing text label issues here and instead save that for DUOS-1218.

src/Routes.js
```diff
@@ -75,6 +76,7 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/chair_console" component={ChairConsole} props={props} rolesAllowed={[USER_ROLES.chairperson]} />
     <AuthenticatedRoute path="/new_chair_console" component={NewChairConsole} props={props} rolesAllowed={[USER_ROLES.chairperson]} />
     <AuthenticatedRoute path="/member_console" component={MemberConsole} props={props} rolesAllowed={[USER_ROLES.member]} />
+    <AuthenticatedRoute path="/new_member_console" component={NewMemberConsole} props={props} rolesAllowed={[USER_ROLES.member]} />
```
rushtong: We should just bite the bullet here and make the path to `member_console` point to `NewMemberConsole`. A follow on PR could replace `MemberConsole` with `NewMemberConsole`.

src/components/dar_table/DarTableActions.js
```diff
@@ -111,7 +117,7 @@ export default function DarTableActions(props) {
           index,
           updateLists,
           isIcon: visibilityOptions.showCancelIcon,
-          isRendered: isElectionOpen(election),
+          isRendered: isElectionOpen(election) && consoleType !== 'member',
           disabled: !isChair && consoleType !== 'manageAccess'
```
rushtong: To prevent future bugs, we should treat these kinds of static string values as enums. Something like:
```
const consoleTypes = {
    MEMBER: 'member',
    MANAGE_ACCESS: 'manageAccess'
}
```
... etc. and then you can compare to those values, i.e.
```
consoleType !== consoleTypes.MEMBER;
```
The trick is to place `consoleTypes` in the right place. Not sure where that should live at the moment, but probably in this component.

src/components/dar_table/DarTableVoteButton.js
```diff
@@ -1,13 +1,18 @@
 import { h } from 'react-hyperscript-helpers';
 import TableTextButton from '../TableTextButton';
+import { isNil } from "lodash/fp";
 
 export default function VoteButton(props) {
-  const {election, history, darReferenceId, disabled} = props;
+  const {targetVotes, election, history, darReferenceId, disabled} = props;
+  const hasVoted = targetVotes.some((vote) => {
+    return !isNil(vote.vote);
+  });
```
rushtong, line 10: What is `targetVotes` here? Does that mean only the votes for the user? Could use some comments to explain what these are. If they are not limited to the user's votes, then the `hasVoted` will be true if any of the DAC members have voted.

src/pages/NewMemberConsole.js
```diff
@@ -0,0 +1,80 @@
+import { useState, useEffect, useCallback } from 'react';
+import { div, h, img } from 'react-hyperscript-helpers';
+import { DAR } from '../libs/ajax';
+import SearchBar from '../components/SearchBar';
+import { Notifications, getElectionDate, processElectionStatus, darSearchHandler } from '../libs/utils';
+import { Styles} from '../libs/theme';
```
rushtong: minor nit: spacing alignment
```suggestion
import { Styles } from '../libs/theme';
```

src/pages/NewMemberConsole.js
```diff
@@ -0,0 +1,80 @@
...
+import DarTable from '../components/dar_table/DarTable';
+import lockIcon from '../images/lock-icon.png';
+import { updateLists as updateListsInit } from '../libs/utils';
+import { tableHeaderTemplate, tableRowLoadingTemplate } from '../components/dar_table/DarTable';
+import DarTableSkeletonLoader from '../components/TableSkeletonLoader';
+import { isNil } from 'lodash';
```
rushtong: nit: prefer lodash/fp for new code and refactoring.

rushtong (commented):
Minor comments inline, but looks good and works well.

src/components/dar_table/DarElectionRecords.js
```diff
@@ -43,7 +43,7 @@ export default function DarElectionRecords(props) {
   const visibleWindow = calcVisibleWindow(currentPage, tableSize, filteredList);
   const dataIDTextStyle = Styles.TABLE.DATA_REQUEST_TEXT;
   const recordTextStyle = Styles.TABLE.RECORD_TEXT;
-  const showVotes = (consoleType === 'chair' || (!isEmpty(extraOptions) && extraOptions.showVote)) ? true : false;
```
raejohanek, line 46: simplify

rushtong (approved):
Thank you for the changes! Looks good, pending fixes for the merge conflicts

JVThomas (approved):
Code looks good! Console is working with constant consoleTypes and labels implemented.

PR #1034, opened 2021-04-21
[DUOS-1219][risk=no] Handle the dataset name validation correctly

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1219

As of https://github.com/DataBiosphere/duos-ui/pull/1023, we handle the `fetchOK` method slightly differently which caused an infinite spinner in the `404` case on the dataset registration page. We really shouldn't be using `fetchOK` since the successful case is actually an HTTP error code so using `fetchAny` is more appropriate.

----

In all cases:

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -467,7 +467,11 @@ export const DataSet = {
   validateDatasetName: async (name) => {
     const url = `${await Config.getApiUrl()}/dataset/validate?name=${name}`;
     try {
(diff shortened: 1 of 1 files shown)


JVThomas (approved):
Looks good, this way dataset validation can work without adversely affecting the session validation flow.

rfricke-asymmetrik (approved):
👍

PR #974, opened 2021-03-31
DUOS-1106[risk=no]Require Library Cards for DAC Chairs to vote 'Yes' on Chair vote

raejohanek:
## SCOPE
- in AccessReview: get library cards from researcher profile and pass the list to DacVotePanel
- in DacVotePanel: pass library cards to VoteAsChair
- inVoteAsChair: determine if the list is empty, if so display error. pass boolean hasLibraryCard to VoteQuestion
- inVoteQuestion if hasLibraryCard is false disable the yes button

## ADDRESSES
https://broadworkbench.atlassian.net/browse/DUOS-1106

Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] PR is labeled with a Jira ticket number and includes a link to the ticket
- [ ] PR is labeled with a security risk modifier [no, low, medium, high] 
- [ ] PR describes scope of changes

In all cases:

- [ ] Get a minimum of one thumbs worth of review, preferably 2 if enough team members are available
- [ ] Get PO sign-off for all non-trivial UI or workflow changes
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1039, opened 2021-04-22
DUOS-1223[risk=no]Confusing options on chair vote page when no LC present

raejohanek:
SCOPE:
- update error message to explain what the user can do
- fix syntax error in ajax call
- replace star imports

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1223

In all cases:

### DIFF
--- a/src/pages/access_review/VoteAsChair.js
+++ b/src/pages/access_review/VoteAsChair.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { div, a, span, hh } from 'react-hyperscript-helpers';
 import { Theme } from '../../libs/theme';
 import { VoteQuestion } from './VoteQuestion';
-import {isNil, head, get, getOr} from 'lodash/fp';
+import {isNil,  get, getOr} from 'lodash/fp';
 import * as moment from 'moment';
 
 const LINK = {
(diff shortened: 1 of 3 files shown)


JVThomas:
@rushtong @solideoglori Quick question: [DUOS-1106](https://github.com/DataBiosphere/duos-ui/pull/974) was implemented specifically in that manner because Chairs were allowed to reject a DAR if the application itself was fundamentally un-approvable (an act that does not require a library card to be present), so why the switch to a full disable?

solideoglori:
@JVThomas we do not want to move to a full disable

rushtong:
Ok, I misunderstood the original requirements. As it stands now, it's confusing as some combinations allow a submit, while others don't while the missing requirements are not spelled out.

src/pages/access_review/VoteAsChair.js
```diff
@@ -62,8 +62,9 @@ export const VoteAsChair = hh(class VoteAsChair extends React.PureComponent {
   };
 
   render() {
-    const { vote, rpVote, onUpdate, matchData, accessElectionOpen, rpElectionOpen, hasLibraryCard } = this.props;
-    const errorMessage = "The Researcher must have a Library Card before this DAR can be approved.";
+    const { vote, rpVote, onUpdate, matchData, accessElectionOpen, rpElectionOpen, libraryCards } = this.props;
+    const hasLibraryCard = !isNil(head(libraryCards));
+    const errorMessage = "The Researcher must have a Library Card before this DAR can be approved. You can still deny the request or vote on question 2.";
```
JVThomas: Minor suggestion, and then the code is good to go.

With your follow up sentence I realized we can make that old first sentence a bit more specific. Something like ```The Researcher must have a Library Card before data access can be granted. You can still deny this request and/or vote on question 2```.

This way the statements describe what you can or can't do for each question

JVThomas (changes requested):

JVThomas (approved):
Alright, PR looks good to go.

PR #1067, opened 2021-04-30
DUOS-1155[risk=no]Add API: get votes by Election IDs, use instead of get votes by reference ID

raejohanek:
Scope:
- pass electionIds to the Votes ajax method instead of just requestId so the votes returned hava much more limited scope
Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1155

In all cases:


src/components/dar_table/DarTable.js
```diff
@@ -121,10 +121,11 @@ export default function DarTable(props) {
   const createElection = async (darId, index) => {
     if (!isNil(darId)) {
       try{
-        const updatedElection = await Election.createDARElection(darId);
-        const votes = await Votes.getDarVotes(darId);
+        const updatedElections = await Election.createDARElection(darId);
+        const electionIds = map(election => election.electionId)(updatedElections);
+        const votes = await Votes.getElectionVotes(darId, electionIds);
```
raejohanek: this is why I made the change to the response of createDataRequestElection on the backend 1555 PR
Besides tests, this is the only place the method is called and it is very beneficial to have it return both elections created

raejohanek: If it only returned the access election we would have to do another query for the RP election, which we could do if that is preferred to the change in the response but I thought this made more sense

rushtong: API changes like this, by definition, are breaking changes. See https://github.com/DataBiosphere/consent/blob/develop/CONTRIBUTING.md#api-changes for more info. New APIs, on the other hand, do not introduce breaking changes and that is the approach we need to take if we want to do something like this.

raejohanek: Oh okay, makes sense. So do you recommend I make a new API with this response change and use that one in the ajax method on the UI and file a ticket for the deprecation/deletion of this one?

rushtong: We should discuss implementation in the ticket - PRs are an awkward place for this conversation.

rushtong (changes requested):
I don't think this is ready for review based on previous discussions. Please do resubmit when it's ready.

rushtong:
Currently blocked on https://github.com/DataBiosphere/consent/pull/1058
Code looks good, so tentatively 👍🏽   on this when we get the API side sorted out.

rushtong (approved):
LGTM 👍🏽

PR #1068, opened 2021-04-30
DUOS-1215[risk=no]Add missing final vote info to Review_Results page

raejohanek:
SCOPE:
For each question summary panel, now separate out votes to show member and chair votes distinctly
Add duos algorithm decision to page
Add final decision decision to page
Add agreement vote to page
Add rp decision to page (and method to calculate it accurately)

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1215

In all cases:

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -705,7 +705,7 @@ export const Match = {
   findMatch: async (consentId, purposeId) => {
     const url = `${await Config.getApiUrl()}/match/${consentId}/${purposeId}`;
     const res = await fetchOk(url, Config.authOpts());
-    return res;
+    return res.json();
   }
 };
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -535,8 +535,8 @@ export const Election = {
     return await res.json();
   },
 
-  findDataAccessElectionReview: async (electionId, isFinalAccess) => {
-    const url = `${await Config.getApiUrl()}/electionReview/access/${electionId}?isFinalAccess=${isFinalAccess}`;
+  findDataAccessElectionReview: async (electionId) => {
+    const url = `${await Config.getApiUrl()}/electionReview/access/${electionId}`;
     const res = await fetchOk(url, Config.authOpts());
     return await res.json();
   },
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -543,8 +543,8 @@ export const Election = {
 
   // RP Election Information. Can be null for manual review DARs.
   // N.B. We get the rpElectionReview from the Access election id, not the rp election id. This is a legacy behavior.
-  findRPElectionReview: async (electionId, isFinalAccess) => {
-    const url = `${await Config.getApiUrl()}/electionReview/rp/${electionId}?isFinalAccess=${isFinalAccess}`;
+  findRPElectionReview: async (electionId) => {
+    const url = `${await Config.getApiUrl()}/electionReview/rp/${electionId}`;
     const res = await fetchOk(url, Config.authOpts());
     if (res.status === 204) {
       return {};
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -705,7 +705,7 @@ export const Match = {
   findMatch: async (consentId, purposeId) => {
     const url = `${await Config.getApiUrl()}/match/${consentId}/${purposeId}`;
     const res = await fetchOk(url, Config.authOpts());
-    return res.json();
+    return res;
   }
 };
 
--- a/src/pages/ReviewResults.js
+++ b/src/pages/ReviewResults.js
@@ -65,9 +65,7 @@ export default function ReviewResults(props) {
     try {
       matchData = await Match.findMatch(consent.consentId, accessElection.referenceId);
       setMatchData(matchData);
-      console.log(matchData);
     } catch (e) {
-      console.log(e);
       Notifications.showError({ text: `Something went wrong trying to get match algorithm results. Error code: ${e.status}` });
     }
   };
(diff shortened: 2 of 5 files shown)


src/libs/ajax.js
```diff
@@ -705,7 +705,7 @@ export const Match = {
   findMatch: async (consentId, purposeId) => {
     const url = `${await Config.getApiUrl()}/match/${consentId}/${purposeId}`;
     const res = await fetchOk(url, Config.authOpts());
-    return res.json();
+    return res;
```
raejohanek: including .json() causes an unexpected end of json input error for me, which is why I removed it but I know that Justin recently did a bug fix on this method part of which was adding .json() back, so I don't mean to mess that up but I don't really know what's going on then if that works well for you guys

src/pages/ReviewResults.js
```diff
@@ -64,6 +58,16 @@ export default function ReviewResults(props) {
...
+    //get the duos matching algorithm decision
+    let matchData;
+
+    try {
+      matchData = await Match.findMatch(consent.consentId, accessElection.referenceId);
+      setMatchData(matchData);
+    } catch (e) {
+      Notifications.showError({ text: `Something went wrong trying to get match algorithm results. Error code: ${e.status}` });
+    }
   };
```
raejohanek, line 71: combined two methods so that I could include the retrieval of matchData and have the two arguments I need in the same place

src/pages/access_review/VoteSummary.js
```diff
@@ -4,8 +4,8 @@ import { Email } from '../../libs/ajax';
 import { Notifications } from '../../libs/utils';
 import { Theme } from '../../libs/theme';
 import { Chart } from 'react-google-charts';
-import * as fp from 'lodash/fp';
 import * as moment from 'moment';
+import {isNil, filter, map} from "lodash/fp";
```
raejohanek, line 9: refactor star imports

src/libs/ajax.js
JVThomas: You need to call res.json() for any request handled with fetchOK in order to parse the json response. 

Here's the Response itself
Screen Shot 2021-04-30 at 1 25 48 PM

And here's the parsed json from that response (```res.json()``` result)
Screen Shot 2021-04-30 at 1 27 35 PM

So if you just pass in the response and save it as ```matchData``` in the components, variables reads like ```matchData.match``` will just return undefined.

That will ultimately affect the formatted DUOS Algorithm result string here (within ```DarApplication.js```)
Screen Shot 2021-04-30 at 1 34 22 PM

raejohanek: okay thank you that makes a lot of sense but any thoughts on why that causes an unexpected end of input error for me and how to get around that

JVThomas: Unexpected end of input usually means a poorly formatted json. Which election is throwing that error?

JVThomas:
On another note, this DAR on dev has a registered Chair/Final vote, but the Committee Chair section remains empty. 

Screen Shot 2021-05-03 at 7 38 38 AM

raejohanek:
> On another note, this DAR on dev has a registered Chair/Final vote, but the Committee Chair section remains empty.
> 
> Screen Shot 2021-05-03 at 7 38 38 AM

nice find I will look into this

raejohanek:
> > On another note, this DAR on dev has a registered Chair/Final vote, but the Committee Chair section remains empty.
> > Screen Shot 2021-05-03 at 7 38 38 AM
> 
> nice find I will look into this

This issue is because of the backend api call, I just merged in my 1215 PR in consent which corrects this api call, so this issue should be resolved shortly

src/libs/ajax.js
raejohanek: for example, DAR-435-A-1

rushtong: Please move ☝🏽  conversation to slack for debugging.

rushtong:
I'm seeing two issues that should probably be addressed. Additionally, the final display needs approval from @solideoglori so let's make sure we get his 👍🏽  on the final result.

### Missing chair votes
I am only seeing DARs that don't have chairperson votes displayed. Not sure why the votes are not coming through, but since this PR is trying to separate those out for display, this should be looked into.

Screen Shot 2021-05-04 at 1 23 32 PM

### API call with `isFinalAccess`
There are two calls to get the same election review, one with `isFinalAccess` and one without. These should be condensed into a single call without that query param:

Screen Shot 2021-05-04 at 1 24 13 PM

JVThomas:
Screen Shot 2021-05-04 at 2 05 54 PM

I can see the chair votes on my end (locally). Rae mentioned that she had recently merged her consent PR, so that recent update may be a factor.

rushtong:
@raejohanek @JVThomas 
> I can see the chair votes on my end (locally). Rae mentioned that she had recently merged her consent PR, so that recent update may be a factor.

I can now see chair votes for DAR-4. DARs 0, 1, and 2 are showing no chair votes, so I think this is just a matter of missing votes on old elections. The second issue of two election review API calls for this page still needs to be looked at.

raejohanek:
> I'm seeing two issues that should probably be addressed. Additionally, the final display needs approval from @solideoglori so let's make sure we get his 👍🏽 on the final result.
> 
> ### Missing chair votes
> I am only seeing DARs that don't have chairperson votes displayed. Not sure why the votes are not coming through, but since this PR is trying to separate those out for display, this should be looked into.
> 
> Screen Shot 2021-05-04 at 1 23 32 PM
> 
> ### API call with `isFinalAccess`
> There are two calls to get the same election review, one with `isFinalAccess` and one without. These should be condensed into a single call without that query param:
> 
> Screen Shot 2021-05-04 at 1 24 13 PM

So if you look at the two calls you will see that they are distinct despite being similar. One retrieves the rpElectionReview (using the access election id) and the other retrieves the accessElectionReview (also using the access election id). So I just removed the query param from the rpElection call

rushtong (approved):
Once tests pass, 👍🏽

JVThomas (approved):
Code looks good, chair votes are now populating at their sections.

PR #1071, opened 2021-05-03
[DUOS-978] Admin Manage Institutions

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-978

**Changes**
- Add New Admin Page for managing institutions
- Add Routing and new Admin Console Box to direct users to new page
- Add Upsert Modal for Institutions
- Add InstitutionTable component for list of institutions

In all cases:


raejohanek (approved):
this page looks great! everything you did is so clean, my only comments are not based on your work, but the mockups on the ticket, I wonder about the set up of the page - with the add modal at the bottom. I kind of prefer the set up of the dataset catalog page how the search bar and add dataset feature are on the same line above the table, what are your/others thoughts on this? Also, do we want a delete institution feature on this page eventually/do we have a ticket for that?

rfricke-asymmetrik:
For the add button I tend to agree with you. My only thought as to the usefulness for having it sit under the table is that it could force the admin to check the table for existing institutions prior to adding, otherwise it would most likely be better up top.

rushtong:
There is a bug with error handling. We need to display the error condition that comes back in the case of error responses. The particular error in this case is problematic (we should never return db specific errors) and will require a fix on the back end, but the UI should be able to recover from this.

Screen Shot 2021-05-04 at 7 45 28 AM

rfricke-asymmetrik:
@rushtong newest commit should resolve the error handling on the front end.

src/pages/AdminConsole.js
```diff
@@ -135,7 +135,18 @@ class AdminConsole extends Component {
...
+                  id: 'btn_manageInstitutions',
+                  url: '/admin_manage_institutions',
+                  color: 'common',
+                  title: 'Manage Institutions',
+                  description: 'Add and manage Institutions',
+                  iconName: 'manage-user',
+                  iconSize: 'default',
+                }),
+              ])
+              //consoleBoxPlaceholder
```
rushtong: nit: rm

rushtong (approved):
Looks great. One minor suggestion inline that shouldn't hold up this PR. Filing some follow-on tickets to address missing functionality.

PR #1072, opened 2021-05-03
DUOS-1243[risk=no] Refactored Step 2 Verification Method for DAR Submission

JVThomas:
Addresses [DUOS-1243](https://broadworkbench.atlassian.net/browse/DUOS-1243)

PR updates step 2 verification on ```DataAccessRequestApplication``` to reflect changes to expected DUL keys on datasets.
(Similar changes were made for the verification procedure on the Step 2 component earlier, this PR now updates the Step 4 verification process).

Test by picking datasets on staging (dev mostly has datasets with the older format), and check if Step 4 verification matches that of Step 2

In all cases:

[DUOS-1243]: https://broadworkbench.atlassian.net/browse/DUOS-1243

### DIFF
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -477,7 +477,7 @@ class DataAccessRequestApplication extends Component {
   verifyStep2() {
     //defined attribute keys for dynamic DUL based questions
     const dulInvalidCheck = () => {
-      const activeQuestions = this.state.formData.activeDULQuestions;
+      const activeQuestions = fp.pickBy((isActive) => isActive)(this.state.formData.activeDULQuestions);
       let result = false;
       //mapping of ontology keys to dar established keys
       const dulQuestionMap = {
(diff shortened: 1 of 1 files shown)


rfricke-asymmetrik (approved):
Looks good to me 👍

src/pages/DataAccessRequestApplication.js
```diff
@@ -477,7 +477,7 @@ class DataAccessRequestApplication extends Component {
   verifyStep2() {
     //defined attribute keys for dynamic DUL based questions
     const dulInvalidCheck = () => {
-      const activeQuestions = this.state.formData.activeDULQuestions;
+      const activeQuestions = fp.pickBy((isActive) => isActive)(this.state.formData.activeDULQuestions);
```
rushtong, line 480: This looks like it's creating a new object from `activeDULQuestions` comprised of any truth-y values, no? Am I reading this correctly? If so, are there no cases where a false-y value should also be validated?

rushtong: Follow up point, it might be more legible to move this assignment to where it's used in the conditional. That way the code is physically closer to where it's used and easier to reason about.

raejohanek (approved):
these changes seem to be working well!

src/pages/DataAccessRequestApplication.js
JVThomas: The active questions is the computed data restrictions based on the given datasets (It is not the same as the form data itself. It basically represents whether or not a question should be visible, it is not the checkbox selection (that's on ```formData```. The true/false value is most likely an introduction from the bulk dataset process. Prior to this datasets would only include keys that apply to it (so there wouldn't be keys with false values).

A false value is no different than the key not being present at all, so it made sense to filter them out since there's no reason to evaluate them through the if/else block.

JVThomas: To follow up, a lot of the activeQuestion wonkiness stems from the fact that the keys didn't match between what's expected from Consent and what comes in via dataset. Thus, activeQuestions has to act as an intermediary: It represents the questions in the dataset nomenclature. If it's present there, then the verify method will translate it over the the DAR nomenclature and look for it in the ```formData```.

rushtong: All of this is reasonable, just wanted to be sure we didn't need to look at false values. I can see a need for bridging the gap between DAR and Consent in a more formalized way - let's have a conversation about that.

rushtong (approved):
👍🏽

src/pages/DataAccessRequestApplication.js
JVThomas: Yes, please! I hated implementing this the first time around!

PR #1083, opened 2021-05-07
DUOS-1251[risk=no] Add advisory 1677|react-scripts to audit-ci allowlist

JVThomas:
Addresses [DUOS-1251](https://broadworkbench.atlassian.net/browse/DUOS-1251)

PR adds ```1677|react-scripts>...``` to allowlist since it does not take outside input, making the marked ReDOS vulnerability moot.

In all cases:


rushtong:
Could the jest entries be simplified with `1677|react-scripts>jest` ... ? We don't even use jest. Or is the concern that some other non-dev library would try to use jest in the production build?

JVThomas:
> Could the jest entries be simplified with `1677|react-scripts>jest` ... ? We don't even use jest. Or is the concern that some other non-dev library would try to use jest in the production build?

There's a current issue on the `audit-ci` repo that describes the allowlist not parsing correctly on partial paths, which is what happened to me. Writing out `1677|react-scripts>jest` still caused audit-ci to throw the advisory (and fail the job), however writing out the entire file paths (for each exception) had it working properly, hence the list.

If the allowlist worked properly for partial paths then `1677|react-scripts>jest` would be the way to go.

rushtong (approved):
> If the allowlist worked properly for partial paths then 1677|react-scripts>jest would be the way to go.

Thanks for digging into this, it's a bit of a painful solution, but 👍🏽

PR #1092, opened 2021-05-13
DUOS-1238[risk=no]Add DAR view link to new DAC Chair Console 'unopened' DARs + text desc.

raejohanek:
SCOPE:
- Add links to review results page on non-open election statuses  (Unreviewed, Approved, Denied, Canceled) for the chair console
- Change -- to 'unreviewed' status for dars with no elections

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1238

In all cases:

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -105,7 +105,7 @@ const Routes = (props) => (
       rolesAllowed={[USER_ROLES.chairperson, USER_ROLES.admin]} />
     <AuthenticatedRoute path="/reviewed_cases" component={ReviewedCases} props={props}
       rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
-    <AuthenticatedRoute path="/review_results/:referenceId" component={ReviewResults} props={props}
+    <AuthenticatedRoute path="/review_results/:referenceId/:status?" component={ReviewResults} props={props}
       rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni, USER_ROLES.chairperson]} />
     <Route path="*" component={NotFound} />
   </Switch>
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -106,7 +106,7 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/reviewed_cases" component={ReviewedCases} props={props}
       rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
     <AuthenticatedRoute path="/review_results/:referenceId" component={ReviewResults} props={props}
-      rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
+      rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni, USER_ROLES.chairperson]} />
     <Route path="*" component={NotFound} />
   </Switch>
 );
--- a/src/pages/ReviewResults.js
+++ b/src/pages/ReviewResults.js
@@ -62,11 +62,13 @@ export default function ReviewResults(props) {
     //get the duos matching algorithm decision
     let matchData;
 
-    try {
-      matchData = await Match.findMatch(consent.consentId, accessElection.referenceId);
-      setMatchData(matchData);
-    } catch (e) {
-      Notifications.showError({ text: `Something went wrong trying to get match algorithm results. Error code: ${e.status}` });
+    if (!isNil(accessElection)) {
+      try {
+        matchData = await Match.findMatch(consent.consentId, accessElection.referenceId);
+        setMatchData(matchData);
+      } catch (e) {
+        Notifications.showError({text: `Something went wrong trying to get match algorithm results. Error code: ${e.status}`});
+      }
     }
   };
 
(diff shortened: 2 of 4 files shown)


src/Routes.js
```diff
@@ -106,7 +106,7 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/reviewed_cases" component={ReviewedCases} props={props}
       rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
     <AuthenticatedRoute path="/review_results/:referenceId" component={ReviewResults} props={props}
-      rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni]} />
+      rolesAllowed={[USER_ROLES.admin, USER_ROLES.alumni, USER_ROLES.chairperson]} />
```
rushtong, line 109: I think this should be open to members as well, but we might not have a code path for that yet.

JVThomas:
`review_results` will always error for unreviewed DARs because the `Match.findMatch()` call needs an election id which they (unreviewed DARs) don't have, so you'll need to figure out how to make the Match call when needed.

Kapture 2021-05-13 at 13 28 47

src/pages/ReviewResults.js
```diff
@@ -62,11 +62,13 @@ export default function ReviewResults(props) {
...
-      setMatchData(matchData);
-    } catch (e) {
-      Notifications.showError({ text: `Something went wrong trying to get match algorithm results. Error code: ${e.status}` });
+    if (!isNil(accessElection)) {
+      try {
+        matchData = await Match.findMatch(consent.consentId, accessElection.referenceId);
+        setMatchData(matchData);
+      } catch (e) {
+        Notifications.showError({text: `Something went wrong trying to get match algorithm results. Error code: ${e.status}`});
+      }
```
raejohanek, line 75: by adding the null check for accessElection this code works for unreviewed dars too

rushtong (changes requested):
I think this PR goes a little further than the ticket requires. The ticket doesn't specifically request that "Unreviewed" link to the Review Results page, only that the string "- -" be replaced with "Unreviewed". I can see the logic in taking that step, but I think it might be scope creep. Additionally, if we do increase the scope to what this PR is trying to do, then ReviewResults will need some enhanced logic because when it's unreviewed, it's trying to get an election for a DAR that doesn't have one and there is a 404 in the network panel that we shouldn't be making.

raejohanek:
> I think this PR goes a little further than the ticket requires. The ticket doesn't specifically request that "Unreviewed" link to the Review Results page, only that the string "- -" be replaced with "Unreviewed". I can see the logic in taking that step, but I think it might be scope creep. Additionally, if we do increase the scope to what this PR is trying to do, then ReviewResults will need some enhanced logic because when it's unreviewed, it's trying to get an election for a DAR that doesn't have one and there is a 404 in the network panel that we shouldn't be making.

I just talked to Jonathan about this ticket this morning and confirmed that this is what we want to do. I need to make the changes to the ticket to reflect that because it was a bit unclear. I'll take a look at the logic and try to simplify things for the unreviewed case

raejohanek:
> > I think this PR goes a little further than the ticket requires. The ticket doesn't specifically request that "Unreviewed" link to the Review Results page, only that the string "- -" be replaced with "Unreviewed". I can see the logic in taking that step, but I think it might be scope creep. Additionally, if we do increase the scope to what this PR is trying to do, then ReviewResults will need some enhanced logic because when it's unreviewed, it's trying to get an election for a DAR that doesn't have one and there is a 404 in the network panel that we shouldn't be making.
> 
> I just talked to Jonathan about this ticket this morning and confirmed that this is what we want to do. I need to make the changes to the ticket to reflect that because it was a bit unclear. I'll take a look at the logic and try to simplify things for the unreviewed case

So I just added an optional param to pass to the review results page if the status is unreviewed, then if that optional param is present in the review results page it does not make the call to find the election, what do you think of this approach?

src/pages/ReviewResults.js
```diff
@@ -13,6 +13,7 @@ const SECTION = {
 
 export default function ReviewResults(props) {
   const darId = props.match.params.referenceId;
+  const status = props.match.params.referenceId;
```
rushtong: This looks like a copy paste error. Shouldn't it be looking for the `status` param?

src/pages/ReviewResults.js
```diff
@@ -41,11 +42,14 @@ export default function ReviewResults(props) {
...
-      //so there is no vote information to display
+    //dars with unreviewed status will not have an election
+    if (isNil(status)) {
+      try {
+        accessElection = await Election.findElectionByDarId(darId);
+      } catch (error) {
+        //access election is null, this is expected for a dar with an unreviewed election status
+        //so there is no vote information to display
+      }
     }
```
rushtong: If we know that `Unreviewed` will have no election, why are we making the query at all? Could you wrap the call with `if (status !== 'Unreviewed')` ?

rushtong (changes requested):
Looking better. See comments inline.

rushtong (approved):
Thanks for the changes. As fallout from this, I created https://broadworkbench.atlassian.net/browse/DUOS-1265.

PR #1093, opened 2021-05-13
DUOS-1229[risk=no]Restore Vote Statistics Download functions

raejohanek:
SCOPE:
- Refactor getFile so the URI is created in the ajax file rather than on the summary votes page
- Tested with the DUOS-1229 PR in Consent and the Download Stats button does cause a file to download when clicked

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1229

In all cases:

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -668,9 +668,14 @@ export const Files = {
 };
 
 export const Summary = {
-  getFile: async (URI, nameFile) => {
+  getFile: async (fileName) => {
+    const URI = `/consent/cases/summary/file?fileType=${fileName}`;
     const url = `${await Config.getApiUrl()}${URI}`;
-    return await getFile(url, nameFile);
+    if (fileName === 'TranslateDUL') {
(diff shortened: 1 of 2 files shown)


src/libs/ajax.js
```diff
@@ -668,9 +668,14 @@ export const Files = {
 };
 
 export const Summary = {
-  getFile: async (URI, nameFile) => {
+  getFile: async (fileName) => {
+    const URI = `/consent/cases/summary/file?fileType=${fileName}`;
     const url = `${await Config.getApiUrl()}${URI}`;
-    return await getFile(url, nameFile);
+    if (fileName === 'TranslateDUL') {
```
JVThomas, line 674: Instead of using an if-else block, you can create an object that maps the keys to the filenames as a const reference (something like `{'TranslateDUL': 'DUL_summary.tsv'...}`). This way you can get the filename via function argument as well as have an easy way to add new key-filename pairings in the future.

JVThomas (approved):
Minor nitpick, but the code can be merged without it.

PR #1099, opened 2021-05-18
DUOS-766[risk=no]UI: Deprecate old Chair Console

raejohanek:
SCOPE:
Delete chair console
Delete all references to the chair console
Delete feature flag
https://broadworkbench.atlassian.net/browse/DUOS-766

In all cases:


src/libs/config.js
```diff
@@ -27,11 +27,6 @@ export const Config = {
 
   getTag: async () => (await getConfig()).tag,
 
-  getFeatureFlag: async (featureName) => {
-    const feature = _.get(await getConfig(), 'features', {});
-    return _.get(feature, featureName, false);
-  },
-
```
raejohanek: unused

config/base_config.json
```diff
@@ -13,6 +13,5 @@
   "powerBiUrl": "",
   "gaId": "",
   "features": {
-    "newChairConsole": false
```
rushtong, line 16: The values in these configs should somewhat align. I recommend that we add the `"newDarUi": true` as well so other configs are consistent.

src/libs/config.js
rushtong: We do still need this, even if we don't have any A/B features to turn on/off. This looks directly at the `features` in the config.json files.

src/Routes.js
```diff
@@ -73,7 +72,6 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/admin_manage_institutions" component={AdminManageInstitutions} props={props} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/summary_votes" component={SummaryVotes} props={props} rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/researcher_console" component={ResearcherConsole} props={props} rolesAllowed={[USER_ROLES.researcher]} />
-    <AuthenticatedRoute path="/chair_console" component={ChairConsole} props={props} rolesAllowed={[USER_ROLES.chairperson]} />
     <AuthenticatedRoute path="/new_chair_console" component={NewChairConsole} props={props} rolesAllowed={[USER_ROLES.chairperson]} />
```
rushtong: Ideally, we should be redirecting folks to the `/chair_console` url, but the backing component would be `NewChairConsole`. Then, after `ChairConsole` is gone, we should rename `NewChairConsole` -> `ChairConsole` so that all traces of `new` are removed since it is no longer new.

config/configmap.yaml
```diff
@@ -24,7 +24,6 @@ data:
       "powerBiUrl": "",
       "gaId": "",
       "features": {
-        "newDarUi": true,
-        "newChairConsole": false
+        "newDarUi": true
```
rushtong: I don't think we need this value either. I believe this was the feature for the new voting screens for a DAR.

rushtong (changes requested):
See comments inline.

config/configmap.yaml
raejohanek: this comment and the one above on base_config seem conflicting, do we want to get rid of the new DarUI flag because we aren't using it anywhere or add it to the other configs so they're consistent?

rushtong: @raejohanek - You are correct, I am in conflict with myself. I think removing `newDarUi` altogether is right answer.

rushtong (approved):
Thank you for the changes!

PR #1107, opened 2021-05-26
DUOS-1276[risk=no] Refactored updateFinalAccessDarVote to include electionId in vote payload

JVThomas:
Addresses [DUOS-1276](https://broadworkbench.atlassian.net/browse/DUOS-1276)

PR simplifies `updateFinalAccessDarVote` to just use the vote payload rather than just filtering out attributes. The filtered attributes were outdated, causing `electionId` to never be included in the vote object. As a result the final vote would always error out with a `404 Vote could not be logged` (since the election could not be found). Since front-end components are already managing the vote payload and the back-end has to validate any incoming payload, it makes sense to simplify the ajax method.

In all cases:

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1008,19 +1008,10 @@ export const Votes = {
   },
 
   updateFinalAccessDarVote: async (requestId, vote) => {
-    var postObject = {};
-    postObject.vote = vote.vote;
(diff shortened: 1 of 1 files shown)


raejohanek (approved):
this is great, thank you, I just filed a Jira ticket for this exact issue I didn't realize you had already addressed it!

PR #1114, opened 2021-06-01
DUOS-1273[risk=no]Dataset metrics: View

raejohanek:
SCOPE:
Until DUOS-1272 is merged in, this PR can only be viewed with local consent spun up on that branch, so I included a video so it is easier to see.

- Create the DatasetStatistics page
- - display info from mockup, link dar titles to dar summaries
- - add new styles to theme.js
- Link page on Dataset Catalog
- Add page to routes
- Add new apiUrl to config 

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1273
[

https://user-images.githubusercontent.com/77015255/120387895-707f4900-c2f8-11eb-97f6-dcf7328db6ce.mov

](url)

In all cases:

### DIFF
--- a/config/base_config.json
+++ b/config/base_config.json
@@ -3,6 +3,7 @@
   "hash": "",
   "tag": "",
   "apiUrl": "",
+  "metricApiUrl": "",
   "ontologyApiUrl": "",
   "clientId": "",
   "errorApiKey": "",
--- a/src/libs/config.js
+++ b/src/libs/config.js
@@ -7,6 +7,8 @@ export const Config = {
 
   getApiUrl: async () => (await getConfig()).apiUrl,
 
+  getMetricApiUrl: async () => (await getConfig()).metricApiUrl,
+
   getOntologyApiUrl: async () => (await getConfig()).ontologyApiUrl,
 
   getFireCloudUrl: async () => (await getConfig()).firecloudUrl,
--- a/config/configmap.yaml
+++ b/config/configmap.yaml
@@ -14,6 +14,7 @@ data:
       "hash": "",
       "tag": "",
       "apiUrl": "",
+      "metricApiUrl": "",
       "ontologyApiUrl": "",
       "clientId": "",
       "errorApiKey": "",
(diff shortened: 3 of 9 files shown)


rushtong:
FYI, this PR will need corresponding changes to:
* https://github.com/broadinstitute/terra-helmfile/blob/master/terra/values/duos/live/dev.yaml
* https://github.com/broadinstitute/terra-helm/tree/master/charts/duos

This is where we keep env-specific config files. In addition, we'll need to make manual changes to the configs in staging/prod so this doesn't break.

src/libs/config.js
```diff
@@ -7,6 +7,8 @@ export const Config = {
 
   getApiUrl: async () => (await getConfig()).apiUrl,
 
+  getMetricApiUrl: async () => (await getConfig()).metricApiUrl,
+
```
rushtong, line 11: To get around the need for helm chart and configmap updates, we could instead calculate this value since we know that it's the same as `apiUrl` minus the `/api` part of the url. However, I think it would be much cleaner if instead we had a root url for the consent server and then tacked on `/api` or whatever was additionally needed. The current approach is a legacy decision that is a bit of a pain right now.

rushtong (approved):
Love the look of the page. Please DO NOT MERGE this until we've manually updated the configmap values for all environments/helm charts and/or refactored this so that we don't need a brand new config url.

config/base_config.json
```diff
@@ -3,6 +3,7 @@
   "hash": "",
   "tag": "",
   "apiUrl": "",
+  "metricApiUrl": "",
```
rushtong, line 6: There is a dev config that will need a real value. The cypress tests use that to start. The tests are working now because there are no tests that make any calls to this url so it is not throwing any errors.

src/libs/config.js
rushtong: Let's tackle this refactor as a separate ticket so we don't need to hold up the PR on this.

config/base_config.json
rushtong: Look at the dev config in `config`

PR #1117, opened 2021-06-02
[DUOS-622][risk=low] Add sorting to table columns

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-622
**Changes**
- Add utility function for creating sort onClick function
- Modify ResearchConsole to use sort function
- Modify the DarTable component to allow for sorting

In all cases:

### DIFF
--- a/src/components/dar_table/DarTable.js
+++ b/src/components/dar_table/DarTable.js
@@ -1,6 +1,6 @@
 import { isEmpty, isNil, assign } from 'lodash/fp';
 import { useState, useEffect, useCallback } from 'react';
-import { div, h } from 'react-hyperscript-helpers';
+import { div, h, span } from 'react-hyperscript-helpers';
 import { Election, User } from '../../libs/ajax';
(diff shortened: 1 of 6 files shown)


raejohanek (approved):
this looks great and works great! my only question is if we want to do sorting on election status? It's not included in this PR and I don't think it would very be helpful to have an alphabetical sort on election status, but I thought I'd ask

JVThomas:
> this looks great and works great! my only question is if we want to do sorting on election status? It's not included in this PR and I don't think it would very be helpful to have an alphabetical sort on election status, but I thought I'd ask

Yeah, like you said alphabetic sort on election status doesn't really offer much compared to something like dates or titles. On top of that the search bar does a better job filtering out statuses anyway, so I don't mind it being excluded here.

JVThomas (approved):
Tried it out locally with a mix of different filters and sorting orders while updating election statuses via buttons. The DARs sort as expected and update accordingly regardless of list manipulation.

PR #1133, opened 2021-06-15
DUOS-1316[risk=no]Beautify NIH IC Webform page UI

raejohanek:
SCOPE:
- fix styling, make consistent coloring, add icon
- Add states so radio buttons and checkboxes are functional
- Add dropdown options so menus are functional
- Implement the ability to create multiple consent groups and render all created groups
- Update question content per Jonathan's guidelines
- Fix dataset registration agreement download button
- The page is meant to be interactive but not fully functional, it is a marketing tool.

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1316

----

- Verify all tests go green

### DIFF
--- a/src/pages/NIHicWebform.js
+++ b/src/pages/NIHicWebform.js
@@ -1535,7 +1535,10 @@ export default function NIHICWebform() {
                       id: 'submission_3months',
                       label: 'within 3 months of last data generated or last clinical visit',
                       defaultChecked: submissionThreeMonths,
-                      onClick: () => setSubmissionThreeMonths(true)
+                      onClick: () => {
+                        setSubmissionThreeMonths(true);
+                        setSubmissionBatches(false);
+                      }
                     }),
 
                     RadioButton({
@@ -1545,7 +1548,10 @@ export default function NIHICWebform() {
                       id: 'submission_batches',
                       label: 'by batches over Study Timeline (e.g. based on clinical trial enrollment benchmarks)',
                       defaultChecked: submissionBatches,
-                      onClick: () => setSubmissionBatches(true)
+                      onClick: () => {
+                        setSubmissionBatches(true);
+                        setSubmissionThreeMonths(false);
+                      }
                     }),
                   ]),
                 ]),
--- a/src/pages/NIHicWebform.js
+++ b/src/pages/NIHicWebform.js
@@ -1,611 +1,264 @@
-import {Component} from 'react';
+
 import {RadioButton} from '../components/RadioButton';
 import {a, div, fieldset, form, h, h3, hr, input, label, span, textarea} from 'react-hyperscript-helpers';
 import Select from 'react-select';
 import {PageHeading} from '../components/PageHeading';
-import {DAR} from '../libs/ajax';
 import AsyncSelect from 'react-select/async';
+import {Styles} from "../libs/theme";
+import addDatasetIcon from '../images/icon_dataset_add.png';
+import {Theme} from "../libs/theme";
+import {useState, useEffect} from "react";
+import { searchOntologies } from "../libs/utils";
+import DataProviderAgreement from "../assets/Data_Provider_Agreement.pdf";
+import {eRACommons} from "../components/eRACommons";
 
-class NIHICWebform extends Component {
+export default function NIHICWebform() {
+  const [multicenter, setMulticenter] = useState();
+  const [individualControlled, setIndividualControlled] = useState();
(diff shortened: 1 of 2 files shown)


rushtong:
The era commons link is not working, I think because `location` is null and never defined.
image

rushtong:
The `nih_ic_webform` seems to be behind the login now. In the original, that is not the case. I don't see this in the ticket, is that intentional?

raejohanek:
> The `nih_ic_webform` seems to be behind the login now. In the original, that is not the case. I don't see this in the ticket, is that intentional?

No, good catch, that is not intentional, how do I change it back? I did not change anything about Routes or permissions

raejohanek:
> The era commons link is not working, I think because `location` is null and never defined.


Yeah, Jonathan said it does not need to be functional, I thought this was redirecting to the fake era login page, if I can get that working do you think that is a good approach?

rushtong:
> > The nih_ic_webform seems to be behind the login now. In the original, that is not the case. I don't see this in the ticket, is that intentional?

> No, good catch, that is not intentional, how do I change it back? I did not change anything about Routes or permissions

I don't see how this is happening in the code, but it does happen for me locally. Will continue to look into it more as I review.

rushtong:
> > The era commons link is not working, I think because location is null and never defined.

> Yeah, Jonathan said it does not need to be functional, I thought this was redirecting to the fake era login page, if I can get that working do you think that is a good approach?

On dev/staging, the link it points to is a mock version of era commons login. On prod, it is a real one. To make it work, you need to define the `location` variable which is where the auth redirect tries to send the user back to. In this case, it's the web form page.

rushtong:
I see why it is now behind the login. Adding the era commons component did this. On that component, it checks the user info: 

```
  componentDidMount = async () => {
    if (this.props.location !== undefined && this.props.location.search !== '') {
      await this.authenticateAsNIHFCUser(this.props.location.search);
    } else {
      await this.getUserInfo();
    }
  };
```

And since that returns a 401 on the /api/me` call, the application catches that and returns you back to the home page. We can't use the era commons component on an un-authed page since it needs access to the authed user.

raejohanek:
> I see why it is now behind the login. Adding the era commons component did this. On that component, it checks the user info:
> 
> ```
>   componentDidMount = async () => {
>     if (this.props.location !== undefined && this.props.location.search !== '') {
>       await this.authenticateAsNIHFCUser(this.props.location.search);
>     } else {
>       await this.getUserInfo();
>     }
>   };
> ```
> 
> And since that returns a 401 on the /api/me` call, the application catches that and returns you back to the home page. We can't use the era commons component on an un-authed page since it needs access to the authed user.

Okay that makes sense, Thanks!   I think we can probably just remove the button since its not functional anyways but I'll ask Jonathan if he would rather have the page un-authed or have the eraCommons button on it as of now

src/pages/NIHicWebform.js
```diff
@@ -665,1250 +289,1578 @@ class NIHICWebform extends Component {
...
-                        disabled: false,
-                      }),
-                    ]),
-                  ]),
+                  eRACommons({
+                    className: 'col-sm-6 col-xs-12 rp-group',
+                    destination: null,
+                    onNihStatusUpdate: () => {},
+                    location: location
+                  })
```
rushtong: This component requires an auth-ed user to be logged in so any non-auth-ed user will be redirected back to the home page. If this page needs to be public, then we need to replace this with a free text field for the user to enter their era commons id.

rushtong (changes requested):
Aside from the era commons issue, the page looks good and seems to work as intended.

raejohanek:
> Aside from the era commons issue, the page looks good and seems to work as intended.

Okay, I checked with Jonathan to see what he thought and per his suggestion, I replaced the eraCommons button with another button that does nothing

JVThomas (changes requested):
One bug, but other than that the page looks to work as expected. Radio buttons in the gif below don't function properly. Once selected they can never be unselected.

Kapture 2021-06-17 at 14 06 25

raejohanek:
> One bug, but other than that the page looks to work as expected. Radio buttons in the gif below don't function properly. Once selected they can never be unselected.

Good catch thank you, I just fixed that bug and pushed the changes

JVThomas (approved):
Bug is fixed, consent group list updates as expected, buttons and text fields update as expected.

rushtong (approved):
👍🏽

PR #1138, opened 2021-06-21
DUOS-1287[risk=no]Remove usages of Deprecated UserFields from UI

raejohanek:
## SCOPE:
Replace all outdated references to profileName and Institution and replace with displayName and institutionId fields on the user
- update researcher query for reviewresults and accessreview, getResearcherProfile API does not have all the information we need, so instead use getUserById
- - update all appsummary user prop references bc new API call used
- - update all appdownloadlink user prop references bc new API call used
- update ResearcherReview.js references to user prop fields
- update DataAccessRequestRenewal.js references to user prop fields
- update DataAccessRequestApplication.js references to user prop fields
- update DarModal.js references to user prop field institution
- update institution reference in ajax method describedar that is used for memberconsole modals

## ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1287

----

- Verify all tests go green

### DIFF
--- a/src/components/ApplicationDownloadLink.js
+++ b/src/components/ApplicationDownloadLink.js
@@ -5,7 +5,7 @@ import {isNil, isEmpty} from 'lodash/fp';
 import { Theme } from '../libs/theme';
 import {Institution} from "../libs/ajax";
 import {useEffect, useState} from "react";
-import {findPropertyValue, UserProperties } from "../libs/utils";
+import {getPropertyValuesFromUser} from "../libs/utils";
 
 const styles = StyleSheet.create({
   page: {
@@ -155,17 +155,8 @@ export default function ApplicationDownloadLink(props) {
   const internalCollaborators = getCollaborators(darInfo, 'internalCollaborators');
   const externalCollaborators = getCollaborators(darInfo, 'externalCollaborators');
   const labCollaborators = getCollaborators(darInfo, 'labCollaborators');
-  const nihUsernameProp = findPropertyValue(UserProperties.NIH_USERNAME, researcherProfile);
-  const linkedInProp = findPropertyValue(UserProperties.LINKEDIN, researcherProfile);
-  const orcidProp = findPropertyValue(UserProperties.ORCID, researcherProfile);
-  const researcherGateProp = findPropertyValue(UserProperties.RESEARCHER_GATE, researcherProfile);
-  const isThePiProp = findPropertyValue(UserProperties.IS_THE_PI, researcherProfile);
-  const piNameProp = findPropertyValue(UserProperties.PI_NAME, researcherProfile);
-  const departmentProp = findPropertyValue(UserProperties.DEPARTMENT, researcherProfile);
-  const cityProp = findPropertyValue(UserProperties.CITY, researcherProfile);
-  const stateProp = findPropertyValue(UserProperties.STATE, researcherProfile);
-  const location = cityProp.concat(", ").concat(stateProp);
-  const emailProp = researcherProfile.email;
+  const researcherProps = getPropertyValuesFromUser(researcherProfile);
+  const location = (researcherProps.city).concat(", ").concat(researcherProps.state);
   // Use PDFViewer during development to see changes to the document immediately
   // return h(PDFViewer, {width: 1800, height: 800}, [
 
@@ -177,23 +168,23 @@ export default function ApplicationDownloadLink(props) {
         h(View, {style: styles.flexboxContainer}, [
           h(SmallLabelTextComponent, {
             label: "NIH eRA Commons ID",
-            text: `${nihUsernameProp}`,
+            text: `${researcherProps.nihUsername}`,
             style: {marginRight: 30}
           }),
           h(SmallLabelTextComponent, {
             label: "LinkedIn Profile",
-            text: `${linkedInProp}`
+            text: `${researcherProps.linkedIn}`
           })
         ]),
         h(View, {style: styles.flexboxContainer}, [
           h(SmallLabelTextComponent, {
             label: "ORC ID",
-            text: `${orcidProp}`,
+            text: `${researcherProps.orcid}`,
             style: {marginRight: 30}
           }),
           h(SmallLabelTextComponent, {
             label: "ResearcherGate ID",
-            text: `${researcherGateProp}`
+            text: `${researcherProps.researcherGate}`
           }),
         ]),
         h(View, {style: styles.flexboxContainer}, [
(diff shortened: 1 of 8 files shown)


src/components/ApplicationDownloadLink.js
```diff
@@ -1,8 +1,11 @@
 import { Page, Document, StyleSheet, View, /*PDFViewer*/ PDFDownloadLink, Text} from '@react-pdf/renderer';
 import {h, span, i} from 'react-hyperscript-helpers';
 import {DataUseTranslation} from '../libs/dataUseTranslation';
-import isEmpty from 'lodash/fp/isEmpty';
+import {find, isEmpty} from 'lodash/fp';
 import { Theme } from '../libs/theme';
+import { isNil }from "lodash";
```
rushtong: We should try to be consistent and pull `isNil` from `lodash/fp`. When in doubt, prefer the fp version.

src/components/ApplicationDownloadLink.js
```diff
@@ -140,7 +155,20 @@ export default function ApplicationDownloadLink(props) {
   const internalCollaborators = getCollaborators(darInfo, 'internalCollaborators');
   const externalCollaborators = getCollaborators(darInfo, 'externalCollaborators');
   const labCollaborators = getCollaborators(darInfo, 'labCollaborators');
-
+  const nihUsernameProp = find({propertyKey: 'nihUsername'})(researcherProfile.researcherProperties);
```
rushtong: I'd love to see all string values like `'nihUsername'` extracted to constants in a util class. Might be a little beyond the scope of this PR, so maybe we can add a tech-debt story to address this. Another way to do this would be to extract all of these into a separate method that handles the extraction by taking in the properties and returning a full object. Something like `const researcherProps = utils.parseProfileProps(researcherProfile) { ... };`

src/components/modals/DarModal.js
```diff
@@ -61,6 +59,22 @@ const requiresManualReview = (darDetails) => {
...
+  const [institution, setInstitution] = useState();
+
+  useEffect(() => {
+
+    const returnInstitution = async (researcher) => {
+      const institutionId = researcher.institutionId;
+      if (isNil(institutionId)) {
+        setInstitution('- -');
+      } else {
+        const institute = await Institution.getById(institutionId);
```
rushtong, line 71: The fact that we need to do this means we should instead be updating the API that gets a researcher to get the full institution alongside all other properties. Can you add a tech-debt story for this? The API we need to update is `/api/user/{userId}`. The UI work will be to try and remove any unnecessary calls to `Institution.getById(institutionId)` once the full institution is part of the user response.

src/pages/ResearcherReview.js
```diff
@@ -73,7 +79,9 @@ class ResearcherReview extends Component {
     }
 
     this.setState(prev => {
-      prev.formData = researcher;
+      prev.user = user;
```
rushtong, line 71: I'm seeing some errors here when clicking on a researcher review link:
```
TypeError: Cannot read property 'name' of undefined
ResearcherReview.render
src/pages/ResearcherReview.js:188
  185 | div({ className: "row margin-top-20" }, [
  186 |   div({ className: "col-lg-12 col-md-12 col-sm-12 col-xs-12" }, [
  187 |     label({ className: "control-label" }, ["Institution Name"]),
> 188 |     div({ id: "lbl_profileInstitution", className: "control-data", name: "profileInstitution", readOnly: true}, [institution.name]),
      | ^  189 |   ])
  190 | ]),
  191 | 
View compiled
▶ 19 stack frames were collapsed.
_callee$
src/pages/ResearcherReview.js:81
  78 |   status = false;
  79 | }
  80 | 
> 81 | this.setState(prev => {
     | ^  82 |   prev.user = user;
  83 |   prev.institution = institution;
  84 |   prev.formData = researcherProps;
View compiled
▶ 5 stack frames were collapsed.
```
Screen Shot 2021-06-22 at 8 47 30 AM

src/pages/access_review/AppSummary.js
```diff
@@ -1,13 +1,15 @@
...
 import { Theme } from '../../libs/theme';
 import { download } from '../../libs/utils';
 import { ApplicationSection } from './ApplicationSection';
 import { StructuredDarRp } from '../../components/StructuredDarRp';
 import { ApplicantInfo } from './ApplicantInfo';
 import { DownloadLink } from '../../components/DownloadLink';
 import { DataUseTranslation } from '../../libs/dataUseTranslation';
-import isNil from "lodash/fp";
+import {isNil} from "lodash";
+import {find} from "lodash/fp";
```
rushtong: similar lodash feedback, stick with `fp` for consistency

src/pages/access_review/AppSummary.js
```diff
@@ -36,10 +38,20 @@ export const AppSummary = hh(class AppSummary extends React.Component {
...
     };
   }
 
+  getInstitutionFromProfile = async () => {
+    const institute = await Institution.getById(this.props.researcherProfile.institutionId);
+    this.setState(prev => {
+      prev.institution = institute.name;
+      return prev;
+    });
+  };
```
rushtong, line 53: Here's another case where we should be getting the full institution from the user API instead of querying for it directly. So, not for this PR, but for the tech-debt one that should make this cleaner.

src/pages/access_review/AppSummary.js
```diff
@@ -50,11 +62,23 @@ export const AppSummary = hh(class AppSummary extends React.Component {
...
+    const isThePiProp = find({propertyKey: 'isThePI'})(researcherProfile.researcherProperties);
+    const piNameProp = find({propertyKey: 'piName'})(researcherProfile.researcherProperties);
+    const piName = isNil(isThePiProp) ? "- -" : isThePiProp.propertyValue ?
+      researcherProfile.displayName : isNil(piNameProp) ? "xx" : piNameProp.propertyValue;
+    const departmentProp = find({propertyKey: 'department'})(researcherProfile.researcherProperties);
+    const department = isNil(departmentProp) ? "" : departmentProp.propertyValue;
+    const cityProp = find({propertyKey: 'city'})(researcherProfile.researcherProperties);
+    const city = isNil(cityProp) ? "" : cityProp.propertyValue;
+    const countryProp = find({propertyKey: 'country'})(researcherProfile.researcherProperties);
+    const country = isNil(countryProp) ? "" : countryProp.propertyValue;
```
rushtong: This is a case where we should have an extracted method to handle this since we're using the same strings as keys.

rushtong (changes requested):
Good work so far, please see comments inline.

src/pages/ResearcherReview.js
```diff
@@ -51,20 +53,37 @@ class ResearcherReview extends Component {
...
+      city: findPropertyValue(UserProperties.CITY, user),
+      state: findPropertyValue(UserProperties.STATE, user),
+      country: findPropertyValue(UserProperties.COUNTRY, user),
+      isThePI: findPropertyValue(UserProperties.IS_THE_PI, user),
+      havePI: findPropertyValue(UserProperties.HAVE_PI, user),
+      piName: "",
+      piEmail: "",
+      pubmedID: findPropertyValue(UserProperties.PUBMED_ID, user),
+      scientificURL: findPropertyValue(UserProperties.SCIENTIFIC_URL, user)
+    };
```
rushtong: This is almost the exact same code as we have in `ApplicationDownloadLink` (line 158) with some minor variations. It would be nice to make a call to a utils method that returned all of this content for the user as a single construct.

src/pages/ResearcherReview.js
rushtong: Looks like this is fixed, thank you!

rushtong (approved):
One optional suggestion inline that would clean things up a bit, but don't let that hold up the PR.

src/pages/ResearcherReview.js
raejohanek: Okay, I just made a helper in utils, so I'm re-requesting review

rushtong (approved):
This looks exactly like what I had envisioned, thank you for the changes!

PR #1143, opened 2021-06-24
DUOS-1326[risk=no]Update Manage Institution page "next" button styling

raejohanek:
SCOPE:
- Make the institution name column much wider so the likelihood of text overflow decreases
- Hide text overflow and show ellipsis instead of letting text go to second line

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1326
----

- Verify all tests go green

### DIFF
--- a/src/libs/theme.js
+++ b/src/libs/theme.js
@@ -266,6 +266,21 @@ export const Styles = {
       justifyContent: "left",
       alignItems: "center",
     },
+    ID_CELL: {
+      width: "5%",
+      margin: "0 1%",
+      display: "flex",
+      justifyContent: "left",
+      alignItems: "center",
+    },
+    INSTITUTION_CELL: {
+      maxWidth: "35%",
+      minWidth: "35%",
+      margin: "0 1%",
+      display: "flex",
+      justifyContent: "left",
+      alignItems: "center",
+    },
     FOOTER: {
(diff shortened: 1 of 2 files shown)


JVThomas (approved):
Looks good, cell height is stable and the table stays at the same height (unless the last page has fewer rows, but that's expected behavior).

src/libs/theme.js
```diff
@@ -266,6 +266,21 @@ export const Styles = {
...
+      alignItems: "center",
+    },
+    INSTITUTION_CELL: {
+      maxWidth: "35%",
+      minWidth: "35%",
+      margin: "0 1%",
+      display: "flex",
+      justifyContent: "left",
+      alignItems: "center",
+    },
```
rushtong, line 283: I have mixed opinions on this being here. I can go either way, so this shouldn't hold up the PR. My thought is that if something is very specific to a single page, then the styling should be there and merged into what we get from the generic theme which effectively overrides or supersedes any conflicting properties. This seems like one of those cases where the institution cell is specific to the manage institutions page. However, if you foresee other pages making use of this style, then I can see the point in adding it here.

rushtong: The downside of putting every possible styling into this theme is that it ends up being a hodge-podge of styles that become very difficult to use.

PR #1146, opened 2021-06-28
DUOS-1068[risk=no] Added AdminManageLC console

JVThomas:
Addresses [DUOS-1068](https://broadworkbench.atlassian.net/browse/DUOS-1068)

PR adds Admin LC console, includes new component for table, page, modal, and ajax links. Also includes slight changes to `ConfirmationModal` and `SearchSelect` to make it compatible with this console (also includes updates to previous implementations of said components).  PR also includes refactors that move some shared functions into `utils`.

PR is tied to [DUOS-1323](https://github.com/DataBiosphere/consent/pull/1131) due to corrections/updates to LC endpoints within consent.

- Verify all tests go green


raejohanek (changes requested):
table looks great, delete, update, and search all work as expected, one change with add Library card - the library cards created using this button do not have the userName field populated, the name is displayed in the selector and the cards have the userId so I think this information is already accessible it just needs to be added to the card before sending the payload to the API to insert. also, this may not require any changes, but these users created using the button don't have the eraCommonsId populated this may not be an issue at all because the eraCommonsId is updated when a user logs in, but I thought I'd point it out in case you think it should be populated here

JVThomas:
> table looks great, delete, update, and search all work as expected, one change with add Library card - the library cards created using this button do not have the userName field populated, the name is displayed in the selector and the cards have the userId so I think this information is already accessible it just needs to be added to the card before sending the payload to the API to insert. also, this may not require any changes, but these users created using the button don't have the eraCommonsId populated this may not be an issue at all because the eraCommonsId is updated when a user logs in, but I thought I'd point it out in case you think it should be populated here

I can pull in the user name (it's under `displayName`), but there's no eraCommonsId being sent in the users payload. This 

> table looks great, delete, update, and search all work as expected, one change with add Library card - the library cards created using this button do not have the userName field populated, the name is displayed in the selector and the cards have the userId so I think this information is already accessible it just needs to be added to the card before sending the payload to the API to insert. also, this may not require any changes, but these users created using the button don't have the eraCommonsId populated this may not be an issue at all because the eraCommonsId is updated when a user logs in, but I thought I'd point it out in case you think it should be populated here

I can pull in the name (it's saved under `displayName`) but the eraCommonsId isn't being sent from the back-end, so I can't fill it in. This can be solved once [DUOS-1095](https://broadworkbench.atlassian.net/browse/DUOS-1095) is implemented. Until then I'll just add in the variable assignment (I'll be checking on `eraCommonsId` for now, but I can update the key later if needed).

JVThomas:
I've updated the branch to include the userName/eraCommonsId additions as well as a bug fix. To recreate the bug (easy to do this before updating your branch locally), save a Library Card on the console, reload the page, and then try to delete the card you just made. The bug will have the delete request go through, but will not have the table update.

raejohanek (approved):
Looks really good!

rushtong:
nit: revert the changes to `package-lock.json` since we don't have any new dependencies. This will get updated along with other library updates.

PR #1150, opened 2021-06-30
DUOS-1093[risk=no]Update DAR to use Researcher's SO

raejohanek:
SCOPE:
This PR adds a 'creatable' drop down menu that users can select their SO from and add their SO to if their SO is not on the list. This PR uses the API created in DUOS-1093 in consent and therefore must be merged after that PR. Since  DUOS-1093 in consent is not merged yet, this PR currently requires that local consent is running on branch DUOS-1093 in order to be tested.

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1093

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -954,7 +954,7 @@ export const User = {
   },
 
   getSOsForCurrentUser: async () => {
-    const url = `${await Config.getApiUrl()}/api/user/SigningOfficials`;
+    const url = `${await Config.getApiUrl()}/api/user/signing-officials`;
     const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), { method: 'GET' }]));
     return res.json();
   }
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -951,6 +951,12 @@ export const User = {
     const url = `${await Config.getApiUrl()}/api/dacuser/status/${userId}`;
     const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(userRoleStatus), { method: 'PUT' }]));
     return res.json();
+  },
+
+  getSOsForCurrentUser: async () => {
+    const url = `${await Config.getApiUrl()}/api/user/SigningOfficials`;
+    const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), { method: 'GET' }]));
+    return res.json();
   }
 
 };
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -34,7 +34,6 @@ class DataAccessRequestApplication extends Component {
       showDialogSubmit: false,
       showDialogSave: false,
       step: 1,
-      allSigningOfficials: [],
       formData: {
         datasets: [],
         darCode: null,
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -188,7 +188,7 @@ class DataAccessRequestApplication extends Component {
   async getSOs()  {
     const allSOs = await User.getSOsForCurrentUser();
     const allSOsOptions = map((user) =>  {
-      return {value: user.displayName, label: user.displayName};
+      return {value: user.userId, label: user.displayName};
     })(allSOs);
     return allSOsOptions;
   }
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -18,6 +18,7 @@ import { isEmpty, isNil, assign } from 'lodash';
 import { isFileEmpty } from '../libs/utils';
 import './DataAccessRequestApplication.css';
 import headingIcon from '../images/icon_add_access.png';
+import {map} from "lodash/fp";
 
 class DataAccessRequestApplication extends Component {
   constructor(props) {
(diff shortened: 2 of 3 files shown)


src/libs/ajax.js
```diff
@@ -951,6 +951,12 @@ export const User = {
     const url = `${await Config.getApiUrl()}/api/dacuser/status/${userId}`;
     const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(userRoleStatus), { method: 'PUT' }]));
     return res.json();
+  },
+
+  getSOsForCurrentUser: async () => {
+    const url = `${await Config.getApiUrl()}/api/user/SigningOfficials`;
```
rushtong: fix url

src/pages/DataAccessRequestApplication.js
```diff
@@ -184,12 +186,22 @@ class DataAccessRequestApplication extends Component {
...
+  }
+
   async init() {
     const { dataRequestId } = this.props.match.params;
     let formData = {};
     const researcher = await User.getMe();
+    const signingOfficials = await this.getSOs();
     this.setState(prev => {
       prev.researcher = researcher;
+      prev.allSigningOfficials = signingOfficials;
```
rushtong, line 203: I think this is an acceptable approach. I think the problem we had in the earlier iteration was that the api call was inside a `useEffect` that was bundled with a lot of other dependencies. If instead it were pulled out into its own separate `useEffect` with an empty array of dependencies, then there would be no triggers on re-render, and it would not re-call the api every time a change happened.

raejohanek: Re-requesting review now that the options load and save as expected

rushtong:
We have a unique problem here with distinguishing between multiple SOs. In the dev Broad case, I added two SOs with different IDs, but the same display name. When one is selected and saved, the UI shows both as selected. We need to update the signing officials endpoint to include a little more distinguishing information, such as email address. The UI will need be able to figure out which one of the two is the selected one from the user id, not the display name.

## Selection
Screen Shot 2021-07-08 at 7 48 06 AM

## After refresh
Screen Shot 2021-07-08 at 7 49 31 AM

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -292,14 +311,17 @@ export default function ResearcherInfo(props) {
...
             div({className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group'}, [
-              input({
-                type: 'text',
-                defaultValue: signingOfficial,
-                name: 'signingOfficial',
-                required: true,
-                className: isEmpty(signingOfficial) && showValidationMessages ? 'form-control required-field-error' : 'form-control',
-                onBlur: (e) => formFieldChange({name: 'signingOfficial', value: e.target.value})
-              }),
+              <Creatable
```
rushtong, line 321: Unrelated to the scope of this PR. We are already using `AsyncSelect` for all of our selection menus. Adding `Creatable` here seems redundant.

rushtong: Update ... ignore this comment - did not realize they are in the same library.

rushtong (approved):
Looks good and works as expected 👍🏽  
On a related note, we might need to revisit the SO's endpoint functionality, but that is out of scope for this PR.

raejohanek:
> We have a unique problem here with distinguishing between multiple SOs. In the dev Broad case, I added two SOs with different IDs, but the same display name. When one is selected and saved, the UI shows both as selected. We need to update the signing officials endpoint to include a little more distinguishing information, such as email address. The UI will need be able to figure out which one of the two is the selected one from the user id, not the display name.

Since we already have the userId which is a distinguishing feature, I think could just use that as the options value on the UI which should solve the issue shown in the picture. Right now each option has the name as the label and the value, but I could keep the label as the name and update the value to be the userId so they are distinct. Also a different potential issue - currently we save the signing official on the dar as just a string representing the name, this could have the same ambiguity. If the SO exists in the system do we want to save their userId on the dar instead? @rushtong

rushtong:
> Right now each option has the name as the label and the value, but I could keep the label as the name and update the value to be the userId so they are distinct. Also a different potential issue - currently we save the signing official on the dar as just a string representing the name, this could have the same ambiguity. If the SO exists in the system do we want to save their userId on the dar instead?

I think this is a good solution. We continue storing whatever is entered as the value in string form on the DAR. If it's an id, we can look for that in the UI select. If it's not an id, we can just display it. It should be an easy check to see if the saved value matches one of the SO ids.

raejohanek:
> I think this is a good solution. We continue storing whatever is entered as the value in string form on the DAR. If it's an id, we can look for that in the UI select. If it's not an id, we can just display it. It should be an easy check to see if the saved value matches one of the SO ids.

Okay, sounds good. Some fallout from this would be now the SO saved on the dar may be a number and not a name, so if we display that information anywhere (I don't see any usages anywhere besides the pages in this PR on the UI, but maybe in the download pdf?) then we may be displaying the userId. Should I file a ticket to update the SO in the downloaded application pdf? @rushtong

rushtong:
> (I don't see any usages anywhere besides the pages in this PR on the UI, but maybe in the download pdf?) then we may be displaying the userId. Should I file a ticket to update the SO in the downloaded application pdf?

Yes, let's file a ticket for that, thank you!

raejohanek:
> Yes, let's file a ticket for that, thank you!
Okay sounds good, re-requesting review no that the issue in the screenshots you posted is resolved and userId is saved in place of name for the SOs that are existing users

rushtong (approved):
Still 👍🏽

PR #1151, opened 2021-07-01
DUOS-1328[risk=no]Use new getUsers by RoleName API for Admins on UI

raejohanek:
SCOPE: 
- update user.list ajax method to point towards new UserResource API created in DUOS-1260 so that must be merged first
- this method is only used on the AdminManageUsers page currently, so add the "Admin" role param there

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1328

----

- Verify all tests go green

### DIFF
--- a/src/pages/AdminManageUsers.js
+++ b/src/pages/AdminManageUsers.js
@@ -8,6 +8,7 @@ import { PaginatorBar } from '../components/PaginatorBar';
 import { SearchBox } from '../components/SearchBox';
 import { User } from '../libs/ajax';
 import manageUsersIcon from "../images/icon_manage_users.png";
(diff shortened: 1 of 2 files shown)


src/pages/AdminManageUsers.js
```diff
@@ -29,7 +29,7 @@ class AdminManageUsers extends Component {
   }
 
   async getUsers() {
-    const users = await User.list();
+    const users = await User.list("Admin");
```
rushtong: Have a look at `utils.js`, `USER_ROLES` where we have constants for these values.
```suggestion
    const users = await User.list(USER_ROLES.admin);
```

rushtong (approved):
One minor optimization, then 👍🏽

PR #1158, opened 2021-07-07
[DUOS-1332][risk=no] Update LC display

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1332

Note to reviewer, suggest hiding whitespace changes.

## Changes
* With the update to the library card model, the display for users that had LCs threw an error. This PR fixes that error.
* Minor update to `ReviewResults` `useEffect` to remove a console warning about unused dependencies.

As a complication of this, we now have slightly different functionality. We only display the LC for the institution the user is a member of. If there are multiple LCs with different institutions for a single user, we will ignore them. Logically, this makes sense since we now associate users to a signing official based on their institution, and LCs are issued by signing officials, so in practice, a user will not have more than one. If they do, they are not relevant as they are associated to an institution the user is not a member of.

----

- Verify all tests go green

### DIFF
--- a/src/pages/ReviewResults.js
+++ b/src/pages/ReviewResults.js
@@ -25,56 +25,54 @@ export default function ReviewResults(props) {
   const [matchData, setMatchData] = useState();
 
   useEffect(() => {
-    setData(darId);
+    const setData = async (darId, status) => {
+      //get information associated with the access request
+      const {datasets, darInfo, consent, researcherProfile} = await getDarData(darId);
+      setDarInfo(darInfo);
+      setConsent(consent);
+      setDatasets(datasets);
+      setResearcherProfile(researcherProfile);
 
-  }, [darId]);
+      //get information about access election and research purpose election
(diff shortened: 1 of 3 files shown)


raejohanek (commented):
this PR fixes the errors that we saw on dev 👍 I think maybe it could use a style update for the library card / institution name display since it looks a little odd with a long institution name, what do you think?
Screen Shot 2021-07-07 at 11 35 33 AM

rushtong:
> it looks a little odd with a long institution name, what do you think?

I agree, the name is a bit long. Let me see if I can make it wrap lines. The update I did do was to center the image/text and pad it a little more so the background green wasn't too jarring.

rushtong:
@raejohanek: Latest fix: 
Screen Shot 2021-07-07 at 11 43 08 AM

raejohanek (approved):
update fixes bug and looks good!

PR #1188, opened 2021-07-20
DUOS-1136[risk=no]SO: Manage Data Access Request component

raejohanek:
SCOPE:
- add dar table to SO Console 
- create SO console type and remove election actions column 
- create skeleton loader for SO
- NOTE: If we think the spacing on the right side of table looks weird/excessive I can update the darTable and skeleton loaders to account for that extra space by widening other columns. 

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1136

----

- Verify all tests go green

### DIFF
--- a/src/components/dar_table/DarElectionRecords.js
+++ b/src/components/dar_table/DarElectionRecords.js
@@ -28,7 +28,9 @@ const goToReviewResults = (dar, history, status) => {
 const electionStatusTemplate = (consoleType, dar, election, recordTextStyle, votes, showVotes, history) =>{
   const linkedStatuses = ['Unreviewed', 'Approved', 'Denied', 'Canceled'];
   const status = processElectionStatus(election, votes, showVotes);
-  const includeLink = (consoleType === consoleTypes.MANAGE_ACCESS || linkedStatuses.includes(status));
+  const includeLink = (consoleType === consoleTypes.MANAGE_ACCESS ||
+    consoleType === consoleTypes.SIGNING_OFFICIAL ||
+    linkedStatuses.includes(status));
   const tag = includeLink ? a : div;
   return tag({
     style: Object.assign({}, Styles.TABLE.ELECTION_STATUS_CELL, recordTextStyle, {
@@ -73,16 +75,18 @@ export default function DarElectionRecords(props) {
       div({style: Object.assign({}, Styles.TABLE.SUBMISSION_DATE_CELL, recordTextStyle)}, [getElectionDate(election)]),
       div({style: Object.assign({}, Styles.TABLE.DAC_CELL, recordTextStyle)}, [dac ? dac.name : '- -']),
       electionStatusTemplate(consoleType, dar, election, recordTextStyle, votes, showVotes, history),
-      h(DarTableActions, {
-        baseStyle: Object.assign({}, Styles.TABLE.ELECTION_ACTIONS_CELL, recordTextStyle),
-        updateLists,
-        openConfirmation,
-        history,
-        electionInfo,
-        index,
-        consoleType,
-        extraOptions
-      })
+      div({isRendered: consoleType !== consoleTypes.SIGNING_OFFICIAL}, [
+        h(DarTableActions, {
+          baseStyle: Object.assign({}, Styles.TABLE.ELECTION_ACTIONS_CELL, recordTextStyle),
+          updateLists,
+          openConfirmation,
+          history,
+          electionInfo,
+          index,
+          consoleType,
+          extraOptions
+        })
+      ])
     ]);
   });
 }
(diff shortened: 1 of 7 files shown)


src/components/dar_table/DarElectionRecords.js
```diff
@@ -73,16 +75,18 @@ export default function DarElectionRecords(props) {
...
-        baseStyle: Object.assign({}, Styles.TABLE.ELECTION_ACTIONS_CELL, recordTextStyle),
-        updateLists,
-        openConfirmation,
-        history,
-        electionInfo,
-        index,
-        consoleType,
-        extraOptions
-      })
+      div({isRendered: consoleType !== consoleTypes.SIGNING_OFFICIAL}, [
```
JVThomas: Adding the `div` here causes the table to break in other pages

Screen Shot 2021-07-21 at 9 15 52 AM

You can add the `isRendered` condition in the config object for `DarTableActions` instead of having the wrapper div.

JVThomas (dismissed):

rushtong:
FYI, I'm seeing that same keys warning on this page:
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `DarTable`. See https://reactjs.org/link/warning-keys for more information.
    at div
    at DarTable (http://localhost:3000/static/js/main.chunk.js:11180:30)
    at div...
```

JVThomas:
It's the table header, the cells need keys. Just add a key attribute to the divs with styles in the `getTableHeaderTemplateWithSort` function (in DarTable)

src/components/dar_table/DarTable.js
```diff
@@ -87,21 +88,26 @@ export const tableHeaderTemplate = [
...
     "Election Status",
     span({ className: 'glyphicon sort-icon glyphicon-sort' })
-  ]),
-  div({style: Styles.TABLE.ELECTION_ACTIONS_CELL}, ["Election Actions"])
+  ])
 ];
+
+export const tableHeaderTemplate = soTableHeaderTemplate.concat(
+  div({style: Styles.TABLE.ELECTION_ACTIONS_CELL}, ["Election Actions"]));
+
```
rushtong: Nit, naming. It looks like we have a set of columns that we're defining as SO columns, then concatenating the election action columns on top of that. By defining the root set of columns as SO columns, we're putting them into a role-based category which isn't how we want to be thinking of them. This will become more convoluted as we look at other role-specific headers, i.e. researchers and data custodians come to mind.

I think this could use some brainstorming. One idea is if `tableHeaderTemplate` were a function, we could pass in an optional role and then granularly build up the divs one by one with checks on the roles.

JVThomas: > Nit, naming. It looks like we have a set of columns that we're defining as SO columns, then concatenating the election action columns on top of that. By defining the root set of columns as SO columns, we're putting them into a role-based category which isn't how we want to be thinking of them. This will become more convoluted as we look at other role-specific headers, i.e. researchers and data custodians come to mind.
> 
> I think this could use some brainstorming. One idea is if `tableHeaderTemplate` were a function, we could pass in an optional role and then granularly build up the divs one by one with checks on the roles.

We could make the cells an opt-in sort of thing by roles. Role constants would store an array of key names (that correspond to specific cells) that would be passed into a function. That function would then use the key names to get the appropriate template cells (stored in an const object in the function) and process them to produce the template.

rushtong (changes requested):
Aside from the header construction, this is working nicely

src/components/dar_table/DarTable.js
raejohanek: Okay I just turned both of the header and row templates into functions that render only the appropriate columns based on consoleType

src/components/dar_table/DarTable.js
```diff
@@ -49,58 +50,61 @@ export const getTableHeaderTemplateWithSort = (sortFunc, descOrder) => {
...
-    div({style: Styles.TABLE.ELECTION_STATUS_CELL}, [
+    div({style: Styles.TABLE.ELECTION_STATUS_CELL, key: "election_status_cell" }, [
       "Election Status"
     ]),
-    div({style: Styles.TABLE.ELECTION_ACTIONS_CELL}, ["Election Actions"])
+    div({style: Styles.TABLE.ELECTION_ACTIONS_CELL, key: "election_actions_cell", isRendered: consoleType !== consoleTypes.SIGNING_OFFICIAL}, ["Election Actions"])
+  ];
+};
+
+export const tableHeaderTemplate = (consoleType) =>  {
```
rushtong, line 67: I had thought of asking if we could make consoleType nullable. That way existing calls that don't need to distinguish on console type wouldn't need to change. However, I think making it required adds clarity to the method, so 👍🏽  for this approach

rushtong (approved):
Thank you for the changes!

PR #1193, opened 2021-07-23
DUOS-1336[risk=no]UI: Consolidate Console Design

raejohanek:
SCOPE: 
- create dropdown menu of consoles
- make links into reusable constants, and add logic for when to render
- remove sublist (manage Dars and manae Dacs) when in the 'your consoles' dropdown  and render them in the main list, slightly indented

ADDRESSES:
- https://broadworkbench.atlassian.net/browse/DUOS-1336

----

- Verify all tests go green

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -230,6 +230,29 @@ class DuosHeader extends Component {
       isSigningOfficial = currentUser.isSigningOfficial;
     }
 
+    const hasTwoOrMoreRoles = () => {
+      let roles = 0;
+      if (isAdmin) {
+        roles += 1;
+      }
+      if (isResearcher) {
+        roles += 1;
+      }
+      if (isMember) {
+        roles += 1;
+      }
+      if (isChairPerson) {
+        roles += 1;
+      }
+      if (isDataOwner) {
+        roles += 1;
+      }
+      if (isSigningOfficial) {
+        roles += 1;
+      }
+      return roles >= 2;
+    };
+
     const dropdownLinks = {
       statistics: {
         'Votes Statistics': {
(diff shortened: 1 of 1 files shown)


JVThomas (dismissed):
Make the `DAC Chair Console` text bolder to make it stick out as a list header and also to show that it's different from the other options (which are links).
Screen Shot 2021-07-26 at 8 19 13 AM

src/components/DuosHeader.js
```diff
@@ -370,7 +370,7 @@ class DuosHeader extends Component {
 
     const chairSubHeader =
       li({ isRendered: isChairPerson, style: {marginTop: '10px', marginBottom: '5px'} }, [
-        span({style: { color: '#00609f', fontSize: '12px', padding: '10px 15px'}}, ['DAC Chair Console'])
+        span({style: { color: '#00609f', fontSize: '12px', fontWeight: '500', padding: '10px 15px'}}, ['DAC Chair Console'])
```
JVThomas: Use a font-weight that's heavier than 500 (600 is considered semi-bold, 800 is bold) for a more noticeable effect.

src/components/DuosHeader.js
```diff
@@ -230,6 +230,29 @@ class DuosHeader extends Component {
...
+        roles += 1;
+      }
+      if (isDataOwner) {
+        roles += 1;
+      }
+      if (isSigningOfficial) {
+        roles += 1;
+      }
+      return roles >= 2;
+    };
```
rushtong: This can definitely be a lot simpler. The `currentUser` has a list of roles on it ... we just need to count the distinct rolenames from that list.

rushtong: Here's how to get all of the distinct role names ... 

```
import { isNil, uniq, map } from 'lodash/fp';
...
      const roleNames = uniq(map("name")(currentUser.roles));
      console.log(JSON.stringify(roleNames));
```
That gives my superadmin user the following on dev:
```
["Alumni","Admin","Researcher","DataOwner","Chairperson","SigningOfficial"]
```

rushtong:
It looks good, works well, but I do have a couple of ideas that might make it better.
* The "DAC Chair Console" header looks clickable. Not sure how to make it not look that way, maybe removing the blue color from the text and bolding it a little more
* Clicking on the header when it's not a link closes the menu. Not sure how to resolve that problem. PR should not be held up on that minor point though.

rushtong (changes requested):
See comments inline. Looks very good!

raejohanek:
> * The "DAC Chair Console" header looks clickable. Not sure how to make it not look that way, maybe removing the blue color from the text and bolding it a little more
> * Clicking on the header when it's not a link closes the menu. Not sure how to resolve that problem. PR should not be held up on that minor point though.

I'm not sure how to resolve that second point either, but I made the text bolder and black, I think it looks better. Do you think it would be helpful if we changed the text to something like DAC Chair Console Options instead of just DAC Chair Console?

rushtong:
@raejohanek 
> Do you think it would be helpful if we changed the text to something like DAC Chair Console Options instead of just DAC Chair Console?

I think the current text is better.

rushtong (approved):
Looks great!

PR #1196, opened 2021-07-26
DUOS-1359[risk=no]Send Reminder link visible in Approved DARs, Review Results

raejohanek:
SCOPE:
- in DarApplication (used by both AccessReview page and ReviewResults page), pass in finalDecision variable (already defined) to VoteSummary component
- in VoteSummary, only render the send reminder link if the finalDescision is "no decision", it can be one of: "no", "yes" or "no decision"

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1359

----

- Verify all tests go green

### DIFF
--- a/src/pages/access_review/VoteSummary.js
+++ b/src/pages/access_review/VoteSummary.js
@@ -155,7 +155,8 @@ export const VoteSummary = hh(
           padding: '1rem',
           textAlign: 'right',
           flexShrink: '0',
-          fontWeight: Theme.font.weight.regular}
+          fontWeight: Theme.font.weight.regular},
+        isRendered: this.props.finalDecision === "No decision"
         }, [a({onClick: () => this.sendReminder(vote)}, ['Send Reminder'])])
       ]);
     };
--- a/src/pages/access_review/VoteSummary.js
+++ b/src/pages/access_review/VoteSummary.js
@@ -156,7 +156,7 @@ export const VoteSummary = hh(
           textAlign: 'right',
           flexShrink: '0',
           fontWeight: Theme.font.weight.regular},
-        isRendered: this.props.finalDecision === "No decision"
+        isRendered: isNil(this.props.accessElection) ? false : this.props.accessElection.status === "Open"
         }, [a({onClick: () => this.sendReminder(vote)}, ['Send Reminder'])])
       ]);
     };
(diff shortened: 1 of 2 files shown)


src/pages/access_review/VoteSummary.js
```diff
@@ -155,7 +155,8 @@ export const VoteSummary = hh(
           padding: '1rem',
           textAlign: 'right',
           flexShrink: '0',
-          fontWeight: Theme.font.weight.regular}
+          fontWeight: Theme.font.weight.regular},
+        isRendered: this.props.finalDecision === "No decision"
```
rushtong: This seems like a very brittle condition. This text can easily change at some point and that will have a downstream effect here that will be difficult to track. I think it makes more sense to condition this on the state of the election. If the data access election is open, we should render the link. It doesn't make sense to send reminders in any other case (closed, canceled, approved, denied, etc.)

rushtong (changes requested):
See comments inline. Works well, but I think we should be passing in the access election and then checking on `isRendered: this.props.election.status === "Open"` instead of the final vote decision string.

src/pages/access_review/VoteSummary.js
```diff
@@ -155,7 +155,8 @@ export const VoteSummary = hh(
           padding: '1rem',
           textAlign: 'right',
           flexShrink: '0',
-          fontWeight: Theme.font.weight.regular}
+          fontWeight: Theme.font.weight.regular},
+        isRendered: isNil(this.props.accessElection) ? false : this.props.accessElection.status === "Open"
```
rushtong, line 159: Have not tested different cases, but this might throw a null error on the election if the user visited this page for an unreviewed election or if the status is null ... not sure if that's a real case. Might be worth checking for validity first.

rushtong (changes requested):
Changes look good, could use some null checking to prevent page-errors.

src/pages/access_review/VoteSummary.js
raejohanek: there are three main cases - unreviewed where the election is null, caught by the null check, (the entire vote boxes are not rendered), or there is an election and it has a status that is either Open or Closed, if there is an election the status is not null

rushtong (approved):
👍🏽

PR #1204, opened 2021-07-30
DUOS-1360[risk=no]Manage Institutes: Add SOs to page

raejohanek:
SCOPE:
- This PR adds a column for SOs to the admin manage institutions page, displays their name and email.
- This PR uses the updated DAO call from DUOS-1360 in Consent so it needs to be tested with that branch
Screen Shot 2021-07-29 at 5 03 20 PM

ADDRESSES:
- https://broadworkbench.atlassian.net/browse/DUOS-1360

----

- Verify all tests go green

### DIFF
--- a/src/components/institution_table/InstitutionTable.js
+++ b/src/components/institution_table/InstitutionTable.js
@@ -74,14 +74,13 @@ export default function InstitutionTable(props) {
       div({style: Styles.TABLE.CONTAINER}, [
         div({style: Styles.TABLE.HEADER_ROW}, [tableHeaderTemplate]),
         filteredList.slice((currentPage - 1) * tableSize, (currentPage * tableSize)).map((inst, index) => {
-          let signingOfficials = '';
-          isNil(inst.signingOfficials) ? signingOfficials = '' : inst.signingOfficials.forEach((user, i) => {
-            signingOfficials = signingOfficials.concat(user.displayName, ' (', user.email, ')');
-            //if there are multiple SOs in the list and this is not the last one, add comma
-            if (inst.signingOfficials.length > 1 && i < (inst.signingOfficials.length - 1)) {
-              signingOfficials = signingOfficials.concat(', ');
-            }
-          });
+          let signingOfficialsList = [];
+          if (!isNil(inst.signingOfficials)) {
+            inst.signingOfficials.forEach((user, i) => {
+              signingOfficialsList.push(user.displayName.concat(' (', user.email, ')'));
+            });
+          }
+          const signingOfficialsText = signingOfficialsList.join(', ');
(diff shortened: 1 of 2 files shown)


src/components/institution_table/InstitutionTable.js
```diff
@@ -72,6 +74,14 @@ export default function InstitutionTable(props) {
       div({style: Styles.TABLE.CONTAINER}, [
         div({style: Styles.TABLE.HEADER_ROW}, [tableHeaderTemplate]),
         filteredList.slice((currentPage - 1) * tableSize, (currentPage * tableSize)).map((inst, index) => {
+          let signingOfficials = '';
+          isNil(inst.signingOfficials) ? signingOfficials = '' : inst.signingOfficials.forEach((user, i) => {
+            signingOfficials = signingOfficials.concat(user.displayName, ' (', user.email, ')');
+            //if there are multiple SOs in the list and this is not the last one, add comma
+            if (inst.signingOfficials.length > 1 && i < (inst.signingOfficials.length - 1)) {
+              signingOfficials = signingOfficials.concat(', ');
+            }
```
rushtong: Another way to approach this is to generate an array of strings, each one representing an SO. Then, just join the array with the `, ` and you'll get the same effect as using an indexed loop.

src/components/institution_table/InstitutionTable.js
```diff
@@ -74,14 +74,13 @@ export default function InstitutionTable(props) {
...
-            signingOfficials = signingOfficials.concat(user.displayName, ' (', user.email, ')');
-            //if there are multiple SOs in the list and this is not the last one, add comma
-            if (inst.signingOfficials.length > 1 && i < (inst.signingOfficials.length - 1)) {
-              signingOfficials = signingOfficials.concat(', ');
-            }
-          });
+          let signingOfficialsList = [];
+          if (!isNil(inst.signingOfficials)) {
+            inst.signingOfficials.forEach((user, i) => {
+              signingOfficialsList.push(user.displayName.concat(' (', user.email, ')'));
```
JVThomas: A cleaner way to write this string is to use template strings. So instead of a concat statement, you can type ``` `${user.displayName} (${user.email})` ```

JVThomas (changes requested):

JVThomas (changes requested):
The code is cleaner, but now we can see the names are awkwardly listed because all of the names are contained within a single span. For the names to be listed (each in their own line), you'll have to output each in their own span tag.

Screen Shot 2021-08-06 at 10 06 46 AM

JVThomas (approved):
Looks good. SOs are listed neatly within the cell

PR #1211, opened 2021-08-03
DUOS-1334[risk=no]User Profile makes too many user calls

raejohanek:
## Please read before reviewing code
- This reduces the number of API calls made on the researcherProfile page to populate information
- - Instead of calling getResearcherProfile and getUserByEmail, the page now only calls getMe
- This removes the API call from the eraCommons component
- - this component was calling getMe, but now the response from getMe is passed in
- - this component is used on the researcherProfle page and the Dar Application page (in ResearcherInfo section) and it was used on the dar renewal page, but instead of updating this page I deleted it on Jonathan's recommendation 
- - this includes a minor change to ResearcherInfo, passing the getMe response to the eraCommons component, and adding an isLoading check to ensure the currrentUser has been received before passing it to the eraCommons component
- This reduces the number of API calls (and simplifies unneeded logic) made on the researcherProfile page to submit information
- - This removes repeated calls to User.update, Researcher.createProperties, and Researcher.updateProperties and replaces them with one call to User.update and one call to Researcher.updateProperties on submit or save. 
- - This removes the numerous submit/save methods with one that works for all scenarios: Deletes saveProperties, saveUser, and updateResearcher (which all duplicated the work of the others). And condenses dialogHandlerSubmit and dialogHandlerSave into dialogHandlerUpdateuser. This eliminates the isResearcher distinction and completed distinction which were distinctions made in the code but the effect of the underlying api calls for all cases were the same.
- This also simplifies classnames, 'col-lg-12 col-md-12 col-sm-12 col-xs-12' = 'col-xs-12' because the styles automatically scale up

TODO: Address warning " A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component."

## Addresses
- https://broadworkbench.atlassian.net/browse/DUOS-1334
- https://broadworkbench.atlassian.net/browse/DUOS-1165 (eliminates isResearcher distinction)

----

- Verify all tests go green


src/libs/utils.js
```diff
@@ -25,7 +25,8 @@ export const UserProperties = {
   COUNTRY: "country",
   RESEARCHER_GATE: "researcherGate",
   PUBMED_ID: "pubmedID",
-  SCIENTIFIC_URL: "scientificURL"
+  SCIENTIFIC_URL: "scientificURL",
+  COMPLETED: "completed"
```
raejohanek: add completed because this property was used on the userprofile page

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -328,12 +328,6 @@ export default function ResearcherInfo(props) {
                 getNewOptionData: (inputValue) => { //formats user input into object for use within Creatable
                   return { displayName: inputValue };
                 },
-                getOptionValue: (option) => { //value formatter for options, attr used to ensure empty strings are treated as undefined
```
raejohanek: unused

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -40,19 +40,14 @@ export default function ResearcherInfo(props) {
...
-    isAnvilUseInvalid
+    isAnvilUseInvalid,
+    currentUser
   } = props;
 
   const navButtonContainerStyle = {
     marginTop: '5rem'
   };
 
-  const soDropDownStyle = {
```
raejohanek: unused

src/pages/ResearcherProfile.js
```diff
@@ -162,6 +126,52 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
 
+  handleChangeDisplayName = (e) => {
+    const value = e.target.value;
+    this.setState(prev => {
+      prev.currentUser.displayName = value;
+      return prev;
+    });
+  };
+
+  handleAdditionalEmailChange = (e) => {
```
raejohanek, line 139: handleAdditionalEmailChange, handleCheckboxChange, are not new they have just been moved to keep all the handle methods in one place

src/pages/ResearcherProfile.js
```diff
@@ -480,41 +402,41 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
-              description: this.state.isResearcher ? 'Please complete the following information to be able to request access to dataset(s)' : ''
+              description: isResearcher ? 'Please complete the following information to be able to request access to dataset(s)' : ''
             }),
             hr({ className: 'section-separator' })
           ]),
-          div({ className: 'col-lg-10 col-lg-offset-1 col-md-10 col-md-offset-1 col-sm-12 col-xs-12 no-padding' }, [
+          div({ className: 'col-md-10 col-md-offset-1 col-xs-12 no-padding' }, [
             form({ name: 'researcherForm' }, [
               div({ className: 'form-group' }, [
-                div({ className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12' }, [
```
raejohanek, line 505: classNames simplified, has no effect

src/libs/utils.js
```diff
@@ -36,7 +37,7 @@ export const findPropertyValue = (propName, researcher) => {
 };
 
 export const getPropertyValuesFromUser = (user) => {
-  let researcherProps = {
+  return {
```
raejohanek, line 41: rm unneeded variable per eslint recommendation

src/pages/ResearcherProfile.js
```diff
@@ -350,124 +322,74 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
-  };
-
-  handleCheckboxChange = (e) => {
-    const value = e.target.checked;
-    this.setState(prev => {
-      prev.profile.checkNotifications = value;
-      return prev;
-    });
+  saveProfile = (event) => {
+    event.preventDefault();
```
raejohanek, line 354: saveProfile and submit are not new they have just been moved to keep the submit/save handling in one place

src/components/eRACommons.js
```diff
@@ -36,10 +36,8 @@ export const eRACommons = hh(class eRACommons extends React.Component {
   authenticateAsNIHFCUser = async (searchArg) => {
     let isFcUser = await this.verifyUser();
     if (!isFcUser) {
-      await User.getMe().then(
-        (response) => isFcUser = this.registerUserToFC(response),
-        () => this.setState({ nihError: true })
-      );
+      isFcUser = this.registerUserToFC(this.props.currentUser);
```
JVThomas: The `.then` function parameters are two functions, the first is a callback function for a success, the second a callback method for an error. So `nihError` being set true only happened when the async request failed.

For your new flow you can keep that setState call, just use it when an error occurs. Easy way to do this is to wrap the function body in a try-catch block, with the setState happening in the catch.

Another note: registerUserToFC is an async function, so you may need the await keyword to have the function wait for the Promise before moving on.

src/libs/utils.js
```diff
@@ -1,12 +1,12 @@
 import Noty from 'noty';
 import 'noty/lib/noty.css';
 import 'noty/lib/themes/bootstrap-v3.css';
-import { forEach } from 'lodash';
-import { DAR, DataSet } from "./ajax";
-import {Theme, Styles } from "./theme";
-import { find, first, map, isEmpty, filter, cloneDeep, isNil, toLower, includes, sortedUniq } from "lodash/fp";
+import {forEach} from 'lodash';
 import _ from 'lodash';
+import {DAR, DataSet} from "./ajax";
```
JVThomas: You can group these ajax module references with the User one below this line

JVThomas (changes requested):
I've made some comments in the code on some corrections, but there's also this bug below...

Kapture 2021-08-05 at 14 43 42

As you can see, you can't clear the eraCommonsId once it's been saved. My guess is that the old code must have called `User.getMe` after the eraCommons clear/log-out to update the currentUser attributes (and thus clearing the saved eraCommonsId).  However that call is now removed, so you'll have to manually update the attribute on the front-end.

src/components/eRACommons.js
raejohanek: good catch, thanks. I also just noticed that in registerUserToFC if the async call errors it already sets the nihError state to true so we don't need to set it again here at all.

src/libs/ajax.js
```diff
@@ -806,6 +806,7 @@ export const Researcher = {
     return await res.json();
   },
 
+  //unused, if we don't want this API I can file a ticket to deprecate the APIs
```
rushtong: Yes, please, file a tech-debt ticket to remove this API

src/libs/ajax.js
```diff
@@ -818,6 +819,7 @@ export const Researcher = {
     return res.json();
   },
 
+  //unused, if we don't want this API I can file a ticket to deprecate the APIs
```
rushtong: Yes, please, file a tech-debt ticket to remove this API

src/libs/utils.js
```diff
@@ -1,12 +1,11 @@
 import Noty from 'noty';
 import 'noty/lib/noty.css';
 import 'noty/lib/themes/bootstrap-v3.css';
-import { forEach } from 'lodash';
-import { DAR, DataSet } from "./ajax";
-import {Theme, Styles } from "./theme";
-import { find, first, map, isEmpty, filter, cloneDeep, isNil, toLower, includes, sortedUniq } from "lodash/fp";
+import {forEach} from 'lodash';
 import _ from 'lodash';
```
rushtong: Not for this PR, but I'd love to clean this file up by migrating all current lodash calls to `lodash/fp`

src/libs/utils.js
rushtong: Good call for this PR, but we should look into removing this field altogether. `Completed` is a calculable field, not something that should be set as a static user property

src/pages/ResearcherProfile.js
```diff
@@ -93,44 +76,25 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
+
+      let profileProperties = getPropertyValuesFromUser(currentUser);
+      prev.profileProperties = profileProperties;
+
       // This ensures that we have a boolean for `checkNotifications`
-      if (!isNil(get(profile, 'checkNotifications', null))) {
-        prev.profile.checkNotifications = get(profile, 'checkNotifications', 'false') === 'true';
+      if (profileProperties.checkNotifications === '') {
+        prev.profileProperties.checkNotifications = false;
       }
```
rushtong: Just a side note here, the `checkNotifications` property refers to sending an additional email to the user's academic/business email property. This is separate and distinct from the optional field of Additional Email, so a user could have up to 3 separate email addresses for notifications. No action needed here.

rushtong (commented):
Going to re-review

raejohanek:
> Going to re-review

Okay I have not had the chance to dig into the authentication issue yet so that is not fixed

src/pages/dar_application/ResearcherInfo.js
```diff
@@ -175,13 +174,14 @@ export default function ResearcherInfo(props) {
...
-          div({ className: 'row no-margin' }, [
+          div({ className: 'row no-margin', isRendered: !isLoading }, [
             eRACommons({
               className: 'col-lg-6 col-md-6 col-sm-6 col-xs-12 rp-group',
               destination: eRACommonsDestination,
               onNihStatusUpdate: onNihStatusUpdate,
               location: location,
-              validationError: showValidationMessages
+              validationError: showValidationMessages,
+              currentUser: currentUser
```
rushtong: In a previous comment, I suggested that you get the current user from `Storage.getCurrentUser()` instead of passing it in as a prop. That would let you revert this and have a single place to get the current user as we know about it.

raejohanek: We could do that, but then we would need to make an API call to get the User with properties because  Storage.getCurrentUser() does not contain user properties and we need the authorization and expiration props

src/components/eRACommons.js
```diff
@@ -55,22 +52,20 @@ export const eRACommons = hh(class eRACommons extends React.Component {
...
-      const eraCommonsId = response.eraCommonsId;
-      this.props.onNihStatusUpdate(nihValid);
-      this.setState(prev => {
-        prev.isAuthorized = isAuthorized;
-        prev.expirationCount = expirationCount;
-        prev.eraCommonsId = isNil(eraCommonsId) ? '' : eraCommonsId;
-        return prev;
-      });
+  getUserInfo = () => {
+    const currentUser = this.props.currentUser;
```
rushtong: Get the user from storage, not props:
```suggestion
    const currentUser = Storage.getCurrentUser();
```

src/components/eRACommons.js
```diff
@@ -111,11 +106,31 @@ export const eRACommons = hh(class eRACommons extends React.Component {
...
+      const nihValid = isAuthorized && expirationCount > 0;
+      const eraCommonsId = response.eraCommonsId;
+      this.props.onNihStatusUpdate(nihValid);
+      this.setState(prev => {
+        prev.currentUser = response;
+        prev.isAuthorized = isAuthorized;
+        prev.expirationCount = expirationCount;
+        prev.eraCommonsId = isNil(eraCommonsId) ? '' : eraCommonsId;
+        return prev;
+      });
```
rushtong, line 132: After having updated the user, we need to update where we store it. Add this line here:
```
      setUserRoleStatuses(response, Storage);
```
That method is in `utils.js` so you'll need an import. Also, we can rename the response to `user` or something that makes more sense.

src/pages/DataAccessRequestApplication.js
```diff
@@ -1047,7 +1047,8 @@ class DataAccessRequestApplication extends Component {
                 isCloudProviderInvalid: this.state.isCloudProviderInvalid,
                 isSigningOfficialInvalid: this.state.isSigningOfficialInvalid,
                 isITDirectorInvalid: this.state.isITDirectorInvalid,
-                isAnvilUseInvalid: this.state.isAnvilUseInvalid
+                isAnvilUseInvalid: this.state.isAnvilUseInvalid,
+                currentUser: this.state.researcher
```
rushtong: rm

rushtong:
There are some deletions that should be in a separate PR. PRs like this that go through rounds of review can hold up other, simpler fixes. Revert:
* src/pages/DataAccessRequestRenewal.js
* src/libs/ajax.js#Researcher.createProperties
* src/pages/DataAccessRequestRenewal.js

src/pages/dar_application/ResearcherInfo.js
rushtong: Then let's make a ticket to store the full user in Storage.setCurrentUser so we have what we need to have. The workarounds we're doing are what's making this a more difficult task.

PR #1212, opened 2021-08-03
DUOS-1389[risk=no]Remove unnecessary API call from Dar Application page

raejohanek:
## SCOPE:
- Remove getPropertiesByResearcherId call and instead use the properties on the already retrieved user from the getMe call

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1389

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -800,6 +800,7 @@ export const PendingCases = {
 
 export const Researcher = {
 
+  //unused
   getPropertiesByResearcherId: async (userId) => {
     const url = `${await Config.getApiUrl()}/api/researcher/${userId}`;
     const res = await fetchOk(url, Config.authOpts());
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -800,7 +800,7 @@ export const PendingCases = {
 
 export const Researcher = {
 
-  //unused
+  //unused do we want to deprecate this API
   getPropertiesByResearcherId: async (userId) => {
     const url = `${await Config.getApiUrl()}/api/researcher/${userId}`;
     const res = await fetchOk(url, Config.authOpts());
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -800,13 +800,6 @@ export const PendingCases = {
 
 export const Researcher = {
 
-  //unused do we want to deprecate this API
-  getPropertiesByResearcherId: async (userId) => {
-    const url = `${await Config.getApiUrl()}/api/researcher/${userId}`;
-    const res = await fetchOk(url, Config.authOpts());
-    return await res.json();
-  },
-
   createProperties: async (researcherProperties) => {
     const url = `${await Config.getApiUrl()}/api/researcher`;
     const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(researcherProperties), { method: 'POST' }]));
--- a/src/pages/DataAccessRequestApplication.js
+++ b/src/pages/DataAccessRequestApplication.js
@@ -233,8 +233,8 @@ class DataAccessRequestApplication extends Component {
     let completed = false;
     if (!fp.isNil(formData.darCode)) {
       completed = '';
-    } else if (rpProperties.completed !== '') {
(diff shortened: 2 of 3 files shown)


src/pages/DataAccessRequestApplication.js
```diff
@@ -185,14 +185,6 @@ class DataAccessRequestApplication extends Component {
     return datasets;
   }
 
-  async getSOs()  {
```
raejohanek, line 188: unused

JVThomas (dismissed):
None of the researcher fields are populating in the DAR application
Screen Shot 2021-08-06 at 10 38 52 AM

Here's the current develop branch in comparison
Screen Shot 2021-08-06 at 10 40 55 AM

NOTE: eraCommonsId works fine, I just wasn't logged in when I took these screenshots.

src/libs/ajax.js
```diff
@@ -800,6 +800,7 @@ export const PendingCases = {
 
 export const Researcher = {
 
+  //unused do we want to deprecate this API
```
rushtong: Yes, we do need to deprecate that API.

src/pages/DataAccessRequestApplication.js
```diff
@@ -213,45 +205,36 @@ class DataAccessRequestApplication extends Component {
       formData = Storage.getData('dar_application') === null ? this.state.formData : Storage.getData('dar_application');
       Storage.removeData('dar_application');
     }
-    let currentUserId = Storage.getCurrentUser().dacUserId;
-    let rpProperties = await Researcher.getPropertiesByResearcherId(currentUserId);
-    formData.darCode = fp.isNil(formData.darCode) ? null : formData.darCode;
-    formData.partialDarCode = fp.isNil(formData.partialDarCode) ? null : formData.partialDarCode;
+    let rpProperties = getPropertyValuesFromUser(researcher);
```
rushtong, line 208: This method doesn't populate the completed property, so add that in.

rushtong (changes requested):
See minor validation bug fix in comment

src/pages/DataAccessRequestApplication.js
```diff
@@ -1028,7 +1011,7 @@ class DataAccessRequestApplication extends Component {
               div({ className: 'dialog-description' },
                 ['Are you sure you want to save this Data Access Request? Previous changes will be overwritten.'])
             ]),
-            div({ isRendered: this.state.step === 1 }, [
+            div({ isRendered: this.state.step === 1 && (this.state.formData.researcher !== '') }, [
```
rushtong: From the console:
```
src/pages/DataAccessRequestApplication.js
  Line 1014:100:  Trailing spaces not allowed  no-trailing-spaces
```

src/pages/DataAccessRequestApplication.js
```diff
@@ -213,45 +205,36 @@ class DataAccessRequestApplication extends Component {
...
+    formData.pubmedId = rpProperties.pubmedID;
+    formData.scientificUrl = rpProperties.scientificURL;
+    formData.userId = researcher.dacUserId;
 
     let completed = false;
     if (!fp.isNil(formData.darCode)) {
       completed = '';
-    } else if (rpProperties.completed !== undefined) {
-      completed = JSON.parse(rpProperties.completed);
+    } else if (rpProperties.completed !== '') {
```
rushtong, line 236: `rpProperties.completed` can still be nil or undefined if `getPropertyValuesFromUser` wasn't able to find a matching property. In fact, I think that's true of all of the `rpProperties` fields, so all of the simplifications above should be reverted.

rushtong (changes requested):
Works well, but we've lost a lot of nil checks that are still necessary.

src/pages/DataAccessRequestApplication.js
raejohanek: if the property is null then the helper returns the empty string.  utils.getPropertyValuesFromUser calls utils.findPropertyValue which does the null checking and replacing

rushtong: Ah, thank you, yes, that works.

rushtong (approved):

PR #1215, opened 2021-08-05
DUOS-1137[risk=no] Added Researcher/Library Card Table to SO Console

JVThomas:
Front-end portion of [DUOS-1137](https://broadworkbench.atlassian.net/browse/DUOS-1137)

PR adds the researcher table to the SO console, allowing SO to create and deactivate library cards for users (or yet to join users) within an institution.

**NOTE:** Certain features are missing and/or not functional due to necessary changes on the back-end, which will be made as a separate PR on Consent. Some of these changes include alterations to how Library Cards are created and destroyed, the alteration of SO User list fetch so that it returns partially empty User models for library cards that are tied to yet-to-join users, and a query that can calculate a User's Active DAR count. 

- Verify all tests go green

### DIFF
--- a/src/components/SimpleButton.js
+++ b/src/components/SimpleButton.js
@@ -1,4 +1,4 @@
-import { div } from 'react-hyperscript-helpers';
+import { button } from 'react-hyperscript-helpers';
 import { useState, useEffect } from 'react';
 
 export default function SimpleButton(props) {
@@ -43,5 +43,5 @@ export default function SimpleButton(props) {
     return baseAttributes;
   };
 
-  return div(getDivAttributes(disabled), [label]);
+  return button(getDivAttributes(disabled), [label]);
 }
--- a/src/components/SimpleButton.js
+++ b/src/components/SimpleButton.js
@@ -2,7 +2,7 @@ import { button } from 'react-hyperscript-helpers';
 import { useState, useEffect } from 'react';
 
 export default function SimpleButton(props) {
-  const { onClick, label, disabled, baseColor, additionalStyle } = props;
+  const { onClick, label, disabled, baseColor, additionalStyle, keyProp } = props;
 
   const updateStyle = (backgroundColor, baseColor, additionalStyle = {}, pointerBool, disabled) => {
     const baseStyle = {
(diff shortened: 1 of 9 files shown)


raejohanek (dismissed):
Screen Shot 2021-08-06 at 9 25 13 AM
The console gives this error when you click add researcher also the issuing of cards does not work. Maybe this should be a draft until your backend PR is in so the functionality of the table can be tested. Search works as expected and table looks good.

src/components/SimpleButton.js
```diff
@@ -18,7 +18,7 @@ export default function SimpleButton(props) {
       cursor: pointerBool ? 'pointer' : 'default'
     };
 
-    const newStyle = Object.assign({}, additionalStyle, baseStyle);
+    const newStyle = Object.assign({}, baseStyle, additionalStyle);
```
rushtong, line 21: Not a request for change, just a comment. My preference is to avoid mutability as much as practically possible which is what `Object.assign` does. This is an example of **safe** usage where the target is an empty object so devs need to be aware of that style. I generally prefer lodash fp merge since it does the same thing, but in a non-mutating manner, and does not have unsafe usages.

src/components/modals/LibraryCardFormModal.js
```diff
@@ -12,11 +12,23 @@ import { Theme } from '../../libs/theme';
 //modal component, used to render row on modal
 const FormFieldRow = (props) => {
   const { card, dropdownOptions, updateInstitution, updateUser, modalType, setCard } = props;
-  const [filteredDropdown, setFilteredDropdown] = useState(dropdownOptions);
+  // signing officials shouldn't be able to pick and choose from solutions
+  // front-end can hide the input by limiting dropdown options to less than two choices
+  // (as a way to impose conditional rendering without adding more prop variables)
+  // however back-end still needs to filter out institutionId
+
+  //NOTE: check if this works as expected
```
rushtong: Does it?

src/components/signing_official_table/SigningOffiicalTable.js
```diff
@@ -0,0 +1,416 @@
...
+        userEmail: researcher.email,
+        institutionId: institutionId
+      },
+      showConfirmationModal
+    });
+
+  return {
+    isComponent: true,
+    id,
+    style: {}, //need to figure out a base style
```
rushtong: nit: rm comment

rushtong (approved):
This looks great! Functional and snappy. I think this could use some pruning of comments, but otherwise, 👍🏽

PR #1221, opened 2021-08-13
[DUOS-1165][DUOS-1344][risk=no] Consolidate user profile calls and unify researcher/admin profile

raejohanek:
## Addresses:
- https://broadworkbench.atlassian.net/browse/DUOS-1165
- https://broadworkbench.atlassian.net/browse/DUOS-1334

## Scope:
- Consolidate user and property calls to User.getMe
- More explicit state-setting for clarity
- Ordering and sorting properties alpha for legibility
- Nil checks and style cleanup

----

- Verify all tests go green


src/pages/ResearcherProfile.js
```diff
@@ -92,37 +90,42 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
-        prev.roles = currentUser.roles;
-      }
-      prev.researcherProfile = profile;
-      prev.profile = profile;
-      // This ensures that we have a boolean for `checkNotifications`
-      if (!isNil(get(profile, 'checkNotifications', null))) {
-        prev.profile.checkNotifications = get(profile, 'checkNotifications', 'false') === 'true';
+        prev.roles = user.roles;
       }
       prev.additionalEmail = user.additionalEmail === null ? '' : user.additionalEmail;
```
JVThomas: `undefined !== null`, so this will pull `user.additionalEmail` even when it's undefined. This will cause the `indexOf` call on around `line 557` to error out. Lodash's `isNil` should work here

src/pages/ResearcherProfile.js
```diff
@@ -592,52 +579,52 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
                       label({ id: 'lbl_profileLibraryCard', className: 'control-label' }, ['Library Cards']),
```
JVThomas, line 577: LibraryCards aren't showing up on the profile page. This could be another outdated feature of the page (I remember it manually fetched LCs with an array of user ids.) Since the profile now gets library cards directly we can just use it directly.

Screen Shot 2021-08-19 at 8 56 25 AM

JVThomas (changes requested):

JVThomas (approved):
Code looks good. Bugs in previous comment no longer show on the profile page

PR #1251, opened 2021-09-07
[DUOS-296][risk=no] Stop building/deploying to deprecated cluster

rushtong:
## Addresses

Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-296

We're migrating from the current dev cluster to a terra dev cluster. This PR turns off the build/deploy to the cluster in the older project so we can be focused on the new one.

----

- Verify all tests go green


JVThomas (approved):
Code makes sense. No reason to keep the older jobs present when we're migrating away to a new cluster.

PR #1255, opened 2021-09-10
[DUOS-1426][risk=no] Query for DARs by SO Role

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1426

See also: https://github.com/DataBiosphere/consent/pull/1259

## Changes
* Minor file rename
* Update the dar query to query by role name. Previously, this would get all dars the user has access to, not necessarily just the ones that the user would see as a signing official

----

- Verify all tests go green


JVThomas (approved):
Code looks good, can confirm that DARs are being fetch correctly on the SO Console.

PR #1268, opened 2021-09-16
[DUOS-1452][risk=no] Dataset Catalog Refactor

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1452

* Converted from class component to function
* Lots of cleanup around item selection
* Fixes select-all datasets -> DAR creation bug
* Updates styling so inactive datasets are non-selectable
* Prevent selection of any inactive datasets
* Removed many unnecessary variables and state checks

----

- Verify all tests go green


src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
+    setDatasetList(selectedDatasets);
   };
 
-  checkSingleRow = (index) => (e) => {
-    let catalog = this.state.dataSetList.catalog;
-    const catalogElement = catalog[index];
-    catalogElement.checked = e.target.checked;
+  const checkSingleRow = (dataset) => (e) => {
+    const checked = e.target.checked;
+    const selectedDatasets = forEach(row => {
```
JVThomas: `forEach` returns the original array's reference. React does state updates on value changes, but if the saved value is the array reference, then changing the array's contents triggers nothing since the reference value is unchanged.

`map` should be enough to trigger the state change since it performs the content manipulation on a new array.

src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
+        // The unselect all case, ensure everything is unselected
+        row.checked = checked;
+      }
+    })(datasetList);
+
+    // Update state
+    setAllChecked(checked);
+    // Note that react seems to understand the arrays are different here without
+    // the need to spread it into a new array.
+    setDatasetList(selectedDatasets);
```
JVThomas, line 251: Technically the arrays are the same here too, but since `allChecked` had a state change the component will re-render regardless. So it's working out but not for the expected reason.

src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
   };
 
-  checkSingleRow = (index) => (e) => {
-    let catalog = this.state.dataSetList.catalog;
-    const catalogElement = catalog[index];
-    catalogElement.checked = e.target.checked;
+  const checkSingleRow = (dataset) => (e) => {
+    const checked = e.target.checked;
```
JVThomas: To be safe, have checked fallback on `false` to avoid `undefined` assignments (which would cause checkboxes to error out since it's expecting a boolean).

src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
+    const selectedDatasets = forEach(row => {
+      if (row.dataSetId === dataset.dataSetId) {
+        if (row.active) {
+          row.checked = checked;
```
JVThomas, line 259: Same checked comment as the previous method.

src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
+        }
+      }
+      row;
+    })(datasetList);
+
+    // Update state - note that we need to ensure react knows the arrays are
+    // different by spreading it. See https://stackoverflow.com/questions/56266575/why-is-usestate-not-triggering-re-render
+    setDatasetList([...selectedDatasets]);
```
JVThomas: If you're using `map` then you won't need to use the spread operator to create a new array.

src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
-                          td({
-                            id: trIndex + '_datasetName', name: 'datasetName',
-                            className: 'cell-size ' + (!dataSet.active ? !!'dataset-disabled' : ''),
-                            style: tableBody
-                          }, [
+              tbody({ isRendered: !isNil(datasetList) && !isEmpty(datasetList) }, [
+                datasetList.filter(searchTable(searchDulText))
+                  .slice((currentPage - 1) * pageSize, currentPage * pageSize)
+                  .map((dataset, trIndex) => {
+                    // console.log(JSON.stringify(dataset.active));
```
JVThomas: Comment cleanup here

src/pages/DatasetCatalog.js
```diff
@@ -148,624 +101,594 @@ class DatasetCatalog extends Component {
...
+        ]),
+        h(TranslatedDulModal,{
+          isRendered: showTranslatedDULModal,
+          showModal: showTranslatedDULModal,
+          dataUse: dataUse,
+          onOKRequest: () => setShowTranslatedDULModal(false),
+          onCloseRequest: () => setShowTranslatedDULModal(false)
+        }),
+
+        ConfirmationDialog({
```
JVThomas, line 610: Doesn't need to be addressed now (should be a different ticket), but this can easily be one modal that updates the action method and text based on the button clicked instead of having four different versions of the same modal.

JVThomas (changes requested):
I left some comments explaining the comments you've left on the selectAll and singleSelect methods. You don't have to implement them since the current functionality works.

The code works as expected, the issue is a matter of expected functionality. When you click the select all button, only the items on the current page is selected. If a user wants to select all the items on a second page, they won't be able to do so. If you hit the button to unselect, ALL checkboxes are unselected.

Kapture 2021-09-24 at 08 13 37


A good example of select all functionality is on gmail. When you hit the button, the select all/unselect all button are only applied to the current page. So if you select all on one page and then turn to an untouched page, the Check All button switches back to an unselected state. 

Gmail has additional states for partial selects, but we can keep it simple. If all are selected, make the Select All button checked. If not, show it as unchecked.

rushtong:
@JVThomas - great feedback all around, will implement!

src/pages/DatasetCatalog.js
rushtong: Done

src/pages/DatasetCatalog.js
rushtong: Done

src/pages/DatasetCatalog.js
rushtong: Done

src/pages/DatasetCatalog.js
rushtong: Since `checked` is now a safe check, this should now be safe as well.

src/pages/DatasetCatalog.js
rushtong: Done

src/pages/DatasetCatalog.js
rushtong: Done

src/pages/DatasetCatalog.js
rushtong: Agreed, this can be cleaned up. Will file a ticket for this.

rushtong: Created https://broadworkbench.atlassian.net/browse/DUOS-1460 to handle this concern.

JVThomas (approved):
Code looks good, the table buttons are working as expected and inaccessible datasets are no longer selectable.

PR #1269, opened 2021-09-16
DUOS-1438[risk=no] Add Demoable Data Custodian Subsection to SO Console

JVThomas:
Addresses [DUOS-1438](https://broadworkbench.atlassian.net/browse/DUOS-1438)

PR adds a Data Custodian sub-component to the SO Console that allows a user to issue/revoke for demo purposes. This means that while the UI will have state updates, there are no updates being saved to the back-end (which means perceived updates will not persist). Demo is only for the SO Console, which means Admin functionality is not available.

This is not meant to be the final version of the sub-component, however the demo will provide a good jumping off point once Data Custodian work begins.
 
- Verify all tests go green

### DIFF
--- a/src/components/modals/DataCustodianFormModal.js
+++ b/src/components/modals/DataCustodianFormModal.js
@@ -1,0 +1,167 @@
+import { useState, useEffect } from 'react';
+import { cloneDeep, isObject, includes, isNil, isEmpty } from 'lodash/fp';
+import { h, div, label } from 'react-hyperscript-helpers';
+import { Styles } from '../../libs/theme';
+import CloseIconComponent from '../CloseIconComponent';
+import Modal from 'react-modal';
+import Creatable from 'react-select/creatable';
+import SimpleButton from '../SimpleButton';
+import { Theme } from '../../libs/theme';
+
+//NOTE: functionality is set for demo purposes on SO Console (no admin functionality)
+//As such, changes to "data" don't persist
+
+const FormFieldRow = (props) => {
+  const {
+    user,
+    dropdownOptions,
+    updateUser,
+    setUser,
+  } = props;
+
+  const nonCustodians = dropdownOptions.filter((option) => {
+    return !option.isDataCustodian;
+  });
+  const [filteredDropdown, setFilteredDropdown] = useState(nonCustodians);
+
+  //filter function for users dropdown
+  const userListFilter = ({ searchTerm, user, setUser, action }) => {
+    let filteredCopy;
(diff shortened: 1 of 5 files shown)


src/pages/SigningOfficialConsole.js
```diff
@@ -73,13 +81,35 @@ export default function SigningOfficialConsole(props) {
...
+            color: 'blue',
+            fontSize: '1.8rem',
+            componentType: tabs.researcher,
+            setSelected: setSelectedTag,
+            selectedType: selectedTag
+          })
+        ]),
+        h(SigningOfficialTable, {isRendered: selectedTag === tabs.researcher, researchers, signingOfficial, unregisteredResearchers, isLoading}, []),
+        h(DataCustodianTable, {isRendered: selectedTag === tabs.custodian, researchers, signingOfficial, unregisteredResearchers, isLoading}, [])
+      ]),
```
rushtong: Can we wrap the new content in a conditional on env? Look at https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/config.js#L6 which should give you the correct environment. See also:
* https://duos-k8s.dsde-dev.broadinstitute.org/config.json
* https://duos-k8s.dsde-staging.broadinstitute.org/config.json
* https://duos-k8s.dsde-prod.broadinstitute.org/config.json

src/components/modals/DataCustodianFormModal.js
```diff
@@ -0,0 +1,167 @@
...
+      getOptionLabel: (option) =>
+        `${option.displayName || 'New User'} (${
+          option.email || 'No email provided'
+        })` || option.email,
+    }),
+  ]);
+  return div({ display: 'flex' }, [template]);
+};
+
+export default function DataCustodianFormModal(props) {
```
rushtong, line 74: Not really feedback for this PR, but just wanted to express my hopes that we can address the modal styling at some point. Will dig around and see if we have any stories in the backlog for this task.

rushtong (commented):
Overall, this looks good. One minor request inline (env-check) and then I'd be happy with merging.

src/pages/SigningOfficialConsole.js
JVThomas: What's the conditional? Don't render on production?

rushtong: Yes, exactly. Staging and dev should be fine.

rushtong (approved):
Thank you for the changes! Pending tests, 👍🏽

PR #1290, opened 2021-10-04
[DUOS-1445][risk=no] Conditional display of institution selection for SOs

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1445

Signing officials are prohibited from updating their institution to prevent users from changing institutions just to issue library cards that might be outside of their normal institution. Signing Officials must go through customer support to accomplish this.

## Changes
* Minor update in background sign-in to reduce code duplication
* Update the rendered institution selection depending on user state.
  * SOs with no institution -> render institution selection
  * Any other user -> render institution selection
  * SOs with institution -> render the existing institution in a disabled state

## SO with Existing Institution
Screen Shot 2021-10-04 at 1 48 44 PM

## Any Other User
Screen Shot 2021-10-04 at 1 48 17 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/BackgroundSignIn.js
+++ b/src/pages/BackgroundSignIn.js
@@ -39,7 +39,7 @@ export default function BackgroundSignIn(props) {
       Storage.setGoogleData({ accessToken: accessToken });
       getUser().then(
         user => {
-          user = Object.assign(user, setUserRoleStatuses(user));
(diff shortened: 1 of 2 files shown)


JVThomas (approved):
Code works great! Tooltip update is a great way to inform SO's why they can't change their institution.

PR #1295, opened 2021-10-06
[DUOS-1461][risk=no] Restrict chair edit for datasets to dac-owned ones

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1461
https://broadworkbench.atlassian.net/browse/DUOS-903

See also https://github.com/DataBiosphere/consent/pull/1292 for back end changes.

Add UI validation to the edit dataset icons for chairpersons. Only functional if the dataset belongs to a DAC that the chairperson has access to.

### Updated Chairperson Catalog View
Note that this covers three cases. 
1. Datasets for which the chairperson has no edit access: all editing icons are disabled
2. Non-deletable datasets for which the chairperson DOES have edit access: icons are enabled, delete is not
3. Deletable datasets for which the chairperson DOES have edit access: icons are enabled, including delete

I have completely removed the link icon for chairs as we will be removing this feature across the board in future epics.

Screen Shot 2021-10-07 at 2 49 34 PM

### Updated Admin Catalog View
This PR also fixes the weird alignment problem some admins would see in https://broadworkbench.atlassian.net/browse/DUOS-903

Screen Shot 2021-10-06 at 5 00 31 PM

----

- Verify all tests go green


rushtong:
NPM Audit failure is a false result. See linked issue for dispute: https://github.com/advisories/GHSA-8p5q-j9m2-g8wr

src/pages/DatasetCatalog.js
```diff
@@ -301,6 +301,17 @@ export default function DatasetCatalog(props) {
     }
   };
 
+  const isEditDatasetEnabled = (dataset) => {
+    if (currentUser.isAdmin) {
+      return true;
+    }
+    // Chairpersons can only edit datasets they have direct access to via their DACs
+    if (currentUser.isChairPerson) {
+      return includes(dataset.dacId)(map('dacId')(filter({name: 'Chairperson'})(currentUser.roles)));
```
JVThomas: `lodash/fp` has a `compose` function that can help make functions like this easier to read

https://hackernoon.com/function-composition-with-lodash-d30eb50153d1

So instead of nested arguments you would have something like...

```
return compose(
    includes(dataset.dacId),
    map('dacId')
    filter({name: 'Chairperson'})
)(currentUser.roles)
```
There's also a `flow` method that has the same end result, it just evaluates the functions differently...
https://www.geeksforgeeks.org/lodash-_-flow-method/

```
return flow(
    filter({name: `Chairperson`}),
    map('dacId'),
    includes(dataset.dacId)
)(currentUser.roles)
```

Basically `flow` executes the innermost function first (listed as the first arg in flow) and applies the result to the outer function (subsequent function in the list)

I don't really have a preference as to what to use, the choice is yours.

JVThomas (approved):
Front-end code looks good, just have one comment (code works without it, so you don't need to address it).

src/pages/DatasetCatalog.js
rushtong: I usually use `flow`, good call! 👍🏽

PR #1315, opened 2021-10-19
[DUOS-1483][risk=no] Fix user profile bugs

rushtong:
## Addresses
Fixes https://broadworkbench.atlassian.net/browse/DUOS-1483

## Changes
* Save user on "Continue later"
* Use user email as the required field in academic business email
* Remove the required from academic/business since it will be defaulted to email
* Clarified requirements for NIH eRA Commons ID
* Removed conditional formatting for Continue Later/Submit buttons

----

- Verify all tests go green

### DIFF
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -126,7 +126,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
 
     return (
       div({ className: this.props.className }, [
-        label({ className: 'control-label' }, ['NIH eRA Commons ID*']),
+        label({ className: 'control-label' }, ['NIH eRA Commons ID ', span({ className: 'italic display-inline' }, ['(Required for DARs)'])]),
         div({
           isRendered: (!this.state.isAuthorized || this.state.expirationCount < 0)
         }, [
--- a/src/components/eRACommons.js
+++ b/src/components/eRACommons.js
@@ -126,7 +126,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
 
     return (
       div({ className: this.props.className }, [
-        label({ className: 'control-label' }, ['NIH eRA Commons ID ', span({ className: 'italic display-inline' }, ['(Required for DARs)'])]),
+        label({ className: 'control-label' }, ['NIH eRA Commons ID*']),
         div({
           isRendered: (!this.state.isAuthorized || this.state.expirationCount < 0)
         }, [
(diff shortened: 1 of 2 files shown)


src/pages/ResearcherProfile.js
```diff
@@ -177,7 +177,7 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
       isThePI = false,
       piEmail = false,
       piName = false,
-      showValidationMessages = false;
+      showValidationMessages = true;
```
JVThomas: You should switch this back to false. The value will flip to true through the validation conditionals below (if the target value is found to be invalid)

MiriamAlter: I thought I undid that change. Good catch.

src/components/eRACommons.js
```diff
@@ -126,7 +126,7 @@ export const eRACommons = hh(class eRACommons extends React.Component {
 
     return (
       div({ className: this.props.className }, [
-        label({ className: 'control-label' }, ['NIH eRA Commons ID*']),
+        label({ className: 'control-label' }, ['NIH eRA Commons ID ', span({ className: 'italic display-inline' }, ['(Required for DARs)'])]),
```
rushtong: We should revert this change for now. The message displayed is true, but this component is also displayed on the DAR application page so adding this text here makes it slightly more confusing there since you're already making a DAR. 

What we should do, in a separate story/PR, is to update this component to take a title as a prop. The places that use this can then provide whatever title text that makes sense for their usage.

Lastly, the term `DAR` is something we use internally to refer to a `Data Access Request`, so using that term here would be confusing to some users, especially international ones.

src/pages/ResearcherProfile.js
```diff
@@ -622,7 +622,7 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
               div({ className: 'form-group' }, [
                 div({ className: 'col-xs-12', style: { 'marginTop': '20px' } }, [
                   label({ className: 'control-label rp-title-question default-color' }, [
-                    'Researcher Identification*',
+                    'Researcher Identification',
```
rushtong: I think we should revert this change as well. The explanation text does a good job of explaining what is required and what isn't.

rushtong (commented):
Nice work, please see comments inline.

rushtong (commented):
Changes look good! 👍🏽 Side note, I can't approve this since I created it originally. Feel free to merge when you're ready.

MiriamAlter (approved):
Leaving a review to enable merging

PR #1328, opened 2021-10-28
[DUOS-1408][risk=no] Overhaul of Researcher Profile page

MiriamAlter:
Issue: https://broadworkbench.atlassian.net/browse/DUOS-1408
- Updated ResearcherProfile.js to React hook format
- Added new dynamic text box to inform researchers of empty required fields
- Simplified previous buttons and dialogs to one all-inclusive "Save" button
- Modified API call to back end if user saves an incomplete profile
- Corrected optional eRA Commons ID field to be PI's rather than user's
- Fixed a bug that stopped library cards from rendering after saving
- Department is no longer a required field

----

- Verify all tests go green


src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
+    havePI: null,
+    piName: '',
+    piEmail: '',
+    piERACommonsID: '',
+    pubmedID: '',
+    scientificURL: '',
+    completed: undefined
+  });
+
+  const [isLoading, setIsLoading] = useState(true);
```
rushtong: Doesn't look like we use `isLoading` any more

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
+    init();
+  }, [props.history]);
 
-  handleChange = (event) => {
-    let field = event.target.name;
-    let value = event.target.value;
+  useEffect(() => {
+    const isValid = (value) => {
+      let isValid = false;
+      if (trim(value.toString()) !== '' && value !== null && value !== undefined) {
```
rushtong: If value is null or undefined, this will throw an error with `value.toString()`. I suggest using Lodash `isNil` and `isEmpty` methods to simplify this.

rushtong: I misspoke here ... `trim` will handle nil and undefined just fine and return an empty string. Maybe wrap that in `isEmpty`?
```suggestion
      if (!isEmpty(trim(value.toString()))) {
```

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
-      prev.profile[field] = value;
-      return prev;
-    }, () => {
-      if (this.state.validateFields) {
-        this.validateUserFields();
+    const isValidNumber = (value) => {
+      let isValid = false;
+      if (value !== 0 && value !== null && value !== undefined) {
+        isValid = true;
       }
```
rushtong: Lodash is your friend here ... https://lodash.com/docs/4.17.15#isNumber

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
-      profileName = true;
-      showValidationMessages = true;
+    if (!isValid(profile.profileName)) {
+      incompletes.push('Name');
     }
 
-    if (!this.isValid(this.state.profile.academicEmail) || this.state.profile.academicEmail.indexOf('@') === -1) {
-      academicEmail = true;
-      showValidationMessages = true;
+    if (!isValid(profile.academicEmail || profile.academicEmail.indexOf('@') === -1)) {
```
rushtong, line 97: This is not a request for this PR, but for a separate one in a separate ticket.

We should have a utility method to validate email addresses. See https://stackoverflow.com/questions/46155/how-to-validate-an-email-address-in-javascript for a potential solution to this. I looked in [Terra's codebase](https://github.com/DataBiosphere/terra-ui), but didn't see anything there that we could use.

rushtong: This is an example of what I would describe as technical debt. There is a correct way to do a particular task, but it is not a direct requirement of the current task, but instead something that we should do to improve the overall health of the codebase.

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
+      incompletes.push('Email Address');
     }
 
-    if (!this.isValidNumber(this.state.institutionId)) {
-      institution = true;
-      showValidationMessages = true;
+    if (!isValidNumber(profile.institutionId)) {
```
rushtong: See https://lodash.com/docs/4.17.15#isNumber

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
-    });
-  };
+    init();
+  }, [props.history]);
 
-  handleChange = (event) => {
-    let field = event.target.name;
-    let value = event.target.value;
+  useEffect(() => {
+    const isValid = (value) => {
```
rushtong: I don't think you need these *valid functions to be inside useEffect. They can exist outside of useEffect since they are all isolated functions that operate on their arguments and don't modify state in any way. The things you DO want in useEffect are the imcompletes and the checks in each of the incomplete fields.

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
-      return prev;
-    });
-  };
+    newFields[field] = value;
 
-  saveProfile = (event) => {
-    event.preventDefault();
-    this.setState({ showDialogSave: true });
-  };
+    if (profile.isThePI === true || profile.isThePI === 'true') {
```
rushtong, line 221: This is a great opportunity for something that checks for [truthy-ness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) but would be smart about the string value `"false"`. Not a request for this PR, but we should have a library method that does something like this in a single call.

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
+      newFields.havePI = '';
+    }
+    if (profile.havePI === false || profile.havePI === 'false' || profile.havePI === '') {
```
rushtong, line 224: Same comment here, but for [falsy-ness](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) instead.

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
-    this.setState({ showDialogSubmit: false });
-    this.props.history.push({ pathname: 'dataset_catalog' });
+  const eraValidate = async () => {
+    const user = await User.getMe();
+    const userProps = getPropertyValuesFromUser(user);
+    const expirationCount = isNil(userProps.eraExpiration) ? 0 : AuthenticateNIH.expirationCount(userProps.eraExpiration);
+
+    return (!isNil(userProps.eraCommonsId) && userProps.eraAuthorized === 'true' && expirationCount >= 0);
+
+    // Temporary fix until eRACommons.js is updated to share re-render info.
```
rushtong: Let's move the comment to the top of the method with a TODO. This is another example of technical-debt which is acceptable to incur, but should be dealt with in a timely manner. Please feel free to create a ticket for this and start on it if you wish.

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
     }
   };
 
-  updateResearcher = async (profile) => {
-    const profileClone = this.cloneProfile(profile);
-    await Researcher.updateProperties(Storage.getCurrentUser().dacUserId, true, profileClone);
-    await this.saveUser();
-    this.setState({ showDialogSubmit: false });
-    this.props.history.push({ pathname: 'dataset_catalog' });
+  const createResearcher = async (profile) => {
```
rushtong: Let's update this method name to reflect what it's really doing. Naming is hard!  I think it should be:
```suggestion
  const createUserProperties = async (profile) => {
```
Additionally, and separately, `Researcher.createProperties(profile);` should be renamed/moved to `User.createProperties(profile);`

src/pages/ResearcherProfile.js
```diff
@@ -1,492 +1,351 @@
...
+    await Researcher.createProperties(profile);
+    await updateUser();
+    props.history.push({ pathname: 'dataset_catalog' });
   };
 
-  saveUser = async () => {
+  const updateResearcher = async (profile) => {
```
rushtong: Similar naming comment applies here.

src/pages/ResearcherProfile.js
rushtong: To be clear, 
> Additionally, and separately, Researcher.createProperties(profile); should be renamed/moved to User.createProperties(profile);

is meant for a separate PR, not this one.

rushtong (commented):
This is my first round of review ... I have not tested functionality yet. Please see comments inline. This is a lot of great work, thank you!

rushtong:
It looks like you'll need to merge the latest version of the develop branch into this one.

src/pages/ResearcherProfile.js
MiriamAlter: I was wondering about that. Fixed

src/pages/ResearcherProfile.js
MiriamAlter: Fixed and simplified

src/pages/ResearcherProfile.js
MiriamAlter: The linter recommended I move those functions into useEffect. They don't get used outside of the incomplete checks so I didn't think it would hurt.

src/pages/ResearcherProfile.js
MiriamAlter: Fixed? Edited, at least

src/pages/ResearcherProfile.js
MiriamAlter: Noted, will post to Jira after next push

src/pages/ResearcherProfile.js
MiriamAlter: Noted, will post to Jira after next push

src/pages/ResearcherProfile.js
MiriamAlter: Noted, will post to Jira after next push

src/pages/ResearcherProfile.js
MiriamAlter: Fixed, will post issue to Jira after next push

src/pages/ResearcherProfile.js
MiriamAlter: Fixed

src/pages/ResearcherProfile.js
rushtong: Still seeing this line in my latest update.

src/pages/ResearcherProfile.js
rushtong: They aren't used outside of the useEffect, that's true, and it's correct that it has no effect on functionality either inside or outside. The linter warning is very strange ... these methods have no dependency on any state values - they're purely functional and could be in an external library if we had more use for them. For example,
```
    const isValidNumber = (value) => {
      let isValid = false;
      if (value !== 0 && value !== null && value !== undefined) {
        isValid = true;
      }
      return isValid;
    };
``` 
Actually can be a call to the lodash function just as easily, so I'm really confused about why the linter is complaining. The point of my comment is that as more things are moved inside of useEffect, it becomes harder to reason about. The things that really need to be there are the state-affecting functions that useEffect has a dependency on, i.e. things that touch `profile` and `incompletes`

src/pages/ResearcherProfile.js
rushtong: I'm still seeing the unfixed version, are your updates pushed to origin?

rushtong (dismissed):
Functionally, this is working perfectly! I found a separate bug related to this with incomplete profiles. If a user deletes a required field, and saves, the form works correctly. If the user then navigates to a DAR, the dar form does not flag that the user is missing required fields. Will file a ticket for this.

Once the updates are merged in, please re-request a 👍🏽  and I'll approve it.

rushtong:
@MiriamAlter This branch needs have the latest changes from `develop` merged into this branch, or be rebased onto `develop`. I find the first option easier and more straightforward if there are a lot of commits in your branch.

src/pages/ResearcherProfile.js
JVThomas: I'm guessing `isValidState` has to be included because it references the state variable `profile` whose value can change, which is why the linter is stating that it needs to be included. However I don't see linter warnings for the other isValid methods, so those can safely be declared outside the useEffect.

src/pages/ResearcherProfile.js
MiriamAlter: Fixed

src/pages/ResearcherProfile.js
MiriamAlter: Fixed

src/pages/ResearcherProfile.js
MiriamAlter: Figured out and fixed

src/pages/ResearcherProfile.js
```diff
@@ -525,531 +356,448 @@ export const ResearcherProfile = hh(class ResearcherProfile extends Component {
...
-                      }, countryNames ),
-                      span({
-                        className: 'cancel-color required-field-error-span',
-                        isRendered: this.state.invalidFields.country && showValidationMessages
-                      }, ['Country is required'])
-                    ])
+              div({ className: 'col-xs-12 no-padding' }, [
+                div({ className: 'row fsi-row-lg-level fsi-row-md-level no-margin' }, [
+                  div({ className: 'col-xs-6' }, [
+                    label({ id: 'lbl_profileDepartment', className: 'control-label' }, ['Department']),
```
rushtong, line 518: It looks like this went from required to non-required with the last update. Can we revert these changes as part of this PR?

MiriamAlter: This was part of the discussion in today's standup. Department was changed to non-required on other parts of the site, and they asked me to change it on this page as well.

rushtong: Gotcha, thanks, wasn't aware of that 👍🏽

rushtong (approved):
I'm happy with this, it's a huge overhaul and makes this page a lot easier to work on 👍🏽

PR #1344, opened 2021-11-05
[DUOS-1498][risk=no] Add standard license

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1498

Adds a standard license text to the repo. See also:
* https://github.com/DataBiosphere/consent/blob/develop/LICENSE.txt
* https://github.com/DataBiosphere/consent-ontology/blob/develop/LICENSE.txt

----

- Verify all tests go green

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,27 @@
+Copyright (c) 2015, Broad Institute, Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2015, Broad Institute, Inc.
```
JVThomas, line 1: Do we need to update the years on this? If we do, it could be something like a range of years (1995-current year)

rushtong: If we do, we need to do it across all repos and services. See, for example, [Orchestration's license](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/LICENSE.txt). If we do, then I'd like to treat that as a separate task so all of our licenses are in sync.

JVThomas (approved):

PR #1352, opened 2021-11-10
DUOS-1493[risk=no] Added Completed Status Helper functions for DAR Application

JVThomas:
Addresses [DUOS-1493](https://broadworkbench.atlassian.net/browse/DUOS-1493)

PR adds helper function to `utils.js` to better evaluate completed status for the `ResearcherInfo` portion of the application. 

`userProperties.completed`, which is normally what should be referred to, is not reliable due to one situation

- User fills everything but the eraCommonsId on the profile -> `userProperties.completed` set to false on profile submission.
- User then visits the DAR application page and fills the `eraCommonsId` there. `eraCommonsId` is updated, but `userProperties.completed` is not (remains false).

Helper function was created because user properties are present within the DAR Application component, therefore the completed status evaluation can be made client side. eraCommonsId is not evaluated since

- value can be updated on the DAR Application page, therefore no alert message needs to be triggered on it being missing
- DAR Application component logic already flags missing inputs as part of it's validation logic, `eraCommonsId` is no exception.

- Verify all tests go green

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -4,7 +4,7 @@ import 'noty/lib/themes/bootstrap-v3.css';
 import { forEach } from 'lodash';
 import { DAR, DataSet } from "./ajax";
 import {Theme, Styles } from "./theme";
-import { find, first, map, isEmpty, filter, cloneDeep, isNil, toLower, includes, sortedUniq, every} from "lodash/fp";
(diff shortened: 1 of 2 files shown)


rushtong (approved):
Very nice cleanup. We also need to add stories for breaking utils up into more streamlined utilities. Will file a ticket for that.

PR #1356, opened 2021-11-16
DUOS-1508[risk=no] Removed unused AccessVotes variable from Access Review

JVThomas:
Addresses [DUOS-1508](https://broadworkbench.atlassian.net/browse/DUOS-1508)

Bugfix for initialization error on `AccessReview`. Root cause was an old variable that was no longer in use (value never initialized) but was still being referenced/passed along in code. It would be sent on return for some functions, but its value was never used. Nevertheless referencing an uninitialized variable will cause the error.

NOTE: This error only shows up on deployed builds. I was unable to reproduce this on a local build. Luckily the dev-k8 domain was able to produce the following output shown below. Production shows a similar output, however the code is minified so variable names are reduced to simple letters.

Screen Shot 2021-11-16 at 7 29 16 AM

image

- Verify all tests go green


rushtong (approved):
Nice find!

PR #1364, opened 2021-11-30
DUOS-1254[risk=no] Initial DAR Collection Review structure with header sub-component

JVThomas:
Partially addresses [DUOS-1254](https://broadworkbench.atlassian.net/browse/DUOS-1254)

PR creates initial structure for the `DarCollectionReview` component along with sub-components for the header/skeleton loaders. PR also includes update to `TabControl` component to render out skeleton loaders if a loading variable is provided to props.

- Verify all tests go green


src/components/TabControl.js
```diff
@@ -1,27 +1,36 @@
...
   return labels.map((label) =>
-    h(SelectableText, {
-      label,
-      key: `${label}-button`,
-      fontSize: '1.8rem',
-      componentType: label,
-      setSelected: setSelectedTab,
-      selectedType: selectedTab
-    })
+    !isLoading ?
```
quazi-h, line 6: Was the previous behavior here to show the SelectableText by default? With the new changes, if isLoading is false would we still show SelectableText?

quazi-h (approved):
The new DarCollectionReview and header components look good to me.

rushtong (approved):
Looks good on screen - no real feedback on anything at this point 👍🏽

src/components/TabControl.js
JVThomas: Yeah the initial setup would always show SelectableText. The idea now is to show SelectableText if `isLoading` is true (comes from the parent controller). If it's not, it'll render the second div, which shows a `div` with an animated gradient (a simple loading effect). If the parent finishes loading, it should update its `isLoading` value to trigger downstream changes on this component.

As a safety measure if `isLoading` doesn't have a value the component will default it as false to ensure that `SelectableText` is rendered.

JVThomas:
setTimeout warning on Codacy is a false positive

PR #1399, opened 2022-01-12
DUOS-1554[risk=no] Added VoteSummaryHeader to DarCollectionReview page

JVThomas:
Addresses [DUOS-1554](https://broadworkbench.atlassian.net/browse/DUOS-1554)

PR creates a Vote Summary Header generated from the DARs and datasets returned by a collection query (query by collection id). Changes include new components for the header itself, a new DAR Collection Utility file that contains the bucket processing logic, updates to dataUseTranslation to account for multiple async requests (in parallel) now being made to ontology, new cypress component tests, and an updated Github-Cypress workflow that now runs components tests after integration tests.

- Verify all tests go green


src/components/common/DataUseVoteSummary/VoteResultContainer.js
```diff
@@ -0,0 +1,70 @@
...
+    flexDirection: 'column',
+    justifyContent: 'space-evenly',
+    width: '10%',
+  };
+  const hyphenatedKey = convertLabelToKey(label);
+  //QUESTION: How will RP votes work from here on out
+  //Currently there is an RP election for each DAR (so by extension one for each dataset)
+  //Therefore there are technically X amount of RP elections, all of which are decided independenly for each other
+  //With the multi-dataset vote feature, will there only be one RP election across the election?
+  //Or will it be the same as before except now we just update all of the RP elections at once on vote submission?
```
rushtong: This needs to be clarified in the ticket before proceeding if it isn't clear.

src/libs/dataUseTranslation.js
```diff
@@ -166,6 +167,113 @@ const getOntologyName = async(obolibraryURL) => {
...
+          //this way you can avoid making repeated requests
+          const ontologyPromises = value.map((ontologyId) => {
+            if (isEmpty(ontologyMap[ontologyId])) {
+              ontologyMap[ontologyId] = getOntologyName(ontologyId);
+            }
+            return ontologyMap[ontologyId];
+          });
+          resp = Promise.all(ontologyPromises).then((ontologies) =>
+            consentTranslations.diseaseRestrictions(ontologies)
+          );
```
rushtong, line 197: We should add a new Ontology endpoint that does this for us. I can see the network traffic from this page exploding pretty quickly.

rushtong: Alternatively, we could do this in Consent by expanding the fields in `dataUse`

src/components/common/DataUseVoteSummary/VoteResultContainer.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (voteTally.true === voteCount) {
+    return true;
+  } else if (voteTally.false === voteCount) {
+    return false;
+  } else if (voteTally.true + voteTally.false === voteCount) {
+    return 'mixed';
+  } else {
+    return 'underReview';
+  }
```
rushtong: Another way to get the true/false/mixed answer is to collect the distinct `vote` values of all `votes` into an array. If that array has a single `true` element, then it's true, similarly if there's a single `false` value, and if there are two elements in that array, it's `mixed`. I've used this strategy at some point on the back end. Your call on this, I'm perfectly fine with the current approach.

src/components/common/DataUseVoteSummary/DataUseVoteSummary.js
```diff
@@ -0,0 +1,71 @@
...
+      const additionalLabelStyle = index === 0 && elementLength > 1? startElementStyle :
+        index === elementLength - 1 && elementLength > 1 ? endElementStyle : middleElementStyle;
+      const { key, votes = {}, isRP } = bucket;
+      const targetAttr = isRP ? 'rp' : 'dataAccess';
+      const finalVotes = flatMap((voteObj) =>
+        !isEmpty(voteObj) ? voteObj[targetAttr].finalVotes : []
+      )(votes);
+      return h(VoteResultContainer, { label: key, finalVotes, additionalLabelStyle }, []);
+    })(row);
+  };
```
rushtong: This is pretty hard for me to decipher, but I think I can grok it. I do think it could use some comments to explain it a little more.

src/utils/DarCollectionUtils.js
```diff
@@ -0,0 +1,120 @@
+import { flow, isEmpty, map, join, filter, forEach, flatMap, toLower, sortBy, isNil } from 'lodash/fp';
+import {translateDataUseRestrictionsFromDataUseArray} from '../libs/dataUseTranslation';
+
+//Initial step, organizes raw data for further processing in later function/steps
+export const generatePreProcessedBucketData = async ({dars, datasets}) => {
+  const dataUses = [];
+  const buckets = {};
+  const datasetBucketMap = {};
```
rushtong, line 8: minor nit ... these const's are mutated. Might be better to use let here for clarity of intention.

rushtong (approved):
See minor comments inline. This looks really good. All feedback here is optional - feel free to ignore or address as you see fit.

src/components/common/DataUseVoteSummary/VoteResultContainer.js
JVThomas: There's an added complication with Under Review, where the election is still ongoing (vote is undefined). So if there's two elements in the mix, it could be `[(True/False), (True/False)]` or it could be `[(True/False), undefined]`. Either way, you'll have to confirm it with an additional loop on the sub-array and then running if checks on the possible combinations.

src/utils/DarCollectionUtils.js
JVThomas: Actually on second thought I want to keep it as a `const`. The elements are being accessed by reference, which I can only guarantee with the Objects being a `const`. If I used a let then it's possible to replace the map/bucket with a new, blank object, which will have adverse effects downstream. 

The code doesn't allow this to happen, but should someone try to do so recklessly eslint will throw a compilation error with `const` compared to`let`.

It's not React like, but this wasn't built to be a React function but a data processing utility function.

src/libs/dataUseTranslation.js
JVThomas: Yeah, anything that lets me do it in one request is vastly preferable to this.

rushtong: Will file a ticket for the API to populate this in consent before it arrives downstream.

PR #1402, opened 2022-01-14
DUOS-1482[risk=no] Implement "Revise" DAR (from researcher console)

quazi-h:
[[Link to JIRA ticket](https://broadworkbench.atlassian.net/browse/DUOS-1482)]
Brief summary of changes:
- Added a new ajax function to resubmit DAR collections.
- Added a new resubmit collection function to the NewResearcherConsole that will be passed to the DarCollectionTable.
- Extended the DarCollectionTable to add a button for resubmitting a canceled DAR.
- Added logic to conditionally show Cancel or Revise buttons depending on the DAR status.
- Had to pull out the code for the confirmation modal to create two separate ones because of differing titles and messages - could be refactored.
- Had started working on the old ResearcherConsole at first - probably should drop those changes since we only want to make changes to the new console.

----

- Verify all tests go green


src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -103,28 +103,67 @@ const CancelCollectionButton = (props) => {
   });
 };
 
+const ResubmitCollectionButton = (props) => {
+  const { collection } = props;
+  return h(SimpleButton, {
+    keyProp: `resubmit-collection-${collection.id}`,
+    label: 'Revise',
+    //todo: should we filter out any statuses that could be resubmitted? is this the correct place for such logic?
+    //disabled: includes(determineCollectionStatus(collection))(['Under Election']),
```
rushtong: I don't think this is necessary here. I can't think of a case where we want to show a disabled "Revise" button. I think that if that were the case, this button should not be on the screen at all.

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -103,28 +103,67 @@ const CancelCollectionButton = (props) => {
...
-      {
-        style: {
-          display: 'flex',
-          justifyContent: 'left'
+  const resubmitButtonTemplate = h(ResubmitCollectionButton, {
+    collection,
+    showConfirmationModal
+  });
+  if(isCollectionCanceled(collection) === true)
+  {
```
rushtong: `isCollectionCanceled` already returns a boolean, so this can be simplified
```suggestion
  if (isCollectionCanceled(collection)) {
```

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -103,28 +103,67 @@ const CancelCollectionButton = (props) => {
...
+        {
+          style: {
+            display: 'flex',
+            justifyContent: 'left'
+          },
+          key: `cancel-collection-cell-${darCollectionId}`
+        },
+        [cancelButtonTemplate]
+      )
+    };
```
rushtong: We have two returns in an if-else block that makes this a little difficult to read. Instead of embedding the object construction inside the if-else, it would be easier to reason about if that were defined outside. For example, we could have two objects defined that could be returned, the cancel object and the revise object, and then simplify the return, something like this:

```
  const cancel = {
    isComponent: true,
    id: darCollectionId,
    label: 'cancel-button',
    data: div(
      {
        style: {
          display: 'flex',
          justifyContent: 'left'
        },
        key: `cancel-collection-cell-${darCollectionId}`
      },
      [h(CancelCollectionButton, {collection, showConfirmationModal})]
    )};

  const revise = {
    isComponent: true,
    id: darCollectionId,
    label: 'resubmit-button',
    data: div(
      {
        style: {
          display: 'flex',
          justifyContent: 'left'
        },
        key: `resubmit-collection-cell-${darCollectionId}`
      },
      [h(ResubmitCollectionButton, {collection, showConfirmationModal})]
    )
  };

  return isCollectionCanceled(collection) ? revise : cancel;
```

src/pages/NewResearcherConsole.js
```diff
@@ -141,6 +141,27 @@ export default function NewResearcherConsole(props) {
...
+      clonedCollections[targetIndex] = resubmittedCollection;
+      setResearcherCollections(clonedCollections);
+      Notifications.showSuccess({text: `Revising collection ${darCode}`});
+    } catch (error) {
+      Notifications.showError({
+        text: 'Error: Cannot revise target collection'
+      });
+    }
+  };
+
```
rushtong, line 164: Somehow, I'm seeing "Revise" for DARs that are under election which should be incorrect. Will dig in to see if I can spot the issue.
Screen Shot 2022-01-18 at 8 30 00 AM

rushtong: OK, the error case here is not the "Revise" button, but the election status. In the case of `DAR-252`, it is `Canceled`, not `Under Election`
Screen Shot 2022-01-18 at 8 33 22 AM

src/pages/ResearcherConsole.js
```diff
@@ -364,6 +386,14 @@ class ResearcherConsole extends Component {
           action: { label: "Yes", handler: this.dialogHandlerCancelDAR }
         }, [div({ className: "dialog-description" }, ["Are you sure you want to cancel this Data Access Request?"]),]),
 
+        ConfirmationDialog({
+          title: 'Revise saved Request?',
+          color: 'review',
+          isRendered: this.state.showDialogReviseDAR,
+          showModal: this.state.showDialogReviseDAR,
+          action: { label: "Yes", handler: this.dialogHandlerReviseDAR }
+        }, [div({ className: "dialog-description" }, ["Are you sure you want to revise this Data Access Request?"]),]),
```
rushtong: The color scheme for the "Yes" button is very difficult to read. We should look for a way to either change the non-hover background color so the white text is visible, or change the non-hover text color.

Screen Shot 2022-01-18 at 8 12 45 AM

rushtong (changes requested):
See comments inline. I think this is pretty close, just a few tweaks and perhaps a logic fix in the election status case.

rushtong (approved):
Thank you for the changes, these look good 👍🏽

PR #1410, opened 2022-01-19
[DUOS-1538][risk=no] Show LCA for adding/issuing LCs to researchers

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1538

## Changes
* Added a new LCA markdown file that can be manipulated in a markdown editor for visual accuracy
* Fetch the markdown and render it using the same technique we do for Privacy Policy. We should migrate the Privacy Policy to use the fetch technique at some point
* Added component tests and docs for running those tests

## Adding New Library Card
Screen Shot 2022-01-25 at 11 15 33 AM

## Issuing LC To Researcher
Screen Shot 2022-01-25 at 11 15 25 AM

----

- Verify all tests go green


JVThomas (approved):
PR looks good. There are some styling adjustments that need to be made for the modal template, but that needs to be done in another ticket (the styles I made for modals way back when don't hold up now).

As for the test, I have no clue why that last one doesn't work. I'm willing to guess that it's an issue with the cypress action library rather than the standalone package. [Here's a scenario where the issue was running the tests in headless mode](https://github.com/cypress-io/cypress/issues/8606#issuecomment-703544094), which may or may not apply here.

rushtong:
Thanks @JVThomas - I'll look more into the headless vs non-headless at some point. For now, will plow forward.

PR #1415, opened 2022-01-25
DUOS-1590[risk=no] ux-merge-researcher-review-edit-user-page

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1590)
Brief summary of changes: 

- add information from Researcher Review page to Edit User modal on the Admin Manage Users page 
----

- Verify all tests go green

### DIFF
--- a/src/pages/ResearcherReview.js
+++ b/src/pages/ResearcherReview.js
@@ -1,12 +1,12 @@
 import { Component } from 'react';
-import { div, hr, label, form, textarea } from 'react-hyperscript-helpers';
+import { div, hr, hh, label, form, textarea } from 'react-hyperscript-helpers';
 import { PageHeading } from '../components/PageHeading';
 import { User } from "../libs/ajax";
 import {getPropertyValuesFromUser} from "../libs/utils";
 import {isNil} from "lodash/fp";
 import { isEmpty } from 'lodash';
 
-class ResearcherReview extends Component {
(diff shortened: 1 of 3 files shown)


src/pages/ResearcherReview.js
```diff
@@ -47,7 +46,7 @@ class ResearcherReview extends Component {
 
   async findResearcherInfo() {
 
-    const user = await User.getById(this.props.match.params.dacUserId);
+    const user = await User.getById(this.props.userId);
```
rushtong, line 49: I think this change breaks the "Review Researcher" button on the "Manage Users" page. I think that button is no longer necessary. Can we remove it for this PR? That would solve the problem that it doesn't work 😄

rushtong (changes requested):
Code looks good, works for what we want it to do. See comment for my one suggestion before this is ready.

src/pages/ResearcherReview.js
shaemarks: I removed the Researcher Review column. There's currently some extra space on the right from where it used to be, but I'm not sure if I should look into that or leave it for now.

rushtong: You can leave that in place, not necessary to shrink it down.

rushtong (approved):
Changes look good, thank you! 👍🏽

PR #1418, opened 2022-01-26
DUOS-1294[risk=no] so-console-dar-collection-table

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-1294)
Summary of changes:

- Add DarCollection table to SO console 
- Remove buttons for canceling and resubmitting DAR collections
- add PI, Institution, and Datasets columns
- style fonts, columns, and sizes for DarCollection table 

----

- Verify all tests go green


src/libs/ajax.js
```diff
@@ -194,6 +194,11 @@ export const Collections = {
     const url = `${await Config.getApiUrl()}/api/collections/${id}`;
     const res = await axios.get(url, Config.authOpts());
     return res.data;
+  },
+  getCollectionByRoleName: async(roleName) => {
```
rushtong: minor naming nit: we should rename this to `getCollectionsByRoleName`

quazi-h (approved):
Incorporating the Dar collection table into the SO console looks good.
Are we restricting the ability to cancel/resubmit a DAR to the researcher console only?

rushtong:
@quazi-broad 
> Are we restricting the ability to cancel/resubmit a DAR to the researcher console only?

Yes, there are different actions that can be taken from different consoles. SOs, in this case, cannot execute any actions. Members have voting and viewing actions, Chairs and Admins have a good deal more actions available to them. However, none of them have the exact same actions that a Researcher has.

JVThomas:
> @quazi-broad
> 
> > Are we restricting the ability to cancel/resubmit a DAR to the researcher console only?
> 
> Yes, there are different actions that can be taken from different consoles. SOs, in this case, cannot execute any actions. Members have voting and viewing actions, Chairs and Admins have a good deal more actions available to them. However, none of them have the exact same actions that a Researcher has.

If we're running into a situation where there's differing combinations of buttons we can just define the actions component on the parent of `DarCollectionTable` and pass that in as a prop for the table to use directly.

This way SO Console, Chair Console, Member Console, and Researcher Console can just define the actions specific to them and the table can just render it directly via `actionCellData` (which you'll have to modify, just assign the component to data)

Or in the instance where we don't want actions, just pass a `true` value as a prop to DarCollectionTable to signify that it shouldn't bother with the column at all.

src/components/dar_collection_table/CollectionConfirmationModal.js
```diff
@@ -0,0 +1,57 @@
...
+      styleOverride: {height: '35%'},
+      closeConfirmation: () => setShowConfirmation(false),
+      title: 'Resubmit DAR Collection',
+      message: `Are you sure you want to resubmit ${collection.darCode}?`,
+      header: getModalHeader,
+      onConfirm: resubmitOnClick
+    });
+
+
+  if(isCollectionCanceled(collection) === true) {
```
JVThomas: The browser's console log is throwing this error below
Screen Shot 2022-01-28 at 11 56 26 AM

It's a bit esoteric, so I'll explain it a bit better. React Functional Components always return a template that is rendered whether or not there's data present or not. What you're doing here is returning the function, not the function result that produces a template.

To remove the error, you can just return the values of the function call (so `resubmitModal() instead of resubmitModal`). To make it even more react like, you can use ternary operators to define the output and return that instead.

`return isCollectionCanceled(collection) === true ? resubmitModal() : cancelModal()`

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -27,147 +28,86 @@ const styles = {
...
   if(!isNil(collections)) {
     return collections.map((collection) => {
-      const { darCollectionId, darCode, createDate } = collection;
+      const { darCollectionId, darCode, createDate, datasets } = collection;
+      const institution = collection.createUser.institution.name;
       /*I want the election-dependent status to be explicit so that the
       researcher knows why they can't cancel the collection*/
       const status = determineCollectionStatus(collection);
       const projectTitle = getProjectTitle(collection);
+      const actionsButton = actionsFlag
```
JVThomas: I should have given a better example when first explaining this. I used `actionsFlag` as a spur of the moment example when I should've used something like `actionsDisabled`. 

The reason why is we want to change this component without breaking its implementation in other components. Right now this table is being used in both the SO Console and the NewResearcherConsole. The "flag" is new, which means neither component is passing in a true value, so the flag is currently false. 

With this implementation both the SO Console and the NewResearcherConsole will not show the buttons at all, SO because it's passing in false, NewResearcherConsole because it hasn't been updated. You could update NewResearcherConsole to include this key, but a better option is to just do something like (this is pseudocode)

`actionsDisabled = true (passed in to SO console as a prop); actionsDisabled ? div : button`

This way you only have to update the SO Table and nothing else.

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -27,147 +28,86 @@ const styles = {
...
 };
 
 
-const processCollectionRowData = (collections, showConfirmationModal) => {
+const processCollectionRowData = (collections, showConfirmationModal, actionsFlag) => {
   if(!isNil(collections)) {
     return collections.map((collection) => {
-      const { darCollectionId, darCode, createDate } = collection;
+      const { darCollectionId, darCode, createDate, datasets } = collection;
+      const institution = collection.createUser.institution.name;
```
JVThomas: Another interesting thing about Users: they may not belong to an institution. When users first join, they are not given an institution since we generate a user account off of their gmail auth log in. The user has to update that value themselves via the User Profile Page.

That said, with or without an institution, a user can still submit DAR drafts and DARs, so doing `collection.createUser.institution.name` might return an error. You'll have to check if institution exists, and if it does, get the name. Otherwise settle on a default value like an empty string or "- -"

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -27,147 +28,86 @@ const styles = {
...
-        },
-        key: `resubmit-collection-cell-${darCollectionId}`
-      },
-      [h(ResubmitCollectionButton, {collection, showConfirmationModal})]
-    )
-  };
-
-  return isCollectionCanceled(collection) ? revise : cancel;
+  const {darCode, name, submissionDate, pi, institution, datasetCount, status, actions} = columnHeaderFormat;
+  return [darCode, name, submissionDate, pi, institution, datasetCount, status, actions];
```
JVThomas, line 85: I just realized something. Not only do the actions buttons change between columns, but also the columns themselves. However that's something we can cover in another refactoring ticket (since it's less about this component specifically and more about prop data abstraction). Just leaving a note here as a point of reference.

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -27,147 +28,86 @@ const styles = {
...
-    status: '15%',
-    actions: '20%'
-  }
+    projectTitle: '20%',
+    submissionDate: '12.5%',
+    pi: '5%',
+    institution: '20%',
+    datasetCount: '7.5%',
+    status: '10%',
+    actions: '7.5%'
```
JVThomas, line 47: Since we're changing the width of the actions column we need to make sure it doesn't break the styling elsewhere. Easy way to do this is to make your own dar locally and then visit the `/new_researcher_console` route and see the changes.

JVThomas (changes requested):
So the code modularization is mostly done with, all that's left is some slight bugs/errors to address.

src/components/dar_collection_table/DarCollectionTable.js
JVThomas: If there's issues, mess around with the values. It's going to primarily be trial and error here. That said, I did mention that we'll deal with column data abstraction in a later ticket, so it'll be best to tackle it there.

JVThomas (approved):
Changes look good. Code modularization makes the tables easier to read, and the column/styling changes creates a good foundation for further abstraction on DarCollectionTable (namely getting columnHeaders as a prop array)

PR #1424, opened 2022-02-02
[DUOS-1601][risk=no] Added ability to select for columns when rendering a DAR collection table

lu-c:
https://broadworkbench.atlassian.net/browse/DUOS-1601

Solution is to allow the user to provide a list of columns they want to include in the dar collection table - Converted the table from the default export to just ONE of the exports. The column options are also exported to allow users to refer to a static set of values

Security risk modifier: medium?


src/components/SimpleTable.js
```diff
@@ -14,7 +14,7 @@ const SkeletonLoader = ({columnRow, columnHeaders, baseStyle, tableSize}) => {
         const style = Object.assign({height: '0.5rem'}, baseStyle, cellStyle);
         return div({style, className: 'text-placeholder', key: `placeholder-row-${i}-cell-${index}`});
       });
-      rowsSkeleton.push(div({style: baseStyle, key: `placeholder-row-${i}-container`}, row));
+      rowsSkeleton.push(div({style: baseStyle, key: `placeholder-row-${i}-container`, className: `placeholder-row-${i}`}, row));
```
JVThomas, line 17: Good call adding a class here. It'll make the DOM element searchable for things like testing.

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -68,25 +68,35 @@ export const styles = {
...
-const columnHeaderFormat = {
-  darCode: {label: 'DAR Code', cellStyle: { width: styles.cellWidth.darCode }},
-  name: {label: 'Project Title', cellStyle: { width: styles.cellWidth.projectTitle }},
-  submissionDate: {label: 'Submission Date', cellStyle: {width: styles.cellWidth.submissionDate }},
-  pi: {label: 'PI', cellStyle: {width: styles.cellWidth.pi }},
-  institution: {label: 'Institution', cellStyle: { width: styles.cellWidth.institution }},
-  datasetCount: {label: 'Datasets', cellStyle: { width: styles.cellWidth.datasetCount }},
-  status: {label: 'Status', cellStyle: {width: styles.cellWidth.status }},
-  actions: {label: 'Action', cellStyle: { width: styles.cellWidth.actions }}
+export const DarCollectionTableColumnOptions = {
```
JVThomas, line 71: Good idea with the map here. It'll help us reference internal values and gives an exportable const for other components to use when defining the columns passed in.

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -100,29 +110,32 @@ const processCollectionRowData = (collections, showConfirmationModal, actionsDis
...
-        cellData.submissionDateCellData({ darCollectionId, createDate }),
-        cellData.piCellData({ darCollectionId }),
-        cellData.institutionCellData({ darCollectionId, institution }),
-        cellData.datasetCountCellData({ darCollectionId, datasets }),
-        cellData.statusCellData({ darCollectionId, status }),
-        actionsButton,
-      ];
+      return columns.map((col) => {
+        switch(col) {
+          case 'darCode': return cellData.darCodeCellData({ darCollectionId, darCode });
```
JVThomas: I wonder if we can tie the cellDataFunctions to `DarCollectionTableColumnOptions` as well. For example, something like...
```
{
   DAR_CODE: {
      <columnStringVarNameHere> : 'darCode',
      cellDataFn: darCodeCellData({darCollectionId, darCode}) 
 }
 ```
 
This way we can 

1. Avoid hard coding string values in more than one place
2. Shorten the function invoked by map
3. Make `DarCollectionTableColumnOptions` as the single source of truth for all things label and cell data processing

JVThomas (approved):
The code works wonderfully! There's one part of the code that I've added a suggestion for. It isn't necessary for implementation, but I would like to hear your thoughts on the matter.

rushtong (approved):
Nice work 👍🏽

src/components/SimpleTable.js
lu-c: That's exactly why I added it! 👍

src/components/dar_collection_table/DarCollectionTable.js
lu-c: I really like this suggestion - I'll push a quick update to this branch before merging it! 👍

PR #1427, opened 2022-02-04
[DUOS-1601] Added table header sorting

lu-c:
Before
Screen Shot 2022-02-04 at 2 45 56 PM

After
Screen Shot 2022-02-04 at 2 41 31 PM

Description:

Added a "sort table" function before calculating the visible table elements - the reason for this is that you cant truncate the list before sorting. This lead to an issue where the row data was extracted and processed after the filtering as well (probably for efficiency!), but then it would be difficult to sort on it again because you didnt know where to pull the data from.

The sort table method is now expecting preprocessed rows of data that are intended to go into the simpleTable, and will sort based off of column index


src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -68,61 +68,113 @@ export const styles = {
...
+  submissionDate: {
+    label: 'Submission Date',
+    cellStyle: { width: styles.cellWidth.submissionDate },
+    cellDataFn: cellData.submissionDateCellData,
+    sortable: true
+  },
+  pi: {
+    label: 'PI',
+    cellStyle: { width: styles.cellWidth.pi },
+    cellDataFn: cellData.datasetCountCellData,
```
JVThomas: I'm guessing this is a copy-paste holdover? It should be the PI data, however I don't think that data is provided from the back-end yet. For now implement it as if it did exist, but have a fallback value (something like `- -` to deal with the missing attribute.

src/libs/utils.js
```diff
@@ -625,11 +625,30 @@ export const getColumnSort = (getList, callback) => {
   };
 };
 
+const sortVisibleTable = ({ list, sort }) => {
```
JVThomas: List isn't defined when `NewResearcherConsole` first mounts, so this method ends up triggering error notifications on that page. Easy way to fix this is to give it a default on the object restructuring statement. `{list = [], sort}` should be enough to fix it.

JVThomas (changes requested):
Code looks good, great job refactoring `SimpleTable` so that it has table sorting built in. That said, there are a few bugs that need to be tackled

src/components/dar_collection_table/DarCollectionTable.js
lu-c: Oh goodness, thank you, this is a great catch! XD Yes, there is already a piCellData method, I'll update it to use that 👍

src/components/SimpleTable.js
```diff
@@ -35,21 +35,38 @@ const OnClickTextCell = ({ text, style, onClick, keyProp }) => {
...
     //style here pertains to styling for individual cells
     //should be used to set dimensions of specific columns
-    return div({style: cellStyle, key: `column-row-${label}`}, [label]);
+    return div({style: cellStyle, key: `column-row-${label}`, className: 'column-header'}, [
+      header.sortable && onSort
+        ? div({
+          style: Styles.TABLE.HEADER_SORT,
+          key: "data_id_cell",
+          className: 'cell-sort',
+          onClick: () => {
```
JVThomas, line 50: For the `NewResearcherConsole` having "Actions" as a sortable column throws errors when you click on the header. Since 'status' already provides that contextual information, we don't need "Actions" to be sortable at all.

Kapture 2022-02-08 at 07 10 03

JVThomas (changes requested):
Good updates on the code so far, only one bug remaining.

src/components/SimpleTable.js
lu-c: I think I pushed up a fix for this now! Thanks for the catch.

I'm going to also try to find the styles for that revise button, heh!

JVThomas (approved):
Everything is in working order and the styling adjustments look great!

PR #1431, opened 2022-02-08
WIP [DUOS-1551][risk=no] DAR Collection Summary View - Application Information

quazi-h:
First attempt at [DUOS-1551](https://broadworkbench.atlassian.net/browse/DUOS-1551)

- Extended ApplicationInformation to pull additional items from darInfo to populate items under the "Application Details" section.
- Extended the AtAGlance component by adding a header for the table and started to write a function to handle pulling the data.
- Added AtAGlanceCellData.js to store properties for the different "At A Glance" fields.
- Need to complete the At A Glance table, and need to figure out how to get signing official email and IT director email for the Application Details
- Need to add IRB protocol link?
- Redo the page header to match the mock?

----

- Verify all tests go green


src/pages/dar_collection_review/AtAGlance.js
```diff
@@ -0,0 +1,150 @@
+import {div, h} from 'react-hyperscript-helpers';
+import {isNil, isEmpty, find, chunk} from 'lodash/fp';
```
rushtong: `isEmpty, find,` look unused here

src/pages/dar_collection_review/AtAGlance.js
```diff
@@ -0,0 +1,150 @@
...
+import {useState} from 'react';
+import SimpleTable from '../../components/SimpleTable';
+import {Styles} from '../../libs/theme';
+import cellData from './AtAGlanceCellData'
```
rushtong: add semi-colon to end
```suggestion
import cellData from './AtAGlanceCellData';
```

src/pages/dar_collection_review/AtAGlance.js
```diff
@@ -0,0 +1,150 @@
...
+
+export const styles = {
+  title: {
+    fontWeight: 800,
+    fontSize: '2.7rem',
+  },
+  baseStyle: {
+    fontFamily: 'Montserrat',
+    fontSize: '1.6rem',
```
rushtong, line 15: I think the fonts look a little large to me. Can we leave this at the default?

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -66,13 +68,31 @@ export default function ApplicationInformation(props) {
...
+    collection,
+    dataUseBuckets,
+    externalCollaborators = '- -',
+    internalCollaborators = '- -',
+    signingOfficial = '- -',
+    itDirector = '- -',
+    signingOfficialEmail = '- -',
+    itDirectorEmail = '- -',
+    internalLabStaff = '- -',
+    anvilStorage = '- -',
```
JVThomas: `anvilStorage`, `localComputing`, and `cloudComputing` all refer to checkbox options on the DAR Form...

Screen Shot 2022-02-09 at 2 08 41 PM

... so for these attributes we can set the default string values to "No"

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -96,7 +116,40 @@ export default function ApplicationInformation(props) {
...
+        ]),
+        div({className: 'information-row', style: styles.row}, [
+          generateLabelSpanContents('Internal Lab Staff', 'internal-lab-staff', internalLabStaff, isLoading),
+          generateLabelSpanContents('Using AnVIL only for storage and analysis', 'anvil-storage', anvilStorage, isLoading),
+        ]),
+        div({className: 'information-row', style: styles.row}, [
+          generateLabelSpanContents('Requesting Permission to use local computing', 'local-computing', localComputing, isLoading),
+          generateLabelSpanContents('Requesting permission to use cloud computing', 'cloud-computing', cloudComputing, isLoading),
+        ]),
+        div({className: 'information-row', style: styles.row}, [
```
JVThomas: Cloud Provider and Cloud Provider Description makes sense if `cloudComputing` is true, so it might make sense to add conditional logic to reflect that (show if true, hide if false)

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -96,7 +116,40 @@ export default function ApplicationInformation(props) {
...
+          generateLabelSpanContents('Signing Official Email', 'signing-official-email', signingOfficialEmail, isLoading),
+          generateLabelSpanContents('IT Director Email', 'it-director-email', itDirectorEmail, isLoading),
+        ]),
+        div({className: 'information-row', style: styles.row}, [
+          generateLabelSpanContents('Internal Lab Staff', 'internal-lab-staff', internalLabStaff, isLoading),
+          generateLabelSpanContents('Using AnVIL only for storage and analysis', 'anvil-storage', anvilStorage, isLoading),
+        ]),
+        div({className: 'information-row', style: styles.row}, [
```
JVThomas: `styles.row` was made with a three column format in mind. Since Application Details is a two-column format, the styling has to be different if it's going to line up with the mockup.

That said the styling shouldn't be that much different since the only thing you'll have to change is the `justifyContent` attribute. Instead of using `flex-start`, you can use `space-between`.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -96,7 +116,40 @@ export default function ApplicationInformation(props) {
...
+          generateLabelSpanContents('Requesting Permission to use local computing', 'local-computing', localComputing, isLoading),
+          generateLabelSpanContents('Requesting permission to use cloud computing', 'cloud-computing', cloudComputing, isLoading),
+        ]),
+        div({className: 'information-row', style: styles.row}, [
+          generateLabelSpanContents('Cloud Provider (description below)', 'cloud-provider', cloudProvider, isLoading),
+          generateLabelSpanContents('', 'row-three-blank', '', false) //blank span to keep row elements in line with those above
+        ]),
+        div({className: 'cloud-provider-description-container'}, [
+          !isLoading ? div({className: 'cloud-provider-description-textbox', style: styles.nonTechSummaryText}, [cloudProviderDescription])
```
JVThomas: nit: I initially named `styles.nonTechSummary` as such because it was the only textbox on the earlier mockups. Since the style is being used for other fields it might make sense to give the variable a more generic name.

JVThomas (commented):
I've left some comments on the Application Details portion of the ticket, still going through the `AtAGlance` portion of the code.

src/pages/dar_collection_review/ApplicationInformation.js
rushtong: Additionally, I wonder if we should show anything at all if there are no values for some of these fields. `externalCollaborators`, `internalCollaborators`, and `internalLabStaff` seem like useless information to put on the screen if they don't exist. Let's clarify with Jonathan about that.

rushtong (changes requested):
Started looking at this yesterday but I think we need some functional clarifications before continuing. See feedback inline.

rushtong:
@quazi-broad Are your newest changes merged? I don't see them here yet.

src/pages/dar_collection_review/ApplicationInformation.js
quazi-h: Whoops, I had added a new style _applicantInfoRow_ but I implemented it in the Applicant Information section instead of the Application Details by accident. Correcting this on my next commit

src/pages/dar_collection_review/AtAGlance.js
```diff
@@ -0,0 +1,150 @@
...
+  const columnHeaderData = () => {
+    const {dataUse, dacFinalDecision, structuredResearchPurpose, dacResearchPurposeAgreement, duosAlgorithmDecision, dacVsDuosAlgorithm} = columnHeaderFormat;
+    return [dataUse, dacFinalDecision, structuredResearchPurpose, dacResearchPurposeAgreement, duosAlgorithmDecision, dacVsDuosAlgorithm];
+  };
+
+  const processAtAGlanceData = ({collection, dataUseBuckets}) => {
+    const rowElementMaxCount = 11;
+    //chunking elements subdivides the entire bucket list to rows of max rowElementMaxCount size
+    //Needed to provide predictable design, can't have a flexbox row of n elements with infinitesimal width
+    const chunkedBuckets = chunk(rowElementMaxCount)(dataUseBuckets);
```
JVThomas: You don't need to use chunkedBuckets or rowElementMaxCount for this function. These lines were only used for the Vote Summary Header (the bar on top that shows the bucket name and the vote result in a colored box) so that there would be a fixed number of data buckets per row. So if there were more than 10 dataUses, chunking would subdivide the elements into rows of 10 max.

With the table width isn't an issue since the columns are fixed and length is a non-factor, so you can just pass the buckets as is.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -92,11 +118,55 @@ export default function ApplicationInformation(props) {
...
+          generateLabelSpanContents('Signing Official Email', 'signing-official-email', signingOfficialEmail, isLoading),
+          generateLabelSpanContents('IT Director Email', 'it-director-email', itDirectorEmail, isLoading),
+        ]),
+        (internalCollaborators.length>0 || anvilStorage) ?
+          div({className: 'information-row', style: styles.applicantInfoRow}, [
+            (internalCollaborators.length>0) ?
+              generateLabelSpanContents('Internal Lab Staff', 'internal-lab-staff', internalLabStaff, isLoading)
+              : generateLabelSpanContents('', 'row-four-blank', '', false), //blank span to keep row elements in line with those above
+            (anvilStorage) ?
+              generateLabelSpanContents('Using AnVIL only for storage and analysis', 'anvil-storage', anvilStorage, isLoading)
```
JVThomas: Boolean values don't render as text in React (annoying, I know) so you'll have to convert it yourself. Odds are this won't just apply to `anvilStorage`.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -92,11 +118,55 @@ export default function ApplicationInformation(props) {
...
+        ]),
+        div({className: 'information-row', style: styles.applicantInfoRow}, [
+          generateLabelSpanContents('Signing Official Email', 'signing-official-email', signingOfficialEmail, isLoading),
+          generateLabelSpanContents('IT Director Email', 'it-director-email', itDirectorEmail, isLoading),
+        ]),
+        (internalCollaborators.length>0 || anvilStorage) ?
+          div({className: 'information-row', style: styles.applicantInfoRow}, [
+            (internalCollaborators.length>0) ?
+              generateLabelSpanContents('Internal Lab Staff', 'internal-lab-staff', internalLabStaff, isLoading)
+              : generateLabelSpanContents('', 'row-four-blank', '', false), //blank span to keep row elements in line with those above
```
JVThomas: So the conditional rendering poses a new problem: What happens if the first element of a row is not rendered, but the second one is?
Screen Shot 2022-02-17 at 12 33 05 PM

If the conditions are right, you'll end up with the above, with data mis-aligned. So instead of the hard-coded template we had before we'll need to make the row generation dynamic.

One approach is to generate the content with a function. Figure out what data needs to be visible and save that data in an array (using `generateLabelSpanContents` to generate the value for each element in the array). From there, you can use the `chunk` method that I referenced earlier to organize them in arrays of two. After that, use a map function to generate a new array that wraps each chunk in the row style. So something like ...

```
const output = chunkedArr.map(chunk => {
    return div({className: 'information-row', style: styles.applicantInfoRow}, chunk)
```
This should give you a template that you can then plug into the component's return statement.

JVThomas (dismissed):
Okay you're making good progress, but there are new issues to consider now that we're conditionally rendering data.

rushtong:
@quazi-broad - Pending Justin's feedback, 👍🏽. I don't see any additional commits after his comments. Please re-request a review if you do merge those in.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +62,25 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// content generated by function
+export function DynamicRowGeneration = (rowElementMaxCount, labelArray) => {
```
JVThomas: You don't need to have `export function` here. A simple `const dynamicRowGeneration` will work here.

You use the `export` keyword if you want the function to be used in another file (so that other file can 'import' the function into its own scope).

Another note: Functions follow camel casing while components (the functions with props) are sort of title cased (each word is capitalized, but there's no spaces between words).

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -92,11 +138,71 @@ export default function ApplicationInformation(props) {
...
           }})
-      ])
+      ]),
+      div({className: 'application-details-container', style: { margin: '2.5rem 0'}}, [
+        div({className: 'applicant-details-subheader', style: styles.title}, ["Application Details"]),
+        // figure out what data needs to be visible and save that data in an array
+        // using generateLabelSpanContents to generate the value for each element in the array
+          // todo: do I include the if logic in here or within the function?
+        DynamicRowGeneration(2, [
+            generateLabelSpanContents('External Collaborators', 'external-collaborators', externalCollaborators, isLoading),
```
JVThomas: Don't generate the labels and spans from the start. You'll need to figure out what fields to show or not show. For example, if `localComputing` and `cloudComputing` are `undefined` or an empty string then there's no reason to generate templates for them at all.

Look at the values directly and see if it makes sense to render them out. If it does, you can add them to an array. Once you've made your array you can use `generateLabelSpanContents` on each element to get your templates, then chunk them afterwards.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +62,25 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// content generated by function
+export function DynamicRowGeneration = (rowElementMaxCount, labelArray) => {
+  // todo: QUESTION - should I use generateLabelSpanContents before or after chunking?
+  // todo: QUESTION - should I apply use this function for the Applicant Information Section as well?
+  // todo: what happens when we have an odd number of elements in an array with an even maxRowElementsValue? do we need to generate a blank span?
```
JVThomas: I don't think you'd have to for Application Details. The individual fields have a fixed width of `48rem`, and the row itself has `justify-content: space-between` css attribute. So if there's an incomplete row at the end (3 elements, 2 row max, end up with 1 element on the end), it'll still fill in the column from the left.

For anything greater than two columns (`maxRowElementsValue > 2`) you would need to add a blank section to ensure that columns in each row line up with each other so the filtering > templating > chunking logic would have to change.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +62,25 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
...
+  //generateLabelSpanContents('', 'row-six-blank', '', false) //blank span to keep row elements in line with those above
+
+  // use the chunk method to organize them in arrays of two
+  const chunkedArr = chunk(rowElementMaxCount)(labelArray);
+
+  // use a map function to generate a new array that wraps each chunk in the row style
+  // template that you can then plug into the component's return statement
+  const output = chunkedArr.map(chunk => {
+    return div({className: 'information-row', style: styles.applicantInfoRow}, chunk);
+  };
```
JVThomas: You need to close the parenthesis from the map function (so it needs to look like `});`

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +62,25 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// content generated by function
+export function DynamicRowGeneration = (rowElementMaxCount, labelArray) => {
+  // todo: QUESTION - should I use generateLabelSpanContents before or after chunking?
+  // todo: QUESTION - should I apply use this function for the Applicant Information Section as well?
```
JVThomas: You don't need to apply it there, the contents on Applicant Information are always visible, so a standard templating approach works there.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +61,26 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// function to generate application details content
+const dynamicRowGeneration = (rowElementMaxCount, appDetailLabels, loading) => {
+  const labelArray = appDetailLabels.map(label => {
+    // only generate elements that are populated
+    if (label[0] && label[0]!='' && label[0].length > 0) {
```
quazi-h: Is there a better way for me to check if a value should be included? I think just making sure there are no null, blank, or 0 length entries should catch everything. Another approach I was thinking of was using the "typeof" function to check each element:
`if(typeof label[0] == 'string') {
    if(label[0] != '') return generateLabelSpanContents()
}
`

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +61,26 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
...
+      return generateLabelSpanContents(label[1], label[2], label[0], loading);
```
quazi-h: I remember you mentioned in a previous comment - Boolean values don't render as text in React (annoying, I know) so you'll have to convert it yourself. Odds are this won't just apply to anvilStorage.

If I was handling different types separately (see above comment) then I could potentially also handle the rendering of text for booleans, right?

src/pages/dar_collection_review/ApplicationInformation.js
JVThomas: The check is even simpler with lodash's [isEmpty](https://lodash.com/docs/4.17.15#isEmpty) function.

It works on objects, arrays, and strings with the same assumption: If the value is undefined/null or if the length is 0 it'll return true, otherwise it'll return false

src/pages/dar_collection_review/ApplicationInformation.js
JVThomas: Yeah you could. Once you know that there's a value you can check to see if you need to do some processing for output. 

Lodash has its [isBoolean](https://lodash.com/docs/4.17.15#isBoolean) method that you can use for booleans. Another option is to use <varName>.toString() and compare values that way. 

NOTE: There's probably even more options, these are just the two I just came up with off the top of my head.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +61,28 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// function to generate application details content
+const dynamicRowGeneration = (rowElementMaxCount, appDetailLabels, loading) => {
+  const labelArray = appDetailLabels.map(label => {
+    // only generate elements that are populated
+    if (!isEmpty(label[0])) {
+      // check to see if processing is required for output (booleans)
+      const spanValue = (typeof label[0] == 'boolean') ? label[0].toString() : label[0];
```
JVThomas: JS has a very odd notion of equality. There are two operators, `==` and `===`. The latter (`===`) is what you would expect equality checks would do in other programming languages. The former(`==`) will try to run an equality check even if the two items are of different types, which leads to all sorts of weird results. 

In short, use `===` (even though `==` would work here)

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +61,28 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// function to generate application details content
+const dynamicRowGeneration = (rowElementMaxCount, appDetailLabels, loading) => {
+  const labelArray = appDetailLabels.map(label => {
+    // only generate elements that are populated
+    if (!isEmpty(label[0])) {
```
JVThomas: Completely forgot to explain this aspect of booleans, so sorry about this. Lodash's `isEmpty` works by checking to see if if there are enumerable keys (Object) or if there is a length attribute associated with the item (arrays, strings). Booleans have neither, so isEmpty will return false for any boolean. 

You'll have check on booleans differently from the other values (which are strings or arrays). `typeof` should be useful for setting up conditional checks.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,6 +61,28 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
   );
 };
 
+// function to generate application details content
+const dynamicRowGeneration = (rowElementMaxCount, appDetailLabels, loading) => {
+  const labelArray = appDetailLabels.map(label => {
```
JVThomas: With `map`, not returning a value doesn't skip over the value, it just assigns undefined for that element in the array, which will have a negative impact on the chunking method that occurs later on. 

What you should do is use lodash's [filter()](https://lodash.com/docs/4.17.15#filter) with `appDetailLabels` so that only the elements with a valid `label[0]` value (non-empty string, non-empty object, non-empty array, booleans) show up. The filter method will return a new array of the appDetailLabels that can be processed with `generateLabelSpanContents`

You have the logic checks set up in your `map` (minus the adjustments stated in the later comments), you just need to use it before this map function to filter for the valid values.

src/pages/dar_collection_review/ApplicationInformation.js
```diff
@@ -53,10 +61,37 @@ const generateLabelSpanContents = (labelValue, key,  spanValue, isLoading) => {
...
+const dynamicRowGeneration = (rowElementMaxCount, appDetailLabels, loading) => {
+  // lodash/fp filter (non-empty string, non-empty object, non-empty array, booleans)
+  const labels = filter(label => {
+    return (typeof label.value === 'boolean' || !isEmpty(label.value));
+  }) (appDetailLabels);
+
+  const labelArray = labels.map(label => {
+    if (typeof label.value === 'boolean') {
+      // Inject "Yes" / "No" for booleans
+      return generateLabelSpanContents(label.title, label.key, (label.value ? 'Yes' : 'No'), loading);
```
quazi-h, line 81: Just double checking - we want to display all booleans regardless of whether they were true or false, correct?

rushtong: That seems right to me. A user can select true/false in multiple DAR questions, so that should be reflected.

JVThomas: For now yes. We haven't figured out whether or not we should only show the true values or show all, so to be safe just stick with logic you have here. We can always update the logic on a new ticket if needed.

rushtong (approved):
Team approved 👍🏽

PR #1442, opened 2022-02-14
[DUOS-1556][risk=no] multi dataset voting vote summary table component

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1556)
Summary of changes:

- Create sortable vote summary table component for dac votes on elections
- Adjust style of table 
- Add cypress component tests for table

----

- Verify all tests go green


src/components/vote_summary_table/VoteSummaryTable.js
```diff
@@ -0,0 +1,130 @@
...
+const columnHeaderData = () => {
+  const { vote, name, date, rationale } = columnHeaderFormat;
+  return [vote, name, date, rationale];
+};
+
+const processVoteSummaryRowData = ({ dacVotes }) => {
+  if(!isNil(dacVotes)) {
+    return dacVotes.map((dacVote) => {
+      const { vote, displayName } = dacVote;
+      const { voteId, createDate, updateDate, rationale } = vote;
```
rushtong: I'm not sure where this model is defined. When I'm looking at the votes that come in from a DAR Collection, they look like this:
```
{
  "dacUserId": 3396, 
  "displayName": "Jonathan Lawson (Admin & DAC Member)", 
  "electionId": 1480, 
  "type": "DAC", 
  "voteId": 20097
}
```
The `vote` value, which is a true|false, can also be null as in the case above, and might not be in the object. I tested this locally by passing in all of the votes for a specific election on a specific dar collection into this component.

rushtong (changes requested):
Table and tests look good. I have a question on the model though. Can we resolve that before moving forward?

src/components/vote_summary_table/VoteSummaryTable.js
shaemarks: What page do you recommend adding the table to for testing locally as I make changes? I was using DarApplication since it displays dar election information , but that seems to use a different vote structure (found within the ElectionReview model) compared to the Vote model

rushtong: Great question - I used the dar collection page. I just used some test code to grab the votes on the first election on the first  dar that I was looking at and fed that into this component.

shaemarks: I was having some problems getting down to the votes level on DarCollectionReview.js to test the table, but I changed the structure expected of a vote to match the Vote model. If there are still issues, could you point me to the test code you used to get the votes?

rushtong: You can see an example here where we dig down into the collection's elections to find votes the user for a user: https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/utils.js#L358

shaemarks: Got it working! Thanks!

rushtong (approved):
Thank you for the changes, this looks great 👍🏽

PR #1444, opened 2022-02-15
DUOS-1632[risk=no] Added error notifications and redirect to DarCollectionReview.js

JVThomas:
Addresses [DUOS-1632](https://broadworkbench.atlassian.net/browse/DUOS-1632)

PR adds error notification and console redirects if the collection fails to initialize on `DarCollectionReview.js`.

- Verify all tests go green

### DIFF
--- a/src/pages/dar_collection_review/DarCollectionReview.js
+++ b/src/pages/dar_collection_review/DarCollectionReview.js
@@ -62,8 +62,6 @@ export default function DarCollectionReview(props) {
 
   useEffect(() => {
     const init = async () => {
-      //NOTE: backend is a bit restrictive, collections by id are currently not visible to anyone other than the creator
-      //Endpoint needs to be adjusted, for now just remove the check on a local instance
       const user = Storage.getCurrentUser();
       try {
         const collection = await Collections.getCollectionById(collectionId);
@@ -91,7 +89,7 @@ export default function DarCollectionReview(props) {
           clearTimeout(timeout);
         }, 500);
       } catch(error) {
-        Notifications.showError({text: "Error fetching DAR collection, redirected you to console page"});
+        Notifications.showError({text: 'Error initializing DAR collection page. You have been redirected to your console'});
         Navigation.console(user, props.history);
       }
     };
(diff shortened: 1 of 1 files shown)


src/pages/dar_collection_review/DarCollectionReview.js
```diff
@@ -62,33 +64,36 @@ export default function DarCollectionReview(props) {
     const init = async () => {
       //NOTE: backend is a bit restrictive, collections by id are currently not visible to anyone other than the creator
       //Endpoint needs to be adjusted, for now just remove the check on a local instance
```
rushtong: nit: this comment is no longer true

src/pages/dar_collection_review/DarCollectionReview.js
```diff
@@ -62,33 +64,36 @@ export default function DarCollectionReview(props) {
...
+        setResearcherProfile(researcherProfile);
+        //setTimeout used to render skeleton loader while sub-components are initializing data for render
+        const timeout = setTimeout(() => {
+          setIsLoading(false);
+          clearTimeout(timeout);
+        }, 500);
+      } catch(error) {
+        Notifications.showError({text: "Error fetching DAR collection, redirected you to console page"});
+        Navigation.console(user, props.history);
+      }
```
rushtong: Technically, this satisfies the underlying problem we've seen. The issue I see is that the only error message this is relevant to is the error from line 69. There's a lot of extra code that could, in theory, throw an error (especially the get user by id call) and in that case, the message doesn't make any sense, nor does the redirection. Not sure if it makes sense to break out the try/catch around just the collection call. Either way, I'm OK with this as a fix.

rushtong (approved):
Minor comments inline ... address as you see fit 👍🏽

src/pages/dar_collection_review/DarCollectionReview.js
JVThomas: I can change the message to "Error initializing DAR Collection page...." to make it more of a catch all (since any issues getting/setting data here means there's something wrong with the Collection and/or User responses).

PR #1454, opened 2022-02-24
DUOS-1622[risk=no] edit user new page no modal

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1622)
Summary of changes:

- Converted modal on admin page for editing existing users into a separate page 
- Added checkbox for adding or removing signing official role for user 
- Added read-only checkbox to show that every user should have a researcher role 
- Created endpoints for api calls to add or remove user roles 
- Removed edit user case from the AddUserModal

_______________

- Verify all tests go green

### DIFF
--- a/src/pages/AdminEditUser.js
+++ b/src/pages/AdminEditUser.js
@@ -66,9 +66,10 @@ export const AdminEditUser = hh(class AdminEditUser extends Component {
     const updatedUser = await User.update(payload, userId);
     await this.updateRolesIfDifferent(userId, this.state.updatedRoles);
 
-    this.setState({
-      user: updatedUser,
-      displayNameValid: updatedUser
+    console.log("Updating the user in state: " + JSON.stringify(updatedUser));
+    this.setState(prev => {
+      prev.user = Object.assign({}, updatedUser);
+      prev.displayNameValid = this.nameRef.current.validity.valid;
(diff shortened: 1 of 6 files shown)


src/pages/AdminEditUser.js
```diff
@@ -0,0 +1,303 @@
...
+    },
+    () => {
+      this.setState({
+        displayNameValid: this.nameRef.current.validity.valid,
+      });
+    });
+  }
+
+  OKHandler = async (event) => {
+    event.persist();
```
JVThomas: As of React 17, event.persist() doesn't do anything, so you can safely remove it.
https://reactjs.org/docs/events.html#overview

src/pages/AdminEditUser.js
```diff
@@ -0,0 +1,303 @@
...
+    event.preventDefault();
```
JVThomas, line 53: You'll need to keep `event.preventDefault()` in because the input/submit fields are in a form tag. Since the default browser behavior is to refresh the page on a form submit, e.preventDefault() stops that from happening.

JVThomas (approved):
Code looks good and the Edit user page works as expected. Added one nitpicking comment, so make the changes as you see fit.

PR #1462, opened 2022-03-08
[DUOS-1555][risk=no] multi dataset voting srp component

shaemarks:
[LInk to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1555)
Summary of changes:

- Create ResearchPropsalVoteSlab, which displays the requested data uses, the research purpose, and an alert box that is rendered if some of the data uses require manual review
- Add a link to the ResearchPropsalVoteSlab to collapse and expand the view of the slab 
- Create data use pills that display the code and description of a data use 
- Add a component for the DataUseAlertBox, which filters and displays the data use descriptions of a translated data use based on if they require manual review 
- Write cypress component tests 
- Attempt an animation with headless ui Transition component to make the expansion and compression of the slab more smooth (I left the Transition in for a second look, but if the animation part still doesn't work, I will remove)

----

- Verify all tests go green


shaemarks:
Note: currently the enter and leave animation modifiers of Transition are not working (still just a standard appear and disappear). I decided to leave it in for the time being in case I'm missing something simple to make it work. If not, I can remove. It could potentially have something to do with using h(Transition, ...  instead of <Transition ...>... .

src/components/collection_voting_slab/DataUseAlertBox.js
```diff
@@ -0,0 +1,51 @@
...
+  return ld.filter(dataUses, dataUse => {
+    return dataUse.manualReview;
+  });
+};
+
+export default function DataUseAlertBox(props) {
+  const {translatedDataUse} = props;
+  const descriptions = dataUseDescriptions(translatedDataUse);
+
+  return div({className: 'data-use-alert-box', style: styles.box, isRendered: !isEmpty(descriptions)}, [
```
rushtong: It looks like this is only used to access the element from tests. Can we instead give it a `data-cy-alert-box` (or something similar) attribute and look for that instead? See https://docs.cypress.io/guides/references/best-practices#Selecting-Elements for a best practice on this subject.

src/components/collection_voting_slab/DataUseAlertBox.js
```diff
@@ -0,0 +1,51 @@
...
+    width: 'fit-content'
+  },
+  text: {
+    fontSize: '1.6rem',
+    fontWeight: '500',
+  },
+  exclamationPoint: {
+    fontSize: '8rem',
+    fontWeight: 'bold'
+  }
```
rushtong: 8 rem translates to about 114 pixels high. The mocks I see have this at 40 pixels high. The result is that the box is very tall. If you bump that down to 3 rem, that brings it to ~43 pixels high which looks a bit cleaner.

src/components/collection_voting_slab/ResearchProposalVoteSlab.js
```diff
@@ -0,0 +1,146 @@
...
+
+    h(Transition, {
+      show: expanded,
+      enter: "transition-opacity ease-linear duration-150",
+      enterFrom: "opacity-0",
+      enterTo: "opacity-100",
+      leave: "transition-opacity ease-linear duration-150",
+      leaveFrom: "opacity-100",
+      leaveTo: "opacity-0",
+    },
```
rushtong: I tried fiddling with this as well and could not get it to work. It might be better to look for a different solution or just ditch it altogether since it's not a primary requirement.

rushtong (commented):
See minor styling comments inline, this looks pretty good!

rushtong (approved):
Thank you for the changes!

PR #1480, opened 2022-03-21
DUOS-1640[risk=low] UI rp, member, and final vote box

shaemarks:
[Link to RP and Member ticket](https://broadworkbench.atlassian.net/browse/DUOS-1640)
[Link to Final vote ticket](https://broadworkbench.atlassian.net/browse/DUOS-1641)
Summary of changes:

- Create component for member or chairperson to submit votes/vote rationales for collections 
- Allow vote and rationale to be updated indefinitely for RP and Member votes
- Disable vote buttons and rationale textbox after single vote for Final votes 
- Display current vote value and rationale if given votes have matching values 
- Write cypress component tests 

----

- Verify all tests go green


src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -0,0 +1,141 @@
...
+    if (!isEmpty(votes)) {
+      const prevVote = votes[0];
+
+      const voteValues = ld.map(votes, vote => vote.vote);
+      if (allMatch(voteValues)) {
+        setVote(prevVote.vote);
+        setSubmitted(true);
+      }
+
+      const rationaleValues = ld.map(votes, vote => vote.rationale);
```
JVThomas: Instead of importing all of `lodash`, you can import the functions that you need for the file.

On line 3 you have 
```
import ld, {isNil} from "lodash";
``` 
and you call map via `ld.map(...`

You can change it to 
```
import {isNil, map} from "lodash"
``` 
and then call the map function directly. The change allows us to only import `isNil` and `map` from `lodash` and nothing else.

Lodash is a pretty big library, so calling in all of the components can add to the file size of our js files (which means larger and slower downloads for users). The selective imports is the easiest way to avoid that.

**Nit**: You can use the `lodash/fp` variant of map instead of the vanilla one. In this situation it doesn't matter which one you use, but it's easier to use and pass around the `fp` version.

The `fp` version has the map in the following format...
```
map((arg) => {codeBlock})(collection)
```
... which means you can do something like this
```
const mapValuesFn = map(() => {<custom map code>});
const mappedArray = mapValues(collection)
```
Basically you can store the map function in a variable and pass it around (other components, other functions, etc), which is useful when you repeat the same processing steps.

src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -0,0 +1,141 @@
...
+      if (allMatch(rationaleValues)) {
+        setRationale(prevVote.rationale);
+      }
+    }
+  }, [votes]);
+
+  const allMatch = (values)  => {
+    return ld.every(values, v => {
```
JVThomas: Same comment as with `map`

src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -0,0 +1,141 @@
...
+      return !isNil(v) && v === values[0];
+    });
+  };
+
+  const updateVote = async (newVote) => {
+    try {
+      const voteIds = ld.map(votes, v => v.voteId);
```
JVThomas: Same import comment as the ones mentioned before

src/components/collection_vote_box/CollectionVoteButton.js
```diff
@@ -0,0 +1,52 @@
...
+  selectedLabelColor: '#FFFFFF'
+};
+
+export default function CollectionVoteButton(props) {
+  const [additionalStyle, setAdditionalStyle] = useState({});
+  const { onClick, label, disabled, isSelected, baseColor, dataCy } = props;
+
+  useEffect(() => {
+    isSelected ? selectedButtonStyle() : defaultButtonStyle();
+  }, [isSelected]);
```
JVThomas: ESLint throws a warning here about missing dependencies in the array (`defaultButtonStyle` and `selectedButtonStyle`). Adding them in here is the right call, however the linter will then throw a new warning about how these two functions need to be defined with `useCallback`, which I'll explain on a comment on that line

src/components/collection_vote_box/CollectionVoteButton.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const defaultButtonStyle = () => {
+    updateStyle(styles.defaultBackgroundColor, styles.defaultLabelColor, false, disabled);
```
JVThomas, line 25: The way `useCallback` (a React hook like useState or useEffect) works is that it'll define and return a callback function if any defined dependencies (listed in the array) changes value.

The reason why the linter is throwing a warning is that this function, alongside selectedButtonStyle, use prop values to execute a function. Since prop values can change on a parent update, the functions' definition needs to be updated as well, so it'll have to be redefined on prop update. 

However you don't want to redefine the function on all prop updates but rather on the prop variables that matter. So for this function if you utilize `useCallback` like so...

```
const defaultButtonStyle = useCallback(() => {
    updateStyle(styles.defaultBackgroundColor, styles.defaultLabelColor, false, disabled);
  }, [disabled]);
```

... you are telling React to only redefine the function when `disabled` changes value. In addition you can reference it and pass it around like you would if you didn't use `useCallback`.

For some additional notes, take a look at this post about when to use `useCallback` or `useMemo` (similar to useCallback, but it calculates a value rather than only function).
https://kentcdodds.com/blog/usememo-and-usecallback

JVThomas: Another note, you need to declare the `useCallback` variables before the `useEffect`. React wants its hooks defined clearly (no conditional definitions) so that it can execute them in order from top to bottom, hence needing the define the functions before they're referenced in `useEffect`.

cypress/component/CollectionSubmitVoteBox/collection_submit_vote_box.spec.js
```diff
@@ -0,0 +1,248 @@
...
+        isDisabled={true}
+      />
+    );
+    cy.stub(Votes, 'updateVotesByIds');
+
+    cy.get('[dataCy=yes-collection-vote-button]').click();
+    cy.get('[dataCy=yes-collection-vote-button]').should('have.css', 'background-color', 'rgb(255, 255, 255)');
+    cy.get('[dataCy=no-collection-vote-button]').should('have.css', 'background-color', 'rgb(255, 255, 255)');
+    cy.get('textarea').should('be.disabled');
+  })
```
JVThomas, line 247: Nit: Missing semicolon

cypress/component/CollectionSubmitVoteBox/collection_submit_vote_box.spec.js
```diff
@@ -0,0 +1,248 @@
...
+        isLoading={true}
+      />
+    );
+    cy.stub(Votes, 'updateVotesByIds');
+
+    cy.get('[dataCy=yes-collection-vote-button]').click();
+    cy.get('[dataCy=yes-collection-vote-button]').should('have.css', 'background-color', 'rgb(255, 255, 255)');
+    cy.get('[dataCy=no-collection-vote-button]').should('have.css', 'background-color', 'rgb(255, 255, 255)');
+    cy.get('textarea').should('be.disabled');
+  })
```
JVThomas, line 230: Nit: Missing semicolon

src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -0,0 +1,141 @@
...
+export default function CollectionSubmitVoteBox(props) {
+  const [vote, setVote] = useState();
+  const [rationale, setRationale] = useState('');
+  const [submitted, setSubmitted] = useState(false);
+  const [disabled, setDisabled] = useState(false);
+  const {question, votes, isFinal, isLoading} = props;
+
+  useEffect(() => {
+    setDisabled(props.isDisabled || (isFinal && submitted) || isLoading);
+  }, [props.isDisabled, isFinal, submitted, isLoading])
```
JVThomas: Nit: missing semicolon

src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -0,0 +1,141 @@
...
+
+  const updateRationale = async () => {
+    try {
+      const voteIds = ld.map(votes, v => v.voteId);
+      await Votes.updateRationaleByIds(voteIds, rationale);
+      Notifications.showSuccess({text: `Successfully updated vote rationale`});
+    } catch (error) {
+      Notifications.showError({text: 'Error: Failed to update vote rationale'});
+    }
+  }
```
JVThomas: Nit: Missing semicolon

JVThomas (changes requested):
Code works functionally and the tests cover the edge cases extensively. Comments are mostly about convention, with the `lodash` comments having higher priority than that about those concerning useCallback (which would be nice to have, but not necessary).

src/components/collection_vote_box/CollectionSubmitVoteBox.js
shaemarks: When I tried switching some of my regular maps to the `lodash/fp` version, I started to experience weird bugs:

- ` map(vote => vote.vote)(votes)` led to all of the vote api calls to fail
- `map(vote => vote.rationale)(votes)` made the vote buttons and rationale textarea to be immediately disabled

I'm not sure why this would be the case because these bugs aren't directly related to the map calls

src/components/collection_vote_box/CollectionVoteButton.js
rushtong: > The reason why the linter is throwing a warning is that this function, alongside selectedButtonStyle, use prop values to execute a function. Since prop values can change on a parent update, the functions' definition needs to be updated as well, so it'll have to be redefined on prop update.

FYI, I'm not seeing any linter errors for this - let's see if we can figure out why different folks are or are not seeing what you're seeing @JVThomas

src/components/collection_vote_box/CollectionVoteYesButton.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const Label = () => {
+    return span({style: styles.label}, [
+      h(CheckCircleOutlined, {style: styles.icon}),
+      "Yes"
+    ]);
+  };
+
+  return h(CollectionVoteButton, {
+    dataCy: "yes-collection-vote-button",
```
rushtong: nit: single quote string values

rushtong (commented):
Pending Justin's feedback, this looks good to me 👍🏽

JVThomas (approved):

PR #1484, opened 2022-03-28
[DUOS-1657][risk=no] Add library card validation to DAR submission

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1657

Adds a validation check on the researcher having a library card.

----

- Verify all tests go green


src/pages/dar_application/DataUseAgreements.js
```diff
@@ -15,7 +15,7 @@ const StepAlertTemplate = (props) => {
 
   return (
     ul({style: ulLinkStyle}, [
-      li({isRendered: props.step1Invalid}, [
+      li({isRendered: props.step1Invalid || props.libraryCardInvalid}, [
```
shaemarks, line 18: Works properly with missing library card, so this comment is potentially out of scope for this ticket: 

When my profile is incomplete (I removed part of my address), I get the alert on the Step 1 page itself to complete my profile, but I do not get this link at the end of the dar application to fix Step 1. I think `props.step1Invalid` doesn't check for all ways a profile may be incomplete. Not sure on the intended behavior, but I thought I would mention it.

shaemarks (commented):
I left a comment on a potentially missing alert at the end of the dar application when I have a library card but an incomplete profile. Otherwise, looks good to me for researchers with and without library cards.

quazi-h (approved):
Changes look good to me and the UI is functioning as expected with these changes. I loaded this branch to a local instance and went through the process of trying to create a DAR both with and without a library card. I was unable to create a DAR without a library card and observed the correct warning text and links when I attempted to do so. I did observe the same thing Shae did - in my case I tried to remove the eraCommonsId. Is that piece supposed to be a required part of the profile?

src/pages/dar_application/DataUseAgreements.js
rushtong: I can reproduce this behavior in dev, so I'm going to file a separate bug ticket for: 
> When my profile is incomplete (I removed part of my address), I get the alert on the Step 1 page itself to complete my profile, but I do not get this link at the end of the dar application to fix Step 1. I think props.step1Invalid doesn't check for all ways a profile may be incomplete.

rushtong: See https://broadworkbench.atlassian.net/browse/DUOS-1685

PR #1493, opened 2022-04-04
DUOS-1293[risk=no] Member Console: Update for DAR Collection

JVThomas:
Addresses [DUOS-1293](https://broadworkbench.atlassian.net/browse/DUOS-1293)

PR adds DAR Collections table to the new member console (can be visited on `dev` via `new_member_console`) and implements member actions on the table. PR also updates `MemberActions` tests where needed.

- Verify all tests go green


rushtong (approved):
New page looks great, no bugs that I could find 👍🏽

src/pages/MemberConsole.js
```diff
@@ -1,198 +1,89 @@
...
+  const [tableSize, setTableSize] = useState(10);
+  const [descendantOrder, setDescendantOrder] = useState(false);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const init = async () => {
+      try {
+        setIsLoading(true);
+        const pendingList = await DAR.getDataAccessManageV2();
```
quazi-h, line 27: It appears that I see more records in the member_console vs the new_member_console (250 vs 50) but I think that's because the old console displays DACs that I do not have permissions to access/action. I believe the behavior I see now with your changes is what we expect.

Perhaps those old functions were not working as inteneded:
    const duls = await PendingCases.findConsentPendingCasesByUser(currentUser.dacUserId);

quazi-h (approved):
Code changes look good and when loading the UI, the dar_collections table within the member console looks good. I observed the correct member actions (Vote/Update Vote) available and did not see any election actions as expected.

src/pages/MemberConsole.js
JVThomas: Most likely it's because we're grouping DARs into collections whereas before we interacted with each DAR individually.  In that regard it would make sense that the record count would decrease.

PR #1494, opened 2022-04-05
[DUOS-1557][risk=low] multi dataset voting grouped dataset vote component

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1557)
Summary of Changes:

Component for DAC Members to cast data access votes and Chairs to view vote information about their DAC.
Rules for expected behavior:

Data Use Pills:
- list of data uses supplied by the bucket itself 

Vote Box:
- Given the votes within this bucket by the current user
- If the vote value (true/false) all match, appropriate vote button is preselected; otherwise, none selected
- If the rationale value among the votes all match, the rationale text area is prefilled; otherwise, filler text visible
- If some of the elections corresponding to the votes are closed, vote box is disabled 
- Chairs can only vote once

Pie Chart:
- Only visible to chairs 
- Displays results from votes within the current user's DAC only 
- Uses the raw counts of the votes (no collapsing); for example, if there are two members voting on 3 elections, there will be 6 votes fed to the pie chart

Vote Summary Table:
- Only visible to chairs
- Displays results from votes within the current user's DAC only 
- Collapses votes if possible: if the same user has the same vote value (true/false) for multiple votes, the vote will only be represented in the table once. If only the date or rationale differs, append date and rationale values together in the same row of the table. If the vote value differs, represent with separate rows in the table.

Datasets Requested:
- Displays only the datasets voted on by the DAC of the current user
- Determined by providing a list of all datasetIds the DAC has access to, a list of the datasetIds in this bucket, and a list of the datasets in the collection

----

- Verify all tests go green


shaemarks:
TESING: 
I tested this component by adding it to `DarCollectionReview.js`, whose page can be accessed via `http://localhost:3000/dar_collection/{#}`. This component is intended to take in the title of the slab, a bucket with DataAccess votes and elections, a list of datasets  within this collection, a list of datasetIds that the DAC of the current user has access to, and a boolean `isChair`.

In `DarCollectionReview.js`, buckets are already produced by 
```
 const processedBuckets = await flow([
          generatePreProcessedBucketData,
          processDataUseBuckets,
  ])({ dars, datasets });
```

To help with testing different scenarios, while coding I created my own examples of the props above: 

The bucket: 
```
const [bucket, setBucket] = useState({
    elections: [
      [
        {dataSetId: 385, electionId: 101, status: 'Open', electionType: 'DataAccess'},
        {electionType: 'RP Vote'}
      ],
      [
        {dataSetId: 415, electionId: 102, status: 'Closed', electionType: 'DataAccess'},
        {electionType: 'RP Vote'}
      ],
      [
        {dataSetId: 103, electionId: 103, status: 'Closed', electionType: 'DataAccess'},
        {electionType: 'RP Vote'}
      ]
    ],
   votes: [
      {
        dataAccess: {
          memberVotes: [
            {dacUserId: 5060, displayName: 'Shae', rationale: 'test1', electionId: 101, voteId: 1, createDate: 1},
            {dacUserId: 5007, displayName: 'Grace',  vote: false, rationale: 'test1', electionId: 101, voteId: 2, createDate: 1},
            {dacUserId: 3365, displayName: 'Julian', vote: true, electionId: 101, voteId: 3, createDate: 1}
          ]
        }
      },
      {
        dataAccess: {
          memberVotes: [
            {dacUserId: 5060, displayName: 'Shae', rationale: 'test2', electionId: 102, voteId: 5, createDate: 2},
            {dacUserId: 5007, displayName: 'Grace',  vote: false, rationale: 'test1', electionId: 102, voteId: 4, createDate: 1},
            {dacUserId: 3365, displayName: 'Julian', vote: false, electionId: 102, voteId: 6}
          ]
        }
      },
      {
        dataAccess: {
          memberVotes: [
            {dacUserId: 5007, displayName: 'Grace', vote: false, electionId: 103, voteId: 7},
            {dacUserId: 3365, displayName: 'Julian', vote: true, rationale: 'test3', electionId: 103, voteId: 8}
          ]
        }
      },
    ],
    dataUses: [
      {code: 'GRU', description: 'Use is permitted for any research purpose'},
      {code: 'HMB', description: 'Use is permitted for a health, medical, or biomedical research purpose'}
    ]
  });
```

In the return statement of the component:
```
 h(MultiDatasetVoteSlab, {
        title: 'GROUP 1',
        bucket,
        dacDatasetIds: [
          385, 415
        ],
        collectionDatasets: [
          {dataSetId: 385, name: 'Dataset 1', datasetIdentifier: 'DUOS-1'},
          {dataSetId: 415, name: 'Dataset 2', datasetIdentifier: 'DUOS-2'},
        ],
        isChair: true
 })
```

And then I navigated to any `http://localhost:3000/dar_collection/` page. If you use this example, replace my name and dacUserId with your own.

shaemarks:
Small bugs I have noticed:
1. When multiple rationales are appended together in the vote summary table, although I add `\n` between different rationales, it is just displayed as a space 
2. Currently `createDate` for votes is a number that is converted into the proper date format within `VoteSummaryTable.js`, so appending multiple dates together doesn't work properly. `VoteSummaryTable.js` currently isn't used anywhere else, so one possible solution is to do the formatting in `MultiDatasetVoteSlab.js` then append the formatted dates together and pass them off as a string to `VoteSummaryTable.js`. However, this will contradict the typical structure of votes

Note: II would also be interested if there is any refactoring that can be done on this component to make it easier to understand.

JVThomas:
> Small bugs I have noticed:
> 
> 1. When multiple rationales are appended together in the vote summary table, although I add `\n` between different rationales, it is just displayed as a space
> 2. Currently `createDate` for votes is a number that is converted into the proper date format within `VoteSummaryTable.js`, so appending multiple dates together doesn't work properly. `VoteSummaryTable.js` currently isn't used anywhere else, so one possible solution is to do the formatting in `MultiDatasetVoteSlab.js` then append the formatted dates together and pass them off as a string to `VoteSummaryTable.js`. However, this will contradict the typical structure of votes
> 
> Note: II would also be interested if there is any refactoring that can be done on this component to make it easier to understand.

I think HTML ignores newline characters by default unless changed via CSS. You may need to use something like `white-space: pre-line` to handle it. 

Here's a link to the CSS documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

As for the second point, refactoring is always okay so as long as you ensure that existing functionality doesn't break. Your approach makes sense, especially since the slab is the source of information for the sub-components.

As for vote structure on the table itself, it's more flexible than you would think. Things like `id` on `VoteSummaryTable` doesn't mean it has to be equal to a specific `voteId`. It's used to assign unique identifiers to DOM elements to allow React to re-render elements selectively and correctly. So you can make that value whatever you want so as long as it isn't shared by any other element. The same can be said of data and label, they're just strings.

Basically don't be too attached with making the table fit a traditional "Vote" structure. Focus on making it a JS object that has the data you need to render it correctly on a table.

src/components/collection_voting_slab/MultiDatasetVoteSlab.js
```diff
@@ -0,0 +1,201 @@
...
+    const votes = !isNil(bucket) ? bucket.votes : [];
+
+    const dacVotes = flow(
+      map(voteData => voteData.dataAccess),
+      filter((dataAccessData) => !isEmpty(dataAccessData)),
+      map(filteredData => filteredData.memberVotes),
+      filter(memberVotes => includes(user.dacUserId, map(memberVote => memberVote.dacUserId)(memberVotes))),
+      flatMap(memberVotes => memberVotes)
+    )(votes);
+    setDacVotes(dacVotes);
```
rushtong: Idea for refactoring ... some of these flow blocks seem very re-usable. It might make sense to try and abstract these to utility methods like, `extractDacUserVotesfromBucket(user, bucket)`, etc. If you do, we should group them somewhere that is DarCollection specific for easier reference by subject/category/name, etc.

rushtong (approved):
This looks great. There is room for extracting some useful functions so that components can be slimmed down a bit - but that's definitely optional.

shaemarks:
Recent change:  I made the Chairperson view of this component use Final votes instead of Member votes for the voting buttons 

Edit: Looking back, I'd like to clarify if the Chairperson view should use Final votes or Chairperson votes for the vote box and what the difference between them is.

JVThomas (approved):
Code and rendered output looks good. There's a chance you'll have to do additional styling adjustments for the component, but we won't know what that entails until `DarCollectionReview` is fully implemented.

rushtong (approved):
Nice utility extractions, thanks for the changes 👍🏽

PR #1497, opened 2022-04-08
DUOS-1594[risk=no] Refactor serial ontology calls to use bulk id param

JVThomas:
Addresses [DUOS-1594](https://broadworkbench.atlassian.net/browse/DUOS-1594)

PR swaps out old ontology request flow (parallel async requests, one for each ontology) with a single API request with a separated ids as an id param.

- Verify all tests go green

### DIFF
--- a/src/libs/dataUseTranslation.js
+++ b/src/libs/dataUseTranslation.js
@@ -1,4 +1,4 @@
-import {isNil, isEmpty, filter, join, concat, clone, uniq, head} from 'lodash/fp';
+import {isNil, isEmpty, filter, join, concat, clone, uniq, head, flow} from 'lodash/fp';
 import { searchOntology } from './ontologyService';
 import { Notifications } from './utils';
 
@@ -162,12 +162,13 @@ export const consentTranslations = {
   }
 };
 
-const getOntologyName = async(obolibraryURL) => {
-  const ontology = await searchOntology(obolibraryURL);
-  return ontology.label;
+const getOntologyName = async(urls) => {
+  const joinedUrls = urls.join(',');
+  const ontology = await searchOntology(joinedUrls);
+  return ontology.map(data => data.label);
 };
 
-export const processRestrictionStatements = async (key, dataUse, ontologyMap = {}) => {
+export const processRestrictionStatements = async (key, dataUse) => {
   let resp;
   let value = dataUse[key];
   /*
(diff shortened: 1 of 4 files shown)


src/libs/dataUseTranslation.js
```diff
@@ -162,12 +162,13 @@ export const consentTranslations = {
...
 };
 
-const getOntologyName = async(obolibraryURL) => {
-  const ontology = await searchOntology(obolibraryURL);
-  return ontology.label;
+const getOntologyName = async(urls) => {
+  const joinedUrls = urls.join(',');
+  const ontology = await searchOntology(joinedUrls);
+  return ontology.map(data => data.label);
 };
```
rushtong: ToL: the new API can accept a shortened version with just the DOIDs which might be something to consider for an optimization. For some of our current insane dev DARs, we have hundreds of ontologies which can translate to thousands of characters if we use the full url (see https://stackoverflow.com/questions/812925/what-is-the-maximum-possible-length-of-a-query-string#:~:text=Although%20officially%20there%20is%20no,a%20max%20of%202048%20characters.)

rushtong: On the Dataset side, there are no problematic cases where the urls would exceed 1024 characters so the dataset registration page is not affected. 
On the DAR side, we'll start hitting problems with ~26 or more datasets (1024/39). The Dar Collection Review page uses this code so there is a risk of breakage here with large DAR Collections.

rushtong (approved):
Code looks good aside from the query string length issue. I'm fine approving this as is and tackling a shortened version of the url in a subsequent ticket/PR.

quazi-h (approved):
Changes look good to me

src/libs/ontologyService.js
```diff
@@ -7,10 +7,22 @@ export async function searchOntology(obolibraryURL) {
...
 }
 
+export function extractDOIDFromUrl(urls) {
+  const doidArr = [];
+  urls.forEach(url => {
+    const startIdx = url.search(/DOID_\d+/);
+    if (startIdx > -1) {
+      doidArr.push(url.slice(startIdx));
+    }
+  });
```
rushtong, line 23: I think this is fine for what we need to do. In the future, it is likely that are DOIDs could have a different prefix. We're hoping to go to the MONDO ontology but we don't have any current plans. If we do, something like https://www.npmjs.com/package/url-parse might be a safer way to grab the identifier which is always the last segment of the url.

rushtong (approved):
Looks good, my only comment is totally optional 👍🏽

src/libs/ontologyService.js
```diff
@@ -13,4 +13,16 @@ export async function searchOntology(obolibraryURL) {
...
+
+  return urls;
```
quazi-h: I was expecting to see doidArr returned here instead of urls.
Isn't this just passing the input through?

quazi-h (approved):
Refactor of how we process the DOID ids/urls looks good to me, just had a question about what was being returned by the new extract function.

src/libs/ontologyService.js
JVThomas: No you're right, it needs to be doidArr.

PR #1513, opened 2022-04-21
[DUOS-1553][risk=low] multi dataset voting chair tab

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1553)
Summary of changes:

- Add MultiDatasetVotingTab for chairs
- Add validation for what tabs should be available to the user based on role and the collection being viewed
- Write tests for DarCollectionUtils methods used for subcomponents 

Testing: 
- As a user without a member or chair role,  when navigating to any `http://localhost:3000/dar_collection/{collection id #}` page, there should be no member or chair vote tabs 
- As a user with only a member role, when navigating to a `http://localhost:3000/dar_collection/{collection id #}` page, a Member Vote tab should appear if the election is associated with your DAC; otherwise, there should be no voting tabs 
- As a user with a chair role, when navigating to a `http://localhost:3000/dar_collection/{collection id #}` page, a Member Vote and Chair Vote tab should appear if the election is associated with your DAC; otherwise, there should be no voting tabs 

Note: the backend has some validation that restricts the viewing of collections based on role, so if your user has lower permissions, the page may not load at all. On the backend, the page will always load for admins, DACs if they have elections associated with the collection, researchers if they created the DAR collection.

Things that could potentially be improved: 
- As a chair, if I cast a vote on the Member Vote tab and then switch to the Chair tab, the change in my member vote is not reflected in the Vote Summary Table or Pie Chart without refreshing the page 
- Depending on how long it takes, loading the tab navigation text can look a little off / not smooth 

----

- Verify all tests go green


quazi-h (approved):
Changes here look good to me.
When testing the UI, everything is behaving as expected:
I wasnt able to see member/chair vote tabs without the correct role, just the Application Information.
When viewing a collection with a member role, I was able to view the Member Vote tab which included the SRP and appropriate vote buckets.
As a chair, I could see the Chair Vote tab which included the pie chart results and detail votes.

JVThomas (approved):
Chair tab looks good. As you mentioned in your ticket, the areas for improvement are things that are easier to tackle once we look at the page as a whole (which is outside the scope of this ticket).

There is a bug with the chair voting functionality, but the cause isn't the chair tab but rather a bug with my bucketing utility function. When corrected locally, the voting functionality works as expected, so I'll make a bug fix ticket to address that.

PR #1517, opened 2022-04-22
[DUOS-1447][risk=no] ToS Acceptance

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1447

Modified the sign in flow so that if a user has not already accepted ToS, we redirect them to a screen that allows for them to do so. If the user has already accepted, they are directed normally to their main console page which was existing behavior. If the user accepts ToS, we post that to Consent/Sam and redirect the user normally. If the user does not, we sign them out and redirect them back to the home page.

Updated the style and layout of the ToS page for easier display alongside the accept/reject button as well as bringing it visually closer what Terra does.

### New ToS page
Screen Shot 2022-04-28 at 1 47 46 PM

### ToS Acceptance page
Screen Shot 2022-04-28 at 1 47 42 PM

----

- Verify all tests go green


src/libs/tosService.js
```diff
@@ -0,0 +1,55 @@
...
+      minHeight: '700px',
+      backgroundImage: `linear-gradient(to right, transparent, white 50%), url(${homeHeaderBackground})`,
+      backgroundRepeat: 'no-repeat'
+    };
+  },
+
+  getContainerStyle: () => {
+    return {
+      margin: '50px',
+      maxWidth: '800px',
```
quazi-h, line 23: I was curious about how we decide on the standard width for containers like this one. When I get to the ToS page my browser window is larger than the container width but it doesn't scale up after the max value. It feels a little weird to me when I resize my window so I was just wondering.

quazi-h (approved):
Changes here look good and the styling and format of the new ToS page look good to me.
I tried testing the flow with the two users I have registered on DUOS. neither of which have accepted ToS.

When signing in:
- Choosing the Reject option correctly signs me out and redirects me to the homepage.
- Choosing the Accept option does update my the tosAccepted status (checked api/sam/register/self/diagostics response on swagger).
- Choosing neither option and attempting to navigate to another page is possible - should it be?
    Ex. Navigating to anything under: Your Consoles, Request Application, etc.
    If I click on the Terms of Service link on the homepage I enter the Tos Flow again.
**Should we be forcing users to accept ToS before getting access to any part of the system?**

I also had a question/comment about the max width on the ToS container (see below).

src/libs/tosService.js
rushtong: I don't think we have a standard for this type of thing. The intent here was to try and create an outer container that would be large enough to display the inner, scrollable content with some comfortable spacing.

rushtong:
> Choosing neither option and attempting to navigate to another page is possible - should it be?
Ex. Navigating to anything under: Your Consoles, Request Application, etc.
If I click on the Terms of Service link on the homepage I enter the Tos Flow again.
Should we be forcing users to accept ToS before getting access to any part of the system?

@quazi-broad - nice catch, this is a bug. We can't let them proceed without accepting ToS. Will look for a way to fix that.

rushtong:
@quazi-broad  PTAL - I've updated the navigation settings so that we set the user login status (and read from that on home page load) so nav links are not visible since the user hasn't completed their logging in yet. Also made some minor style updates per Yvonne's feedback.

quazi-h (approved):
Awesome, the new changes prevent me from bypassing the ToS page. I even tried going around the nav links and instead trying to go directly to a dar_collection review page but was taken back to the homepage and signed out.

JVThomas:
Is there anyway to have the TOS check/redirect happen before the console redirect? It's a bit weird to have a console page load before getting redirected to the TOS.
Kapture 2022-05-04 at 13 45 17

rushtong:
> Is there anyway to have the TOS check/redirect happen before the console redirect? It's a bit weird to have a console page load before getting redirected to the TOS.

I can't reproduce this ... not sure how you got to it. The tos acceptance redirect does happen before going to any other page. Look at the SignIn component - we never go anywhere UNLESS tos is already accepted. If it isn't, we go straight to the acceptance page. I wonder if some other component is doing something outside of sign in.

JVThomas (approved):

PR #1538, opened 2022-04-29
[DUOS-1777][risk=no] add bypass for lc requirement on dar submission

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1777)
Summary of Changes:
- Add field in DAR form data to track if checkbox for NIH dataset use only is checked
- Change hasLibraryCards to libraryCardReqSatisified and modify to include status of check box 
- Add tests

[Link to Consent PR](https://github.com/DataBiosphere/consent/pull/1559)
PR above needs to be used as the backend in order to save the status of the checkbox

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DataAccessRequest/researcher_info.spec.js
+++ b/cypress/component/DataAccessRequest/researcher_info.spec.js
@@ -66,7 +66,8 @@ describe('Researcher Info', () => {
   });
 
   it('renders the missing library cards alert correctly', () => {
-    mount(<WrappedResearcherInfo {...props}/>);
+    const mergedProps = {...props, ...{checkNihDataOnly: false}};
+    mount(<WrappedResearcherInfo {...mergedProps}/>);
     cy.get('[dataCy=researcher-info-missing-library-cards]').should('be.visible');
     cy.get('[dataCy=researcher-info-profile-unsubmitted]').should('not.exist');
     cy.get('[dataCy=researcher-info-profile-submitted]').should('not.exist');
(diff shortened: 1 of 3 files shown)


rushtong:
Something in the validation is triggering Step 1 being invalid, even with the `I am exclusively applying for NIH/NHGRI data (ex. GTex)` checked, which is preventing an otherwise complete DAR from being able to be submitted. Digging!

rushtong:
The `checkNihDataOnly` value doesn't seem to be sent to the back end when clicking `Save`. This also means it isn't being re-populated when the DAR is refreshed, so the user has to re-check that selection.

shaemarks:
> The `checkNihDataOnly` value doesn't seem to be sent to the back end when clicking `Save`. This also means it isn't being re-populated when the DAR is refreshed, so the user has to re-check that selection.

Is this bug occurring even with the [https://github.com/DataBiosphere/consent/pull/1559](https://github.com/DataBiosphere/consent/pull/1559) used as the backend?

rushtong:
>> The checkNihDataOnly value doesn't seem to be sent to the back end when clicking Save. This also means it isn't being re-populated when the DAR is refreshed, so the user has to re-check that selection.

> Is this bug occurring even with the https://github.com/DataBiosphere/consent/pull/1559 used as the backend?

Aha, no, I thought 1559 was merged already - will re-test with that branch.

rushtong:
With the right back end in place, I am able to save the form correctly. Still seeing this for submission though:
> Something in the validation is triggering Step 1 being invalid, even with the `I am exclusively applying for NIH/NHGRI data (ex. GTex)` checked, which is preventing an otherwise complete DAR from being able to be submitted. Digging!

rushtong (approved):

JVThomas (approved):
Tested locally in conjunction with Consent, the checkbox allows researchers to proceed with the DAR application without an LC.

PR #1541, opened 2022-05-02
[DUOS-1784] populate the pi in console pages

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1784)
Summary of changes:

Populate PIs in `DarCollectionTable`, using:
- the `createUser`'s `displayName` if their `isThePI` property is true
- the `createUser`'s `piName` property if their `isThePI` property is false
- `- -` as filler if `isThePI` is false and `piName` is empty 

----

- Verify all tests go green

### DIFF
--- a/src/components/dar_collection_table/DarCollectionTable.js
+++ b/src/components/dar_collection_table/DarCollectionTable.js
@@ -1,6 +1,6 @@
 import { useState, useEffect, Fragment, useCallback } from 'react';
 import { div, h } from 'react-hyperscript-helpers';
-import { isNil, isEmpty, find } from 'lodash/fp';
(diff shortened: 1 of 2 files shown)


rushtong (approved):
This looks good to me 👍🏽 
~One thing I notice is that we don't show the PI on chair or member consoles. We do show them in the mocks, so will file a separate ticket for that.~ Edit: ignore that, was looking at the wrong, older console pages.

quazi-h (approved):
Changes look good to me, I see the PI is now populated in the DarCollection table when I check the new_member_console and new_chair_console pages

PR #1544, opened 2022-05-03
[DUOS-1782][risk=no] filter vote summary header

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1782)
Summary of changes:
- Only display  vote summary header boxes for buckets that the current user has votes in 
- Move filtering logic on buckets from `MultiDatasetVotingTab` to `DarCollectionReview` since it is now used both for `MultiDatasetVotingTab` and `DataUseVoteSummary` (the vote summary header)
- Update tests

Notes:
- I allowed RP buckets to pass the filter without checking for votes since there is an existing bug where RP elections aren't always created. The main reason I can think of for why someone would want to check for votes in the RP bucket is to not display any vote summary header boxes for an admin navigating to this page. It shouldn't really affect other roles since the `DarCollectionReview` page has restrictions on when it will load based on the current user's roles

----

- Verify all tests go green

### DIFF
--- a/src/utils/DarCollectionUtils.js
+++ b/src/utils/DarCollectionUtils.js
@@ -206,7 +206,7 @@ export const extractUserRPVotesFromBucket = (bucket, user, isChair) => {
     map(voteData => voteData.rp),
     filter((rpData) => !isEmpty(rpData)),
     flatMap(filteredData => isChair ?
-      concat(filteredData.finalVotes, filteredData.chairpersonVotes) :
+      filteredData.chairpersonVotes :
       filteredData.memberVotes),
     filter(vote => vote.dacUserId === user.dacUserId)
   )(votes);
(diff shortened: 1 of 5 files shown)


src/utils/DarCollectionUtils.js
```diff
@@ -206,7 +206,7 @@ export const extractUserRPVotesFromBucket = (bucket, user, isChair) => {
     map(voteData => voteData.rp),
     filter((rpData) => !isEmpty(rpData)),
     flatMap(filteredData => isChair ?
-      concat(filteredData.finalVotes, filteredData.chairpersonVotes) :
+      filteredData.chairpersonVotes :
```
quazi-h, line 209: Nit: Update comment for this function

quazi-h (approved):
Changes look good to me and I am seeing the correct set of vote buckets for collections now.

JVThomas (approved):
LGTM

PR #1547, opened 2022-05-09
[DUOS-1694][risk=no] Workaround for create date parse failure

rushtong:
## Addresses
Bug found while looking into https://broadworkbench.atlassian.net/browse/DUOS-1694

When we try to send a creation date for a vote to the server, it will fail if it's not in the right format. This bug should be fixed there too, but that is a soon-to-be-deprecated endpoint so it's easier to just fix the payload here.

----

- Verify all tests go green


JVThomas (approved):
Looks good to me!

shaemarks (approved):
Looks good!

PR #1561, opened 2022-05-18
[DUOS-1740][risk=no] Single Quote Rule

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1740

Add a single quote rule to eslint

----

- Verify all tests go green


shaemarks (approved):
When using double quotes instead of single quotes, the text was highlighted with a warning and had a shortcut to swap with single quotes

.eslintrc
```diff
@@ -23,6 +23,7 @@
     "array-bracket-spacing": "warn",
     "no-unused-vars": "warn",
     "no-extra-semi": "warn",
+    "quotes": ["warn", "single"], // Checks for single quote usage where possible
```
JVThomas: Change it to `"quotes": ["warn", "single", {"avoidEscape": true}]`. It's okay to use double quotes if the string has  apostrophes in it, otherwise you have to manually escape each apostrophe in the string.

JVThomas (changes requested):

.eslintrc
rushtong: Thank you, feedback applied, PTAL

JVThomas (approved):

PR #1569, opened 2022-05-25
DUOS-1771[risk=no] Created Read Only View of DAR Collections for Admins

JVThomas:
Addresses [DUOS-1771](https://broadworkbench.atlassian.net/browse/DUOS-1771)

PR creates a new route that shows a read-only review page of the DAR Collection for admins. PR adds a link to the console page accessible by clicking on the DAR code of any collection in the DAR Collection table. `DarCollectionReview` is also updated to account for an `adminPage` boolean set as a prop on the route to allow for admin specific features for the review page (no filtering datasets or votes by DAC or user, buttons and textfields are disabled, no member tab visible).

- Verify all tests go green


src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -98,9 +98,9 @@ export default function CollectionSubmitVoteBox(props) {
 
   const VoteSubsectionHeading = () => {
     const heading = isFinal ?
-      'Your Vote* (Vote and Rationale cannot be updated after submitting)' :
+      `${!adminPage ? 'Your Vote* (Vote and Rationale cannot be updated after submitting)' : 'Vote*'}` :
       'Your Vote*';
```
rushtong: This logic is working on the dataset bucket vote slab, but not on the RP vote slab. Screenshot:

Screen Shot 2022-05-31 at 9 16 03 AM

rushtong (commented):
Page looks and functions great. One minor bug wrt RP vote slab inline.

src/components/collection_voting_slab/MultiDatasetVoteSlab.js
```diff
@@ -79,17 +79,18 @@ export default function MultiDatasetVoteSlab(props) {
...
         type: 'danger',
-        isRendered: !allOpenElections
+        isRendered: !adminPage && !allOpenElections
       }),
       h(CollectionSubmitVoteBox, {
         question: 'Should data access be granted to this applicant?',
         votes: currentUserVotes,
         isFinal: isChair,
-        isDisabled: readOnly || isEmpty(currentUserVotes) || !allOpenElections,
+        isDisabled: adminPage || readOnly || isEmpty(currentUserVotes) || !allOpenElections,
```
shaemarks, line 88: I think the ticket mentions not showing the vote buttons/rationale textbox at all for the admin page. I can see the use of showing them in the case that the admin is also a chair that voted on this collection, but I'm wondering if it would make more sense to just not show the buttons since that is not the majority of cases

JVThomas: Currently the review page shows the final vote through the buttons (highlighted background color, white if vote not submitted), the rationale through the textfield, and the member votes through the table. The admin should be able to see everything and I think the current template does that very well. 

Removing the buttons and rationale just takes information away, so it's just easier to make everything read-only and to ensure that no data filtering processes are fired for the admin view.

shaemarks (approved):
Code and tests look good to me! It may be worth updating the comments above the utils functions where `adminPage` was added as a parameter, but great otherwise

rushtong (approved):

PR #1571, opened 2022-05-25
[DUOS-1826][risk=no] dar status shows open when all elections are closed

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1826)
Summary of changes:
- fix bug where incorrect statuses were shown by filtering out RP elections in `determineCollectionStatus`
- Add tests for `determineCollectionStatus` and `processElectionStatus`
- Update string equivalence conditionals to use lowercase version of string 

----

- Verify all tests go green

### DIFF
--- a/cypress/component/utils/utils.spec.js
+++ b/cypress/component/utils/utils.spec.js
@@ -1,7 +1,8 @@
 /* eslint-disable no-undef */
-import { getSearchFilterFunctions, formatDate } from '../../../src/libs/utils';
-import { cloneDeep } from 'lodash/fp';
+import { getSearchFilterFunctions, formatDate, darCollectionUtils } from '../../../src/libs/utils';
+import {cloneDeep, toLower} from 'lodash/fp';
 import { forEach } from 'lodash';
+const { determineCollectionStatus } = darCollectionUtils;
 
 const collectionsSkeleton = [
   {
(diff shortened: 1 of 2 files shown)


cypress/component/utils/utils.spec.js
```diff
@@ -17,6 +18,53 @@ const collectionsSkeleton = [
   }
 ];
 
+const collectionWithMixedElectionStatuses = {
+  dars: {
+    0: {
```
lu-c, line 23: Why key this by index? Would an array be a better object for `dars`?

src/libs/utils.js
```diff
@@ -66,9 +66,10 @@ export const darCollectionUtils = {
             }
             return [];
           } else {
-            //if elections exist, filer out elections based on relevant ids
+            //if elections exist, filter out elections based on relevant ids
+            //only Data Access elections impact the status of the collection
             //NOTE: Admin does not have relevantIds, DAC roles do
-            const electionArr = Object.values(elections);
+            const electionArr = filter(election => toLower(election.electionType) === 'dataaccess')(Object.values(elections));
```
lu-c, line 72: LGTM, like the change for lower 👍

lu-c (approved):
LGTM

cypress/component/utils/utils.spec.js
rushtong: Real dar collection data is a map of dar reference id to dar. Similarly, elections are a map of election id to election, etc.

rushtong (approved):
LGTM 👍🏽

PR #1572, opened 2022-05-26
[DUOS-1834][risk=no] Added localstorage setting for updating the default sort for DataCollectionTables.

lu-c:
1. Updated the default sort field from title to SUBMISSION_DATE
2. Store settings when the user sorts to localstorage
3. Read this value and use that on init of the table component

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,7 +23,7 @@
     "array-bracket-spacing": "warn",
     "no-unused-vars": "warn",
     "no-extra-semi": "warn",
-    "quotes": ["warn", "single", {"avoidEscape": true}], // Checks for single quote usage where possible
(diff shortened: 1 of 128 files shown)


shaemarks (approved):
Changes look good to me! Sorting correctly defaults to descending submission date, and sort settings are maintained as the user navigates the site. 

For users with multiple roles, sort settings are seen across all consoles (ex: a change in the admin DarCollectionTable sorting is also seen in the DAC member DarCollectionTable), but I'm not sure that's something worth worrying about

quazi-h (approved):
I observed the sort on pages with Dar Collection to default to submission date as expected. Changes to sort on one page also persist when looking at other pages with Dar Collections. Changes look good to me!

PR #1573, opened 2022-05-27
[DUOS-1834][risk=no] Updated style of sorting icons

lu-c:
To give more indication as to which column is sorted on right now. Decided to do this work because it was confusing to work on DUOS-1834 while not really being able to immediately tell what the sort was

**Before**
Screen Shot 2022-05-27 at 3 23 21 PM

**AFTER**
Screen Shot 2022-05-27 at 3 21 15 PM
Screen Shot 2022-05-27 at 3 21 20 PM
Screen Shot 2022-05-27 at 3 21 25 PM

----

- Verify all tests go green

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -630,6 +630,28 @@ small {
   color: #2FA4E7;
 }
 
+.cell-sort .sort-container {
+  display: flex;
+  flex-direction: column;
+}
+
+.cell-sort .sort-container .sort-icon {
+  font-size: 20px;
+  opacity: 0.4;
+}
+
+.cell-sort .sort-container .sort-icon.active {
+  opacity: 1;
+}
+
+.cell-sort .sort-container .sort-icon.sort-icon-up {
+  transform: translateY(7px);
+}
+
+.cell-sort .sort-container .sort-icon.sort-icon-down {
+  transform: translateY(-7px);
+}
+
 .cell-sort .sort-icon {
   color: #777777;
   font-size: 10px;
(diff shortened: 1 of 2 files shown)


src/index.css
```diff
@@ -630,6 +630,28 @@ small {
...
+}
+
+.cell-sort .sort-container .sort-icon.sort-icon-up {
+  transform: translateY(7px);
+}
+
+.cell-sort .sort-container .sort-icon.sort-icon-down {
+  transform: translateY(-7px);
+}
+
```
rushtong, line 654: Can we move this to an inline style object in `SimpleTable.js`? We have been trying to move away from using index.css for a few different reasons, chief among them being the difficulty in refactoring styles when they end up used across many components.

rushtong (commented):
New icons look great. Let's also get some feedback from Yvonne wrt our current invision mocks. See feedback inline re: css.

src/index.css
lu-c: Is that decision up for discussion at all? 😂 I have some ideas and preferences for organizing styles that I've used before in the past, but I also don't want to cramp the team's style if this being done for bigger reasons.

I'm happy to make the changes here for this PR to get it closed out faster btw (don't want to just leave things lingering while waiting on larger decisions)!

rushtong: All decisions are always open for discussion! Feel free to ignore this suggestion and we can schedule a conversation with the larger team.

rushtong (approved):
👍🏽

shaemarks (approved):
Looks good to me!

src/index.css
lu-c: Ok, if you don't mind then, I'll leave the styles in place! The reasoning for that being that the cell-sort styles were already in there, there are other parts of the application using them, and it might be weird / unintuitive to have sort styling in multiple places.

PR #1578, opened 2022-05-31
[DUOS-1665][risk=low] Refactored nav bar admin tab

lu-c:
**IMPORTANT**: 
In order to see these changes locally, you will have to enable this:

Copy paste this into the javascript console of the site:
`localStorage.setItem('feature-flag:navstyle', 'DUOS-1668')`

**Changes**:
1. Copy + pasted contents of `DuosHeader.js` to `DuosNavigation.js` - Since the mobile view isnt going to change, we needed to keep some of the old content. Also kept DuosHeader around while this feature would break things (until the overall update is complete. The code has been marked with a TODO to remove. When completed, we can also completely remove `DuosHeader.js` and switch over to exclusively `DuosNavigation.js`
2. Added `NavigationTabsComponent` to the file that handles the bulk of the changes for the desktop view update.
3. updated some of the styles to use flexbox instead of float

**Risks**
1. Will forget clean up / merge the double duos headers in the future
2. This is a half feature, and doesn't completely work. Tried to mitigate this by using the localstorage to feature flag this. It should basically never be visible to users at present.

**BEFORE**
Screen Shot 2022-05-31 at 7 01 13 PM

**AFTER**
Screen Shot 2022-05-31 at 6 40 38 PM
Screen Shot 2022-05-31 at 6 40 30 PM
Screen Shot 2022-05-31 at 6 41 00 PM

----

- Verify all tests go green


src/components/DuosHeaderOld.js
```diff
@@ -0,0 +1,746 @@
+import {Component, Fragment, useState} from 'react';
```
lu-c, line 1: This entire file is the original untouched `DuosHeader.js` file, which we can get rid of later

rushtong:
Functionally, this looks great. The minimized version of the nav disappears the background, so we'll need to deal with that at some point but should not hold up PR on this point.

Screen Shot 2022-06-01 at 11 19 51 AM

src/components/DuosHeader.js
```diff
@@ -8,21 +8,59 @@ import MenuIcon from '@material-ui/icons/Menu';
...
   drawerPaper: {
     backgroundColor: '#00243c',
     color: 'white',
     fontFamily: 'Montserrat'
+  },
+  mainTab: {
+    padding: '0 25px',
+    fontSize: '16px',
+    textTransform: 'none',
+    fontFamily: '\'Montserrat\', sans-serif',
```
rushtong: nit: this also works:
```suggestion
    fontFamily: 'Montserrat, sans-serif',
```

rushtong:
Could use a little more padding at the end of the list of links. This happens when the window is at the smallest width before it switches to the right hand vertical nav:

Screen Shot 2022-06-01 at 11 32 22 AM

rushtong (approved):
This looks great! There are a couple of minor nits that should be considered optional for this PR (see comments inline and in main PR). As we move forward with other consoles, let's look for opportunities to refactor common links/elements into separate components.

lu-c:
> Functionally, this looks great. The minimized version of the nav disappears the background, so we'll need to deal with that at some point but should not hold up PR on this point.
> 
> Screen Shot 2022-06-01 at 11 19 51 AM

I think I fixed this issue! Thank you for catching it :)

rushtong:
Not for this PR - I think we need a new ticket for this. We will need to update the minimized nav panel to have sub-nav links for the admin console since the admin console page is going away. I've added a screenshot for an example of what we could do.

Screen Shot 2022-06-02 at 11 18 42 AM

rushtong (approved):
Changes look good, thank you! 👍🏽

shaemarks (approved):
Looks great!

PR #1580, opened 2022-06-01
[DUOS-1827][risk=no] chair console update and cancel buttons

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1827)
Summary of changes:
- filter out RP elections in `ChairActions` when determining which buttons should appear 
- update logic for `isCanceledEnabled` so that the cancel button only appears if there exists an open but no closed data access elections that are relevant to the chairperson 
- fix bug where a "canceled" status was displayed all lowercase in the `DarCollectionTable `
- update / add tests 

----

- Verify all tests go green


shaemarks:
Looking back at the ticket, the first bullet point only mentions filtering out RP elections for the "Update Vote" button, whereas currently these changes will filter out RP elections to determine which buttons appear for all types of actions. Which of these interpretations is correct?

Only filtering out RP elections for the Vote button example:  if all elections are open except for the RP election, the Open button would appear and the Cancel button would not

rushtong (approved):
Changes look great, thank you!

shaemarks:
### **Documenting the cases possible if the status of RP elections is considered when determining if Open and Cancel buttons appear vs always filtering out RP elections to determine actions buttons:**
- My overall opinion is that it makes the most sense to always filter out RP elections since it is consistent with how the status column of the DarCollectionTable is determined. The status column shouldn't consider RP elections because doing so will mess with  single dataset collections. Currently this PR does filter out RP elections for determining the chair action buttons 

**Rules about when buttons appear:**
The Open button appears if there are any dars without elections or there are any non-open elections 
The Cancel button appears if there are no non-open elections and there is at least one open election



**So if RP elections are included, these cases become true:**

The Open button appears if the DataAccess elections are all open but the RP election is not (either closed or canceled)
- I don't think the canceled RP condition should occur unless there was a bug 
- I don't know if we are still closing RP elections, but they at least exist in legacy code. Since RP votes aren't final, it would still be possible for the user to click the Vote button, navigate to the DarCollectionReview page, and change their RP vote without needing to re-open the elections on the collection 
- status column in DarCollectionTable would say "Open" but the open button would also appear 


The Cancel button appears if all DataAccess elections are canceled and only the RP election was open 
- I'm not sure how probable this is to occur, if at all, unless there was a bug
- status column in DarCollectionTable would say "Canceled" but the Cancel button would also appear 

The Cancel button disappears if all DataAccess elections are open and only the RP election is closed 
- meaning at that point, either the DAC must continue with the elections or use the Open button to cancel and reopen the elections, after which point the Cancel button would reappear  
- status column in DarCollectionTable would say "Open" but Cancel button would not appear, whereas a collection of all open elections (including RP) would have an "Open" status and a Cancel button. This might be confusing from the user's side since it isn't clear why the buttons are different 

Already addressed in the Jira ticket, but RP elections should not be considered for the Vote button since this would create situations where the status column of the DarCollectionTable says "Approved" or "Denied" but there is a "Update Vote" button because the RP election is Open. If the chairperson clicks the Update Vote button, they are taken to a page where only the RP election can be voted on

quazi-h (approved):
Changes look good to me, your notes on the different cases and scenarios for when buttons should appear was also very helpful to me, thanks for writing that up!

PR #1581, opened 2022-06-01
[DUOS-1779][risk=no] Update Confirmation Modal Styling

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1779

## Changes

* Minor updates to the confirmation dialog to use the current button styles for all of our console pages.
* Updated usages of the confirmation dialog where necessary to tweak for the new fonts/buttons/etc.
* Also addressed minor bug fixes found during review.

### New: 

Screen Shot 2022-06-01 at 2 14 52 PM

Screen Shot 2022-06-01 at 2 17 23 PM

### Old

Screen Shot 2022-06-01 at 2 16 28 PM

----

- Verify all tests go green

### DIFF
--- a/src/components/dar_collection_table/CollectionConfirmationModal.js
+++ b/src/components/dar_collection_table/CollectionConfirmationModal.js
@@ -32,7 +32,6 @@ export default function CollectionConfirmationModal(props) {
   const cancelModal =
     h(ConfirmationModal, {
       showConfirmation,
-      styleOverride: {height: '35%'},
       closeConfirmation: () => setShowConfirmation(false),
       title: 'Cancel DAR Collection',
       message: `Are you sure you want to cancel ${collection.darCode}?`,
--- a/src/components/dar_collection_table/CollectionConfirmationModal.js
+++ b/src/components/dar_collection_table/CollectionConfirmationModal.js
@@ -52,7 +52,6 @@ export default function CollectionConfirmationModal(props) {
 
   const openModal = h(ConfirmationModal, {
     showConfirmation,
-    styleOverride: { height: '35%' },
     closeConfirmation: () => setShowConfirmation(false),
     title: 'Open DAR Collection',
     message: `Are you sure you want to open ${collection.darCode}?`,
--- a/src/components/dar_collection_table/CollectionConfirmationModal.js
+++ b/src/components/dar_collection_table/CollectionConfirmationModal.js
@@ -35,7 +35,7 @@ export default function CollectionConfirmationModal(props) {
       closeConfirmation: () => setShowConfirmation(false),
       title: 'Cancel DAR Collection',
       message: `Are you sure you want to cancel ${collection.darCode}?`,
-      header: getModalHeader,
+      header: getModalHeader(),
       onConfirm: cancelOnClick
     });
 
@@ -46,7 +46,7 @@ export default function CollectionConfirmationModal(props) {
       closeConfirmation: () => setShowConfirmation(false),
       title: 'Revise DAR Collection',
       message: `Are you sure you want to revise ${collection.darCode}?`,
-      header: getModalHeader,
+      header: getModalHeader(),
       onConfirm: reviseOnClick
     });
 
@@ -55,7 +55,7 @@ export default function CollectionConfirmationModal(props) {
     closeConfirmation: () => setShowConfirmation(false),
     title: 'Open DAR Collection',
     message: `Are you sure you want to open ${collection.darCode}?`,
-    header: getModalHeader,
+    header: getModalHeader(),
     onConfirm: openOnClick,
   });
 
--- a/src/components/dar_drafts_table/DarDraftTable.js
+++ b/src/components/dar_drafts_table/DarDraftTable.js
@@ -97,7 +97,7 @@ const ResumeDraftButton = (props) => {
 
 //helper function to create id string for error messages
 const getIdentifier = ({id, data}) => {
-  return !isEmpty(data) ? (data.projectTitle || data.tempDarCode) : id;
+  return !isEmpty(data) ? (data.projectTitle || data.partialDarCode) : id;
 };
 
 //sub-component that renders draft delete button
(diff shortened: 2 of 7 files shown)


shaemarks (commented):
When I was looking at the different places the Confirmation Modal is used, I noticed that the Delete Draft DAR Modal from the Researcher Console has the incorrect title (see image below), and that Deactivate Library Card Modal on the Signing Official Console is taller than expected. Both of these bugs seem to be from before this PR, but I thought I would mention it. Changes look good otherwise

Delete Draft Modal

rushtong:
> When I was looking at the different places the Confirmation Modal is used, I noticed that the Delete Draft DAR Modal from the Researcher Console has the incorrect title (see image below), and that Deactivate Library Card Modal on the Signing Official Console is taller than expected. Both of these bugs seem to be from before this PR, but I thought I would mention it. Changes look good otherwise

Nice catches - I looked at these usages too and missed those bugs. Will address in this PR.

rushtong:
@shaemarks - Updated, PTAL, thank you!

src/components/dar_drafts_table/DarDraftTable.js
```diff
@@ -243,11 +243,10 @@ export default function DarDraftTable(props) {
     }),
     h(ConfirmationModal, {
       showConfirmation,
-      styleOverrise: { height: '35%' },
       closeConfirmation: () => setShowConfirmation(false),
       title: 'Delete Draft DAR',
       message: `Are you sure you want to delete DAR draft ${getIdentifier({id: selectedDraft.id, data: selectedDraft.data})}`,
-      header: getModalHeader(),
+      header: getModalHeader,
```
shaemarks: I'm realizing now that the example I was looking at with "Title" as the header was actually for a DAR draft named "Title" (sorry about that!). Now I'm not seeing a header at all, so I think the original code was correctly making the header the name of the draft

rushtong: Hah, will revert that change then!

rushtong: There was a second bug on that page where it was not finding the partial dar code correctly. Also fixed that.

src/components/dar_drafts_table/DarDraftTable.js
```diff
@@ -97,7 +97,7 @@ const ResumeDraftButton = (props) => {
 
 //helper function to create id string for error messages
 const getIdentifier = ({id, data}) => {
-  return !isEmpty(data) ? (data.projectTitle || data.tempDarCode) : id;
+  return !isEmpty(data) ? (data.projectTitle || data.partialDarCode) : id;
```
rushtong, line 100: Existing bug found during review.

shaemarks (commented):
The modals used in the `DarCollectionTable` currently don't have headers because the
 `header: getModalHeader` lines need parentheses after them. See line 38 of `CollectionConfirmationModal` as an example.

rushtong:
> The modals used in the DarCollectionTable currently don't have headers because the
header: getModalHeader lines need parentheses after them. See line 38 of CollectionConfirmationModal as an example.

Updated, thank you!

shaemarks (approved):

src/components/dar_collection_table/CollectionConfirmationModal.js
```diff
@@ -32,11 +32,10 @@ export default function CollectionConfirmationModal(props) {
   const cancelModal =
     h(ConfirmationModal, {
       showConfirmation,
-      styleOverride: {height: '35%'},
       closeConfirmation: () => setShowConfirmation(false),
       title: 'Cancel DAR Collection',
       message: `Are you sure you want to cancel ${collection.darCode}?`,
-      header: getModalHeader,
+      header: getModalHeader(),
```
rushtong, line 38: Existing bug found during review.

src/components/dar_collection_table/CollectionConfirmationModal.js
```diff
@@ -47,17 +46,16 @@ export default function CollectionConfirmationModal(props) {
       closeConfirmation: () => setShowConfirmation(false),
       title: 'Revise DAR Collection',
       message: `Are you sure you want to revise ${collection.darCode}?`,
-      header: getModalHeader,
+      header: getModalHeader(),
```
rushtong, line 49: Existing bug found during review.

src/components/dar_collection_table/CollectionConfirmationModal.js
```diff
@@ -47,17 +46,16 @@ export default function CollectionConfirmationModal(props) {
...
     });
 
   const openModal = h(ConfirmationModal, {
     showConfirmation,
-    styleOverride: { height: '35%' },
     closeConfirmation: () => setShowConfirmation(false),
     title: 'Open DAR Collection',
     message: `Are you sure you want to open ${collection.darCode}?`,
-    header: getModalHeader,
+    header: getModalHeader(),
```
rushtong, line 58: Existing bug found during review.

src/components/modals/ConfirmationModal.js
```diff
@@ -1,37 +1,64 @@
...
 const ConfirmationModal = (props) => {
   const {showConfirmation, closeConfirmation, title, message, header, onConfirm, styleOverride = {}} = props;
   const closeFn = () => closeConfirmation();
 
-  const confirmButton = button({className: 'cell-button hover-color', onClick: onConfirm}, ['Confirm']);
-  const computedStyle = Object.assign({}, Styles.MODAL.CONFIRMATION, styleOverride);
+  const duosBlue = 'rgb(0, 96, 159)';
+  const duosBlueHover = 'rgb(9,72,183)';
+
+  const SecondaryButton = styled(Button)(() => ({
```
lu-c, line 17: I think these styles would be great candidates for the global styles declaration, however we choose to handle those in the future! :)

rushtong: Agreed - I was pulling these out explicitly for that reason.

src/components/modals/ConfirmationModal.js
```diff
@@ -1,37 +1,64 @@
...
+    },
+  }));
+
+  const PrimaryButton = styled(Button)(({theme}) => ({
+    fontFamily: 'Montserrat, sans-serif',
+    color: theme.palette.getContrastText(duosBlue),
+    backgroundColor: duosBlue,
+    borderRadius: '4px',
+    fontSize: '1.45rem',
+    '&:hover': {
```
lu-c, line 36: this really is useful, and addresses some of my concerns about using js styles ::)

rushtong: I wasn't able to use regular css with the `Button` components ... had to go this route.

lu-c (approved):
LGTM

PR #1589, opened 2022-06-03
[DUOS-1666][risk=no] Added DAC console nav tabs

lu-c:
- Added nav tabs for DAC. I think these are the right locations for the children links, plz correct me if it's wrong!

----

- Verify all tests go green


shaemarks:
There are two types of roles for users on DACs: Chairs and Members. DAC Chairs can do everything members can do with some additional permissions / responsibilities. Currently only users with an admin or chairperson role can access the `/manage_dac` page, so I'd assume chairs would need links to  `this.state.dacChairPath`, `/manage_dac`, and `/dataset_catalog`, whereas members would only need links to `this.state.dacMemberPath` and `/dataset_catalog`.

src/components/DuosHeader.js
```diff
@@ -539,10 +539,24 @@ class DuosHeader extends Component {
         link: this.state.researcherPath,
         search: 'researcher_console'
       },
+      isChairPerson && {
+        label: 'DAC Console',
```
JVThomas: It might be a good idea to have distinct names for the Chair and Member consoles. A user can be a Chairperson for one DAC and be a Member in a separate DAC. In this scenario both consoles will show up (as it should), however they'll both be labeled as `DAC Console` which can be confusing.

JVThomas (dismissed):
One small change regarding the console link labeling (see comment above), but other than that the nav bar updates look good.

lu-c:
Updated the label!

quazi-h (approved):
DAC console nav tabs look good here

src/components/DuosHeader.js
```diff
@@ -539,10 +539,24 @@ class DuosHeader extends Component {
...
+        search: 'chair_console',
+        children: [
+          { label: 'Manage DARs', link: this.state.dacChairPath },
+          { label: 'Datasets', link: '/dataset_catalog' },
+          { label: 'DAC Members', link: '/manage_dac' }
+        ]
+      },
       isMember && {
-        label: 'DAC Member Console',
+        label: 'DAC Console',
```
JVThomas: Could you label this as `DAC Member Console`? Reason being is that `DAC Console` is a bit ambiguous, especially when next to `DAC Chair Console`

JVThomas (approved):
One minor adjustment, but other than that the nav bar links are good to go

src/components/DuosHeader.js
lu-c:
```suggestion
        label: 'DAC Member Console',
```

lu-c: Updated!

PR #1590, opened 2022-06-03
[DUOS-1732] UI collection read only views for researcher

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1732)
Summary of Changes:
- created new Route with path `/dar_application_review/{collection_id}`

- Updated `DataAccessRequestApplication` so that when a collectionId is provided via the path, the dar application is populated with formData from that collection. This is very similar to the existing view of completed DARs, but updated to include all datasets in the collection rather than the single dataset of a single DAR 

- Updated the `reviewCollection` function for the `NewResearcherConsole` to use `/dar_application_review/{collection_id}` instead of an arbitrary DAR application within the collection. This corrects the list of datasets displayed in the application 

- Switched lodash functions to fp versions for consistency and to get rid of the * import.
          - Function calls with just the prefix `fp` deleted shouldn't be at risk of breaking due to these changes as long as the specific function is imported. 
          - `assign` and `get` with a default value involve more of a change from the regular lodash to the lodash/fp versions. Since lodash/fp doesn't have as robust documentation as regular lodash, I tested the functions with simple examples to check my understanding of their functionality 
          - lodash/fp `assign` does not mutate the given objects (whereas the regular lodash function does involve mutation)
          - lodash/fp `get` does not offer an option for a default value, so I used `getOr` as a replacement, which uses the syntax `getOr(default, path, object)`

Currently this view is accessed through the Review buttons in the `DarCollectionTable` of the `NewResearcherConsole`. An alternative design would be to add a link to the darCode in the table similar to what is done for admins and DAC members.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DarCollectionTable/chair_actions.spec.js
+++ b/cypress/component/DarCollectionTable/chair_actions.spec.js
@@ -17,9 +17,10 @@ const nonVoteableDars = {
   1: {
     elections: {
       1: {
+        electionType: 'DataAccess',
         status: 'Open',
         votes: {
-          dacUserId: 2
+          1: {dacUserId: 2}
         }
       }
     },
@@ -31,6 +32,7 @@ const votableDars = {
   1: {
     elections: {
       1: {
+        electionType: 'DataAccess',
         status: 'Open',
         votes: {
           1: {dacUserId: 2}
@@ -43,6 +45,7 @@ const votableDars = {
   2: {
     elections: {
       3: {
+        electionType: 'DataAccess',
         status: 'Open',
         votes: {
           2: {dacUserId: 1}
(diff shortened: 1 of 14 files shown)


rushtong:
On step 4, I think we can always enable the LCA link since it doesn't add any edit-ability to the DAR:

Screen Shot 2022-06-06 at 1 04 33 PM

rushtong (approved):
Looks and works perfectly, thank you 👍🏽 
Consider the PR comment as optional.

shaemarks:
I think github responded weirdly to that last commit due to all the indentation changes and commonalities between different sections of the file. To clarify what the change is, l removed the fieldset line that disables the download if `!isNil(darCode)`

quazi-h (approved):
I am able to access the read only view of the DAR when clicking on the Review button for a DAR on the researcher console. Everything looks good to me.

PR #1596, opened 2022-06-07
[DUOS-1649][risk=no] Added sorting to the data catalog table

lu-c:
- Added clickable table header to sort the table.
- Pulled out some of the sorting field values into the default dataset obj, so the sort keys match with the data fields (just to make sorting easier)
- Added one line of a global style to update the styling of the "cell-sort" class - I confirmed with the simpletable pages that this doesnt break anything, but if this is used anywhere else, I'm not aware of it, and it might've changed things.

----

- Verify all tests go green


src/index.css
```diff
@@ -622,6 +622,11 @@ small {
   font-weight: 500;
 }
 
+.cell-sort {
```
lu-c, line 625: I confirmed with the simpletable pages that this doesnt break anything, but if this is used anywhere else, I'm not aware of it, and it might've changed things.

src/pages/DatasetCatalog.js
```diff
@@ -275,15 +297,11 @@ export default function DatasetCatalog(props) {
 
   const findPropertyValue = (dataSet, propName, defaultVal) => {
     const defaultValue = isNil(defaultVal) ? '' : defaultVal;
-    return span({}, [
-      getOr(defaultValue)('propertyValue')(find({ propertyName: propName })(dataSet.properties))
-    ]);
+    return getOr(defaultValue)('propertyValue')(find({ propertyName: propName })(dataSet.properties));
```
lu-c, line 301: Removed the span wrapping the element (so i can sort on the string value itself). This doesn't seem to have an effect on the rendering of the table

quazi-h (approved):
I am able to sort by least/greatest for numerical fields and alphabetically for text fields on the dataset catalog with these changes. I tested with multiple users and the sorting is working across different roles.

rushtong:
The DAC population seems to be broken in this PR. Compare the DAC column from here and on dev, dev has populated DACs.

rushtong:
The sorting works as expected. However, this introduces a significant amount of column shifting that is very confusing. With this, we absolutely need some form of fixed column widths to reduce the jarring effect.

rushtong (dismissed):
No issues with coding, but the DAC column is not working locally.

lu-c:
> The DAC population seems to be broken in this PR. Compare the DAC column from here and on dev, dev has populated DACs.

Ah, this was a timing issue - Tried fixing it by moving the awaits around (getDatasets now waits on getDacs!). I think this solves the issue! Thanks for catching it 👍

rushtong (approved):
LGTM 👍🏽

PR #1597, opened 2022-06-07
[DUOS-1768][risk=no] Filter Dataset Count on Chair Console

quazi-h:
[DUOS-1768](https://broadworkbench.atlassian.net/browse/DUOS-1768)

Summary of changes:
- Removed datasets variable and call to getUserRelevantDatasets from the NewChairConsole (handling filtering within getCollectionsByRoleName).
- Populate the relevantDatasets variable from collection.datasets in the DarCollectionTable.
- Passing through the relevantDatasets variable to the dataset count function in DarCollectionTableCellData.

Open Questions:
- Need to add a relevant test (DarCollectionServiceTest?)

----

- Verify all tests go green

### DIFF
--- a/src/pages/NewChairConsole.js
+++ b/src/pages/NewChairConsole.js
@@ -11,7 +11,6 @@ import { cancelCollectionFn, openCollectionFn, updateCollectionFn } from '../uti
 export default function NewChairConsole(props) {
   const [collections, setCollections] = useState([]);
   const [filteredList, setFilteredList] = useState([]);
-  const [relevantDatasets, setRelevantDatasets] = useState();
   const [isLoading, setIsLoading] = useState(true);
   const searchRef = useRef('');
   const filterFn = getSearchFilterFunctions().darCollections;
@@ -31,7 +30,6 @@ export default function NewChairConsole(props) {
           Collections.getCollectionsByRoleName('chairperson')
         ]);
         setCollections(collections);
-        setRelevantDatasets(collections);
         setFilteredList(collections);
         setIsLoading(false);
       } catch(error) {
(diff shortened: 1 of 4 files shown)


src/pages/NewChairConsole.js
```diff
@@ -27,12 +27,11 @@ export default function NewChairConsole(props) {
...
       try {
-        const [collections, datasets] = await Promise.all([
-          Collections.getCollectionsByRoleName('chairperson'),
-          User.getUserRelevantDatasets()
+        const [collections] = await Promise.all([
+          Collections.getCollectionsByRoleName('chairperson')
         ]);
         setCollections(collections);
-        setRelevantDatasets(datasets);
+        setRelevantDatasets(collections);
```
shaemarks: We shouldn't be setting the `relevantDatasets` to the collections, but the fact that this doesn't seem to mess up the `DarCollectionTable` makes me wonder if we still need the `relevantDatasets` state variable in the `NewChairConsole` component at all. I think it's only use here is to be passed to the `DarCollectionTable`, but the `DarCollectionTable` seems to replace it with `collection.datasets` in `processCollectionRowData` before ever using it

JVThomas: Yep, that'll work. You'll have to update `NewMemberConsole` to do what you've done here on `NewChairConsole` and expand the `consent` update to include members. If you do that then you can remove `relevantDatasets` as a prop from `DarCollectionTable`.

rushtong:
Do we need any UI update at all if the datasets (and therefore counts) are being filtered from the API?

shaemarks:
> Do we need any UI update at all if the datasets (and therefore counts) are being filtered from the API?

I think you're right. If the collection itself already has the datasets filtered, I don't think that the `DatasetsRequestedPanel` would need to do any additional filtering to get the relevant datasets for the buckets. Also, if which buckets are created from the collection is determined mostly by the datasets, we may be able to simplify other types of bucket filtering.

rushtong:
The functionality here was resolved with https://github.com/DataBiosphere/consent/pull/1600. Closing this PR.

PR #1599, opened 2022-06-08
[DUOS-1828][risk=no] vote summary component filter votes

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1828)
Summary of changes:
- Filtered out votes from outside of the current user's DAC when `adminPage` is false so that irrelevant votes do not change the vote result 
- Simplified `DarCollectionUtils` functions with similar filtering based on `adminPage` 
- Added tests for filtering of votes on the component 

----

- Verify all tests go green


src/utils/DarCollectionUtils.js
```diff
@@ -16,7 +16,7 @@ import {
   concat,
   findIndex,
   cloneDeep,
-  groupBy
+  groupBy, flatten
```
lu-c: minor nitpick:
```suggestion
  groupBy,
  flatten
```
To just add a single newline chara here

src/utils/DarCollectionUtils.js
```diff
@@ -167,36 +167,53 @@ export const extractDacDataAccessVotesFromBucket = (bucket, user, adminPage) =>
...
-      (memberVotes) => includes(
-        user.dacUserId,
-        map(memberVote => memberVote.dacUserId)(memberVotes)
-      )
-    )(memberVotesArrays);
+    memberVotesArrays = filterVoteArraysForUsersDac(memberVotesArrays, user);
   }
-  memberVotesArrays = flatMap((memberVotes) => memberVotes)(memberVotesArrays);
-  return memberVotesArrays;
+  return flatten(memberVotesArrays);
```
lu-c, line 173: Oh this is def the right call!

lu-c (approved):
LGTM!

src/utils/DarCollectionUtils.js
```diff
@@ -167,36 +168,53 @@ export const extractDacDataAccessVotesFromBucket = (bucket, user, adminPage) =>
...
+// Applies filter to arrays of votes grouped by election and
+// only keeps arrays where at least one vote has the dacUserId of the provided user
+const filterVoteArraysForUsersDac = (voteArrays = [], user) => {
+  const userIdsOfVotes = (votes) => {
+    return map(vote => vote.dacUserId)(votes);
+  };
+
+  return filter(
+    voteArray => includes(user.dacUserId, userIdsOfVotes(voteArray))
+  )(voteArrays);
```
rushtong, line 217: I'm not clear on this method. The intention, I think, should be to filter on the user's DAC, not the user. The logic, as I understand it, should look something like this:

1. Find the user's DAC id(s)
2. Filter the votes for users who have the same DAC id(s)

rushtong (changes requested):
Please correct me if I'm wrong here, but I think this PR filters too many votes from being displayed in the vote summary bar. This will find all votes for a user, but not those of other users in the same data access committee.

src/utils/DarCollectionUtils.js
shaemarks: **Documenting the task and possible approaches**

Here we want all votes of a specific type (member, chairperson, final) in a particular bucket that belong to the DAC(s) of the current user. This type of filtering is seen in the vote summary header and the pie chart/vote summary table on the chair tab. 

Within a bucket, bucket.votes is structured as follows:
```
votes: [
  0:  { 
        rp: {   //all votes within this object have the same electionID
             chairpersonVotes: [ vote array filled with individual vote objects],
             finalVotes: [{...}, {...}, {...}, ...],
             memberVotes: [{...}, {...}, {...}, {...}, {...}, {...}, ...],
        },
        //if this is not the rp bucket, another object like the rp one above will be present
        dataAccess: {   //all votes within this object have the same electionID
             chairpersonVotes: [{...}, {...}, {...}, ...],
             finalVotes: [{...}, {...}, {...}, ...],
             memberVotes: [{...}, {...}, {...}, {...}, {...}, {...}, ...],
        },
    } ,
  1: { .... }
...
]
```


With the current approach, since each array of vote objects has the same `electionId` for all votes, we see if any votes in the array have the `dacUserId` of the current user. If so, we conclude that the vote array is for an election for the dac of the current user and keep the votes in the array. 
Considerations:
- Relies on the structure of the vote arrays rather than things directly observable in the data itself 
- I'm not sure how switching from the 1 dar : 1 dataset to 1 dar : N datasets model or other changes could potentially impact things down the line 
- requires more background knowledge on the structure of the data 
- requires fewer API calls and fewer checks as we filter the votes 

For the alternative approach described in your comment, we can 
1) iterate through the roles of the current user and retrieve the DAC id(s) associated with the roles, if present 
2) Use the `/api/dac/{dacId}` endpoint to get the DAC(s)
3) map through the chairperson and member arrays of the DAC(s) and get their `dacUserId`s
4) when filtering the votes, for each vote, get the `dacUserId` associated with it and see if the id is present in the list of `dacUserId`s from step 3. If so, keep; discard otherwise 
Considerations:
- relies on retrievable properties in the data, so more flexible to changes in structure 
- easier to figure out how it works for someone reading the code
- if we put the api call closer to when the filtering occurs, we would hit the same endpoint multiple times and get the same result; if we put the api call in a higher level component, it might not be immediately intuitive why we are passing down a list of `dacUserId`s 
- filtering votes requires checking the `dacUserId` associated with each individual vote against another list of `dacUserId`s (a list against another list vs a list against a single value)


Hopefully this helps the discussion about which approach is preferred

rushtong: > With the current approach, since each array of vote objects has the same electionId for all votes, we see if any votes in the array have the dacUserId of the current user. If so, we conclude that the vote array is for an election for the dac of the current user and keep the votes in the array.

I'm comfortable with your current approach. I was incorrect to think the filtering would filter out votes from other users within the scope of the current election.

rushtong (approved):
👍🏽

PR #1602, opened 2022-06-09
[DUOS-1852][DUOS-1868][risk=no] hide alert if in readOnly mode.

lu-c:
Sorry, wasn't able to find a DAR to test this on, this is just a pure code change that I *assume* works... no reason for it not to?

What I did to try to FIND one to test this on, i went to `/admin_review_collection`, clicked on one that had "Closed: 2", then clicked on the Chair Vote tab. The ticket suggests finding one that is approved or denied, but that doesn't seem to be one of the statuses in the table - I assume that a "Closed" dar was closed because it was either approved or denied. 

Anyway, that's the logic I went through here. I think this fulfills teh ticket tho, but I'm sorry I couldnt do it right!

----

- Verify all tests go green

### DIFF
--- a/src/components/collection_voting_slab/MultiDatasetVoteSlab.js
+++ b/src/components/collection_voting_slab/MultiDatasetVoteSlab.js
@@ -79,7 +79,7 @@ export default function MultiDatasetVoteSlab(props) {
       h(Alert, {
         title: 'Voting is disabled since not all elections are open.',
         type: 'danger',
-        isRendered: !adminPage && !allOpenElections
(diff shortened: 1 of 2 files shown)


rushtong:
Any "Closed" election status should also work for this story. So, if you tested that case, then you tested the other cases as well.

rushtong (approved):
👍🏽

lu-c:
> Any "Closed" election status should also work for this story. So, if you tested that case, then you tested the other cases as well.

Haha, perfect, thank you XD

phendriksen100 (approved):

PR #1603, opened 2022-06-09
[DUOS-1855][risk=no] fixed error when loading data collection with "no elections"

lu-c:
- Always pushes the DAR reference Ids when making the dsde-dev match-purpose, so we are not making a blank request
- Tested it with a trash ID of `sdfsdf`, and confirmed it doesnt 404, but will just return a blank array

----

- Verify all tests go green


rushtong (approved):
Once the test failure is addressed, this looks good 👍🏽

src/utils/DarCollectionUtils.js
```diff
@@ -267,10 +267,16 @@ export const getMatchDataForBuckets = async (buckets) => {
         ).referenceId;
         return isNil(dataAccessReferenceId);
       });
+
       if(!isNil(dataAccessReferenceId)) {
         idsArr.push(dataAccessReferenceId);
         electionIdBucketMap[dataAccessReferenceId] = bucket;
       }
+
+      dars.forEach((dar) => {
```
JVThomas: Instead of adding the dar reference ids here, why not just add an if-block around the match fetch logic (around line 284)?

If the idsArray is empty then we shouldn't have to make the fetch at all, even more so if sending the dar referenceIds in the GET request just results in a blank array.

JVThomas (dismissed):
I left one comment above that should work and prevent the UI from making unnecessary API calls.

src/utils/DarCollectionUtils.js
lu-c: Ah, we could do that too, I suppose - I was following the language on the ticket that said it should be searching by darId as well. I assumed there would be some data out there that would have data for the darId because of that, but that assumption may be wrong! Just to be sure, would it be possible to ever get data returned when requesting by the reference id of the dar?

JVThomas (approved):
Address the Codacy error and the code will be good to go!

PR #1605, opened 2022-06-10
DUOS-1776[risk=no] Update Member Vote table on DAR Collection Review Chair Tab

JVThomas:
Addresses [DUOS-1776](https://broadworkbench.atlassian.net/browse/DUOS-1776)

PR fetches for Collection data when switching to the Chair tab to ensure up to date Member vote data for Chair display. PR also includes several styling and logic adjustments to provide loading animations when the fetch occurs.

- Verify all tests go green

[DUOS-1776]: https://broadworkbench.atlassian.net/browse/DUOS-1776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rushtong (approved):
Looks good! 👍🏽

src/components/vote_summary_table/VoteSummaryTable.js
```diff
@@ -86,7 +87,7 @@ function nameCellData({name = '- -', voteId, label = 'name'}) {
 
 function dateCellData({date, voteId, label = 'date'}) {
   return {
-    data: date,
+    data: formatDate(date),
```
shaemarks: To account for legacy data where dates or rationales of votes may differ for a single user, we send the votes through `collapseVotesByUser` from `DarCollectionUtils`. In the case of multiple dates, they need to be formatted individually before being appended, which occurs on line 339 of `DarCollectionUtils`. When I voted on the member tab and switched to chair, everything worked properly except for the date not being displayed, so I think adding a call to `formatDate` here is messing with it.

shaemarks (commented):
Everything works great except for the date not being displayed in the VoteSummaryTable when voting on the member tab and switching to the chair tab. See other comment for context

src/components/vote_summary_table/VoteSummaryTable.js
JVThomas: I think it might be more than that. There's also the fact that the votes on the Collection have an `updateDate` and not a `createDate`.

Screen Shot 2022-06-14 at 9 59 52 AM

To be safe I'm going to check on both `createDate` and `updateDate` on the collapse function (`updateDate` being primary, `createDate` as a fallback) before pushing them into the array. I'll also change the name of the array to `lastUpdates` to avoid confusion with the `updateDate` or `createDate` terms since each date can be either one of them.

rushtong: Side note: In consent, `updateDate` is null until a vote has been cast. When we create votes for elections in a DAR Collection, we always set a `createDate` value.

shaemarks (approved):
Thanks for the changes! I'm observing proper updates when voting on the member tab and switching to the chair tab, and after voting, the date appears in the table next to my vote. I'm not seeing a date in the table before casting a vote (from `createDate`, assuming those dates are generated upon election creation) but that may be better suited for a separate ticket if that is an issue

JVThomas:
> Thanks for the changes! I'm observing proper updates when voting on the member tab and switching to the chair tab, and after voting, the date appears in the table next to my vote. I'm not seeing a date in the table before casting a vote (from `createDate`, assuming those dates are generated upon election creation) but that may be better suited for a separate ticket if that is an issue

Yeah I noticed that as well. I am checking for both `updateDate` and `createDate` so I'm not sure if it's a front-end or back-end issue right now. In terms of the PR I think it's a non-issue. The dates don't show up on un-submitted votes; they are only visible when the vote is submitted. Thinking about it from a Chair perspective I can see that making sense. 

Nevertheless the issue does warrant investigation, I just think it needs to be a separate ticket (like you said).

PR #1610, opened 2022-06-13
[DUOS-1853][risk=no] update read-only vote button styles

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1853)
Summary of changes:
- Updated logic for `VoteSubsectionHeading` within `CollectionSubmitVoteBox.js` so that read-only scenarios show a sentence detailing the selected vote rather than the usual "My Vote*" heading 
- Hide voting buttons for read-only scenarios 
- Update cypress tests

Notes:
- the admin page view (accessed by clicking linked dar codes in the DarCollectionTable) always uses text instead of vote buttons based on the chairperson / final votes 
- the read-only view for DACs (accessed by clicking linked dar codes in the DarCollectionTable when elections are closed) always uses text based on their votes, which can be YES / NO / NOT SELECTED
- when a chairperson submits a vote on a bucket with datasets (not RP), the vote buttons disappear and text appears reflecting their vote since it cannot be changed once submitted. This closes the election, so the member tab will also have their buttons replaced with text for this bucket since voting is disabled 

----

- Verify all tests go green


src/components/collection_vote_box/CollectionSubmitVoteBox.js
```diff
@@ -48,7 +71,7 @@ export default function CollectionSubmitVoteBox(props) {
   const {question, votes, isFinal, isApprovalDisabled, isLoading, adminPage} = props;
 
   useEffect(() => {
-    setIsVotingDisabled(props.isDisabled || (isFinal && submitted) || isLoading);
+    setIsVotingDisabled( props.isDisabled || (isFinal && submitted) || adminPage);
   }, [props.isDisabled, isFinal, submitted, isLoading]);
```
JVThomas: Nit: Add `adminPage` to the dependency array. Even if the value doesn't change after component load, adminPage is a prop and the `useEffect` depends on its value.

JVThomas (approved):
One nitpick, but other than that the code looks good

rushtong (approved):
👍🏽

rushtong (approved):
Still 👍🏽

JVThomas (approved):
Looks good to me

PR #1624, opened 2022-06-15
[DUOS-1881][risk=no] Added functionality to "request help" button

lu-c:
- Added contact us modal on "request help" click
- Removed existing "request help" dropdown (cause it had nothing in it anyway)
- Changed styles for modal so it will show up better in mobile / shrunken screens

before
image

after
Screen Shot 2022-06-15 at 3 29 21 PM

Before
image

After
Screen Shot 2022-06-15 at 3 31 38 PM

----

- Verify all tests go green

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -61,6 +61,12 @@ const styles = {
     color: '#00609f',
     minHeight: '65px',
     fontWeight: 'bold'
+  },
+  navButton: {
+    background: 'transparent',
+    color: 'white',
+    border: 'none',
+    minHeight: '80px'
   }
 };
 
@@ -179,7 +185,7 @@ const NavigationTabsComponent = (props) => {
     makeNotifications,
     navbarDuosIcon, duosLogoImage, DuosLogo, navbarDuosText,
     currentUser, signOut, isLogged,
-    contactUsButton, supportrequestModal,
+    contactUsButton, showRequestModal, supportrequestModal,
     tabs, initialTab, initialSubTab
   } = props;
   const [selectedMenuTab, setSelectedMenuTab] = useState(false);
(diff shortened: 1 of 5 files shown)


src/components/DuosHeader.js
```diff
@@ -262,29 +268,17 @@ const NavigationTabsComponent = (props) => {
...
-            ])
-          ]),
+
+        button({
+          onClick: showRequestModal,
+          style: styles.navButton
+        }, [
+          div({ id: 'help', style: { whiteSpace: 'nowrap' } }, [
+            'Request Help',
+            span({ className: 'caret caret-margin' }),
```
shaemarks: Since this is no longer a dropdown, I think it would make more sense to remove the caret.

src/components/DuosHeader.js
```diff
@@ -262,29 +268,17 @@ const NavigationTabsComponent = (props) => {
...
-              ]),
-            ])
-          ]),
+
+        button({
+          onClick: showRequestModal,
+          style: styles.navButton
+        }, [
+          div({ id: 'help', style: { whiteSpace: 'nowrap' } }, [
+            'Request Help',
```
shaemarks: The label is "Request Help" on the full navbar view and "Contact Us" on the compressed hamburger menu view even though they trigger the same modal. Would it be better to make this consistent across both views?

shaemarks (commented):
Contact Us modal looks good and responds well to different page sizes. I left a couple of small comments, but they aren't defined in the ticket so potentially get a second opinion on them

src/components/DuosHeader.js
JVThomas: I think it makes sense to keep them the same. It keeps parity between the two navbars and doesn't require the user to figure out what the link label will be on certain views.

src/components/DuosHeader.js
lu-c: gotcha, can get rid of this

src/components/DuosHeader.js
lu-c: Would you guys prefer the language to be always "contact us" or always "request help" or a mix of both? What if we changed the title in the modal so it changes based on the language of the link you clicked on to open it?

JVThomas (approved):
Changes look good. Link shows up on both navigation bars and the labelling is consistent.

shaemarks (approved):

PR #1632, opened 2022-06-22
[DUOS-1876][risk=no] updated table css for mobile view on vote summary card

lu-c:
BEFORE
Screen Shot 2022-06-22 at 1 07 03 PM
Screen Shot 2022-06-22 at 1 07 11 PM

AFTER
Screen Shot 2022-06-22 at 1 04 12 PM
Screen Shot 2022-06-22 at 1 04 24 PM

----

- Verify all tests go green

### DIFF
--- a/src/components/CollectionAlgorithmDecision.js
+++ b/src/components/CollectionAlgorithmDecision.js
@@ -11,10 +11,9 @@ export default function CollectionAlgorithmDecision(props) {
     style: Object.assign(
       {
         padding: '4%',
-        flex: 1,
-        display: 'flex',
         justifyContent: 'space-around',
-        fontFamily: 'Montserrat'
+        fontFamily: 'Montserrat',
+        width: '50%'
       },
       styleOverride
     )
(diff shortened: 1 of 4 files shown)


rushtong (approved):
LGTM 👍🏽

shaemarks (commented):
Changes look good! Should we do anything to the sizing of the pie chart under the Structured Research Purpose slab? Since there's no algorithm decision for the rp vote, we only show the pie chart, which previously filled up the whole space because of the flex box. Now that the pie chart width is set to 50%, the right half of that section is just blank space

Before:
pie chart before


After:
pie chart after

lu-c:
> Changes look good! Should we do anything to the sizing of the pie chart under the Structured Research Purpose slab? Since there's no algorithm decision for the rp vote, we only show the pie chart, which previously filled up the whole space because of the flex box. Now that the pie chart width is set to 50%, the right half of that section is just blank space
> 
> Before: pie chart before
> 
> After: pie chart after

Updated to fix this issue - it'll now check if the second column is going to be rendered, and provides the width part as a style override (imitating how it was handled in the right box)

shaemarks (approved):

PR #1644, opened 2022-06-28
[DUOS-1886][risk=no] Updated condition to use env for the new duos header

lu-c:
Removed that one line :) Ezpz!

----

- Verify all tests go green

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -74,7 +74,7 @@ function App() {
   };
 
   // TODO: Remove this when all the new navigation features are available
-  const header = env === 'dev' || env === 'staging'
+  const header = env !== 'prod'
     ? h(DuosHeader, { onSignOut: signOut })
     : h(DuosHeaderOld, { onSignOut: signOut });
 
--- a/src/App.js
+++ b/src/App.js
@@ -74,7 +74,7 @@ function App() {
   };
 
   // TODO: Remove this when all the new navigation features are available
-  const header = localStorage.getItem('feature-flag:navstyle') === 'DUOS-1668' //eslint-disable-line
+  const header = env === 'dev'
     ? h(DuosHeader, { onSignOut: signOut })
     : h(DuosHeaderOld, { onSignOut: signOut });
 
--- a/src/App.js
+++ b/src/App.js
@@ -74,7 +74,7 @@ function App() {
   };
 
   // TODO: Remove this when all the new navigation features are available
-  const header = env === 'dev'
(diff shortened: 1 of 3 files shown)


lu-c:
Going to head off some of the routing comments - I think I can guess what some of the feedback is going to be, and would like to respectfully disagree ;D

**The old admin console page and route should be removed**: The new navbar is a desktop only feature, the mobile view still has a link to the old admin console page. I don't think there's a single link we can add that would work to replace this link, and the solution is to update the navbar so it's compatible with mobile view

src/App.js
```diff
@@ -74,7 +74,7 @@ function App() {
   };
 
   // TODO: Remove this when all the new navigation features are available
-  const header = localStorage.getItem('feature-flag:navstyle') === 'DUOS-1668' //eslint-disable-line
+  const header = env === 'dev'
```
JVThomas: We're going to be enabling this on everything but prod, so you'll need to say `env !== 'prod'`

JVThomas: You'll also need to update the Navigation methods on `utils.js` (lines 257 - 307) as well as the Admin test on `auth.cy.js`

lu-c: Looking at the test in `auth.cy.js`, I don't actually think it needs to change. The text that they're looking for, "Admin Console" should actually still be visible in the new header - Is that right? Am I understanding thsi correctly?

JVThomas: The current test is running on the mobile view as seen in the screenshot below.

Screen Shot 2022-06-29 at 10 31 31 AM

Because of that, the 'Admin Console' label on the navbar doesn't show up. This means you have a couple of ways to approach updating the test. You could either update it so that it checks on the desktop view or you can update the `cy.contains` statement to look for something else on the page.

lu-c: Unfortunately, there's no other text on that page that reallyyyyy indicates the page is part of the admin console... Ok, thanks for working with me on this! ^_^ I'll try forcing the desktop view

JVThomas: Yeah the integration tests are 1) old and 2) extremely minimal.  Feel free to make the tests more detailed so that anyone can look at the test and know for sure that it's testing for stuff that's uniquely admin.

quazi-h (approved):
Change looks pretty straight forward and my local instance is displaying the new navbar. Updated logic should enable the new navbar on dev + staging.

rushtong (approved):
Looks good!

JVThomas (approved):

PR #1651, opened 2022-06-30
[DUOS-1692] Add reject ToS button

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1692

Adds 'Reject' button the ToS page, which allows a user to reject the terms after they've already logged in and accepted.

Requires backend on [this branch](https://github.com/DataBiosphere/consent/tree/DUOS-1691-reject-tos) as a new endpoint had to be created.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1137,6 +1137,11 @@ export const ToS = {
     const url = `${await Config.getApiUrl()}/api/sam/register/self/tos`;
     const res = await axios.post(url, {}, Config.authOpts());
     return res.data;
+  },
+  rejectToS: async () => {
+    const url = `${await Config.getApiUrl()}api/sam/register/self/tos`;
+    const res = await axios.delete(url, Config.authOpts());
+    return res.data;
   }
 };
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1139,7 +1139,7 @@ export const ToS = {
     return res.data;
   },
   rejectToS: async () => {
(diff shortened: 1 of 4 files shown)


src/libs/ajax.js
```diff
@@ -1137,6 +1137,11 @@ export const ToS = {
     const url = `${await Config.getApiUrl()}/api/sam/register/self/tos`;
     const res = await axios.post(url, {}, Config.authOpts());
     return res.data;
+  },
+  rejectToS: async () => {
+    const url = `${await Config.getApiUrl()}api/sam/register/self/tos`;
```
rushtong: Missing a slash:
```suggestion
    const url = `${await Config.getApiUrl()}/api/sam/register/self/tos`;
```

rushtong (changes requested):
Looks and works perfectly for a signed in user. We need to hide this for users who have not signed in yet, so this needs that one tweak (and path fix in comments) before moving forward.

connorlbark:
should be fixed now! slash added + button not rendered if logged out.

rushtong (approved):
Thank you for the changes 👍🏽

quazi-h (approved):
Changes look good to me and I was able to use the reject ToS button when signed in and I do not see it when signed out.

UI behavior is working as expected - when clicking the reject button, I am automatically signed out and when signing back in I am routed directly to the ToS page where I must accept to be able to access the UI. The reject button is grayed out if I have not made a decision yet and if I choose to reject here, I am signed out.

PR #1654, opened 2022-07-02
[DUOS-1909][risk=no] Migrate from User.dacUserId to User.userId

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1909

----

- Verify all tests go green


lu-c (approved):
Looks straightforward! LGTM

src/components/data_custodian_table/DataCustodianTable.js
```diff
@@ -310,16 +310,16 @@ export default function DataCustodianTable(props) {
   };
 
   const removeDataCustodian = async (selectedResearcher, researchers) => {
-    const { displayName, email, dacUserId } = selectedResearcher;
-    const searchableKey = !isNil(dacUserId) ? 'dacUserId' : 'email';
+    const { displayName, email, userId } = selectedResearcher;
+    const searchableKey = !isNil(userId) ? 'dacUserId' : 'email';
```
quazi-h, line 314:
```suggestion
    const searchableKey = !isNil(userId) ? 'userId' : 'email';
```
Is the logic here correct? if the userId is not Nil, we should be using it for the searchableKey, right?

quazi-h (commented):
Changes look good, I could only find one reference to dacUserId in logic that did not involve Vote/Match, added a comment below, in the DataCustodianTable code.

I also found the following in Routes.js:
**line 75:**
AuthenticatedRoute path="/admin_edit_user/:dacUserId" component={AdminEditUser} props={props} rolesAllowed={[USER_ROLES.admin]}

**line 117:**
AuthenticatedRoute path="/researcher_review/:dacUserId" component={ResearcherReview} props={props} rolesAllowed={[USER_ROLES.admin]}

I wasn't sure if those two lines above also needed updated references to userId but the first of those routes (admin_edit_user) is working correctly. The second (researcher_review) throws an error when I try to navigate to the URL.
I tried "http://localhost:3000/researcher_review/5052" for example but saw the error:
**TypeError: Cannot read properties of undefined (reading 'researcherProperties')**
I'm not sure if I am just accessing that researcher_review page incorrectly.

src/components/data_custodian_table/DataCustodianTable.js
rushtong: Yes, I think this logic still holds true here.

rushtong:
> Changes look good, I could only find one reference to dacUserId in logic that did not involve Vote/Match, added a comment below, in the DataCustodianTable code.
> 
> I also found the following in Routes.js: **line 75:** AuthenticatedRoute path="/admin_edit_user/:dacUserId" component={AdminEditUser} props={props} rolesAllowed={[USER_ROLES.admin]}
> 
> **line 117:** AuthenticatedRoute path="/researcher_review/:dacUserId" component={ResearcherReview} props={props} rolesAllowed={[USER_ROLES.admin]}
> 
> I wasn't sure if those two lines above also needed updated references to userId but the first of those routes (admin_edit_user) is working correctly. The second (researcher_review) throws an error when I try to navigate to the URL. I tried "http://localhost:3000/researcher_review/5052" for example but saw the error: **TypeError: Cannot read properties of undefined (reading 'researcherProperties')** I'm not sure if I am just accessing that researcher_review page incorrectly.

Thanks @quazi-broad. I'll update the two route paths. `researcher_review` looks like an orphaned page now - there's no way to get to it from the UI, so I'm not going to try and fix that page. I'll file a ticket to remove that page/route.

PR #1658, opened 2022-07-05
[DUOS-1910][risk=no] Duos mobile header

lu-c:
**After:**
Logged out:
Screen Shot 2022-07-05 at 1 50 27 PM

Logged in:
Screen Shot 2022-07-05 at 1 50 49 PM

----

- Verify all tests go green


package.json
```diff
@@ -66,7 +64,7 @@
     "google-auth-library": "8.1.0",
     "html-webpack-plugin": "4.5.2",
     "react-error-overlay": "6.0.9",
-    "react-scripts": "4.0.3",
+    "react-scripts": "^2.1.3",
```
rushtong: Is this intentional? I can foresee github complaints about this as there are a ton of security issues with older transient dependencies in react-scripts

connorlbark:
The navbar itself looks great and works really intuitively! My only nitpick in how it works is how it handles links to the same page. For example, there are multiple ways to get to the Dataset Catalog page. However, no matter how you get there, the navbar will show the 'Admin Console->Dataset Catalog' tab (for me), even if I tapped the 'Researcher Console->Data Catalog' button. It could be that the expected behavior here would be to stay on the Researcher Console tab.

I also noticed there were some bugs if you make the page the size of a smartphone. Do we anticipate anyone using DUOS on a mobile phone, or is that not a use case we're supporting?

rushtong (approved):
Looks great! I do have one suggestion that could be in a follow-up ticket. Once a link is clicked, the nav panel should no longer display. If that's easy enough to do in this PR, please do, otherwise please file a separate story for that.

rushtong:
> I also noticed there were some bugs if you make the page the size of a smartphone. Do we anticipate anyone using DUOS on a mobile phone, or is that not a use case we're supporting?

FYI, @connorlbark, we do not officially support mobile/tablet usage. That is a fantastic idea, but our user base isn't there yet.

connorlbark:
> Looks great! I do have one suggestion that could be in a follow-up ticket. Once a link is clicked, the nav panel should no longer display. If that's easy enough to do in this PR, please do, otherwise please file a separate story for that.

this would solve my concern as well! so I will approve with the same condition.

connorlbark (approved):

PR #1666, opened 2022-07-08
DUOS-1781[risk=no] Updated styling for placeholder components on the dar collection review header

quazi-h:
Ticket: [DUOS-1781](https://broadworkbench.atlassian.net/browse/DUOS-1781)

- Updated styling on placeholder components for the dar collection ReviewHeader and DataUseVoteSummary.
- Updated the number of vote result placeholders to 2 since that is the minimum that will be shown.

----

- Verify all tests go green


shaemarks (commented):
Placeholder component sizes are a lot closer to the actual page components 👍 
It might be worth updating `TabControl` as well since the placeholder is a lot shorter than the actual tab

quazi-h:
Updated styling in TabControl so the placeholder looks similar to the "Application Information" tab.
When testing the changes with different collections I noticed that it's still somewhat abrupt when viewing a collection with no DataUseVote buckets. I attempted to write some logic in DataUseVoteSummary - where the loadingTemplate is called: but was unable to get it working - I think because the dataUseBuckets/chunkedBuckets variable is not available when isLoading is True.
line 96 in DataUseVoteSummary:
`(dataUseBuckets.length > 0) ? loadingTemplate : null;`

shaemarks (approved):

JVThomas (approved):
Changes look good to me as well

PR #1672, opened 2022-07-13
[DUOS-1815][risk=no] New Researcher Profile Page

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1815

Notes:
- Calls new `PUT /api/user/me` endpoint on profile page - endpoint created at https://github.com/DataBiosphere/consent/pull/1649 , so must use that backend branch for testing.
- Changes the Institution and Signing Official search bars to allow freetext entry (press enter after searching) separately from selecting a pre-existing institution. If set, will update `suggestedInstitution`/`suggestedSigningOfficial` instead of sending the ids. 
- The page is different if you are a research or a signing official. Signing officials cannot change their institution once set. Researchers may change their institution and signing officials anytime.

----

- Verify all tests go green


lu-c:
Hi!

I'm seeing some differences in the branch between the mock ups - Am I running it wrong?

1. "An eRA Commons ID will be required to submit a dar." <-- The eRA commons id is a link in the mock up, but seems like just text here
2. The checklist with "what are you trying to do" isn't showing up - Reading the ticket, I think it might be out of scope?

src/pages/ResearcherProfile.js
```diff
@@ -79,72 +46,28 @@ export default function ResearcherProfile(props) {
...
+        setSigningOfficialList((institution.signingOfficials ? institution.signingOfficials : []));
+      });
     }
+  }, [profile.institutionId]);
 
-    setIncompleteFields(incompletes);
-    setResearcherFieldsComplete(incompletes.length === 0);
-  }, [profile]);
+  const hasInstitution = () => {
+    return (isNumber(profile.institutionId) && profile.institutionId !== 0) || (profile.suggestedInstitution !== undefined && profile.suggestedInstitution !== '');
```
lu-c, line 58: Screen Shot 2022-07-14 at 2 34 52 PM

This method isn't updating, can we check why this is happening?

I added this console log, and confirmed that my profile DOES have an `institutionId: 231` value, but it's not updating for some reason!

src/pages/ResearcherProfile.js
```diff
@@ -350,31 +169,56 @@ export default function ResearcherProfile(props) {
           type: 'text',
           disabled: true,
```
lu-c, line 170: Should this institution input always be set to `disabled: true`?

src/pages/ResearcherProfile.js
connorlbark: heya! do you mean why is institution greyed out? it's probably because that account is a signing official, which means you cannot update your institution. i had to use a second account to test this PR which was just a researcher

connorlbark:
> Hi!
> 
> I'm seeing some differences in the branch between the mock ups - Am I running it wrong?
> 
> 1. "An eRA Commons ID will be required to submit a dar." <-- The eRA commons id is a link in the mock up, but seems like just text here
> 2. The checklist with "what are you trying to do" isn't showing up - Reading the ticket, I think it might be out of scope?

oops, you're right about the link eRA commons link! and yes, i believe there is another ticket for the 'what are you trying to do' box

src/pages/ResearcherProfile.js
connorlbark: yep - this is in the case that the user is a signing official and has already been assigned an institution. in this case, a user can never change their institution.

src/libs/ajax.js
```diff
@@ -860,8 +860,27 @@ export const User = {
...
   },
 
+  updateSelf: async (user) => {
+    const url = `${await Config.getApiUrl()}/api/user`;
+    // We should not be updating the user's create date, associated institution, or library cards
+    let filteredUser = flow(
+      cloneDeep,
+      unset('updatedUser.createDate'),
+      unset('updatedUser.institution'),
+      unset('updatedUser.libraryCards')
```
rushtong: I think we're mixing concepts here. The object passed in shouldn't actually be a full `user` object, but instead a custom object with a limited set of fields as defined in the server side API.

src/pages/ResearcherProfile.js
```diff
@@ -272,74 +105,60 @@ export default function ResearcherProfile(props) {
...
-    setStateNames(stateNames);
-  };
+    const payload = currentUserUpdate;
 
-  const showIncompleteFields = () => {
-    const listIncompleteFields = incompleteFields.map((field) =>
-      li({ key: field }, [field])
-    );
-    return ul({}, [listIncompleteFields]);
+    let updatedUser = await User.updateSelf(payload);
```
rushtong, line 118: If you go up to line 105, the payload here is derived from `Storage.getCurrentUser()` so it has lots of extra information that we don't really want to be sending as a payload in this API call. I would suggest that we build a generic payload object and populate it with the fields we care about.

connorlbark:
note: just found a bug in the backend code for getting signing officials which breaks the frontend, investigating and will see if i can fix
edit: has been fixed in the linked backend pull request

src/pages/ResearcherProfile.js
connorlbark: great point! now sends a custom payload object.

src/pages/ResearcherProfile.js
lu-c: OH, ok, yup - That's... i am indeed a signing official. 😂 Ok! Problem solved! XD

lu-c (approved):
Looks good! :)

shaemarks:
Not sure about the intended behavior, but I was able to delete my profileName and save it as empty, and then when I tried to submit a DAR, part 1 of the application wouldn't load

shaemarks:
This is probably better suited for a different ticket, but I could only see profile name changes appear in the topleft corner of the header if I sign out and sign back in

rushtong:
> This is probably better suited for a different ticket, but I could only see profile name changes appear in the topleft corner of the header if I sign out and sign back in

I agree, we could do a better job of refreshing the user saved in storage. There are other scenarios where this happens as well, such as role additions/removals

rushtong:
> Not sure about the intended behavior, but I was able to delete my profileName and save it as empty, and then when I tried to submit a DAR, part 1 of the application wouldn't load

@connorlbark Please do add a validation check for this - we should not be setting the user name to "".

src/components/SearchSelectOrText.js
```diff
@@ -0,0 +1,114 @@
...
+        className: 'dropdown-menu select-dropdown-menu'
+      }, [
+        input({
+          type: 'text',
+          placeholder: searchPlaceholder ? searchPlaceholder : 'Search...',
+          className: 'search-bar',
+          onChange:() => handleSearch(searchTerms),
+          onKeyDown: (e) => {
+            if (e.key == 'Enter') {
+              selectManual(searchTerms.current.value);
```
shaemarks, line 101: If possible, it would be nice if manually typed values could also be set onBlur since my intuition as a user is to type the value and then click off of the textbox

src/components/SearchSelectOrText.js
```diff
@@ -0,0 +1,114 @@
...
+  const [filteredList, setFilteredList] = useState(props.options);
+
+  const searchTerms = useRef('');
+
+  useEffect(() => {
+    if (props.value) {
+      setCurrentSelection(props.value);
+      const item = props.options.filter(i => i.key === props.value);
+      if (item && item.length) {
+        setCurrentDisplay(item[0].displayText);
```
shaemarks: optional nit: alternative could be using `find` function from lodash/fp to get the first element that meets the condition, or undefined if none exist.
Ex: 
```
import {find} from 'lodash/fp';



const item = find(i => i.key === props.value)(props.options);
if (item) {
    setCurrentDisplay(item.displayText);
}
```

src/pages/ResearcherProfile.js
```diff
@@ -1,75 +1,40 @@
...
-    piName: '',
-    piEmail: '',
-    piERACommonsID: '',
-    pubmedID: '',
-    scientificURL: '',
+    suggestedInstitution: undefined,
+    selectedSigningOfficialId: undefined,
+    suggestedSigningOfficial: undefined,
+    eraCommonsId: undefined,
     completed: undefined
```
shaemarks: I don't see completed used anywhere on this page anymore but here

connorlbark:
added validation to make sure the username is at least four characters (that seems reasonable to me, let me know if there is a preference for another length or just ensuring it's not empty)

shaemarks (commented):
I left a couple of small suggestions, but looks and works great!

connorlbark:
added:
- validation to the signing official (makes sure it's a valid email if you're suggesting)
- updates manual on blur (so if you click off) or if you hit enter

src/components/SearchSelectOrText.js
connorlbark: good tip thank you!

src/components/SearchSelectOrText.js
connorlbark: added!

shaemarks (approved):

PR #1674, opened 2022-07-13
[DUOS-1897][risk=no] Added institutions list to edit admin user page

lu-c:
Before:
Screen Shot 2022-07-13 at 10 27 32 AM

After:
Screen Shot 2022-07-13 at 10 27 17 AM

ONE THING in the ticket i did not address:
- Remove Academic/Business email input field. We're not modifying that field any longer
^ I did not see this on the admin-edit-user page.

----

- Verify all tests go green


src/pages/ResearcherReview.js
```diff
@@ -70,13 +69,9 @@ export const ResearcherReview = hh(class ResearcherReview extends Component {
...
           form({ name: 'researcherForm', noValidate: true }, [
             div({ className: 'row form-group margin-top-20' }, [
-              div({ className: 'col-lg-6 col-md-6 col-sm-6 col-xs-12' }, [
+              div({ className: 'col-lg-12 col-md-12 col-sm-6 col-xs-12' }, [
                 label({ className: 'control-label' }, ['Full Name']),
                 div({ id: 'lbl_profileName', className: 'control-data', name: 'profileName', readOnly: true }, [user.displayName])
-              ]),
-              div({ className: 'col-lg-6 col-md-6 col-sm-6 col-xs-12' }, [
-                label({ className: 'control-label' }, ['Academic/Business Email Address']),
-                div({ id: 'lbl_profileAcademicEmail', className: 'control-data', name: 'profileAcademicEmail', readOnly: true}, [formData.academicEmail]),
```
lu-c, line 79: Should I revert this change? This removes the academic email from display only

src/pages/ResearcherProfile.js
```diff
@@ -386,22 +386,6 @@ export default function ResearcherProfile(props) {
                 }),
               ]),
 
-              div({ className: 'col-xs-12' }, [
```
lu-c, line 389: Deleted this input block for academic/business email because the ticket asked us to... but! The input form is *always* disabled, so I don't see a harm to leaving it there. Shoudl this change be reverted?

src/pages/AdminEditUser.js
```diff
@@ -60,6 +69,7 @@ export const AdminEditUser = hh(class AdminEditUser extends Component {
       userId: userId,
       displayName: this.state.displayName,
       emailPreference: this.state.emailPreference,
+      institutionId: this.state.institutionId
     };
     const payload = {updatedUser: user};
```
shaemarks: Currently when I update a user's institution, display name, etc, and hit save, changes are reverted once I refresh. I think the issue is wrapping the updated properties in `{updatedUser: user}` since the new endpoint expects json that can be converted directly to `UserUpdateFields`

src/pages/ResearcherProfile.js
shaemarks: Probably best to get a second opinion, but my impression based on [DUOS-1818](https://broadworkbench.atlassian.net/browse/DUOS-1818) is that the intention is to remove all usages of academic email in the ui and then get rid of it altogether in consent

shaemarks (dismissed):
Dropdown and search select for institutions looks good! Only issue is that I'm not seeing the changes to a user persist after refreshing the page / leaving and returning

src/pages/ResearcherProfile.js
shaemarks: Probably best to get a second opinion, but my impression based on [DUOS-1818](https://broadworkbench.atlassian.net/browse/DUOS-1818) is that the intention is to remove all usages of academic email in the ui and then get rid of it altogether in consent

src/pages/AdminEditUser.js
rushtong: @shaemarks is correct here - the payload format is different for `/api/user/${userId}` than it is for `/api/dacuser/${userId}`. See [PUT /api/user/${userId}](https://consent.dsde-dev.broadinstitute.org/#/User/put_api_user__userId_) for what the payload should look like. Only values that are populated in the payload are updated, so anything not there will be ignored.

rushtong: For a little bit of history, the payload for `/api/dacuser/${userId}` used to take all sorts of arguments. We used to be able to swap one user for another for certain roles which was a huge headache. We've simplified it quite a bit on both sides, but the `{updatedUser: user}` structure is a holdover from that time.

lu-c: Gotcha, I have updated this to work!

shaemarks (approved):
Thanks for the changes! Institution SearchSelect works great, and updates are now being saved

quazi-h (approved):
As an admin when I go to the edit user page, I see the drop down selector with the list of all the available institutions. When editing a user that is already assigned to an institution, I see that institution pre-populated as the selection. Changes look good to me!

PR #1675, opened 2022-07-13
DUOS-1887[risk=med] Enable nav refactor for Prod with updated env conditionals

quazi-h:
[DUOS-1887](https://broadworkbench.atlassian.net/browse/DUOS-1887)
- Updated the app to exclusively use the new DUOS header.
- Updated utils to remove conditional logic that considers environment for navigation.
- Removed conditional logic that considers environment for DataCustodianTable visibility in the Signing Official Console.
- Fixed a no-unused-vars from the MDV changes.

----

- Verify all tests go green

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -259,9 +259,8 @@ export const Navigation = {
         :user.isChairPerson ? '/new_chair_console'
           : user.isMember ? '/new_member_console'
             : user.isResearcher ? '/dataset_catalog'
-              : user.isDataOwner ? '/data_owner_console'
-                : user.isAlumni ? '/summary_votes'
-                  : '/';
+              : user.isAlumni ? '/summary_votes'
+                : '/';
     history.push(page);
   },
   console: async (user, history) => {
@@ -270,9 +269,8 @@ export const Navigation = {
         : user.isChairPerson ? '/new_chair_console'
           : user.isMember ? '/new_member_console'
             : user.isResearcher ? '/new_researcher_console'
-              : user.isDataOwner ? '/data_owner_console'
-                : user.isAlumni ? '/summary_votes'
-                  : '/';
+              : user.isAlumni ? '/summary_votes'
+                : '/';
     history.push(page);
   }
 };
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -6,7 +6,6 @@ import { DAR, DataSet } from './ajax';
 import {Theme, Styles } from './theme';
 import { each, flatMap, flatten, flow, forEach, get, getOr, indexOf, uniq, values, find, first, map, isEmpty, filter, cloneDeep, isNil, toLower, includes, sortedUniq, every, pick, capitalize } from 'lodash/fp';
 import {User} from './ajax';
-import {Config} from './config';
 
 export const UserProperties = {
   ACADEMIC_EMAIL: 'academicEmail',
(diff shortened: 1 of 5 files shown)


quazi-h:
Open Questions:
- When looking for other instances of env conditionals, I found background sign in is disabled for prod staging. That should remain that way, correct?
- Should I remove the DuosHeaderOld code?
- How do I get to the Data Custodian table? I wasn't able to access the DataCustodianTable - need reviewers to take a closer look at that to make sure the logic I updated is working correctly.

src/pages/AdminConsole.js
```diff
@@ -18,7 +18,6 @@ class AdminConsole extends Component {
       showModal: false,
       showAddUserModal: false,
       dulUnreviewedCases: 0,
-      darUnreviewedCases: 0,
```
quazi-h, line 21: I removed darUnReviewedCases from this code because it was presenting a no-unused-vars when compiling.
It was necessary for the pre MDV code. Is it okay for me to include these changes here?

rushtong: Yes, that's fine, good catch.

src/libs/utils.js
```diff
@@ -255,51 +254,25 @@ export const setUserRoleStatuses = (user, Storage) => {
...
-                : user.isDataOwner ? '/data_owner_console'
-                  : user.isAlumni ? '/summary_votes'
-                    : '/';
-    }
+    let page =
+      user.isAdmin ? '/admin_manage_dar_collections'
+        :user.isChairPerson ? '/new_chair_console'
+          : user.isMember ? '/new_member_console'
+            : user.isResearcher ? '/dataset_catalog'
+              : user.isDataOwner ? '/data_owner_console'
```
shaemarks:
```suggestion
```

nit: `user.isDataOwner` and the `/data_owner_console` route have both been removed, so this line isn't needed 

Also, I think we can remove the "new" prefix from these paths (ex: `/new_chair_console` -> `/chair_console`) now that both Routes exist and point to the new consoles. `DuosHeader.js` also has 
 ```
 this.setState({dacChairPath: '/new_chair_console'});
    this.setState({dacMemberPath: '/new_member_console'});
    this.setState({researcherPath: '/new_researcher_console'});
```
, not sure about anywhere else

File: src/libs/utils.js [hunk 3]
```diff
@@ -255,51 +254,25 @@ export const setUserRoleStatuses = (user, Storage) => {
...
-                  : user.isDataOwner ? '/data_owner_console'
-                    : user.isAlumni ? '/summary_votes'
-                      : '/';
-    }
+    let page =
+      user.isAdmin ? '/admin_manage_dar_collections'
+        : user.isChairPerson ? '/new_chair_console'
+          : user.isMember ? '/new_member_console'
+            : user.isResearcher ? '/new_researcher_console'
+              : user.isDataOwner ? '/data_owner_console'
```
shaemarks: Same as above

shaemarks (commented):
New navbar was present and working well for all environments. Some small comments about updating pathnames, but looks good!

shaemarks:
> * How do I get to the Data Custodian table? I wasn't able to access the DataCustodianTable - need reviewers to take a closer look at that to make sure the logic I updated is working correctly.

You shouldn't have to worry about getting the DataCustodianTable visible - we removed tabs to it on the signing official console since it wasn't displaying any information, just a duplicated table of researchers. We've been keeping around the component though with the intention of repurposing it later for data submitters, I believe. Maybe it would be useful to leave a comment so that if someone looks at that code in the future they'll have this context?

src/libs/utils.js
rushtong: FYI, @shaemarks, @quazi-broad ... the `new` changes could happen in this PR, but we also have a story for specifically this: https://broadworkbench.atlassian.net/browse/DUOS-1823

rushtong:
> When looking for other instances of env conditionals, I found background sign in is disabled for prod staging. That should remain that way, correct?

@quazi-broad Correct, we only want background sign-in enabled for certain testing envs

rushtong:
> Should I remove the DuosHeaderOld code?

@quazi-broad - That's not necessary for this ticket. I think it's a good idea to add it to https://broadworkbench.atlassian.net/browse/DUOS-1823 though.

src/libs/utils.js
rushtong: +1 to this suggestion

rushtong (approved):
Looks good to me 👍🏽 Pending Shae's feedback on removing the data owner console links, this looks good to go.

src/libs/utils.js
quazi-h: Removed the data_owner_console, leaving the renaming "new_" routes for [1823](https://broadworkbench.atlassian.net/browse/DUOS-1823)

shaemarks (approved):

PR #1677, opened 2022-07-14
[DUOS-1764][risk=no] Loop over all dataset ids instead of assuming the first

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1764

See also: https://github.com/DataBiosphere/consent/pull/1648

This PR refactors the outdated assumption that a submitted DAR will always have a single dataset id. In all cases where we looked for the first element of the `datasetIds` array in `data`, first look in the `dar` itself. This PR tries to maintain the same logical conditions but for the whole list instead of just the head of the array.

----

- Verify all tests go green


shaemarks:
When I submited a collection with two datasets, inside of the collection there were two dars. For each of these dars, `datasetIds` was an array with 1 dataset id, but `data.datasetIds` and `data.datasets` inside of each dar contained both datasets. Is this what is expected?

dar datasetIds (2)

dar data datasetIds

rushtong:
> When I submited a collection with two datasets, inside of the collection there were two dars. For each of these dars, `datasetIds` was an array with 1 dataset id, but `data.datasetIds` and `data.datasets` inside of each dar contained both datasets. Is this what is expected?

In short, yes, I think this is expected.

Longer answer ... when we `submit` a DAR (on the server side), we still split that DAR into one DAR per dataset requested. Each individual DAR should then have a single dataset id associated with it at the DAR reference id level. At the `data` level of the dar, I don't think we do anything specific so it keeps in place whatever was initially created when the DAR was still a draft. However, this all happens on the server side so this PR can't have any effect on the final result of whatever we do in consent.

In a future PR, when we stop splitting the DAR up on submission, then the `datasetIds` on both the DAR and the data should be in alignment.

shaemarks (approved):
Looks good to me!

JVThomas (approved):
Changes looks good to me

PR #1678, opened 2022-07-15
[DUOS-1893][risk=no] Added submit help ticket for feature request

lu-c:
Jira: https://broadworkbench.atlassian.net/browse/DUOS-1893

- Added "submit help ticket" at time of profile update

Screen Shot 2022-07-14 at 9 13 57 PM

----

- Verify all tests go green


src/pages/ResearcherProfile.js
```diff
@@ -385,400 +304,185 @@ export default function ResearcherProfile(props) {
...
-                    label({ id: 'lbl_profileCity', className: 'control-label' }, ['City*']),
-                    input({
-                      id: 'profileCity',
-                      name: 'city',
-                      type: 'text',
-                      className: 'form-control',
-                      defaultValue: profile.city,
-                      onBlur: handleChange
-                    })
+              div({ className: 'row no-margin'}, [
```
lu-c: https://getbootstrap.com/docs/4.0/layout/grid/

These changes here are a lot of spacing changes only - There was a css issue that was being caused by all of these not being in the proper bootstrap format, which should be a class with "row" that contains any classes with "col-[size]-[num]". Without the "row" container, all fo the objects shared the same container, and you couldnt set the background color of one full-width col without affecting multiple at the same time.

src/pages/ResearcherProfile.js
```diff
@@ -108,6 +149,51 @@ export default function ResearcherProfile(props) {
     return updatedUser;
   };
 
+  const sendActions = async () => {
```
connorlbark: nit: rename to `sendSupportTicket`; the noun `action` here is unclear to me. maybe `desiredAction`? in addition, this function seems to do two things: check if there are any actions, then submit them. another cleanup thing might be to do:
```js
if (userHasDesiredActions()) {
   await sendSupportTicket();
}
```
just to make the function a bit more readable

connorlbark (commented):
not sure exactly how to test that the actual support ticket went through, but it seems to function exactly as I would expect, and, upon inspecting the objects send to the support ticket, appears to send the correct information. just the one naming nit and it looks good to me, pending someone checking that support tickets are indeed successfully created.

src/pages/ResearcherProfile.js
lu-c: i kind of split it out into two diff methods, lemme know if this works better!

I changed the variable name from actions -> supportRequests... I think i made it actions in the first place because I wasn't sure if they were all supposed to be support ticket requests, or if they would trigger other actions in the app!

JVThomas (approved):
UI changes line up with the mocks and the response is returning a 200.

src/pages/ResearcherProfile.js
connorlbark: thank you!!

connorlbark (approved):
LGTM!

PR #1694, opened 2022-07-20
[DUOS-1949][risk=no] Select all checkbox only selects visible datasets

connorlbark:
## Addresses 

https://broadworkbench.atlassian.net/browse/DUOS-1949

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetCatalog.js
+++ b/src/pages/DatasetCatalog.js
@@ -403,20 +403,15 @@ export default function DatasetCatalog(props) {
             }
           },
         }, [
-
-          div({style: {position: 'absolute'}}, [
-            span({
-              style: {
-                marginTop: '4px',
-                fontSize: '.8em',
-              }
-            }, ['Select Visible']),
-          ]),
-
-          form({ className: 'pos-relative' }, [
+          form({
+            className: 'pos-relative',
+            style: {
+              marginBottom: '10px',
+            }
+          }, [
             div({ className: 'checkbox check-all' }, [
               input({ checked: allOnPageSelected(), type: 'checkbox', 'select-all': 'true', className: 'checkbox-inline', id: 'chk_selectAll', onChange: selectAllOnPage }),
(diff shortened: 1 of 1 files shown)


rushtong (changes requested):
Functionally, this works great! The ticket also describes labeling the select-all checkbox. Can we add that to this PR? Thank you!

src/pages/DatasetCatalog.js
```diff
@@ -251,24 +250,35 @@ export default function DatasetCatalog(props) {
     return true;
   };
 
+  const filteredAndPaginatedDatasetList = () => {
+    return datasetList.filter(searchTable(searchDulText)).slice((currentPage - 1) * pageSize, currentPage * pageSize);
+  };
+
+  const allOnPageSelected = () => {
+    return filteredAndPaginatedDatasetList().every((row) => {
+      return row.checked || (!row.active);
```
shaemarks, line 262: The Select Visible checkbox appears checked when the datasets are still loading since this condition is technically true. You can also make sure that the dataset list is not empty here to fix it

src/pages/DatasetCatalog.js
```diff
@@ -251,24 +250,35 @@ export default function DatasetCatalog(props) {
...
+    });
+  };
+
   const selectAll = (e) => {
```
shaemarks: nit: change name to reflect that only the visible datasets are being selected (ex: selectAllOnPage)

shaemarks (changes requested):
I left a couple of minor comments, but otherwise looks good!

connorlbark:
> Functionally, this works great! The ticket also describes labeling the select-all checkbox. Can we add that to this PR? Thank you!

I'm not sure I like the look of the table with a label, personally:

image

What do you think?

rushtong (approved):
Thanks for the changes 👍🏽

shaemarks (approved):

PR #1698, opened 2022-07-21
[DUOS-1943] redirect to edit user page after saving user edits

lu-c:
https://broadworkbench.atlassian.net/browse/DUOS-1943

----

- Verify all tests go green


shaemarks (commented):
Looks good to me! I would suggest adding a try catch and using Notifications.showError if something goes wrong since currently it looks like you'll be redirected without knowing that the change failed to go through

connorlbark (approved):
looks good pending the try/catch shae mentioned!

lu-c:
> Looks good to me! I would suggest adding a try catch and using Notifications.showError if something goes wrong since currently it looks like you'll be redirected without knowing that the change failed to go through

Updated the code so if it fails, you don't redirect at all

shaemarks (approved):

PR #1699, opened 2022-07-21
[DUOS-1903][risk=low] Modify Chair Manage DAR Collections to use Summary Endpoint

lu-c:
- Modified chair manage DAR collections to use summary endpoint
- This will break the other pages that use the DAR collection table object

... sorry :(

----

- Verify all tests go green


src/components/dar_collection_table/ChairActions.js
```diff
@@ -101,17 +89,17 @@ export default function ChairActions(props) {
...
+  const [voteEnabled, setVoteEnabled] = useState(props.voteEnabled);
   const [voteLabel, setVoteLabel] = useState('Vote');
 
   //if there's something wrong with the collection it's best to fail gracefully
   //use this function to hide buttons on processing err
   const updateStateOnFail = () => {
     setOpenEnabled(false);
     setCancelEnabled(false);
-    setVoteLabel(false);
+    setVoteEnabled(false);
```
connorlbark: unsure: do we also need to keep `setVoteLabel(false)` here?

src/components/dar_collection_table/DarCollectionTableCellData.js
```diff
@@ -133,15 +133,20 @@ export function statusCellData({status = '- -', darCollectionId, label = 'status
...
       actionComponent = h(AdminActions, {collection, showConfirmationModal});
       break;
     case consoleTypes.CHAIR:
-      actionComponent = h(ChairActions, {collection, showConfirmationModal, goToVote, relevantDatasets});
+      actionComponent = h(ChairActions, {
+        collection, showConfirmationModal, goToVote, relevantDatasets,
+        openEnabled: actions.includes('Open'),
+        cancelEnabled: actions.includes('Cancel'),
+        voteEnabled: actions.includes('Vote')
+      });
```
connorlbark: I personally don't have a strong opinion, but since we're being told what actions are acceptable from the backend, would it make sense to consolidate `AdminActions`, `ChairActions`, etc., into one `Actions` class which displays the actions the backend says are available? That way the frontend is more separate from the backend.

connorlbark (commented):
Looks great and functioned well when I tested it! If we decide to keep the separate actions classes, then I think it's ready to have stuff merged in.

src/components/dar_collection_table/DarCollectionTableCellData.js
lu-c: I considered doing that initially, but the ChairActions object was really complex, with further logic that controls what the state of the button should be (based on failure of request). I wasn't sure if the other Action components would also have a lot of specific logic to their state, but assumed they had to since they were split out in this way in the first place 🤔

src/components/dar_collection_table/ChairActions.js
lu-c: good catch, adding this back!

lu-c:
> Looks great and functioned well when I tested it! If we decide to keep the separate actions classes, then I think it's ready to have stuff merged in.

What would you think about adding the combined actions as a separate function? I think it needs more investigation than the single ticket for conversion of the summary endpoint requires, and is a slight bit of scope creep from the original ask

connorlbark:
> > Looks great and functioned well when I tested it! If we decide to keep the separate actions classes, then I think it's ready to have stuff merged in.
> 
> What would you think about adding the combined actions as a separate function? I think it needs more investigation than the single ticket for conversion of the summary endpoint requires, and is a slight bit of scope creep from the original ask

I definitely agree this might be getting into some scope creep! @rushtong what do you think?

rushtong:
> > > Looks great and functioned well when I tested it! If we decide to keep the separate actions classes, then I think it's ready to have stuff merged in.
> > 
> > 
> > What would you think about adding the combined actions as a separate function? I think it needs more investigation than the single ticket for conversion of the summary endpoint requires, and is a slight bit of scope creep from the original ask
> 
> I definitely agree this might be getting into some scope creep! @rushtong what do you think?

Agreed on the scope issue. I suggest we address refactoring the actions in a separate PR/ticket, depending on what seems the most practical.

JVThomas (dismissed):
Tested locally, consoles are loading the summaries quickly. Actions go through successfully and re-renders the table with the updated summary model. Search bar looks to be working as intended.

src/components/dar_drafts_table/DarDraftTable.js
```diff
@@ -0,0 +1,254 @@
...
+        partialDarCodeCell({partialDarCode, id}),
+        projectTitleCell({projectTitle, id}),
+        updateDateCell({updateDate: updateDate || createDate, id}),
+        actionCellData({targetDraft: draft, showConfirmationModal, history})
+      ];
+    });
+  }
+};
+
+export default function DarDraftTable(props) {
```
shaemarks: I think this file was deleted while we worked on this feature branch and adding it back in here is unintentional

src/pages/AdminManageDarCollections.js
```diff
@@ -26,7 +26,7 @@ export default function AdminManageDarCollections() {
   useEffect(() => {
     const init = async() => {
       try {
-        const collectionsResp = await Collections.getCollectionsByRoleName('admin');
+        const collectionsResp = await Collections.getCollectionSummariesByRoleName('admin');
```
shaemarks:
```suggestion
        const collectionsResp = await Collections.getCollectionSummariesByRoleName(USER_ROLES.admin);
```

nit: use existing user role constant

src/pages/ChairConsole.js
```diff
@@ -29,7 +29,7 @@ export default function ChairConsole(props) {
     const init = async() => {
       try {
         const [collections, datasets] = await Promise.all([
-          Collections.getCollectionsByRoleName('chairperson'),
+          Collections.getCollectionSummariesByRoleName('chairperson'),
```
shaemarks:
```suggestion
          Collections.getCollectionSummariesByRoleName(USER_ROLES.chairperson),
```

src/pages/MemberConsole.js
```diff
@@ -27,7 +27,7 @@ export default function MemberConsole(props) {
     const init = async () => {
       try {
         const [collections, datasets] = await Promise.all([
-          Collections.getCollectionsByRoleName('member'),
+          Collections.getCollectionSummariesByRoleName('member'),
```
shaemarks:
```suggestion
          Collections.getCollectionSummariesByRoleName(USER_ROLES.member),
```

src/pages/ResearcherConsole.js
```diff
@@ -60,40 +30,9 @@ export default function ResearcherConsole() {
...
-        collectionArray,
-        errorMsg,
-        'Failed to fetch Data Access Request Drafts'
-      );
-      if(!isEmpty(errorMsg)) {
-        Notifications.showError({text: errorMsg.join('\n')});
-      }
-      setResearcherCollections(collectionArray);
-      setFilteredList(collectionArray);
+      const collections = await Collections.getCollectionSummariesByRoleName('Researcher');
```
shaemarks:
```suggestion
      const collections = await Collections.getCollectionSummariesByRoleName(USER_ROLES.researcher);
```

src/components/dar_collection_table/DarCollectionTableCellData.js
```diff
@@ -107,9 +104,9 @@ export function institutionCellData({institution = '- -', darCollectionId, label
   };
 }
 
-export function datasetCountCellData({datasets = [], darCollectionId, label = 'datasets'}) {
+export function datasetCountCellData({datasetIds = [], darCollectionId, label = 'datasets'}) {
   return {
-    data: datasets.length > 0 ? datasets.length : '- -',
+    data: datasetIds.length > 0 ? datasetIds.length : '- -',
```
shaemarks: optional nit: dar collection summaries have a `datasetCount` field that could be used instead

shaemarks (commented):
Dar Collection Table appearance and functionality looks great! I think `DarDraftTable` wasn't meant to be added here, and I left some minor suggestions for using the USER_ROLES constants more consistently, but overall 👍

shaemarks (approved):
Looks good to me! Tested the DarCollectionTables on all consoles and actions.

JVThomas (approved):
Tested locally, feature branch looks to be in good working condition

PR #1704, opened 2022-07-22
[DUOS-1941][risk=no] Track # selected datasets, allow filtering to selected

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1941

I also added the ability to filter according to the datasets which have been selected: see discussions in the #duos-ux slack channel for context.

----

- Verify all tests go green


rushtong (approved):
Love the new display option 👍🏽

JVThomas (approved):
Changes look good. Select All works as you would expect and the Show Selected Datasets shows all visible datasets. The two buttons also work well with each other. I can click Show Selected and unselect "Select All" and see both the visible table and the selected dataset count update accordingly.

PR #1705, opened 2022-07-22
[DUOS-1823][risk=no] Remove Unused Console/Voting Pages

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1823

I looked through the calls in both the old and new pages and did not find any endpoints which became newly unused, so there is no corresponding backend PR.

----

- Verify all tests go green


JVThomas:
There are two other pages that are no longer used in UI. In `Routes.js`, there are routes assigned to two components, `ReviewResults.js` and `AccessReview.js`. Both of these pages refer to (respectfully) the summary and voting pages for individual DARs. Since we've moved over to Collection based operations these pages are no longer accessible and thus no longer in use.

Take a look at those pages and the components used in them. Odds are a lot of those can be removed, but check to be sure.

rushtong (approved):
This looks good, thank you! There are definitely newly unused pages, but I'll file separate tickets for those individually

JVThomas (approved):
Since deletion/route removals will be handled in a different ticket this PR is good to go!

PR #1708, opened 2022-07-26
[DUOS-1905][risk=no] Update Researcher Console to use Summary Endpoint

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1905

In dev, the summary endpoint does not return reference ids. We need them so that draft dar collections can be resumed. The backend change to add that is here: https://github.com/DataBiosphere/consent/pull/1670

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #1710, opened 2022-07-27
[DUOS-1904] Added uniform Actions component

lu-c:
Not sure if deleted anything important - But I basically just removed anything complicated whatsoever, and assumed that those changes were made on the back end when we make the request for the new summary

----

- Verify all tests go green


src/components/dar_collection_table/Actions.js
```diff
@@ -0,0 +1,118 @@
...
+        padding: '10px 5px',
+        justifyContent: 'flex-start',
+        alignItems: 'center',
+      },
+    },
+    [
+      h(SimpleButton, openButtonAttributes),
+      h(SimpleButton, voteButtonAttributes),
+      h(SimpleButton, updateButtonAttributes),
+      h(TableIconButton, cancelButtonAttributes),
```
connorlbark, line 115: for researchers, there is also: `Resume` `Review` and `Revise`: there were some backend changes needed to implement the `Resume` button. perhaps https://github.com/DataBiosphere/duos-ui/pull/1708 could be pulled in and then you could take in the changes added there to add the action here.

lu-c: Yeah at some point we will need to add those other options in. I was just working incrementally within this branch, since only chair and member are the only ones using the new summary.

I would have to look into the researcher actions component to understand it tho, since it's different! And I'd want to make sure I'm carrying any working bits here. For right now, I'd prefer to keep this branch targeted for just the chair actions then, and add in the researcher stuff after this has been approved! I'm afraid that converting the chair -> generic actions meant I deleted something important/necessary, and this is like a sanity check to make sure it's still going okay

shaemarks (approved):
Changes look good to me and clean up a lot of code! Deleted code in ChairActions, AdminActions, etc just looks like button attributes and logic to determine actions, so it shouldn't be a problem getting rid of it

JVThomas (approved):
Changes look good! With the actions being processed on the back-end simplifying the actions makes sense. Buttons are showing up as expected on Admin, Chair, and Member consoles.

PR #1711, opened 2022-07-27
DUOS-1990[risk=no] Updated bucket logic to account for upcoming changes to DAR-Election relation

JVThomas:
Addresses [DUOS-1990](https://broadworkbench.atlassian.net/browse/DUOS-1990)

PR updates the bucket logic to account for upcoming changes to the DAR-Election relation on consent. Instead of assuming election organization by dar, the logic will instead organize elections by the dataset it is associated with and perform downstream data manipulation from there.

- Verify all tests go green

### DIFF
--- a/src/utils/DarCollectionUtils.js
+++ b/src/utils/DarCollectionUtils.js
@@ -59,24 +59,26 @@ export const generatePreProcessedBucketData = async ({dars, datasets}) => {
       buckets[dataUseLabel] = {
         dars: [],
         datasets: [],
-        dataUse: restrictions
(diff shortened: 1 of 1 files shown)


rushtong (approved):
👍🏽 This works with my DUOS-1764 branch. I will admit that I can't follow a lot of the logic here, but it does break out the elections & votes into the appropriate buckets when viewed at the component level.

shaemarks (approved):
Looks good to me! Bucketing works correctly with both multi-dar collections and single-dar collections created with the DUOS-1764 branch

PR #1719, opened 2022-07-29
[DUOS-1885][risk=no] Update Manage User page styling

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1885

----

- Verify all tests go green

### DIFF
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -8,12 +8,10 @@ export default function SearchBar(props) {
   const { handleSearchChange } = props;
   const searchTerms = useRef('');
 
-  const width = props.width ? props.width : '100%';
-  const margin = props.margin ? props.margin : '';
   const button = props.button ? props.button : div();
 
-  const style = props.style ? props.style : {
-    width: width,
+  const style = {
+    width: '100%',
     border: '1px solid #cecece',
     backgroundColor: '#f3f6f7',
     borderRadius: '5px',
@@ -21,7 +19,7 @@ export default function SearchBar(props) {
     paddingLeft: '2%',
     fontFamily: 'Montserrat',
     fontSize: '1.5rem',
-    margin: margin,
+    ...props.style
   };
 
   return div({className: 'right-header-section', style: Styles.RIGHT_HEADER_SECTION}, [
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -7,15 +7,21 @@ import { defaultTo } from 'lodash/fp';
 export default function SearchBar(props) {
   const { handleSearchChange } = props;
   const searchTerms = useRef('');
+
+  const width = props.width ? props.width : '100%';
+  const margin = props.margin ? props.margin : '';
+  const button = props.button ? props.button : div();
+
   const style = props.style ? props.style : {
-    width: '100%',
+    width: width,
     border: '1px solid #cecece',
     backgroundColor: '#f3f6f7',
     borderRadius: '5px',
     height: '4rem',
     paddingLeft: '2%',
     fontFamily: 'Montserrat',
-    fontSize: '1.5rem'
+    fontSize: '1.5rem',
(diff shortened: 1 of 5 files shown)


src/components/SearchBar.js
```diff
@@ -7,15 +7,21 @@ import { defaultTo } from 'lodash/fp';
 export default function SearchBar(props) {
   const { handleSearchChange } = props;
   const searchTerms = useRef('');
+
+  const width = props.width ? props.width : '100%';
```
lu-c: This seems kind of clunky if we wanted to update multiple style features in the future - Can I recommend we change this overall code to be more like:

```suggestion
  const style = {
    width: '100%',
    border: '1px solid #cecece',
    backgroundColor: '#f3f6f7',
    borderRadius: '5px',
    height: '4rem',
    paddingLeft: '2%',
    fontFamily: 'Montserrat',
    fontSize: '1.5rem',
    ...props.style
 }
```

Then in the section in AdminManageUser where SearchBar is instantiated, we update it to

```suggestion
  h(SearchBar, {
    ...
    style: {
      width: '60%',
      margin: '0 3% 0 0'
    }
  })
```

This will allow us to override all of the styles in the search bar without having to extract each one directly first

lu-c (approved):
I think a lot of the changes are moving code around? Hard to see the specific changes made for the ticket, but I think it looks okay! Ran branch locally, styling looks good 👍

src/components/manage_users_table/ManageUsersTable.js
```diff
@@ -0,0 +1,190 @@
...
+    color: '#7B7B7B',
+    fontFamily: 'Montserrat',
+    fontSize: '1.2rem',
+    fontWeight: 'bold',
+    letterSpacing: '0.2px',
+    textTransform: 'uppercase',
+    backgroundColor: 'B8CDD3',
+    border: 'none'
+  }),
+  cellWidth: {
```
JVThomas, line 36: **Optional** 

So the skeleton loader looks a bit awkward because the column widths are 33.33%. Since the loader just uses the same values for the graphic the bars overlap with each other. If you give the the cellWidth some space in between (and update the columnStyle appropriately, the loaders will look a lot cleaner.

One example: Adjusting the cellWidths so that the sum is equal to `94%` gives you `6%` space for margins, or `1%` on each side of a column. From there you can just update the `cellStyles` on `columnHeaderConfig` so that its something like `cellStyle: {width: <insert width here>, margin: '0% 1%'}`

src/libs/utils.js
```diff
@@ -576,7 +576,20 @@ export const getSearchFilterFunctions = () => {
         includes(lowerCaseTerm, toLower(phrase))
       )([partialDarCode, ...(projectTitle.split(' ')), (updateDate || createDate)]);
       return !isNil(matched) && (draft !== false || draft !== 'false');
-    })(targetList)
+    })(targetList),
+    users: (term, targetList) => {
+      const lowerCaseTerm = toLower(term);
+      const isMatch = (userField) => includes(lowerCaseTerm, toLower(userField));
+
+      return filter(user => {
```
JVThomas, line 584: You'll need to add the user's roles alongside displayName and email as part of the searchable terms (since they're visible on the table).

JVThomas (changes requested):
The table update is looking good! There are some adjustments to be made (one optional styling adjustment, one essential function adjustment) but other than that I see no issues with the PR.

src/components/SearchBar.js
connorlbark: this is a much smarter solution! thank you

src/components/manage_users_table/ManageUsersTable.js
connorlbark: good catch!

JVThomas (approved):
Thanks for the updates, the page looks great!

PR #1724, opened 2022-08-01
[DUOS-1987][risk=no] Redirect to profile page if profile incomplete

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1987

Should redirect when:
- User who has not completed profile accepts ToS
- User who has accepted ToS but not completed profile logs in.

----

- Verify all tests go green

### DIFF
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -6,7 +6,7 @@ import {Alert} from './Alert';
 import {ToS, User} from '../libs/ajax';
 import {Config} from '../libs/config';
 import {Storage} from '../libs/storage';
-import {Navigation, setUserRoleStatuses, researcherProfileComplete} from '../libs/utils';
+import {Navigation, setUserRoleStatuses} from '../libs/utils';
 import loadingIndicator from '../images/loading-indicator.svg';
 import {Spinner} from './Spinner';
 import ReactTooltip from 'react-tooltip';
@@ -42,12 +42,6 @@ export default function SignIn(props) {
     if (!isEmpty(userStatus) && !tosAccepted) {
       await Storage.setUserIsLogged(false);
       history.push('/tos_acceptance');
-      return;
-    }
-
-    const {isResearcher} = Storage.getCurrentUser();
-    if (isResearcher && !researcherProfileComplete(user)) {
-      history.push('/profile');
       return;
     }
     redirect(user);
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -42,9 +42,9 @@ export default function SignIn(props) {
     if (!isEmpty(userStatus) && !tosAccepted) {
       await Storage.setUserIsLogged(false);
       history.push('/tos_acceptance');
-      return;
+    } else {
+      redirect(user);
     }
-    redirect(user);
   };
 
   const onSuccess = async (response) => {
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -6,7 +6,7 @@ import {Alert} from './Alert';
 import {ToS, User} from '../libs/ajax';
 import {Config} from '../libs/config';
 import {Storage} from '../libs/storage';
-import {Navigation, setUserRoleStatuses} from '../libs/utils';
+import {Navigation, setUserRoleStatuses, researcherProfileComplete} from '../libs/utils';
 import loadingIndicator from '../images/loading-indicator.svg';
 import {Spinner} from './Spinner';
 import ReactTooltip from 'react-tooltip';
@@ -42,9 +42,15 @@ export default function SignIn(props) {
     if (!isEmpty(userStatus) && !tosAccepted) {
       await Storage.setUserIsLogged(false);
       history.push('/tos_acceptance');
-    } else {
-      redirect(user);
+      return;
     }
+
+    const {isResearcher} = Storage.getCurrentUser();
+    if (isResearcher && !researcherProfileComplete(user)) {
+      history.push('/profile');
+      return;
+    }
+    redirect(user);
   };
 
   const onSuccess = async (response) => {
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -772,12 +772,3 @@ export const completedResearcherInfoCheck = (properties) => {
   } = properties;
   return !isNil(institutionId);
 };
-
-export const researcherProfileComplete = (user) => {
-  const {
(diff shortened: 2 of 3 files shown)


src/libs/utils.js
```diff
@@ -772,3 +772,12 @@ export const completedResearcherInfoCheck = (properties) => {
...
   return !isNil(institutionId);
 };
+
+export const researcherProfileComplete = (user) => {
+  const {
+    institutionId,
+    selectedSigningOfficialId,
+  } = getPropertyValuesFromUser(user);
+
+  return !isNil(institutionId) && !isNil(selectedSigningOfficialId);
```
connorlbark: Is this how we want to check if a profile is complete? Right now, checks that there is a valid institution id and selected sigining official id; in the case that the user suggests a signing official / institution, their profile is still considered incomplete; this seems right to me, but wanted to make sure.

connorlbark: thought: is the expected behavior for a user which is a signing official / any other non-researcher to always skip this ?

JVThomas: > Is this how we want to check if a profile is complete? Right now, checks that there is a valid institution id and selected sigining official id; in the case that the user suggests a signing official / institution, their profile is still considered incomplete; this seems right to me, but wanted to make sure.

I talked with @solideoglori about this and we both decided not to worry about a "complete" profile and only have the redirect occur on users that are registering with DUOS for the first time. Depending on the role you may not care about SO or Institution, so nil checks on both won't be enough (For example, DAC roles don't care about SO or Institution, SO requires an institution but not SO, and a research needs both SO and institution). At the same time the system allows for multiple roles which just adds complexity.

From a product perspective the odds of someone signing out of the profile page without doing anything is relatively low. If a user does do that, the only features that would require checks are the DAR Form and the SO console. The DAR page already performs that check and provides the user with a message and link to the profile page. For the SO we only assign that role if the user has an institution so we're okay there as well.

**TLDR**: Don't worry about a "complete" profile, just perform the redirect for new users. If the user already exists, just have the normal redirects occur.

src/pages/TermsOfServiceAcceptance.js
```diff
@@ -23,6 +23,12 @@ export default function TermsOfServiceAcceptance(props) {
     const user = await Storage.getCurrentUser();
     await TosService.acceptTos();
     await Storage.setUserIsLogged(true);
+
+    if (user.isResearcher && !researcherProfileComplete(user)) {
+      history.push('/profile');
+      return;
+    }
```
rushtong: I don't think we need to check for `user.isResearcher` here - all users are always researchers (that's our default user role).

rushtong: I'm on the fence about the `researcherProfileComplete` check. I'm leaning to sending the user to the profile page whenever they accept ToS, regardless of completeness.

1. If a user has never selected an institution, both approaches work
2. If a user has previously filled in an institution, then rejected ToS, I think we want to send them to the profile page so they can see the on-boarding form again if they re-accept ToS. In this case, the current approach doesn't work.

src/components/SignIn.js
```diff
@@ -42,9 +42,15 @@ export default function SignIn(props) {
     if (!isEmpty(userStatus) && !tosAccepted) {
       await Storage.setUserIsLogged(false);
       history.push('/tos_acceptance');
-    } else {
-      redirect(user);
+      return;
+    }
+
+    const {isResearcher} = Storage.getCurrentUser();
+    if (isResearcher && !researcherProfileComplete(user)) {
```
rushtong: Same feedback here as below, `isResearcher` should be true for all users.

rushtong: I think this is the right place for `researcherProfileComplete`

rushtong (commented):
I'm proposing a change to the ToS redirect in comments below.

src/pages/TermsOfServiceAcceptance.js
lu-c: I don't have a strong opinion here, but I think when I use other websites, I would expect to be redirected to whatever my main home view would be when accepting the ToS. Like, when I've already had to do an extra unexpected click, I just want to see the content, and this extra profile step would be only mildly annoying (and hopefully only seen once :P)

lu-c:
I have a few questions for clarity for this ticket:
1. Is the ToS the same thing as the onboarding form?
2. The ticket didnt describe anything about redirecting to the profile at all - was this addressed in a separate conversation?
3. Would you want to consider a localstorage approach? The ticket says "when a new user signs into duos", so I think it sounds like something that would make sense for tracking that ("`has the user logged in before` AND (`already seen the onboarding form` OR `profile completed`)? if no, redirect to onboarding form, if yes, then send them to the dataset catalog instead)

rushtong:
> I have a few questions for clarity for this ticket:
> 
> 1. Is the ToS the same thing as the onboarding form?
> 2. The ticket didnt describe anything about redirecting to the profile at all - was this addressed in a separate conversation?

I can preemptively answer a couple of these

1. No, they're different things. The on-boarding form is on the profile page. The ToS is the page that shows up if you select "Terms of Service" in the footer, or have not accepted Terms of Service yet (or have accepted, then rejected ToS)
2. Since the onboarding form is on the profile page, that's why we're redirecting there.

rushtong:
> 3\. ("`has the user logged in before` AND (`already seen the onboarding form` OR `profile completed`)

We can answer some of these pretty easily. When a user signs in, we make a couple of calls to the back end to figure out some details.

1. Has the user logged in before? If the user has accepted ToS, then we know they've logged in before.
2. Has the user seen the onboarding form? This is not clear. If the user has accepted ToS or if they've selected an institution, then we know that they have. However, we don't always know if they haven't (sorry for the double negative). If a user has accepted ToS, then rejected ToS, that user looks just like one who's never signed in before. 
3. Is the user's profile completed? That's an easy one - they've selected an institution. We should heed Justin's comment: 

> TLDR: Don't worry about a "complete" profile, just perform the redirect for new users.

connorlbark:
Great discussion here!

Taking in feedback: removed the profile complete check on sign on, and always redirects to the profile page on ToS acceptance. This makes the ticket very simple - no more need for checking profile completeness, or any sort of logic.

rushtong (approved):
👍🏽

JVThomas (approved):

PR #1737, opened 2022-08-05
[DUOS-1967][risk=no] Only show registered users in `Add New Researcher` modal

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1967

----

- Verify all tests go green


shaemarks (commented):
Approach makes sense to me. The only thing I'm wondering is if there are liability reasons for needing the lca text available on the page

rushtong:
@connorlbark - let's bring this up with Jonathan/Pamela. From my poor memory, we do want to be able to add a library card for an unregistered user - that was an original design goal. There was some logic somewhere that would look for unregistered user LCs and link them up if a new user signed in with the email associated to the LC.

connorlbark:
> @connorlbark - let's bring this up with Jonathan/Pamela. From my poor memory, we do want to be able to add a library card for an unregistered user - that was an original design goal. There was some logic somewhere that would look for unregistered user LCs and link them up if a new user signed in with the email associated to the LC.

Talked to Pamela - so you're right that the ability to type in a freetext email is desirable, but the ability to see users without insitutions is not. I will add back the `Add New Researcher` popup to this PR but remove the users without institutions.

src/components/signing_official_table/SigningOfficialTable.js
```diff
@@ -198,6 +198,10 @@ const generateLcaText = (text) => {
   </ReactMarkdown>;
 };
 
+const onlyResearchersWithoutCardFilter = (researcher) => {
+  return isNil(researcher.libraryCards);
```
shaemarks: I can imagine an edge case where a researcher belongs to this institution, doesn't have a library card from this institution, but does have another library card (since admins can technically give researchers library cards from any institution). In that case, they'll appear in the table but not the dropdown in the modal. Not sure how big of an issue this is though

rushtong: Great catch @shaemarks ... and you're right. There will be cases where a user will have library cards from other institutions.

connorlbark: changed the filter here and also the filter in the table itself to check that the SO's institution id the same as the card's institution id

src/components/signing_official_table/SigningOfficialTable.js
```diff
@@ -115,7 +115,7 @@ const LibraryCardCell = ({
 }) => {
   const id = researcher.userId || researcher.email;
   const card = !isEmpty(researcher.libraryCards)
-    ? researcher.libraryCards[0]
+    ? find((card) => card.institutionId == institutionId)(researcher.libraryCards)
```
rushtong: Do you want `===` here?

src/pages/SigningOfficialResearchers.js
```diff
@@ -20,13 +19,7 @@ export default function SigningOfficialResearchers() {
       try {
         setIsLoading(true);
         const soUser = await User.getMe();
-        const soPromises = await Promise.all([
-          User.list(USER_ROLES.signingOfficial),
-          User.getUnassignedUsers()
```
rushtong, line 25: This was the only usage of `User.getUnassignedUsers()` so we can remove it, along with deprecating the API method in consent: `GET /api/user/institution/unassigned`

src/pages/SigningOfficialResearchers.js
```diff
@@ -20,13 +19,7 @@ export default function SigningOfficialResearchers() {
...
-        ]);
-        const researcherList = soPromises[0];
-        const unregisteredResearchers = soPromises[1];
```
rushtong, line 28: ~Just a note that the naming here was incorrect, these were not unregistered users, they were registered users with no institution selection.~

Naming is hard. The API call was actually returning unregistered users with a LC for the calling user's institution, so `User.getUnassignedUsers()` is what is mis-named here.

rushtong:
FYI, the modal has a couple of issues. The prompt text and selection lets me add new users via email. It would be nice to clean up the button alignment as well if possible.

Screen Shot 2022-08-10 at 7 47 59 AM

Screen Shot 2022-08-10 at 7 48 52 AM

rushtong (changes requested):
Looking good, see comments inline.

connorlbark:
> FYI, the modal has a couple of issues. The prompt text and selection lets me add new users via email. It would be nice to clean up the button alignment as well if possible.

@rushtong I messaged Pamela about this. Here is what she said about the popup: `I think the popup is still helpful because I believe SOs can type in the email of an unregistered user and issue them a Library Card. I think that is still a good feature`. I believe the ability to enter freetext emails is good, however the users which displayed when searching were the issue. 

I am confused though on the `getUnassignedUsers` call; we don't need this call, right? In other words, the current behavior of the modal is right, but the feedback/changers here are to update the styling?

rushtong:
> @rushtong I messaged Pamela about this. Here is what she said about the popup: I think the popup is still helpful because I believe SOs can type in the email of an unregistered user and issue them a Library Card. I think that is still a good feature. I believe the ability to enter freetext emails is good, however the users which displayed when searching were the issue.

I may have been confused about the goal of the ticket/PR - I was under the impression that the intention was to **STOP** creating Library Cards for users who are not already registered with the signing official's institution. I would be happy to be wrong here, but it does add some complications to the new designs of Signing Official pages. If so, ignore my feedback about the modal and free text usage.

> I am confused though on the `getUnassignedUsers` call; we don't need this call, right? In other words, the current behavior of the modal is right, but the feedback/changers here are to update the styling?

`User.getUnassignedUsers()` calls `GET /api/user/institution/unassigned` which returns a [list of users who do not have an institution](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java#L152). If we're not querying for those users, we don't need to make that call here.

src/pages/SigningOfficialResearchers.js
connorlbark: it's also used in `SigningOfficialDataSubmitters.js`; is that deprecated code? (also signing official console, but that is being deprecated)

connorlbark:
> I may have been confused about the goal of the ticket/PR - I was under the impression that the intention was to **STOP** creating Library Cards for users who are not already registered with the signing official's institution. I would be happy to be wrong here, but it does add some complications to the new designs of Signing Official pages. If so, ignore my feedback about the modal and free text usage.

Sounds good to me! We can also open a dialogue with Pamela/Jonathan if you think the behavior should be different.

src/pages/SigningOfficialResearchers.js
rushtong: Looks like I was incorrect ... it is used in three places in develop right now, not sure how I missed them.
* SigningOfficialConsole
* SigningOfficialDataSubmitters
* SigningOfficialResearchers

src/components/signing_official_table/SigningOfficialTable.js
connorlbark: yep! you're totally right

src/components/signing_official_table/SigningOfficialTable.js
```diff
@@ -198,6 +198,15 @@ const generateLcaText = (text) => {
   </ReactMarkdown>;
 };
 
+const onlyResearchersWithoutCardFilter = (institutionId) => (researcher) => {
+  const cards = researcher.libraryCards;
+  if (isEmpty(cards)) {
+    return true;
+  }
+
+  return isNil(find((card) => card.institutionId == institutionId)(researcher.libraryCards));
```
shaemarks:
```suggestion
  return isNil(find((card) => card.institutionId === institutionId)(researcher.libraryCards));
```

shaemarks (approved):
Looks good to me! Styling on modal looks much better!

rushtong (approved):

PR #1743, opened 2022-08-09
[DUOS-1935][risk=no] Add data submission page

connorlbark:
## Addressees

https://broadworkbench.atlassian.net/browse/DUOS-1935

----

- Verify all tests go green

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -38,6 +38,7 @@ import MemberConsole from './pages/MemberConsole';
 import TermsOfService from './pages/TermsOfService';
 import TermsOfServiceAcceptance from './pages/TermsOfServiceAcceptance';
 import { HealthCheck } from './pages/HealthCheck';
+import DataSubmissionForm from './pages/DataSubmissionForm';
 
 const Routes = (props) => (
   <Switch>
@@ -97,6 +98,8 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/dataset_statistics/:datasetId" component={DatasetStatistics} props={props}
       rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/tos_acceptance" component={TermsOfServiceAcceptance} props={props} rolesAllowed={[USER_ROLES.all]} />
+    <AuthenticatedRoute path="/data_submission_form" component={DataSubmissionForm} props={props} rolesAllowed={[USER_ROLES.all]} />
+
     <Route path="*" component={NotFound} />
   </Switch>
 );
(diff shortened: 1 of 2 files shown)


src/pages/DataSubmissionForm.js
```diff
@@ -0,0 +1,49 @@
...
+              'Register a Dataset',
+            ]),
+            div(
+              {
+                style: {
+                  fontFamily: 'Montserrat',
+                  fontSize: '1.6rem'
+                },
+              },
+              ['View Dataset submissions and status']
```
connorlbark: This is the description in the mock, but this doesn't feel accurate to me? There aren't submissions (plural)

src/Routes.js
```diff
@@ -97,6 +98,8 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/dataset_statistics/:datasetId" component={DatasetStatistics} props={props}
       rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/tos_acceptance" component={TermsOfServiceAcceptance} props={props} rolesAllowed={[USER_ROLES.all]} />
+    <AuthenticatedRoute path="/data_submission_form" component={DataSubmissionForm} props={props} rolesAllowed={[USER_ROLES.all]} />
```
JVThomas: The route should only be visible when env === dev or env === local. We don't want it to be accessible anywhere else.  Use the line below to do just that

`{(props.env === 'local' || props.env === 'dev') && <AuthenticatedRoute path="/data_submission_form" component={DataSubmissionForm} props={props} rolesAllowed={[USER_ROLES.all]} />}`

src/pages/DataSubmissionForm.js
JVThomas: Good point. I also think the status may not be needed either since it's a submission form. For now feel free to either keep it or adjust it as needed. We can always update the text down the line on a later ticket.

JVThomas (changes requested):
One change required for routes, the other comment is more of a suggestion. Other than that the ticket looks good.

src/Routes.js
connorlbark: good catch!! thank you

JVThomas (approved):
Changes look good! The Dataset Registration page only shows up on local and dev. This should be enough of a base for the sub-component work.

src/pages/DataSubmissionForm.js
```diff
@@ -0,0 +1,49 @@
+import { useState } from 'react';
+import {div, img} from 'react-hyperscript-helpers';
+import lockIcon from '../images/lock-icon.png';
+import {Styles} from '../libs/theme';
+
+
+export const DataSubmissionForm = () => {
+
+  // NOTE: remove after adding components
+  /* eslint-disable no-unused-vars */
```
lu-c, line 10: :D

lu-c (approved):
lets gooo

PR #1718, opened 2022-07-28
[DUOS-1904] Added researcher actions to Actions component

lu-c:
----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #1744, opened 2022-08-10
DUOS-2014[risk=no] Updated Researcher Actions functionality to account for DarCollectionSummary

JVThomas:
Addresses [DUOS-2014](https://broadworkbench.atlassian.net/browse/DUOS-2014)

PR updates the functions around Researcher actions to account for the new DarCollectionSummary format. PR also includes import corrections to `applyHoverStyle` in utils

- Verify all tests go green


src/libs/utils.js
```diff
@@ -1,7 +1,7 @@
 import Noty from 'noty';
 import 'noty/lib/noty.css';
 import 'noty/lib/themes/bootstrap-v3.css';
-import {map as nonFPMap} from 'lodash';
+import {map as nonFPMap, forEach as nonFPForEach} from 'lodash';
```
lu-c: would you feel okay with leaving these as the original names? It seems less confusing that way

JVThomas: So there's two forEach's being used in this file, one from 'lodash/fp' and one from 'lodash'. The reason being is that with the fp version, the function is `forEach(collection, (value) => //code here)` where as in the regular version its `forEach(value, key) => //insert code here)(collection)`.

As for why we need two versions, remember the error Shae pointed out here: https://github.com/DataBiosphere/duos-ui/pull/1718#pullrequestreview-1067388069

It's being caused by this function in  `utils.js`

`export const applyHoverEffects = (e, style) => {
  nonFPForEach((key, value) => {
    e.target.style[key] = value;
  })(style);
};`

What had happened was that the normal `forEach` that was used here was replaced with the `fp` variant. The difference in the iterating function causes the function to error out. 

So in short, we need both functions, but they can't be named the same variable, hence the `nonFPForEach`.

That said, I still need to update the iterating function to match the non-fp version.

lu-c: Ok! Didnt realize it was because both lodash and lodash/fp were beign used.... That being said, I still think the name is confusing (just because it wasn't immediately obvious to me what it was "nonFP" was supposed to mean. Don't know what to do about that though

connorlbark: maybe it could be `lodashForEach` and `lodashFpForEach` ? that makes both mirror each other.

connorlbark: maybe it could be `lodashForEach` and `lodashFpForEach` ? that way they mirror each other

rushtong: Another way to solve this is to import a prefix for either the non-fp or the fp version. In the past, I've used `import _ from 'lodash';` and `import * as fp from 'lodash/fp';`, for example

shaemarks (commented):
Actions are all functioning as expected 👍 

I think the header of the `CollectionConfirmationModal` needs to be updated as well - currently it always says undefined as the project title 
Screenshot (87)

connorlbark (approved):
worked well in ui! other than the weird naming of the forEach & modal issue, 👍

src/libs/utils.js
JVThomas: Yeah but that would cause Webpack to import the entire `lodash` library on file bundling, and lodash is a pretty big library. Even if the code is minified, I'd rather not import methods that are never used for a particular file.

JVThomas: > maybe it could be `lodashForEach` and `lodashFpForEach` ? that way they mirror each other

That works for me

src/components/dar_collection_table/CollectionConfirmationModal.js
```diff
@@ -7,7 +7,7 @@ import {getProjectTitle} from './DarCollectionTable';
 export default function CollectionConfirmationModal(props) {
   const {collection, showConfirmation, setShowConfirmation, cancelCollection, reviseCollection, openCollection, consoleAction, deleteDraft} = props;
 
-  const getModalHeader = () => {
+  const getModalHeader = (collection) => {
     if(!isNil(collection)) {
       return `${collection.darCode} - ${getProjectTitle(collection)}`;
```
shaemarks: Optional nit: I don't see `getProjectTitle` used anywhere but here, so maybe inline `collection.name` here and delete the function

shaemarks (approved):
Changes look good to me!

connorlbark (approved):
👍 changes since i last reviewed look good to me

lu-c (approved):

PR #1749, opened 2022-08-12
[DUOS-1950][risk=no] Added study information component to data submission form page

lu-c:
----

- Verify all tests go green


src/components/SearchSelectOrText.js
```diff
@@ -22,7 +22,7 @@ export const SearchSelectOrText = (props) => {
       if (item) {
         setCurrentDisplay(item.displayText);
       }
-    } else {
+    } else if (props.freetextValue) {
```
lu-c, line 25: Not sure if this changes the searchSelectOrText behavior too much, it seemed okay to me, but could use confirmation!

cypress/component/DataSubmissionStudyInformation/ds_study_information.spec.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable no-undef */
+import React from 'react';
+import { mount } from 'cypress/react';
+import { cloneDeep } from 'lodash/fp';
+import DataSubmissionStudyInformation from "../../../src/components/data_submission/ds_study_information";
```
rushtong, line 5: Codacy has a fix for the issue: [Strings must use singlequote.](https://app.codacy.com/gh/DataBiosphere/duos-ui/pullRequest?prid=9974908)
```suggestion
import DataSubmissionStudyInformation from '../../../src/components/data_submission/ds_study_information';
```

cypress/component/DataSubmissionStudyInformation/ds_study_information.spec.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-undef */
+import React from 'react';
+import { mount } from 'cypress/react';
+import { cloneDeep } from 'lodash/fp';
+import { User } from '../../../src/libs/ajax';
+import DataSubmissionStudyInformation from "../../../src/components/data_submission/ds_study_information";
```
rushtong: Codacy has a fix for the issue: [Strings must use singlequote.](https://app.codacy.com/gh/DataBiosphere/duos-ui/pullRequest?prid=9974908)
```suggestion
import DataSubmissionStudyInformation from '../../../src/components/data_submission/ds_study_information';
```

src/components/data_submission/ds_study_information.js
```diff
@@ -0,0 +1,145 @@
...
+import { User } from '../../libs/ajax';
+import { FormField, FormTable } from './ds_common';
+
+const updateUserAndFields = async ({ setUser, onChange }) => {
+  const me = await User.getMe();
+  setUser(me);
+  onChange({key: 'dataSubmitterId', value: me.dacUserId});
+};
+
+export default function DataSubmissionStudyInformation(props) {
```
connorlbark, line 9: should this file be snake case too? maybe this file should be `DataSubmissionStudyInformation`

src/pages/DataSubmissionForm.js
```diff
@@ -23,26 +25,24 @@ export const DataSubmissionForm = () => {
...
           ]),
         ]
       ),
     ]),
+
+    form({}, [
+      h(DataSubmissionStudyInformation, {
+        onChange: ({ key, value }) => {
+          /* eslint-disable no-console */
+          console.log('StudyInfo OnChange:', key, value);
```
connorlbark: use `setFormData`?

src/components/data_submission/ds_common.js
```diff
@@ -0,0 +1,341 @@
...
+*    * checkboxType
+* required, disabled
+* placeholder, defaultValue,
+* style (for the formControl component)
+* inputStyle (for the input element)
+* onChange,
+* errors, setErrors,
+* formInfo, setFormInfo
+*/
+export const FormField = (config) => {
```
connorlbark: I really like the idea of making all of these generic components! Should we move this to somewhere where they can be used more broadly? Like just in the `Components` folder with a more descriptive name? Maybe in a separate ticket we could try to normalize the creation of preexisting form fields (in e.g. the researcher profile), as I find that consistent styling has been an issue - this would help mitigate that in forms.

src/components/data_submission/ds_common.js
```diff
@@ -0,0 +1,341 @@
...
+          ...formCol,
+          id: `${id}-${i}-${formCol.id}`,
+          path: `${id}.${i}.${formCol.id}`,
+          hideTitle: true, ariaDescribedby: `${id}-${formCol.title}`,
+          onChange, errors, setErrors, formInfo, setFormInfo
+        });
+      }));
+    }),
+    // add new button
+    div({ style: { display: 'flex', width: '100%', justifyContent: 'flex-end' } }, [
```
connorlbark: what is the add new button for? maybe this should exist outside of `FormTable` since the component could be used for more things.

src/components/data_submission/ds_common.js
```diff
@@ -0,0 +1,341 @@
...
+    setFormInfo(updatedFormInfo);
+  }
+};
+
+//---------------------------------------------
+// Form Controls
+//---------------------------------------------
+const formInput = (config) => {
+  switch (config.type) {
+    case 'select': return formInputSelect(config);
```
connorlbark: maybe we could make these strings constants and export them so that you don't ever have to use a raw string to specify things

connorlbark (commented):
haven't ran it yet - but it looks great! biggest thing is i think it would be cool to make the FormField stuff more generic and able to used in more contexts.

src/pages/DataSubmissionForm.js
lu-c: I think it ultimately should, but the ticket said to just make it an onChange method that will have a separate ticket to set

src/components/data_submission/ds_common.js
lu-c: The "Add new" button adds a new row to the formTable, so its functionality is tied pretty strictly to that!

src/components/data_submission/ds_common.js
lu-c: Good idea, i'll add that!

src/components/data_submission/dsStudyInformation.js
```diff
@@ -0,0 +1,145 @@
+import { useState, useEffect } from 'react';
+import { h, h2, div } from 'react-hyperscript-helpers';
+import { Notifications, isEmailAddress } from '../../libs/utils';
+import { User } from '../../libs/ajax';
+import { FormFieldTypes, FormField, FormTable } from './ds_common';
+
+const updateUserAndFields = async ({ setUser, onChange }) => {
+  const me = await User.getMe();
+  setUser(me);
+  onChange({key: 'dataSubmitterId', value: me.dacUserId});
```
lu-c: Should I be using `dacUserId` or `userId`?

src/components/data_submission/ds_study_information.js
lu-c: Updated!

src/components/SearchSelectOrText.js
lu-c, line 25: Not sure if this changes the searchSelectOrText behavior too much, it seemed okay to me, but could use confirmation!

src/components/data_submission/ds_common.js
lu-c: I have so many thoughts about this! We should try to do as much as possible through css for normalizing the appearances of form fields, but where there is a necessary html component (like checkboxes and sliders), i would definitely want to create generic form components for it!

I think the components folder is really hard to browse for existing items though, and could use a little restructuring? Like, generic reuseable components should be grouped, and specific (potentially single purpose use) components shoudl also be grouped separately.

Maybe we could make a common/ colder inside components? Or it could exist outside of components as it's own `common/` too?

cypress/component/DataSubmissionStudyInformation/ds_study_information.spec.js
```diff
@@ -0,0 +1,302 @@
+/* eslint-disable no-undef */
+import React from 'react';
+import { mount } from 'cypress/react';
+import { cloneDeep } from 'lodash/fp';
+import { User } from '../../../src/libs/ajax';
+import DataSubmissionStudyInformation from '../../../src/components/data_submission/ds_study_information';
```
connorlbark: maybe the overall folder could be called `DataSubmission` so all the component tests for the overall functionality are in one place ?

lu-c: I'm sorry, I don't understand your suggestion - What do you mean?

src/components/data_submission/ds_study_information.js
```diff
@@ -0,0 +1,147 @@
...
+      try {
+        updateUserAndFields({ setUser, ...props });
+      } catch (error) {
+        Notifications.showError({text: 'Error: Unable to retrieve user data from server'});
+      }
+    };
+
+    init();
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []); // need the eslint to force useEffect to only run on initialization
```
JVThomas: Ignoring the exhaustive-deps rule is frowned upon. Here's a thread on Github with gaeron (Dan Abramov, software engineer on the React team) explaining his concerns
https://github.com/facebook/create-react-app/issues/6880

Basically turning the rule off causes developers to miss out on code smells or logical issues with the method in the useEffect itself. In this case, we see that assigning the required dependencies results in an infinite re-render loop.

So what's causing the issue? The two dependencies on this useEffect are `onChange` and `updateUserAndFields`. `updateUserAndFields` is defined in this component, so the only way that can trigger this function to re-fire is if the props are updated. And the only thing on props right now is `onChange`. So somehow, onChange is being updated when this useEffect is fired, causing the component to re-render and fire the useEffect again, resulting in the infinite loop.

Okay, so let's look at the parent component `DataSubmissionForm` and see what's up there. onChange looks to be an anonymous function that is defined directly on the component as a prop. This means that if `DataSubmissionForm` re-renders, the `onChange` function is re-defined, meaning the props update, triggering the re-render.

That said, what's causing the parent component to re-render? Well the onChange is updating `formData` with a cloned value. A cloned value means a new object reference. React triggers re-renders if props or state variables update their values. If the variable is an object, then React will compare object references (not the contents) to see if they are the same value. 

In this instance a cloned copy is a new object which means a different reference. This will trigger a re-render on the parent and start the infinite loop you're seeing. To avoid the complete page re-render, update the target keys on `formData`.

I know React promotes immutability, but it does so with regards to `this.state` itself (meaning you shouldn't directly update state unless using a hook). Updating `formData` here is fine, especially considering that each sub-component is going to rely on formData as a source of truth for data and rendering triggers.

For further reference, here's another post explaining the above in greater detail
https://blog.logrocket.com/immutability-in-react-ebe55253a1cc/

lu-c: Thanks for the explanation! Updated the component to not have an empty deps array, and cleaned up the sideeffects fallout!

src/components/forms/forms.js
```diff
@@ -0,0 +1,404 @@
...
+    : null;
+};
+
+const extractValidators = (config) => {
+  const { required, validator } = config;
+  const ret = [];
+
+  if (required) {
+    ret.push(validateRequired);
+  }
```
lu-c, line 33: created this function bc long term, it might be useful to add different types of common validators (like Array.minLength)

src/components/data_submission/ds_study_information.js
```diff
@@ -0,0 +1,144 @@
...
+      id: 'species',
+      title: 'Species',
+      onChange
+    }),
+    h(FormField, {
+      id: 'piName',
+      title: 'Principal Investigator Name',
+      onChange
+    }),
+    user && h(FormField, {
```
JVThomas: Nit: The props object for hyperscript functions can process an `isRendered` key for conditional rendering. so you can do something like...

```
h(FormField, {
      id: 'dataSubmitterEmail',
      title: 'Data Submitter Email',
      defaultValue: user?.email,
      disabled: true,
      isRendered: !isEmpty(user),
      onChange
    })
```

and have the same effect.

src/pages/DataSubmissionForm.js
```diff
@@ -1,15 +1,25 @@
...
 
 export const DataSubmissionForm = () => {
 
   // NOTE: remove after adding components
   /* eslint-disable no-unused-vars */
   const [formData, setFormData] = useState({});
 
+  const onChange = ({ key, value }) => {
+    /* eslint-disable no-console */
+    console.log('StudyInfo OnChange:', key, value);
```
JVThomas: Nit: stray console.log

JVThomas:
The current functionality allows users to remove all elements on the File Type list.

Screen Shot 2022-08-16 at 11 27 54 AM

----

Screen Shot 2022-08-16 at 11 36 24 AM


Since we require there to be at least one element on the list (with all inputs filled), you'll need to either show an error if the list is empty or prevent them from removing all elements. You can approach this in two ways...

1. Only show the close/delete icon if there's more than one item
2. Show a validation error if there are no elements present in the list (show the error box below the table).

Considering that the other form components allow for validators, I think it's reasonable for the FormTable to take in validation props that it can apply to itself (like min size of `enter size here`). The fields themselves should also have validators defined to check if the input is empty.

src/components/forms/forms.js
```diff
@@ -0,0 +1,393 @@
...
+        },
+        type: 'button',
+        onClick: () => {
+          const formValueClone = cloneDeep(formValue);
+          formValueClone.push({});
+          setFormValue(formValueClone);
+          onChange({ key: id, value: formValueClone });
+        }
+      }, [
+        'Add New Filetype',
```
connorlbark: this string should be a prop so that the field form can be used generically

src/components/data_submission/dsStudyInformation.js
rushtong: `userId` ... dac user id is deprecated in all of the APIs.

src/components/forms/forms.js
lu-c: done!

src/pages/DataSubmissionForm.js
lu-c: i actually kind of liked it there... just while there was no other confirmation that the code was working as intended. Would you be okay with keeping this until the feature is completed?

JVThomas: Sure, I was worried Codacy was going to freak out about it, but apparently not

lu-c:
Found a bug with deleting rows (value updates, but not ui), going to tackle that next!

src/components/forms/forms.js
```diff
@@ -288,14 +292,20 @@ const formInputSlider = (config) => {
...
   const {
-    id, title, hideTitle, description,
+    id, title, hideTitle, description, type,
     defaultValue, style, validators
   } = config;
 
   const [error, setError] = useState();
-  const [formValue, setFormValue] = useState(defaultValue || '');
+  const [formValue, setFormValue] = useState(type?.defaultValue || '');
   const required = validators ? validators.findIndex(v => v === FormValidators.REQUIRED) !== -1 : false;
```
connorlbark: `validators.includes(FormValidators.REQUIRED)`?

connorlbark: same comment other usages of this

src/pages/DataSubmissionForm.js
```diff
@@ -14,9 +14,9 @@ export const DataSubmissionForm = () => {
   /* eslint-disable no-unused-vars */
   const [formData, setFormData] = useState({});
 
-  const onChange = ({ key, value }) => {
+  const onChange = ({ key, value, isValid }) => {
```
connorlbark, line 17: I like the is valid approach!

connorlbark (approved):
Everything works super well locally! Couple nits but overall I would not mind getting this merged so the rest of us can use the FormField work (which is awesome: great work on that). My only concern is that I was confused on the MULTITEXT fields. I didn't realize I had to hit `Enter` for it to update / submit the value. Maybe there should be a small '+' or something on the righthand side? I would be ok with tackling this now or by doing it in a separate small ticket so that we can free up others to work on the form.

src/components/forms/forms.js
lu-c: updated!

lu-c:
> Everything works super well locally! Couple nits but overall I would not mind getting this merged so the rest of us can use the FormField work (which is awesome: great work on that). My only concern is that I was confused on the MULTITEXT fields. I didn't realize I had to hit `Enter` for it to update / submit the value. Maybe there should be a small '+' or something on the righthand side? I would be ok with tackling this now or by doing it in a separate small ticket so that we can free up others to work on the form.

Yeah, sure, I can add a button for it! Still waiting on a second approval, but I won't let that hold this pr back. ;) Will merge whether or not the button exists, and figure out adding that button when it can

JVThomas (approved):
The code looks great! The form is validating the fields as you would expect. The form components themselves will be extremely useful for the rest of the form (and any other form in the future).

PR #1756, opened 2022-08-17
[DUOS-2027] Fixed signing on always redirecting you to the first available console page

lu-c:
Just that when you are on a page for too long, maybe you were coding in your IDE, your session times out, and the website asks you to log back in. When you log in though, instead of redirecting you back to the page you were on, you find yourself now on "/admin_manage_dar_collections", and you can't even hit the back button to go back to where you were from before the site FORCED you to leave

I know there's definitely a better way to do this (without relying on query parameters in the url bar, and just using the router itself), but i have wasted enough unauthorized time on this unasked for ticket

fast, but dirty. please let this go through XP This just gets annoying during development!

----

- Verify all tests go green

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -218,14 +218,9 @@ export const setUserRoleStatuses = (user, Storage) => {
 
 export const Navigation = {
   back: async (user, history) => {
-    const queryParameters = Object.fromEntries(
-      window.location.search
-        .replace('?', '')
-        .split('&')
-        .map(queryString => queryString.split('='))
-    );
+    const queryParams = new URLSearchParams(window.location.search);
     let page =
-      queryParameters.redirectTo ? queryParameters.redirectTo
+      queryParams.get('redirectTo') ? queryParams.get('redirectTo')
         : user.isAdmin ? '/admin_manage_dar_collections'
           : user.isChairPerson ? '/chair_console'
             : user.isMember ? '/member_console'
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -218,13 +218,20 @@ export const setUserRoleStatuses = (user, Storage) => {
 
 export const Navigation = {
   back: async (user, history) => {
+    const queryParameters = Object.fromEntries(
+      window.location.search
+        .replace('?', '')
+        .split('&')
+        .map(queryString => queryString.split('='))
+    );
     let page =
-      user.isAdmin ? '/admin_manage_dar_collections'
-        :user.isChairPerson ? '/chair_console'
-          : user.isMember ? '/member_console'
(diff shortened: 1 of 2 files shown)


src/libs/utils.js
```diff
@@ -218,13 +218,20 @@ export const setUserRoleStatuses = (user, Storage) => {
 
 export const Navigation = {
   back: async (user, history) => {
+    const queryParameters = Object.fromEntries(
+      window.location.search
+        .replace('?', '')
+        .split('&')
+        .map(queryString => queryString.split('='))
+    );
```
connorlbark: is there a utility/library function to do this? just for safety purposes

src/libs/utils.js
```diff
@@ -218,13 +218,20 @@ export const setUserRoleStatuses = (user, Storage) => {
...
-            : user.isResearcher ? '/dataset_catalog'
-              : user.isAlumni ? '/summary_votes'
-                : '/';
+      queryParameters.redirectTo ? queryParameters.redirectTo
+        :user.isAdmin ? '/admin_manage_dar_collections'
+          :user.isChairPerson ? '/chair_console'
+            : user.isMember ? '/member_console'
+              : user.isResearcher ? '/dataset_catalog'
+                : user.isAlumni ? '/summary_votes'
+                  : '/';
```
connorlbark:
```suggestion
      queryParameters.redirectTo ? queryParameters.redirectTo
        : user.isAdmin ? '/admin_manage_dar_collections'
          : user.isChairPerson ? '/chair_console'
            : user.isMember ? '/member_console'
              : user.isResearcher ? '/dataset_catalog'
                : user.isAlumni ? '/summary_votes'
                  : '/';
```
nit

src/libs/utils.js
lu-c: looks like there's a native browser way to do it 👍

connorlbark (approved):
👍

rushtong:
I think we have a related ticket to this: https://broadworkbench.atlassian.net/browse/DUOS-1501
Please review to see if there's overlap here.

lu-c:
> I think we have a related ticket to this: https://broadworkbench.atlassian.net/browse/DUOS-1501 Please review to see if there's overlap here.

Oh yeah, this is definitely the same thing! I'm adding it as a url hash because I couldn't figure out the appropriate react-router way to do it... That's perfect! :D

rushtong (approved):
This works as designed. One thing I'd like to see is the ability to navigate to a pasted link in the address bar by remembering the URL, signing the user in, and then redirecting to the remembered URL. This PR doesn't quite get us to that point and we will definitely need this for TDR/Catalog integration.

PR #1766, opened 2022-08-22
[DUOS-1669][risk=no] Added data submitter console

lu-c:
Updated to add data submitter header link, console page.

Currently, it seems like the data submitter role does not yet exist (at least from the consent swagger-ui that role id was not indicated under the users tab), but this pr could be changed to make it fit when if there are discrepancies!

Overall, there were a few more changes to this PR to try to make updating roles simpler, by moving consolidating them a little bit more.

- Moved the tab configs out of the component level so it is only rendered once (since it is a static variable atm)
- Added field "isRendered" to the tab config to remove the manual declarations of roles (ie: `isChairPerson` and `isResearcher`... etc)
- Added new header tab config for data submitter
- Added data submitter role to the utils
- Modified redirection navigation route to default to use the tabConfigs to determine which page to load when the user first logs in (ie: first console, first link). States that did not exist as part of the console were kept at the ends.

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -67,6 +67,72 @@ const styles = {
   }
 };
 
+export const headerTabsConfig = [
+  {
+    label: 'Admin Console',
+    link: '/admin_manage_dar_collections',
+    children: [
+      { label: 'DAR Requests', link: '/admin_manage_dar_collections' },
+      { label: 'Dataset Catalog', link: '/dataset_catalog' },
+      { label: 'DACs', link: '/manage_dac' },
+      { label: 'Statistics', link: '/summary_votes' },
+      { label: 'Users', link: '/admin_manage_users' },
+      { label: 'Institutions', link: '/admin_manage_institutions' },
+      { label: 'Library Cards', link: '/admin_manage_lc' }
+    ],
+    isRendered: (user) => user.isAdmin
+  },
+  {
+    label: 'SO Console',
+    link: '/signing_official_console/researchers',
+    children: [
+      { label: 'Researchers', link: '/signing_official_console/researchers' },
+      { label: 'DAR Requests', link: '/signing_official_console/dar_requests' }
+    ],
+    isRendered: (user) => user.isSigningOfficial
+  },
+  {
+    label: 'Researcher Console',
+    link: '/researcher_console',
+    search: 'researcher_console',
+    children: [
+      { label: 'DAR Requests', link: '/researcher_console' },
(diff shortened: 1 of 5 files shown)


connorlbark:
I like the changes here! Makes it a lot cleaner - do we want to also add the ability to make someone a data submitter from the admin console? If so, I could see that being here or in another ticket - what do you think?

lu-c:
> I like the changes here! Makes it a lot cleaner - do we want to also add the ability to make someone a data submitter from the admin console? If so, I could see that being here or in another ticket - what do you think?

What is the admin console? Is that in teh UI? Or do you mean through the swagger ui? If the former, I'd like to clear it with the team first, just to see if there's some security concern, but it is a good idea! (always prefer doing stuff through a pretty ui 😊 )

connorlbark:
> What is the admin console? Is that in teh UI? Or do you mean through the swagger ui? If the former, I'd like to clear it with the team first, just to see if there's some security concern, but it is a good idea! (always prefer doing stuff through a pretty ui 😊 )

Ohhh sorry I meant the admin edit user page! it would just make it easier to assign people to being data submitters. maybe let's ask tmrw at standup !

src/components/DuosHeader.js
```diff
@@ -67,6 +67,72 @@ const styles = {
...
+  },
+  {
+    label: 'DS Console',
+    link: '/data_submission_form',
+    search: 'data_submission_form',
+    children: [
+      { label: 'Datasets', link: '/data_submission_form' }
+    ],
+    isRendered: (user) => user.isDataSubmitter
+  }
```
shaemarks, line 133: I think we should move the researcher console tab config to the end of this array if we want Data Submitters to see the Data Submitter console first. Every user has a researcher role, so the `find` function that gets the first console would hit the researcher console first.

shaemarks (commented):
Changes look great, and code is a lot cleaner! However, I wasn't able to give myself the Data Submitter role to test locally (the role does exist on the backend as of [DUOS-1974](https://broadworkbench.atlassian.net/browse/DUOS-1974), but it looks like a little more backend work is necessary for users to receive the role)

src/components/DuosHeader.js
lu-c: gotcha, I agree! Moving the researcher block to the end of the array 👍

shaemarks (approved):
Changes look good to me! I was able to give myself the DataSubmitter role via the `PUT /api/user/{userId}` endpoint and access the console

connorlbark (approved):
👍 works locally on my user who is a data submitter!

PR #1770, opened 2022-08-23
[noticket][risk=no] Added null pointer catches for variables on the statistics page

lu-c:
loading a dar can lead to errors, and when it does, it error toasts and then shows the user the full stack of the error (which is probably bad). Now it will toast, but then show a blank page.

----

- Verify all tests go green


connorlbark (approved):
👍

shaemarks (approved):
LGMT! Tested locally, and the dataset statistics page with blank fields was displayed instead of the site crashing

PR #1747, opened 2022-08-11
[DUOS-1964][risk=no] WIP Add DataAccessGovernance section to Data Submission Form

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1964

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #1771, opened 2022-08-23
WIP [DUOS-1958][risk=no] "NIH and AnVIL use" sub-component for data submission form 

shaemarks:
Addresses [DUOS-1958](https://broadworkbench.atlassian.net/browse/DUOS-1958)

----

- Verify all tests go green


src/components/forms/forms.js
```diff
@@ -232,6 +255,35 @@ const formInputSelect = (config) => {
   ]);
 };
 
+const formInputRadioButtons = (config) => {
```
connorlbark, line 238: I also added this in mine! I haven't added tests yet but mine also has the ability to have 'Other' with text fields and such. Not sure how exactly we should merge them. My code is here https://github.com/DataBiosphere/duos-ui/pull/1747 if you want to check it out. It's a bit more complicated but it would work here too.

src/components/forms/forms.js
```diff
@@ -232,6 +255,35 @@ const formInputSelect = (config) => {
...
+        disabled,
+        description: option.label,
+        checkedStyle: {
+          backgroundColor: '#0948B7',
+          border: '5px solid white',
+          boxShadow: '#0948b7 0 0 0 1px'
+        }
+      })
+    ),
+    error && div({ className: `error-message fadein`}, [
```
lu-c: Since this repeats multiple times through the code, can we pull this out into a separate function? Sorry, that was my bad!

src/components/forms/forms.js
```diff
@@ -30,6 +33,10 @@ export const FormValidators = {
   REQUIRED: {
     isValid: (value) => value !== undefined && value !== null && value !== '',
     msg: 'Please enter a value'
+  },
+  EMAIL: {
```
lu-c, line 37: This is a great addition!

src/components/forms/forms.js
```diff
@@ -62,15 +69,30 @@ const normalizeValue = (value) => {
   return value;
 };
 
+const clearDependentFormFields = (config) => {
+  const {onChange, dependentFormFields = []} = config;
+  /*
+   * Clears fields for questions that are rendered conditionally, setting them to their default value
```
lu-c: I think of "clear" as to set it to blank, like `''`, while "reset" would return to the default value - There's a real value to having both of these functions! What would you think of adding a param to allow you to completely clear to `type.defaultValue`?

src/components/forms/forms.js
lu-c: these merges are going to be rough! I added some radio buttons too, but yall wrote way better code XD

src/pages/DataSubmissionForm.js
```diff
@@ -41,7 +42,8 @@ export const DataSubmissionForm = () => {
     ]),
 
     form({}, [
-      h(DataSubmissionStudyInformation, { onChange })
+      h(DataSubmissionStudyInformation, { onChange }),
+      h(DataSubmissionNihAnvilUse, { onChange, formData })
```
lu-c: The study information component doesn't take in any existing formData - should it? (is the purpose to set values for a saved submission form?)

src/components/forms/forms.js
shaemarks: Yeah, I think I'll copy over the radio buttons from #1747 to my branch to make merging easier since Connor's version can work for both of us

src/pages/DataSubmissionForm.js
shaemarks: Thanks for the feedback!

You can ignore the formData prop for now! 
I was trying to work out some bugs with my ticket and that was left over from my experimentation. I think I found a workaround, but let me know if you have ideas on a better approach (I'll comment closer to the actual code to provide context)

src/pages/DataSubmissionForm.js
```diff
@@ -14,7 +15,7 @@ export const DataSubmissionForm = () => {
   const onChange = ({ key, value, isValid }) => {
     /* eslint-disable no-console */
     console.log('StudyInfo OnChange:', key, value, isValid);
-    set(key, value, formData);
+    set(formData, key, value);
```
shaemarks, line 18: Not sure why, but I couldn't get the lodash/fp version of set to actually update formData (but the regular lodash function seems to be working fine). I tested by adding `console.log(formData)` after the call to set, and with the lodash/fp version it always printed `{} `

src/components/data_submission/DataSubmissionNihAnvilUse.js
```diff
@@ -0,0 +1,87 @@
...
+      dependentFormFields: [
+        {id: 'submittingToAnvil', type: FormFieldTypes.RADIO, shouldClear: ({value}) => value === I_DID},
+        {id: 'dbGaPPhsID', type: FormFieldTypes.TEXT},
+        {id: 'dbGaPStudyRegistrationName', type: FormFieldTypes.TEXT},
+        {id: 'embargoReleaseDate', type: FormFieldTypes.TEXT},
+        {id: 'sequencingCenter', type: FormFieldTypes.TEXT},
+      ],
+      validators: [FormValidators.REQUIRED],
+      onChange,
+      setFormValueParent: setNihAnvilUse
```
shaemarks, line 52: My ticket requires different questions to be rendered based on the option selected for the `nihAnvilUse`, but the `formValue` state variable is contained within the `FormField` itself. To address this, I'm also tracking the value of `nihAnvilUse` in the `DataSubmissionNihAnvilUse` component.

Past Attempts:
 I tried experimenting getting the value from `formData.nihAnvilUse`, but the changes didn't register in my component since the same formData object is being used. I also tried having formData be a state variable, and in onChange, I would deepClone the current contents of formData, set the given key value pair, and call setFormData with the new object. However, this caused the hook in `DataSubmissionStudyInformation` to be called way too often. 

Current Approach:
Currently, I am passing `setNihAnvilUse` as an optional `setFormValueParent` prop to FormField, which is called with the new form field value in `onFormInputChange`. This ensures that when the `formValue` is changed in the FormField, it is also changed in this component

src/components/data_submission/DataSubmissionNihAnvilUse.js
```diff
@@ -0,0 +1,87 @@
...
+      id: 'nihAnvilUse',
+      title: 'Will you or did you submit data to the NIH?',
+      type: FormFieldTypes.RADIO,
+      selectOptions: [
+        {label: I_DID, value: I_DID},
+        {label: I_WILL, value: I_WILL},
+        {label: NO, value: NO},
+      ],
+      dependentFormFields: [
+        {id: 'submittingToAnvil', type: FormFieldTypes.RADIO, shouldClear: ({value}) => value === I_DID},
```
shaemarks: I originally had all of the dependent form fields be cleared when `nihAnvilUse` is changed, but there are two values of `nihAnvilUse`  where the `submittingToAnvil` question appears. So, if I switched between `nihAnvilUse === 'I will'` and `'nihAnvilUse' === 'No'`, `submittingToAnvil` in formData is cleared but the selected radio buttons weren't. One possible solution is allowing the `dependentFormFields` to have a `shouldClear` condition (if not provided, assume clear on any change). Here, `submittingToAnvil` is only cleared in the formData when the question is no longer rendered, so that the value in formData and the selected radio button are consistent with each other

src/components/forms/forms.js
lu-c: Actually... I don't know, the more I'm thinking about it, the more I'm wondering if this is necessary! Should handling behavior of dependent form components be a job of the global forms functions? The parent formField could handle clearing child form values through the `onChange` option that's already available.

connorlbark:
closing this - redone in https://github.com/DataBiosphere/duos-ui/pull/1800

PR #1773, opened 2022-08-24
[DUOS-1752][risk=no] Step 1 for new data access request application page

lu-c:
- Added new route to the request page (all changes take place there instead) - you can see it here: https://local.broadinstitute.org:3000/dar_application_new/746c463d-75d9-4288-ba1d-6e69b0c08e59
- Mostly updates form components into our new ones, and then restyles some things.

----

- Verify all tests go green


src/components/forms/forms.js
```diff
@@ -285,29 +279,55 @@ const formInputSlider = (config) => {
...
+
+  return radioOptions.map(({value, displayText}) => {
+    return RadioButton({
+      key: `${id}-${value}`,
+      value,
+      defaultChecked: formValue === value,
+      onClick: (event) => onFormInputChange(config, value),
+      description: displayText,
+      style: { fontFamily: 'Montserrat', color: '#1f3b50'}
+    })
```
rushtong, line 295: Codacy has a fix for the issue: [Trailing spaces not allowed.](https://app.codacy.com/gh/DataBiosphere/duos-ui/pullRequest?prid=10042654)
```suggestion
    })
```

src/components/forms/forms.js
```diff
@@ -285,29 +279,55 @@
...
```
rushtong, line 295: Codacy has a fix for the issue: [Missing semicolon.](https://app.codacy.com/gh/DataBiosphere/duos-ui/pullRequest?prid=10042654)
```suggestion
    });  
```

src/components/forms/forms.js
```diff
@@ -285,29 +279,55 @@
...
+
+  return radioOptions.map(({value, displayText}) => {
+    return RadioButton({
+      key: `${id}-${value}`,
+      value,
+      defaultChecked: formValue === value,
+      onClick: (event) => onFormInputChange(config, value),
+      description: displayText,
+      style: { fontFamily: 'Montserrat', color: '#1f3b50'}
+    });
```
rushtong, line 295: Codacy has a fix for the issue: [Trailing spaces not allowed.](https://app.codacy.com/gh/DataBiosphere/duos-ui/pullRequest?prid=10042654)
```suggestion
    });
```

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,641 @@
...
+          name: 'anvil-use-approval-status',
+          checked: option.value === anvilUse,
+          value: option.value,
+          onChange: () => (formFieldChange({name: 'anvilUse', value: option.value}))
+        }),
+        span({ className: 'radio-check'}),
+        span({ className: 'radio-label'}, [option.label])
+      ])
+    )
+  ]);
```
connorlbark: could this use the formfield component that I added?

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,641 @@
...
+                div({className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group'}, [
+                  input({
+                    type: 'checkbox',
+                    id: 'cloud-requested',
+                    name: 'cloudUse',
+                    className: 'checkbox-inline rp-checkbox',
+                    disabled: !isNil(darCode),
+                    required: true,
+                    checked: cloudUse,
+                    onChange: (e) => formFieldChange({name: 'cloudUse', value: e.target.checked})
```
connorlbark: use formfield?

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,641 @@
...
+                  disabled: !isNil(darCode),
+                  onBlur: (e) => formFieldChange({name: 'cloudProviderDescription', value: e.target.value}),
+                  name: 'cloudProviderDescription',
+                  id: 'cloudProviderDescription',
+                  rows: '6',
+                  required: true,
+                  placeholder: 'Please describe the type(s) of cloud computing service(s) you wish to obtain (e.g PaaS, SaaS, IaaS, DaaS)'
+                    + ' and how you plan to use it (them) to carry out the work described in your Research Use Statement (e.g. datasets to be included, process for data transfer)'
+                    + ' analysis, storage, and tools and/or software to be used. Please limit your statement to 2000 characters',
+                  maxLength: 2000
```
connorlbark: migrate this to a formfield?

connorlbark (dismissed):
looks great, just like the mock!!
couple notes:
- selecting a signing official doesn't seem to work
- when you select no on the cloud use statement, the textbook looks like it has the wrong styling

src/pages/dar_application/ResearcherInfo_new.js
lu-c: it could! I didn't want to rewrite it, and figured it would be in soon! XD

src/pages/dar_application/ResearcherInfo_new.js
lu-c: Added a new Textarea formfield component!

src/pages/dar_application/ResearcherInfo_new.js
lu-c: Updated both of these inputs to formfield checkboxes!

JVThomas:
> Opening up a draft to get some opinions!
> 
> I did this UI component first here, but then ... I think I didnt read the description of the ticket thoroughly enough, and I am now at this issue where
> 
> > All logic for these form elements should follow the same logic we have in the current DAR form. Styling should be updated to reflect that of the mocks.
> 
> The logic for the original form is so complex, I'm scared that trying to redo it means I'm going to lose something. I think the proper solution is to trash this effort, and just modify the css / html portions of the original researcher component... However, my concerns with that are that the ticket specifically says to create a new component instead of modify an existing one.
> 
> Not sure what to do next? While this draft is up, I'm going to be migrating these component elements back into the original researcher component. Please please let me know if i shouldn't be doing that tho!
> 
> Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.
> 
> * Label PR with a Jira ticket number and include a link to the ticket
> * Label PR with a security risk modifier [no, low, medium, high]
> * PR describes scope of changes
> * Get a minimum of one thumbs worth of review, preferably two if enough team members are available
> * Get PO sign-off for all non-trivial UI or workflow changes
> * Verify all tests go green
> * Test this change deployed correctly and works on dev environment after deployment

If you think you can keep the existing logic and simply adjust the styling then that's fine. One thing to note though is that the page is now switching to a scrolling model where

1.  The side bar updates the selection as you scroll through the form
2. Clicking on a step on the side bar will cause the form to scroll to that step.

Also keep in mind that the DAR form is in active use on production as it is the primary action for any researcher in the system. Directly changing the existing DAR Form code could compromise its current functionality if redesigns break already working functionality. 

By creating a separate page, route, and sub-components we can avoid breaking a key feature by working on it on dev/local dedicated route. To make that task easier you can make copies of existing components (or build new ones from scratch) and make changes as needed when tackling the UI tickets.

Once the entire UI update is done, it's just a matter of updating the routes to use the new page and removing the old components once we've determined that the new feature works as expected.

It's why I'm not too comfortable with the current changes as is, especially if we're going to merge it against `develop`

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,638 @@
...
+                //   onBlur: (e) => formFieldChange({name: 'cloudProviderDescription', value: e.target.value}),
+                //   name: 'cloudProviderDescription',
+                //   id: 'cloudProviderDescription',
+                //   rows: '6',
+                //   required: true,
+                //   placeholder: 'Please describe the type(s) of cloud computing service(s) you wish to obtain (e.g PaaS, SaaS, IaaS, DaaS)'
+                //     + ' and how you plan to use it (them) to carry out the work described in your Research Use Statement (e.g. datasets to be included, process for data transfer)'
+                //     + ' analysis, storage, and tools and/or software to be used. Please limit your statement to 2000 characters',
+                //   maxLength: 2000
+                // })
```
connorlbark: nit: delete

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,638 @@
...
+                collaborators: internalCollaborators,
+                collaboratorKey: 'internalCollaborators',
+                collaboratorLabel: 'Internal Collaborator',
+                deleteBoolArray: (new Array(internalCollaborators.length).fill(false)),
+                showApproval: false})
+            ])
+          ])
+        ]),
+        div({className: 'form-group'}, [
+          div({className: 'row no-margin'}, [
```
connorlbark: nit: delete these divs?

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,638 @@
...
+                div({className: 'col-lg-12 col-md-12 col-sm-12 col-xs-12 rp-group'}, [
+                  h(FormField, {
+                    id: 'localUse',
+                    disabled: !isNil(darCode),
+                    validators: [FormValidators.REQUIRED],
+                    type: FormFieldTypes.CHECKBOX,
+                    toggleText: 'I am requesting permission to use local computing to carry out the research described in my Research Use Statement',
+                    checked: localUse,
+                    onChange: ({ key: name, value }) => formFieldChange({name, value})
+                  })
```
connorlbark: are threse mutually exclusive? i.e. should these be in a radio group or is it possible to need both local and cloud computing?

lu-c:
closing this pr because it looks like too many changes are necessary

src/pages/dar_application/ResearcherInfo_new.js
lu-c: I think it could be possible to need both! This is the logic from the original request page, so I'd like to keep it the same unless otherwise stated, but this section was also not shown in the mockups

lu-c:
ALRIGHT LET'S DO THIS AGAIN

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,497 @@
...
+              `You must submit `, profileLink, ` and obtain a `, libraryCardLink,
+              ` from your Signing official before you can submit a Data Access Request.`
+            ])
+          })
+        ]),
+
+        h2('Step 1: Researcher Information'),
+
+        div({className: 'dar-application-row'}, [
+          h(FormField, {
```
JVThomas, line 102: You can disable the field here since we're pre-filling the field with the user profile name

src/pages/dar_application/ResearcherInfo_new.js
JVThomas: You can select both. A researcher could theoretically handle some computations locally and more complex calculations on cloud based systems, so the form should facilitate selecting both. The text fields for more info should only apply to cloud computing since it's a third party service relative to us (Broad) and the researcher's institution.

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,497 @@
...
+            validators: [FormValidators.REQUIRED],
+            ariaLevel: ariaLevel + 1,
+            onChange: ({key: name, value,}) => formFieldChange({name, value}),
+            defaultValue: itDirector
+          })
+        ]),
+
+        div({className: 'dar-application-row'}, [
+          div([
+            h(FormField, {
```
JVThomas, line 252: So the choice doesn't visually persist on the component when switching steps or reloading the page. That said, the selection is saved against `formData` which leaves me to believe that the radio button form component isn't handling default values properly

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,497 @@
...
+              ariaLevel: ariaLevel + 2,
+              onChange: ({key: name, value}) => formFieldChange({name, value}),
+              defaultValue: checkCollaborator
+            })
+          ]),
+        ]),
+
+        div({className: 'dar-application-row'}, [
+          h(FormField, {
+            id: `principal_investigator`,
```
JVThomas: The id is snake cased but the component is expecting the defaultValue from a camel cased variable name. So you're saving the variable to formData correctly, it's not reading the init/updated value correctly. It's why when you switch tabs the value goes away. Save and read from the camelCased name to ensure proper state reads/writes.

That said, be aware that the back-end currently doesn't handle `principalInvestigator` (since it's new), so don't be surprised if you don't see the data after saving and reloading.

rushtong: We have `piName` as a value in the current schema: https://github.com/DataBiosphere/consent/blob/develop/src/main/resources/data-submitter-schema_v1.json#L62 

Let's track any modifications necessary to the schema so that we can tackle them as a unit. I lean towards naming that makes the most sense from a UI dev perspective.

src/pages/dar_application/ResearcherInfo_new.js
JVThomas: You can select both. More often than not researchers/labs will leverage third-party services to handle more complex and/or niche operations.

src/Routes.js
```diff
@@ -74,8 +75,12 @@ const Routes = (props) => (
     {/* Order is important for processing links with embedded dataRequestIds */}
     <AuthenticatedRoute path="/dar_application/:dataRequestId" component={DataAccessRequestApplication} props={props}
       rolesAllowed={[USER_ROLES.researcher]} />
+    <AuthenticatedRoute path="/dar_application_new/:dataRequestId" component={DataAccessRequestApplicationNew} props={props}
```
JVThomas: Add a conditional statement to the new routes so that they only show up if `env` is either `dev` or `local`

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,499 @@
...
+                { name: 'no', text: 'No' }
+              ],
+              validators: [FormValidators.REQUIRED],
+              ariaLevel: ariaLevel + 1,
+              orientation: 'horizontal',
+              onChange: ({key: name, value}) => {
+                const normalizedValue = value && value.selected === 'yes';
+                formFieldChange({name, value: normalizedValue});
+              },
+              defaultValue: { selected: anvilUse ? 'yes' : 'no' }
```
connorlbark: by default, this is no when you open the page. however, to get the conditional fields underneath it to render, you need to actually click no. I think by default the radio shouldn't be checked at all (i.e. value should be null by default).

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,499 @@
...
+              }
+            },
+          }),
+        ]),
+
+        div({className: 'dar-application-row'}, [
+          h(FormField, {
+            id: 'itDirector',
+            description: 'I certify that the individual listed below is my IT Director',
+            placeholder: 'IT Director',
```
connorlbark: Specify if this should be email, firstname lastname, etc?

connorlbark (commented):
couple minor things! but looking very good!!

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -0,0 +1,499 @@
...
+                    disabled: !isNil(darCode),
+                    validators: [FormValidators.REQUIRED],
+                    type: FormFieldTypes.CHECKBOX,
+                    toggleText: 'I am requesting permission to use cloud computing to carry out the research described in my Research Use Statement',
+                    defaultValue: cloudUse,
+                    ariaLevel: ariaLevel + 2,
+                    onChange: ({ key: name, value }) => formFieldChange({name, value})
+                  })
+                ])
+              ]),
```
connorlbark: should this be below the following checkbox? It's a little bit confusing that the conditional fields are separated from what causes them to render. I think it would make sense to move this one down or the conditional fields up.

JVThomas: Ah, good point (and something that I should've caught/adjusted with the old form as well). For this ticket the easiest thing to do is to render the local checkbox first, then the cloud computing. This way the cloud fields are, relatively speaking, grouped together, while keeping the checkbox options and input fields separate.

src/pages/dar_application/ResearcherInfo_new.js
lu-c: updated this to work, i think?

src/pages/dar_application/ResearcherInfo_new.js
lu-c: I just moved the checkboxes into the same parent group - But I dont know, I feel like it makes more sense to keep the entire radio group together than to spread it out (with the conditional renders between).

JVThomas: I should've been more clear (sorry about that). I was just asking to switch the order of the checkboxes so that local is first and cloud computing is second. This way, when you do click on cloud computing, the text inputs to populate cloud computing information will render below it, as opposed to the current situation where you'll have awkward form placement where the cloud form fields show up below the local radio option 

See pic from the old DAR form of the scenario I want to avoid

Screen Shot 2022-09-12 at 1 56 44 PM

JVThomas:
Currently you cannot enter spaces on any of the fields for the Cloud Provider inputs (name, type, textbox). I'm hitting the space bar but there's no spaces being entered.

My guess is that the `normalizeValue` function in `formComponents.js` is trimming the spaces off as the user types it in, which is going to happen since the function is firing off `onChange`. 

So if the user wants to type "Google Cloud Platform", and has currently typed "Google " (notice the space), the onChange will fire the normalize function which then trims the currently trailing space.

One thing you can do is to normalize the value when the input loses focus (onBlur). This way you know for sure that the user has finished typing and can safely process the string.

lu-c:
> Currently you cannot enter spaces on any of the fields for the Cloud Provider inputs (name, type, textbox). I'm hitting the space bar but there's no spaces being entered.
> 
> My guess is that the `normalizeValue` function in `formComponents.js` is trimming the spaces off as the user types it in, which is going to happen since the function is firing off `onChange`.
> 
> So if the user wants to type "Google Cloud Platform", and has currently typed "Google " (notice the space), the onChange will fire the normalize function which then trims the currently trailing space.
> 
> One thing you can do is to normalize the value when the input loses focus (onBlur). This way you know for sure that the user has finished typing and can safely process the string.

Oof, I see that problem too, but I'm not entirely convinced it's the same issue. The other FormField components are updating onChange as well, but the "Principal Investigator" input is still allowing spaces. There must be something different in the set up that I'm missing, but I will take a look at that right now!

JVThomas (approved):
With the latest commit inputs are now being rendered properly and the values are being saved (exception of Principal Investigator, which is a new field and can be adjusted to have its variableName match back-end expectations).

quazi-h (approved):
The new component for Step 1 of the DAR process looks great! Tested locally and all the functioning input fields are working well and functioning as expected.

PR #1774, opened 2022-08-25
[DUOS-1752][risk=no] Updated forms with common components

lu-c:
Tied to ticket only because a lot of these changes were necessary for that ticket.

- Reorganized css
- Added select_creatable with the intent of replacing the existing select... But I didn't want to get rid of the existing select yet since I think other components are using it. The select_creatable feels like a more powerful approach to the select dropdown with an input field for searching, their documentation has a lot of customization options available! It's being used in other places in the application already, so I'm thinking this is probably a good shot? At the very least, the options allow us to get the error state (for required) to function more similarly to the regular form text input
- 

----

- Verify all tests go green


connorlbark:
love the work here!!! i was thinking of ways to make formfield easier to use. one of those things i think would be to separate out the `FormField` component into its own file, where the file contains: documentation, type enum, and the component itself. Then the guts of how that component works is in a `formTypes.js` (or similar) file. The reason is that I think  separating the guts of how the component works would make it easier for the end developer to figure out what props to pass (based on only documentation) and doesn't have to trudge through a bunch of code to find that documentation. kind of minor but yeah just a thought!

also: big fan of using react-select!! for my ticket i also needed one which was not creatable; maybe you could pass in a `creatable` prop and if false it creates a normal react-select component? Same with `multi`

lu-c:
> love the work here!!! i was thinking of ways to make formfield easier to use. one of those things i think would be to separate out the `FormField` component into its own file, where the file contains: documentation, type enum, and the component itself. Then the guts of how that component works is in a `formTypes.js` (or similar) file. The reason is that I think separating the guts of how the component works would make it easier for the end developer to figure out what props to pass (based on only documentation) and doesn't have to trudge through a bunch of code to find that documentation. kind of minor but yeah just a thought!
> 
> also: big fan of using react-select!! for my ticket i also needed one which was not creatable; maybe you could pass in a `creatable` prop and if false it creates a normal react-select component? Same with `multi`

Okay, we can do that! Do you think you could make a pr with your radio button / select changes too? It would be nice to work it all together before we continue on building forms on top of them :D

connorlbark:
> Okay, we can do that! Do you think you could make a pr with your radio button / select changes too? It would be nice to work it all together before we continue on building forms on top of them :D

can do!!

src/components/RadioButton.js
```diff
@@ -7,7 +7,6 @@ export const RadioButton = (props) => {
     fontSize: 15,
     lineHeight: '2rem',
     color: 333,
-    fontFamily: '\'Roboto\', sans-serif',
```
rushtong, line 10: Thank you!

src/components/forms/formComponents.js
```diff
@@ -0,0 +1,387 @@
...
+//---------------------------------------------
+// Form Controls
+//---------------------------------------------
+export const formInputGeneric = (config) => {
+  const {
+    id, title, type, disabled,
+    placeholder,
+    inputStyle, ariaDescribedby,
+    formValue, error, setError
+  } = config;
```
rushtong, line 72: Is there no way to enforce values for config? `id` for example is required, but `type` and `placeholder` is not. Also, there is a one-of condition for `placeholder` and `title`. Perhaps some default values might be necessary? I'm not opposed to this as is, but I do think there will be hard-to-debug bugs when devs forget to add all the right configs for the specific form entry they're using.

rushtong (approved):
Looks like a nice re-organization. I am concerned about `configs` being easily broken, but without something more comprehensive, like form-element-specific configs, it might not be a reasonable lift to implement.

src/components/forms/formComponents.js
```diff
@@ -0,0 +1,387 @@
...
+    onChange: (option) => {
+      if (isStringArr) {
+        if (isMulti) {
+          // string result, multiple options
+          onFormInputChange(config, option.map((o) => o.displayValue));
+          setFormValue(option);
+          return;
+        }
+        // string result, only one option
+        onFormInputChange(config, option.displayValue);
```
shaemarks: When I click the 'x' button, onChange is triggered and I receive this error:
Screenshot (105)

src/components/forms/formComponents.js
```diff
@@ -0,0 +1,387 @@
...
+      disabled
+    }),
+    label({
+      className: `regular-checkbox ${error ? 'errored' : ''}`,
+      htmlFor: `cb_${id}_${toggleText}`,
+    }, [toggleText])
+  ]);
+};
+
+export const formInputSlider = (config) => {
```
shaemarks, line 364: Not necessarily a fix for this PR, but the public visibility input slider has a default of true, but onChange isn't being called until the user interacts with the slider (so it is not initially being set in the formData). Since this is a required boolean value, I would expect the backend to store it as false if no value is sent over in the formData json

src/components/forms/formComponents.js
connorlbark: possibly, we could have this information in the FormField type declaration: 
```js
SELECT: {
  ...,
  requiredProps: ['id'],
  customPropValidation: (props) => { 
    if (!isNil(props.title) && !isNil(props.placeholder)) {
      throw 'both title and placeholder cannot be specified';
    }
  }
}
```

thoughts?

src/components/forms/formComponents.js
```diff
@@ -0,0 +1,387 @@
...
+            isRendered: !disabled
+          })
+        ]);
+      })
+    )
+  ]);
+};
+
+// Using react-select/creatable - Passing config directly through!
+export const formInputSelect = (config) => {
```
shaemarks, line 181: For some reason, when I select a File Type, I'm not seeing the text displayed (but onChange is being called and formValue is being set)

shaemarks (approved):
Thanks for all this really useful work! See comments inline for a couple of bugs I found, but I know that this is an important PR for unblocking other tickets, and most of this PR is reorganizing existing code, so I understand if it would be more helpful to have the bug fixes as separate tickets 

Note: on `DataSubmissionForm.js`, we're using` lodash/fp set` in `onChange`, which doesn't mutate the object, so the changes aren't sticking. We can fix this by either using the regular `lodash set` or doing` formData = set(key, value, formData)` with `lodash/fp` (tested by adding `console.log(formData)` to `onChange`)

src/components/forms/formComponents.js
rushtong: Something like that should work, but maybe that's a little more than this PR is intended to do? Might be a good follow-on ticket.

rushtong:
> Note: on `DataSubmissionForm.js`, we're using` lodash/fp set` in `onChange`, which doesn't mutate the object, so the changes aren't sticking. We can fix this by either using the regular `lodash set` or doing` formData = set(key, value, formData)` with `lodash/fp` (tested by adding `console.log(formData)` to `onChange`)

I believe we're using the pattern `Object.assign(obj1, obj2, ...)` for cases where we intentionally want to mutate objects.

src/components/forms/formComponents.js
lu-c: That's a really great suggestion, Conor! Would like to add that in the next update :D

connorlbark:
recent changes look good - do we want to make a new ticket for the follow on stuff and then merge this for now?

edit: oop it looks like you're still working on stuff - lmk when you're finished and i will rereview.

src/components/forms/formComponents.js
lu-c: Good to know! The mockups show this as a default true value (even tho it doesnt match the back end), maybe when we define the form data, that's where we should set it instead!

src/components/forms/formComponents.js
lu-c: Fixed! Thank you!

PR #1776, opened 2022-08-29
[DUOS-1752][risk=no] Common radio and select forms

connorlbark:
I cleaned up the way that the select component worked (I personally think it's cleaner, but let me know what you think!). I also added the radio group component, with the ability to have a textbox input..

----

- Verify all tests go green


cypress/component/Forms/forms.spec.js
```diff
@@ -314,7 +423,7 @@ describe('FormField - Tests', () => {
           'Analytical', 'Prospective', 'Retrospective',
           'Case report', 'Case series', 'Cross-sectional',
           'Cohort study'
-        ]
+        ].map((opt) => {return {displayName: opt, displayValue: opt};})
```
lu-c, line 424: I kind of want to keep this version without requiring the developer to provide the object, and allow them to provide just the string, since I think it's probably going to be the most common way to do this

src/components/forms/formComponents.js
```diff
@@ -247,9 +236,99 @@ export const formInputCreatable = (config) => {
...
+            }
+          }, [
+            h(RadioButton, {
+              id: id+'_'+optionId,
+              name: id+'_'+optionId,
+              key: idx,
+              defaultChecked: !isNil(formValue) && formValue.selected === option.name,
+              onClick: () => onFormInputChange(config, {
+                selected: option.name,
+                value: (option.type === 'string'? '' : true),
```
lu-c: I think the boolean here is throwing me off a little - when would a radio button be a boolean value? I feel like then that means we are using the wrong form element, because it should be either a checkbox or a slider.

UX advantage of the slider/checkbox would be that the user doesnt have to move their mouse to change their selection. Accessibility for keyboard-only users would not have to tab between different items, and it's also easier for teh screen reader to display the status of the selection.

I think this is maybe something to consider! Maybe it would be worth just making the radiogroup values always a string? And then if there is a fallout boolean effect, we can pass through an onchange method that would update a separate state variable with the true/false.

src/components/forms/formComponents.js
```diff
@@ -247,9 +236,99 @@ export const formInputCreatable = (config) => {
...
+              style: {
+                fontFamily: 'Montserrat',
+                fontSize: '14px',
+              },
+              description: option.text,
+              disabled,
+            }),
+            input({
+              isRendered: option.type === 'string' && formValue.selected === option.name,
+              id: id+'_'+optionId+'_text_input',
```
lu-c:
```suggestion
              id: `${id}_${optionId}_text_input`,
```

src/components/forms/formComponents.js
```diff
@@ -247,9 +236,99 @@ export const formInputCreatable = (config) => {
...
+        {
+          className: 'radio-group',
+          id: id,
+        },
+        options.map(((option, idx) => {
+          const optionId = (!isNil(option.id) ? option.id : option.name);
+
+          return div({
+            style: {
+              key: idx,
```
lu-c: move `key` outside of the `style` tag

src/components/forms/formComponents.js
```diff
@@ -247,9 +236,99 @@ export const formInputCreatable = (config) => {
...
+              margin: '1.0rem 0 0.5rem 0',
```
lu-c: Can we add these styles to the forms.css instead?

src/components/forms/formComponents.js
```diff
@@ -247,9 +236,99 @@ export const formInputCreatable = (config) => {
...
+            }
+          }, [
+            h(RadioButton, {
+              id: id+'_'+optionId,
+              name: id+'_'+optionId,
```
lu-c:
```suggestion
              id: `${id}_${optionId}`,
              name: `${id}_${optionId}`,
```

src/components/forms/formComponents.js
```diff
@@ -175,59 +177,46 @@ export const formInputMultiText = (config) => {
...
 };
 
+// Using react-select/creatable - Passing config directly through!
 export const formInputSelect = (config) => {
   const {
-    id, title, disabled, error, setError,
+    id, title, disabled, required, error, setError,
     selectOptions, placeholder, ariaDescribedby,
-    creatableConfig
+    formValue, creatable, isMulti, setFormValue,
```
lu-c:
```suggestion
    formValue, setFormValue,
    isCreatable, isMulti,
```
Reorganizing the inputs so similar things are together
Renamed creatable -> isCreatable so it's easier to tell the value should be a bool

lu-c: (deleted)
Jusssst saw the selectConfig, nm

src/components/forms/formComponents.js
```diff
@@ -175,59 +177,46 @@ export const formInputMultiText = (config) => {
...
 
   const isStringArr = isString(selectOptions[0]);
   const normalizedOptions = isStringArr
     ? selectOptions.map((option) => { return {key: option, displayValue: option }; })
     : selectOptions;
 
-  return h(Creatable, {
+  return h(component, {
     key: id,
+    id: id,
```
lu-c:
```suggestion
    id,
```

cypress/component/Forms/forms.spec.js
connorlbark: oh yeah sorry it does work that way i just added this before i added that feature !

src/components/forms/formComponents.js
connorlbark: ok yeah so i understand how this is confusing. i mean the the OPTION itself is boolean, but the radio group isn't necessarily boolean. if that makes sense. so, for example, in a group of buttons, you might have:

- Option A (boolean) <- A is either selected or not selected, thus boolean
- Option B (boolean) <- B is either selected or not selected, thus boolean
- Other (string) <- Other must be specified, thus string

The string option adds a textbox once clicked.

So, if you click A, it will report that A was set true. If you select Other it will report that Other was set to an empty string. Like a normal radio group, these are mutually exclusive values.

Again, not sure if that's the best naming convention, but that was my logic behind it!

connorlbark: One thing we could do is only have the `value` field when the option is a string. That might clarify your confusion.

So, if A selected: `{selected:'Option A'}`, if Other selected: `{selected: 'Other', value: ''}`. Thoughts?

lu-c: I like how the last way works `{selected: "option A" / "Other"}`! That's how i'd expect it to work vs the booleans, since the selected value should be updating a single field, not multiple

connorlbark: ok! great. makes sense to me too!

lu-c (approved):

PR #1781, opened 2022-08-31
[DUOS-2020][risk=no] Show Match Failure Reasons in the DAC voting screen

quazi-h:
[DUOS-2020](https://broadworkbench.atlassian.net/browse/DUOS-2020)

- Added a div for the failureReasons in the CollectionAlgorithmDecision.
- Updated algorithmResult in DarCollectionUtils to include failureReasons.

To test, look at the Chair Vote tab for dar_collection 493.
This collection has multiple failure reasons:
Screen Shot 2022-08-31 at 8 39 26 AM
I was trying to format the string somehow - using ".join('\n') directly in the span worked but caused errors for collections that did not have a failure reason. I may try adding a formatMethod in utils.js specifically for a list of strings.

Any suggestions would be helpful.
Also need to figure out a good way to test this.

----

- Label PR with a JIRA ticket number and include a link to the ticket
- Verify all tests go green

### DIFF
--- a/src/components/CollectionAlgorithmDecision.js
+++ b/src/components/CollectionAlgorithmDecision.js
@@ -4,7 +4,7 @@ import{ isNil } from 'lodash/fp';
 
 export default function CollectionAlgorithmDecision(props) {
   const {algorithmResult = {}, styleOverride = {}} = props;
-  const { createDate, id, result } = algorithmResult;
+  const { createDate, id, result, failureReasons} = algorithmResult;
 
   const containerProps = {
     id: `collection-algorithm-id-${id}`,
@@ -30,6 +30,10 @@ export default function CollectionAlgorithmDecision(props) {
         div({style: {fontSize: '1.5rem'}}, [
           span({id: `collection-${id}-date-label`, style: {paddingRight: '1%'}}, ['Date:']),
           span({id: `collection-${id}-date-value`, style: {fontWeight: 400}}, [!isNil(createDate) ? formatDate(createDate) : 'N/A'])
+        ]),
+        div({style: {fontSize: '1.5rem'}}, [
+          span({id: `collection-${id}-reason-label`, style: {paddingRight: '1%'}}, ['Reason:']),
+          span({id: `collection-${id}-reason-value`, style: {fontWeight: 400}}, [failureReasons.join('\n') || 'N/A'])
         ])
       ])
     ])
--- a/src/components/CollectionAlgorithmDecision.js
+++ b/src/components/CollectionAlgorithmDecision.js
@@ -34,7 +34,7 @@ export default function CollectionAlgorithmDecision(props) {
         div({style: {fontSize: '1.5rem'}}, [
           span({id: `collection-${id}-reason-label`, style: {paddingRight: '1%'}}, ['Reason:']),
           span({id: `collection-${id}-reason-value`, style: {fontWeight: 400}},
-            [!isEmpty(failureReasons) ? failureReasons.map((r) => p([r])) : 'N/A'])
+            [!isEmpty(failureReasons) ? failureReasons.map((r, idx) => p({key: idx}, [r])) : 'N/A'])
         ])
       ])
     ])
--- a/src/components/CollectionAlgorithmDecision.js
+++ b/src/components/CollectionAlgorithmDecision.js
@@ -33,7 +33,7 @@ export default function CollectionAlgorithmDecision(props) {
         ]),
         div({style: {fontSize: '1.5rem'}}, [
           span({id: `collection-${id}-reason-label`, style: {paddingRight: '1%'}}, ['Reason:']),
-          span({id: `collection-${id}-reason-value`, style: {fontWeight: 400}}, [!isNil(failureReasons) ? failureReasons : 'N/A'])
+          span({id: `collection-${id}-reason-value`, style: {fontWeight: 400}}, [!isNil(failureReasons) ? failureReasons.join(' ') : 'N/A'])
         ])
       ])
     ])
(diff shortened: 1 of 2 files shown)


connorlbark:
maybe you could do something like:

```js
failureReasons.map((r, idx) => span({key: idx}, [r]))
```

(i think you need the key, not sure)

That way you can do more complex styling so it looks more natural. Maybe if you want newlines, you could use `p` instead of `span`.

quazi-h:
> maybe you could do something like:
> 
> ```js
> failureReasons.map((r, idx) => span({key: idx}, [r]))
> ```
> 
> (i think you need the key, not sure)
> 
> That way you can do more complex styling so it looks more natural. Maybe if you want newlines, you could use `p` instead of `span`.

That was very helpful, thank you! I was able to implement your suggestion and it worked without the key.

connorlbark:
Do we want to separate them by newlines? It looks a little cluttered. I am ok w spaces though - if that's preferred then I will approve !

quazi-h:
> Do we want to separate them by newlines? It looks a little cluttered. I am ok w spaces though - if that's preferred then I will approve !

I agree that newlines make it look less cluttered. When playing around and switching between using p and span, I came up with these 2 options:

`failureReasons.map((r) => p([r]))`
using p

`failureReasons.map((r) => span([r + '\n']))`
using span + ' n'

Using either the p or span + '\n' gets the individual reasons on their own line. Is either of these preferable or within convention?

connorlbark:
> `failureReasons.map((r) => p([r]))`

I like this one (the first one)!

src/components/CollectionAlgorithmDecision.js
```diff
@@ -30,6 +30,11 @@ export default function CollectionAlgorithmDecision(props) {
         div({style: {fontSize: '1.5rem'}}, [
           span({id: `collection-${id}-date-label`, style: {paddingRight: '1%'}}, ['Date:']),
           span({id: `collection-${id}-date-value`, style: {fontWeight: 400}}, [!isNil(createDate) ? formatDate(createDate) : 'N/A'])
+        ]),
+        div({style: {fontSize: '1.5rem'}}, [
+          span({id: `collection-${id}-reason-label`, style: {paddingRight: '1%'}}, ['Reason:']),
```
lu-c, line 35: Comment with no action requested: I think padding with percentages are kind of risky and leave the code vulnerable to responsiveness problems in the future, personally. I see that this line of code is following the convention in this file, so I .. am not asking that you update it (or all of it), but just wanted to try pointing out in case others agree.

And if they do, i can help updating these paddings in a future pr?

src/components/CollectionAlgorithmDecision.js
```diff
@@ -30,6 +30,11 @@ export default function CollectionAlgorithmDecision(props) {
...
+          span({id: `collection-${id}-reason-value`, style: {fontWeight: 400}},
+            [!isNil(failureReasons) ? failureReasons.map((r) => span([r + ' '])) : 'N/A'])
```
lu-c: what would you think of replacing `isNil` with `isArray`? Might be safer since failureReasons then calls an array function.

Also, consider adding a length check? If failureReasons == [], it would render nothing, instead of 'n/a'.

I think with the map, you'll also have to add the { key: r } to prevent react's key rendering issue as well

src/components/CollectionAlgorithmDecision.js
rushtong: > And if they do, i can help updating these paddings in a future pr?

I agree Christine, please consider this task on the menu.

src/components/CollectionAlgorithmDecision.js
lu-c:
```suggestion
            !isArray(failureReasons) && failureReasons.length > 0 ? failureReasons.map((r) => span({ key: r }, [r + ' '])) : ['N/A']
```
like this?

connorlbark: it would be just `isArray(failureReasons)` right, instead of `!isArray(...)`?

lu-c: oh yes! my brain 2 slow! Thank you!

src/components/CollectionAlgorithmDecision.js
JVThomas: ~~Percentage and `em/rem` are the de-facto measurements for responsive sizing though since the measurements are always relative to the parent container. Pixels are an absolute measurement and do not scale with screens of varying resolutions (due to differences in a device's pixel density), so you're forced to designate media breakpoints in CSS.~~

Edit: Just realized you were mentioning percentages for padding specifically, scratch everything above

src/components/CollectionAlgorithmDecision.js
JVThomas: You can use lodash's `isEmpty` instead of `isArray`, which do a nil check and length check on any collection-like object.

JVThomas: Same as above, use lodash's `isEmpty` method to do the nil and empty check for you

quazi-h: Implemented isEmpty to do both checks, thank you!

quazi-h: Implemented isEmpty to do both checks.

lu-c (approved):
Looks good!

src/components/CollectionAlgorithmDecision.js
quazi-h: Created a new ticket to address this: [DUOS-2081](https://broadworkbench.atlassian.net/browse/DUOS-2081)

connorlbark (approved):
👍

src/components/CollectionAlgorithmDecision.js
quazi-h: Adding key to resolve the React warning in the console for trying to call map without a key:
`Warning: Each child in a list should have a unique "key" prop.
Check the render method of `CollectionAlgorithmDecision`. See https://reactjs.org/link/warning-keys for more information.
    at p
    at CollectionAlgorithmDecision (https://local.broadinstitute.org:3000/main.fe753e4fe3cd895880c3.hot-update.js:25:37)`
    Thanks @lu-c!

PR #1787, opened 2022-09-02
[DUOS-1759][risk=no] Updated step selector for data access request page

lu-c:
Lots of ugliness in the code to accomodate keeping the existing page as it is!

----

- Verify all tests go green


src/pages/DataAccessRequestApplication.js
```diff
@@ -952,7 +960,57 @@ class DataAccessRequestApplication extends Component {
...
+              orientation: 'vertical',
+              TabIndicatorProps: {
+                style: { background: '#2BBD9B' }
+              },
+              onChange: (event, step) => {
+                this.goToStep(step);
+              }
+            }, [
+              h(Tab, {
+                // key: `dar-request-step-1`,
```
connorlbark: is this comment needed? if not, maybe delete.

connorlbark (approved):
the tabs themselves look good!

lu-c:
sorry, didnt read the ticket thoroughly enough, will continue work on the behavior of this today (this pr just focused on the UI)

lu-c:
IT'S READY AGAIN, lots of changes this time - read the ticket more closely, this is lookin' pretty sweet ;)

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -215,6 +230,8 @@ class DataAccessRequestApplicationNew extends Component {
       prev.formData = merge(prev.formData, formData);
       return prev;
     });
+
+    window.addEventListener('scroll', this.onScroll);
```
lu-c: Shoot, i didnt realize codacy wouldnt like this... I don't know how to do this part of the ticket without adding an event listener though. :[ Let me brainstorm

rushtong: If codacy is giving us false positives here, we can change that. I'm not sure what the best practice is, but codacy shouldn't be the primary roadblock.

lu-c: TIL you can bypass codacy warnings by adding disabling eslint on the same lines - would this be okay? XD I guess I'll let the reviewers weigh in hehe

connorlbark (approved):
This is really cool! It's very satisfying! There's some weird behavior around step 4 on my laptop (seems to prefer staying on 3 or 5) but clicking each of them works as expected and this is likely just some weirdness with heights and such.

src/pages/DataAccessRequestApplication_new.js
JVThomas: There's nothing wrong with your function. It's a textbook definition of a debounce with a reset when the function is called repeatedly. 

Codacy's issue is that it's not context aware of its warnings. Generally `setTimeout` is frowned upon for anything regarding data since defining a fixed time delay on async operations (where time to completion is indeterminate) can be dangerous. In this context though the UI changes are independent of anything else going on in the component, so it's not a problem.

If you care about Codacy warnings, try looking into `lodash's` (vanilla, not fp) `debounce` function and see if that does anything (though odds are it'll be doing the same things under the hood). 

That said, I have no idea about `eventListener` warnings. If Codacy is giving you warnings about `onScroll` then you can ignore them.

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -1062,7 +1094,12 @@ class DataAccessRequestApplicationNew extends Component {
               })
             ]),
 
-            div({ isRendered: this.state.step === 4 }, [
+            div({id: 'step-4', className: 'step-container'}, [
```
JVThomas: You don't have to worry about the Progress Report component right now. The mocks have it set up, but we won't be implementing the feature on the back-end for a while due to Data Submitter and the DAR Form UI redesign being a higher priority. 

The funky behavior in the scroll is probably because the div is so small that the typical user's scroll movement will shift this component in and out of your designated windowTop threshold during the debounce period.

JVThomas (approved):
Overall the code looks good. I've left two comments (the Codacy related one is optional) for you to look over, but the left navbar is working smoothly.

lu-c:
> This is really cool! It's very satisfying! There's some weird behavior around step 4 on my laptop (seems to prefer staying on 3 or 5) but clicking each of them works as expected and this is likely just some weirdness with heights and such.

I think this is because step 4 is literally like 30 pixels large or something, so it's really hard to catch - when this component is fleshed out and implemented, we won't see this issue anymore! :)

src/pages/DataAccessRequestApplication_new.js
lu-c: ok i will remove this for now

PR #1791, opened 2022-09-07
[DUOS-1964][risk=no] Add DataAccessGovernance section to Data Submission Form 

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1964

Replaced the old one (closed that PR) because it had so many conflicts that it was easier to make a new branch from develop.

----

- Verify all tests go green


src/components/data_submission/DataAccessGovernance.js
```diff
@@ -0,0 +1,168 @@
...
+
+  const startEditConsentGroup = (idx) => {
+    const newConsentGroupsState = cloneDeep(consentGroupsState);
+    newConsentGroupsState[idx].editMode = true;
+    setConsentGroupsState(newConsentGroupsState);
+  };
+
+  return div({
+    style: {
+      padding: '50px 0',
```
quazi-h: I don't have access to the invision mock so I'm not sure if it's intentional but the space between the DataSubmissionStudyInformation and DataAccessGovernance sub-components looks kinda large to me. I think you'd have to change this padding and the padding in [ds_study_information](https://github.com/DataBiosphere/duos-ui/blob/6e5a46d32d51cf9e974aa9657d621d4212779318/src/components/data_submission/ds_study_information.js#L34) to make that gap smaller. This is purely a nit based on my own opinion so please let me know if I am way off!

src/components/forms/formComponents.js
```diff
@@ -341,23 +359,57 @@ export const formInputRadioGroup = (config) => {
...
     input({
       type: 'checkbox',
-      id: `cb_${id}_${toggleText}`,
-      checked: formValue,
+      id: id,
+      checked: (valueType === 'string' ? isString(formValue) : formValue),
       className: 'checkbox-inline',
       'aria-describedby': ariaDescribedby,
-      onChange: (event) => onFormInputChange(config, event.target.checked),
+      onChange: (event) => {
```
quazi-h: On my local instance I wasn't actually able to check any of the options under **Consent Secondary Data Use Terms**. I was digging around and saw that the EditConsentGroup uses FormFieldTypes.CHECKBOX which uses this formInputCheckbox components but I think that we may have it disabled for now, is that correct?

src/components/forms/formComponents.js
```diff
@@ -45,11 +49,13 @@ const normalizeValue = (value) => {
 };
 
 const onFormInputChange = (config, value) => {
```
quazi-h, line 42: I'm getting a TypeError on the data_submission_form page when I get to the Data Access Governance, Data Access Committee section and I choose a DAC from the dropdown and then use the "X" to cancel my selection. See screenshot of error and stack trace:
TypeError
TypeError Stack Trace

src/components/data_submission/consent_group/EditConsentGroup.js
```diff
@@ -0,0 +1,249 @@
...
+            id: 'hmb',
+            name: 'hmb',
+            text: 'Health/Medical/Biomedical Research Use',
+          },
+          {
+            id: 'diseaseSpecificUse',
+            name: 'diseaseSpecificUse',
+            text: 'Disease-Specific Research Use',
+            type: 'string',
+            placeholder: 'Please enter one or more diseases',
```
JVThomas: This field should function like the field on the DAR form or `DatasetRegistration.js, line 1129` in the sense that an autofill list should be popping up as you type in the input field.

src/components/forms/formComponents.js
connorlbark: fixed!!

connorlbark:
I am not sure why npm audit is failing (bootstrap requirement didn't change?) but this should be ready for re-review

Changes:
- Search for diseases in ontology 
- More idiomatic / cleaner form field usage
- Date validation + field for date 

The select field has become more complex. It allows both asynchronous and synchronous usage. This could be split into a `SELECT` and an `ASYNCSELECT` but there was a lot of overlapping functionality, so I preferred combining them.

src/components/data_submission/consent_group/EditConsentGroup.js
connorlbark: copied over functionality from the DAR form. if the user selected 'disease 1' and 'disease 1' the form will have the string: `disease 1, disease 2`, since the schema just has a string there. however, if we wanted to store the diseases as a list of strings, that would be an easy fix

src/components/data_submission/DataAccessGovernance.js
connorlbark: updated styling to be a bit more consistent / smaller spacing!

src/components/data_submission/consent_group/EditConsentGroup.js
rushtong: The schema is expecting an array of disease strings: https://github.com/DataBiosphere/consent/blob/6ec50f6b9ddc1649e17b1f0522107eea3084dddb/src/main/resources/dataset-registration-schema_v1.json#L305

connorlbark: oh great I will make that change. makes the code easier actually

connorlbark: done, thank you for the quick catch !

quazi-h:
Looking good - I can now select items under "Secondary Data Use Terms".
With the new changes however, section **1.1 Data Custodian** is grayed out and I can't enter anything into the form field. I'm also unable to enter anything for section **1.10 Data Access Committee** and the drop down doesn't have any DACs listed. If I remember correctly this was working before.

connorlbark:
> Looking good - I can now select items under "Secondary Data Use Terms". With the new changes however, section **1.1 Data Custodian** is grayed out and I can't enter anything into the form field. I'm also unable to enter anything for section **1.10 Data Access Committee** and the drop down doesn't have any DACs listed. If I remember correctly this was working before.

Hmm, curious. The data access committee dropdown works fine locally. Is it possible there is an error on you hitting the DAC endpoint on consent?

Also, I believe the data submitter (I assume you mean submitter? the custodian field also works for me locally) field is supposed to be greyed out - it's prefilled with the current user's information.

quazi-h (approved):
Took another look and everything on the data_submission_form page is working correctly. Sorry, I was mistakenly looking at the dataset_registration page (Researcher Console -> Dataset Catalog -> Add Dataset) when I was reviewing it last.

src/components/data_submission/consent_group/EditConsentGroup.js
```diff
@@ -0,0 +1,385 @@
...
+      placeholder: 'Data Location(s)',
+      validators: [FormValidators.REQUIRED],
+      onChange,
+    }),
+
+    h(FormField, {
+      id: idx+'_url',
+      name: 'url',
+      title: 'Data URL',
+      validators: [FormValidators.REQUIRED, FormValidators.URL],
```
JVThomas, line 379: Minor issue on the URL Form validator. `www.` isn't always representative of a valid url. For example, `www.duos.org` or `https://www.duos.org` are invalid urls, but `duos.org` or `https://duos.org` will get us to the DUOS production page.

connorlbark: Wait so what is the desired behavior here?

`duos.org` should pass, and `www.duos.org` should not?

JVThomas: No, `www.` shouldn't be a requirement at all. You can try this out with the DUOS example I listed. Typing in `www.duos.org` will give you a 'Site can't be reached' but `duos.org` will get you to our production site.

`www` is not a hard requirement for websites. It's up to the domain administrator to determine whether or not they want their website accessible by pre-prending `www` or not.

The very first website (which still works) also operates under the same premise. Pre-prending it with `www.` will not work
http://info.cern.ch/

connorlbark: `https://duos.org` should work; I am currently using javascript's inbuilt `new URL()` functionality to test this. I believe the only catch is that you have to specify `http://` or `https://` for it to parse properly. If we want `duos.org` to pass I can change the functionality to assume `https://` schema always.

JVThomas: It's not a matter of functionality, its the issue of User Experience. 

The average user does not type in `https://www.google.com`, they type in `google.com`. From there the browser executes the request under `https` for you, and DNS will redirect you to `www.google.com`.

Now imagine I'm a researcher for DUOS. I type in `duos.org` because that's what I always do when I visit that site.

Screen Shot 2022-09-23 at 8 18 57 AM

Validator is now telling me that the url is invalid and provides me `https://www.google.com` as a valid url. Me, the user, will then input `https://www.duos.org` to match the example given. I've seen `http` and `www` before, so I would assume that its a matter of providing the full URL rather than the shorthand that I use in my day to day.

Screen Shot 2022-09-23 at 8 20 45 AM

Great, the validator isn't complaining anymore. The problem is the URL is wrong and any user who views that url on the UI will copy and paste that link on their browser only to find themselves at a page that does not exist.

The application essentially prompted the user to input an invalid URL in the field, meaning we're potentially guiding users to input incorrect information.

connorlbark: so, would be assuming `https://` always and showing `duos.org` as an example url be enough to fix the user experience?

connorlbark: URL validation now is simpler (from user perspective) and allows `duos.org`. Provides `duos.org` as an example.

JVThomas: Yep this works. This way we're allowing the user to input the url without influencing them in anyway. If the value ends up being wrong we can either point people to SOs or contact them (SOs) directly and tell them to fix it.

JVThomas (approved):
Thanks for the changes! The code is good to go

PR #1792, opened 2022-09-07
[DUOS-2073][risk=no] Add validation to FormField props

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2073

Also included: 
- Cleanup to make the prop naming of `SELECT` clearer (imo)

Now, if you provide a prop which is unknown to formfield, an exception will be thrown. SImilarly, fields which are required, such as `id` or `selectOptions` will also throw an exception. For select, validation will also occur to ensure that the `selectOptions` passed are valid.

----

- Verify all tests go green

### DIFF
--- a/src/components/forms/formUtils.js
+++ b/src/components/forms/formUtils.js
@@ -1,0 +1,47 @@
+import {FormFieldTypes} from './forms'
+import { isNil, isEmpty, isString, isArray } from 'lodash/fp';
+
+const alwaysRequiredProps = ['id'];
+const alwaysPossibleProps = [
+  'name',
+  'disabled',
+  'description',
+  'title',
+  'ariaLevel',
+  'defaultValue',
+  'hideTitle',
+  'style',
+  'validators',
+  'onChange',
+  'type'
+];
+
+export const validateFormProps = (props) => {
+  const type = (!isNil(props.type) ? props.type : FormFieldTypes.TEXT)
+  
+  const requiredProps = type.requiredProps || [];
+  const possibleProps = type.possibleProps || [];
+
+  requiredProps.push(...alwaysRequiredProps);
+  possibleProps.push(...alwaysPossibleProps);
+  possibleProps.push(...requiredProps);
+
+  const propKeys = Object.keys(props);
+
+  propKeys.forEach((prop) => {
+    if (!possibleProps.includes(prop)) {
+      throw `unknown prop ${prop}`;
+    }
+  });
+
+  requiredProps.forEach((requiredProp) => {
+    if (!propKeys.includes(requiredProp)) {
+      throw `prop ${requiredProp} is required`;
+    }
+  });
+
+  if (!isNil(type.customPropValidation)) {
+    type.customPropValidation(props);
+  }
+
+} 
--- a/src/components/forms/formUtils.js
+++ b/src/components/forms/formUtils.js
@@ -1,35 +1,21 @@
-import {FormFieldTypes} from './forms';
-import { isNil, isFunction } from 'lodash/fp';
-
-const alwaysRequiredProps = ['id'];
-const alwaysPossibleProps = [
-  'name',
-  'disabled',
-  'description',
-  'title',
-  'ariaLevel',
-  'defaultValue',
-  'hideTitle',
-  'style',
-  'validators',
-  'onChange',
-  'type'
-];
+import { FormFieldTypes, commonRequiredProps, commonOptionalProps } from './forms';
+import { isNil, isFunction, isArray, isEmpty, isString } from 'lodash/fp';
+import { isEmailAddress } from '../../libs/utils';
 
 export const validateFormProps = (props) => {
   const type = (!isNil(props.type) ? props.type : FormFieldTypes.TEXT);
 
   const requiredProps = type.requiredProps || [];
-  const possibleProps = type.possibleProps || [];
+  const optionalProps = type.optionalProps || [];
 
-  requiredProps.push(...alwaysRequiredProps);
-  possibleProps.push(...alwaysPossibleProps);
-  possibleProps.push(...requiredProps);
+  requiredProps.push(...commonRequiredProps);
+  optionalProps.push(...commonOptionalProps);
+  optionalProps.push(...requiredProps);
 
   const propKeys = Object.keys(props);
 
   propKeys.forEach((prop) => {
-    if (!possibleProps.includes(prop)) {
+    if (!optionalProps.includes(prop)) {
       throw `unknown prop ${prop}`;
     }
   });
@@ -43,4 +29,80 @@ export const validateFormProps = (props) => {
   if (isFunction(type.customPropValidation)) {
     type.customPropValidation(props);
   }
+};
+
+
+export const customSelectPropValidation = (props) => {
+  if (!isArray(props.selectOptions)) {
+    throw 'prop \'selectOptions\' must be an array';
+  }
+
+  if (isEmpty(props.selectOptions)) {
+    throw '\'selectOptions\' cannot be empty';
+  }
+
+  const isStringArr = isString(props.selectOptions[0]);
+
+  props.selectOptions.forEach((opt) => {
+    if (isStringArr) {
+      if (!isString(opt)) {
+        throw 'all values in \'selectOptions\' must be string typed';
+      }
+
+      return;
+    }
+
+    if (isNil(opt.displayText)) {
+      throw 'every value in \'selectOptions\' needs a \'displayText\' field';
+    }
+  });
+};
+
+export const customRadioPropValidation = (props) => {
+  if (!isArray(props.options)) {
+    throw '\'options\' prop must be an array.';
+  }
+
+  props.options.forEach((opt) => {
+    if (isNil(opt.name)) {
+      throw 'each option in \'options\' prop must have a \'name\' field';
+    }
+
+    if (isNil(opt.text)) {
+      throw 'each option in \'options\' prop must have a \'text\' field';
+    }
+
+    if (!isNil(opt.type) && !['boolean', 'string'].includes(opt.type)) {
+      throw `radio group option types can be 'boolean' or 'string', not: '${opt.type}'`;
+    }
+  });
+};
+
+export const updateSelectDefaultValue = ({ selectOptions, defaultValue }) => {
+  const isStringArr = isString(selectOptions[0]);
+  return isStringArr
+    ? { key: defaultValue, displayText: defaultValue }
+    : defaultValue;
+};
+
+export const requiredValidator = {
+  isValid: (value) => value !== undefined && value !== null && value !== '',
+  msg: 'Please enter a value',
+};
+
+export const urlValidator = {
+  isValid: (val) => {
+    try {
+      new URL(val);
+    } catch (_) {
+      return false;
+    }
+    return true;
+  },
+  msg: 'Please enter a valid url (e.g., https://www.google.com)',
+};
+
+export const emailValidator = {
+  isValid: isEmailAddress,
+  msg: 'Please enter a valid email address (e.g., person@example.com)',
 };
--- a/src/components/forms/formUtils.js
+++ b/src/components/forms/formUtils.js
@@ -1,5 +1,5 @@
-import {FormFieldTypes} from './forms'
-import { isNil, isEmpty, isString, isArray } from 'lodash/fp';
+import {FormFieldTypes} from './forms';
+import { isNil } from 'lodash/fp';
 
 const alwaysRequiredProps = ['id'];
 const alwaysPossibleProps = [
@@ -17,8 +17,8 @@ const alwaysPossibleProps = [
 ];
 
 export const validateFormProps = (props) => {
-  const type = (!isNil(props.type) ? props.type : FormFieldTypes.TEXT)
-  
+  const type = (!isNil(props.type) ? props.type : FormFieldTypes.TEXT);
+
   const requiredProps = type.requiredProps || [];
   const possibleProps = type.possibleProps || [];
 
@@ -44,4 +44,4 @@ export const validateFormProps = (props) => {
     type.customPropValidation(props);
   }
 
-} 
+};
--- a/src/components/forms/formUtils.js
+++ b/src/components/forms/formUtils.js
@@ -75,13 +75,6 @@ export const customRadioPropValidation = (props) => {
   });
 };
 
-export const updateSelectDefaultValue = ({ selectOptions, defaultValue }) => {
-  const isStringArr = isString(selectOptions[0]);
-  return isStringArr
-    ? { key: defaultValue, displayText: defaultValue }
-    : defaultValue;
-};
-
 export const requiredValidator = {
   isValid: (value) => value !== undefined && value !== null && value !== '',
   msg: 'Please enter a value',
--- a/src/components/forms/formUtils.js
+++ b/src/components/forms/formUtils.js
@@ -72,9 +72,6 @@ export const customRadioPropValidation = (props) => {
       throw 'each option in \'options\' prop must have a \'text\' field';
     }
 
-    if (!isNil(opt.type) && !['boolean', 'string'].includes(opt.type)) {
-      throw `radio group option types can be 'boolean' or 'string', not: '${opt.type}'`;
-    }
   });
 };
 
--- a/src/components/forms/formUtils.js
+++ b/src/components/forms/formUtils.js
@@ -1,5 +1,5 @@
 import {FormFieldTypes} from './forms';
-import { isNil } from 'lodash/fp';
+import { isNil, isFunction } from 'lodash/fp';
 
 const alwaysRequiredProps = ['id'];
 const alwaysPossibleProps = [
@@ -40,8 +40,7 @@ export const validateFormProps = (props) => {
     }
   });
 
-  if (!isNil(type.customPropValidation)) {
+  if (isFunction(type.customPropValidation)) {
     type.customPropValidation(props);
   }
-
 };
--- a/src/components/forms/forms.js
+++ b/src/components/forms/forms.js
@@ -8,7 +8,6 @@ import {
   requiredValidator,
   urlValidator,
   emailValidator,
-  updateSelectDefaultValue,
 } from './formUtils';
 import {
   formInputGeneric,
@@ -28,6 +27,7 @@ export const commonOptionalProps = [
   'description',
   'title',
   'ariaLevel',
+  'ariaDescribedby',
   'defaultValue',
   'hideTitle',
   'style',
@@ -36,9 +36,9 @@ export const commonOptionalProps = [
   'type'
 ];
 
-// ----------------------------------------------- //
-// ======       MAIN FORM FIELD TYPES       ====== //
-// ----------------------------------------------- //
+// ----------------------------------------------------------------------------------------------------- //
+// ======                                  MAIN FORM FIELD TYPES                                  ====== //
+// ----------------------------------------------------------------------------------------------------- //
 export const FormFieldTypes = {
   MULTITEXT: {
     defaultValue: [],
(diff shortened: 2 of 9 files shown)


src/components/forms/forms.js
```diff
@@ -14,24 +15,131 @@ import {
...
-  SLIDER: { defaultValue: false, component: formInputSlider },
-  RADIO: { defaultValue: null, component: formInputRadioGroup },
-  TEXT: { defaultValue: '', component: formInputGeneric },
-  NUMBER: { defaultValue: '', component: formInputGeneric },
+  MULTITEXT: {
+    defaultValue: [],
+    component: formInputMultiText,
+    requiredProps: [],
+    possibleProps: [
+      'title',
```
lu-c: should possible props always contain these common props? The ones that are possible for all form field types?

src/components/forms/forms.js
```diff
@@ -14,24 +15,131 @@ import {
...
+      'ariaDescribedby',
+      'inputStyle'
+    ],
+  },
+  SLIDER: {
+    defaultValue: false,
+    component: formInputSlider,
+    requiredProps: [],
+    possibleProps: [
+      'toggleText',
```
lu-c, line 58: Should this array also contain the common props like "title" and "description"?

src/components/forms/formUtils.js
```diff
@@ -0,0 +1,46 @@
+import {FormFieldTypes} from './forms';
```
lu-c: With the inclusion of the new `formUtils.js` file, I think we should consider moving the form behavior functions there, like form validation, and maybe `FormFieldTypes`? I know we made the forms.js file in the first place to make the options easier to identify, but it feels like it should belong in a utils XD

src/components/forms/formUtils.js
```diff
@@ -0,0 +1,46 @@
...
+import { isNil, isFunction } from 'lodash/fp';
+
+const alwaysRequiredProps = ['id'];
+const alwaysPossibleProps = [
```
lu-c: what would you think of renaming this to `commonProps` or something instead? I'm imagining like


```suggestion
const commonProps = [
    id: { type: 'string', required: true },
    name: { type: 'string' },
    description: { type: 'string' },
    disabled: { type: 'bool' }
    ...
];
```

lu-c:
The config looks siiick, way nicer than just relying on the config extracts in the formComponents file!

I think the formFieldTypes props declaration is a little confusing to me - I'm not sure if the common/allPossibleProps are meant to included or excluded from the list, since it seems be kind of both in there. It's my preference to exclude all of the common fields from each type declaration, and only include the ones specific to that formField.

What do you think?

connorlbark:
> I think the formFieldTypes props declaration is a little confusing to me - I'm not sure if the common/allPossibleProps are meant to included or excluded from the list, since it seems be kind of both in there. It's my preference to exclude all of the common fields from each type declaration, and only include the ones specific to that formField.

Oh oops! You are correct, that is what I meant to do. I think I was on autopilot when I was making those. I will clean them up to just be the ones specific to those types.

src/components/forms/formUtils.js
connorlbark: sure! works for me.

src/components/forms/formUtils.js
connorlbark: Yeah! I agree. I think FormFieldTypes itself should stay in `Forms.js` because it's an easy place of reference, but moving custom validation to a `selectOptionsCustomValidation` function in utils could be a nice way to clean up the types to not be so large, and just be a central way to see how all these form components work. Will do.

src/components/forms/formUtils.js
connorlbark: do you think typechecking would be good here? i changed the name to something i think a bit better, but kept optional and required separate: `commonOptionalProps`, but i am not sure if type checking is overkill here or a good inclusion. if we want to do that i can merge all props configs like you suggested

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
...
-* disabled: bool
-* placeholder, defaultValue,
-* style (for the formControl component)
-* inputStyle (for the input element)
-* onChange: func({key: '', value: '', isValid: boolean}),
-* validators: [{isValid: func, msg: ''}]
-* Accessibility: (defaults blank)
-*    * ariaDescribedby
-*    * ariaLevel
-*/
```
connorlbark, line 94: also: deleted this because (imo) this information is more easily ascertained by looking at the configs. However, I would understand if we wished to keep this as well.

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
...
+    component: formInputGeneric,
+    requiredProps: [],
+    optionalProps: [
+      'placeholder',
+      'inputStyle',
+      'ariaDescribedby',
+    ],
+  },
   CHECKBOX: {
     defaultValue: (config) => (config?.valueType === 'string' ? '' : false),
```
lu-c, line 95: i honestly can't remember, how / when would we want a checkbox to have a string value? If this is a common usecase, would you mind adding a comment in the code for future devs to understand too?

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
...
+      //  {displayText: 'Dac 1', dacId: 213}
+      // ]
+    ],
+    optionalProps: [
+      'isCreatable', // allows user to input their own
+      'isMulti',
+      'placeholder',
+      'ariaDescribedby',
+      'required',
+      'selectConfig',
```
lu-c, line 124:
```suggestion
      'selectConfig', // react-select or react-select/creatable config
```

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
...
+      //   ^^^ can pass other fields in as extra info, e.g.:
+      //  {displayText: 'Dac 1', dacId: 213}
+      // ]
+    ],
+    optionalProps: [
+      'isCreatable', // allows user to input their own
+      'isMulti',
+      'placeholder',
+      'ariaDescribedby',
+      'required',
```
lu-c: Should this be a field? vs `FormValidators.REQUIRED`?

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
...
-      }
-      return true;
-    },
-    msg: 'Please enter a valid url (e.g., https://www.google.com)'
-  },
-  EMAIL: {
-    isValid: isEmailAddress,
-    msg: 'Please enter a valid email address (e.g., person@example.com)'
-  }
+  REQUIRED: requiredValidator,
```
lu-c, line 145: Instead of exporting all of these functions from formUtils, can we just put the entire FormValidators object there please?

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
 
 import './forms.css';
 
+export const commonRequiredProps = ['id'];
+export const commonOptionalProps = [
+  'name',
+  'disabled',
+  'description',
+  'title',
+  'ariaLevel',
```
lu-c, line 30:
```suggestion
  'ariaLevel',
  'ariaDescribedby',
```

src/components/forms/forms.js
```diff
@@ -13,80 +21,122 @@ import {
...
+      'ariaDescribedby',
+    ],
+  },
   CHECKBOX: {
     defaultValue: (config) => (config?.valueType === 'string' ? '' : false),
-    component: formInputCheckbox
+    component: formInputCheckbox,
+    requiredProps: [],
+    optionalProps: [
+      'placeholder',
```
lu-c: Should checkboxes have a placeholder?
EDIT: i would get rid of inputStyle as well? not sure if we have areas where we're doing that

src/components/forms/formUtils.js
```diff
@@ -0,0 +1,108 @@
...
+      throw 'each option in \'options\' prop must have a \'text\' field';
+    }
+
+    if (!isNil(opt.type) && !['boolean', 'string'].includes(opt.type)) {
+      throw `radio group option types can be 'boolean' or 'string', not: '${opt.type}'`;
+    }
+  });
+};
+
+export const updateSelectDefaultValue = ({ selectOptions, defaultValue }) => {
```
lu-c: Would recommend moving this function back into the forms config declaration, just to reduce the distance between these. The other config functions (for like defaultValue) are being defined there too, i dont know, it just keeps it consistent. I can understand if you want to keep it here tho!

src/components/forms/formUtils.js
```diff
@@ -0,0 +1,108 @@
...
+  const isStringArr = isString(selectOptions[0]);
+  return isStringArr
+    ? { key: defaultValue, displayText: defaultValue }
+    : defaultValue;
+};
+
+export const requiredValidator = {
```
lu-c, line 78:
```suggestion
//---------------------------------------------
// Form Validators
//---------------------------------------------
export const requiredValidator = {
```

src/components/forms/forms.js
connorlbark: agreed - i was just copying over what was there but this is a good opportunity to normalize these

src/components/forms/forms.js
quazi-h: I think when the checkbox option is "Other" and to allow the user to input text - came across that when reviewing the Data Governance sub-component in another PR.

connorlbark: Yes, Quazi is correct. I will add a comment here + on the field itself; this might be confusing though, and I am open to alternate suggestions.

src/components/forms/forms.js
connorlbark: these are used iff valueType is string; i will try to make this more clear!

src/components/forms/forms.js
connorlbark: Hmm, I have an idea. What if for both radio and checkbox there was a `conditionalFormField` prop, which, if passed, makes the checkbox enter into a setup where the checkbox controls the conditional rendering of a formField, such as an input box or something else. I suggest this because in another ticket, I need to conditionally render a more complex field than just an input box. Thoughts on this approach?

lu-c: i think any sort of more complex conditional rendering shouldnt be handled as part of the global forms components, and should be handled with the methods that are rendering them.

Instead of `FormField({ conditionalFormField: (condition, template) => ...`, i would rather `FormField({ onChange: () => show/hide stuff`

I think it would help keep the code cleaner and easier to understand long term, since writing all of the code to handle the conditions / checks could get complex, and will largely be unnecessary out of the uses of that one specifically rendered formField

connorlbark: I think I agree for checkboxes, though I do think this logic is common enough that having it be a part of the checkbox could be useful. However, for radio groups, this might need to be addressed - since, e.g., there is a radio group in https://github.com/DataBiosphere/duos-ui/pull/1791 which requires conditional rendering - if you check disease specific use, between the disease specific use option and populations, origins, and ancestry option a searchbar needs to come up. In that case, do you think a conditionalFormField would make sense? I am not sure how else to inject that field inside of the radio group.

For radio groups, it could just be as simple as `renderIfSelected: h(...)` within each of the option declarations:

```
{
   ...,
   key: 'asdf',
   renderIfSelected: p({}, ['Selected!']),
}
```

lu-c: I don't know, it just seems too complex to be nesting formField configs inside formField configs, especially when we can still use `isRendered` and `onChange` handles all of the functionality in the same way. I don't want to be controlling multiple types of objects, or nesting long giant objects, it's harder (for me!) to read.

Interested in seeing what some other opinions are then!

src/components/forms/forms.js
connorlbark: I kind of like them being here, because they demonstrate the functionality of the formfields. To me, a `formUtils` folder contains all of the 'backend' code which a developer who is just trying to use the components doesn't need to know; since it's forward facing, I think it should stay in `Forms.js`. That's just my thought though

src/components/forms/forms.js
JVThomas: I'm with Christine on this. Things like `otherText` (a textbox that shows up if `Other` is selected), can easily be controlled by a different form field with conditional rendering. Those additional fields are often mapped to a variable/key separate from that of the radio button/checkbox.

Take the disease list selection for instance. A similar feature exists in the DAR Form on Step 2. With it, the radio button value is saved to one key, but the follow up value (textbox, list, whatever) is saved to another. The diseases radio button is saved to `diseases` (don't remember the exact name), but the list of diseases itself is saved on `ontologies`

Benefits to this can be applied to the front-end and back-end. For one, the types behind each key are set it stone. A radio button will have to equal a string, a checkbox will have to be true/false, multi-select will have to be an array, etc. If there's a form element down the road that is conditionally rendered on one of these values, I can do a simple equality check vs something like `if(isFunction) {do one thing}, else if(isString) {do something else}...` and so forth. In terms of the back-end, variables of a class need to have their types declared. Having a variable, which is mapped on the UI to a checkbox/radio button that can assign either a `String` or `Boolean` just adds complexity that Java (which is not as flexible as JS) now has to handle.

connorlbark: I agree with all of this, but I think functionally speaking this is a bit different than what I was suggesting (which is different than the original code which was commented on). Currently, I have the base code pushed up for my idea, but I don't have an example of how it would work. Let me flesh out a sketch of how I was intending on using it on the data access governance page:



```js
const [diseaseSpecificUseText, setDiseaseSpecificUseText] = useState(...)


h(FormField, {
   type: RADIO,
    ...
   options: [
      ...
      {
         key: "diseaseSpecificUse",
         renderIfSelected: h(FormField, ... form for searching different diseases, onchange updates diseaseSpecificUseText ... )
      }
      ....
   ],
   onChange({}) => {... if (selected == 'diseaseSpecificUse') then setField('diseaseSpecificUse', diseaseSpecificUseText)...},
}

```

This would put a searchbar directly underneath the diesase specific use radio button, and it's on the end developer to do state updates to keep that information. This means the radio operates as one expects, and the user is able to see the searchbar inside of the radio group as they would expect. This also means the values are separated, so there aren't weird type checking issues like you were worried about.

JVThomas: Approaching this from a high-level view, it's a choice between ...
```
h(FormField), {
  type: RADIO,
  ...
  options: [
    //define other options for parent here
    {
      key: 'diseaseSpecificUse',
      renderIfSelected: h(FormField, {
        type: SELECT,
        ...
        options: [
          //define options for child here
        ],
        onChange: ({})
      })
    }
  ],
  onChange: () => {... if(selected === 'diseaseSpecificUse') then ...}
}
```

and

```
h(FormField), {
  type: SELECT,
  ...
  options: [
    //define options here
  ]
  onChange: () => {/*update parent question value here*/}
},

FormField, {
  type: SELECT,
  ...
  options: [
    //define options here
  ],
  onChange: () => { //update select list here},
  isRendered: diseaseSpecificUse
}
```

Both are truncated definitions, but even so there's one noticeable difference. With the first example the sub-component is now a prop value to the parent component (regardless of whether you attach it to `options` or or pass it in as another prop variable), meaning that if any value on the parent prop is updated, the entire component is re-rendered. Since the child component is a part of the parent, its forced to re-render as well. If the child has any specific logic that fires on initialization, it will fire off once again on its re-render.

You can work around the above situation by using a `useMemo` hook to define the child component so that it re-renders on specific variable updates, but now you're in a situation where you've defined two components (with one passed in as a child component to the other) alongside a `useMemo` variable.

The second option doesn't have to worry about re-rendering outside of the `isRendered` condition and input updates. If the `options`, `onChange function definition`, `disabled`, or any other prop on the non-dependent (parent) component changes, the dependent (child) component will be unaffected (unless the input value of the parent is used to calculate/compute/update the input value of the child).

So then the question now is whether or not there's any significant gains to writing the code this way. Reasons could be anything from performance, testability, organization, readability, etc. From a high-level perspective I don't see any noticeable differences between version 1 and version 2 other than having the child component nestled under the parent in a container div. In terms of readability and performance version 1 takes a hit because we now have to deal with the complexity of nested components, which is why I lean towards version 2. If there was something more like isolating complex calculations or manipulations then I could lean towards 1, but right now we're just accomplishing the same effect in version 2 with more overhead instead of less.

connorlbark: I totally agree with what you're saying about complexity, and if they rendered the exact same HTML then I would prefer the second option. However, the reason I am pushing for this option specifically is becuase the mock looks like so:

image

I think for the end user having the searchbar be inside of the radio group is a better look for the option. If we don't nest the searchbar inside of the radio group's formfield, then having the search bar be a seamless part of the radio group isn't possible. If we're okay with putting the searchbar below the radio group, then I totally agree and will remove this functionality; however, most of this work has been in effort to figure out how to accomplish the nested searchbar.

JVThomas: Ah the form component is a template for the entire group, not the individual component. The part I'm having trouble rectifying is this line

``` If we don't nest the searchbar inside of the radio group's formfield, then having the search bar be a seamless part of the radio group isn't possible```

The radio buttons don't have to exist as part of a group template, they can exist as individual elements. UI libraries and company style guides will always develop individual `RadioButton` components explicitly for edge cases where you need custom formats and layouts. This way you can style and arrange them as you see fit for the page in question. You want the search bar/input/textfield to appear above, left, right, below? Easy to handle if the elements are drawn out individually in HTML. But if it's passed in as a prop? To handle the custom positioning you'll need even more custom props.

This isn't to say that the Group template is bad, it's still useful for situations where there are no extra fields popping in and out. But for custom layouts the accepted approach is to manually write out the fields.

Here's Atlassian's documentation on radio buttons:
https://atlassian.design/components/radio/examples

Here's Material-UI's take on radio buttons (standalone)
https://mui.com/material-ui/react-radio-button/#standalone-radio-buttons

Here's Material-UI's approach on RadioGroups
https://mui.com/material-ui/react-radio-button/#radio-group

connorlbark: Ok this these are all great points. I think I will remove this functionality so that this PR can get merged in. In my `DataAccessGovernance` PR I will work on a FormField which is just a singular radio button.

connorlbark: Thank you for all the documentation and thorough discussion. This is super helpful & a lot of great resources.

connorlbark:
@JVThomas @quazi-broad this PR should be ready for a re-review! removed the conditional rendering and (as mentioned above) I will create an individual radio button formfield component for the DataAccessGovernance section. I also merged in develop (with the Step 1 DAR form) which required some non-trivial edits to get these working toegether.

quazi-h (approved):
New changes look good. The validation should help us catch bugs faster in the future.

JVThomas (dismissed):
There's a regression in the checkbox functionality. I can no longer select or unselect checkboxes on `ResearcherInfo_new.js` (the cloud/local computing questions), whereas the `develop` branch can.

NOTE: There's also a bug where the cloud provider text fields still show up when 'Yes' is selected for the anVIL use question, but that bug existed before this branch, so I'll add a bugfix ticket to address it.

Kapture 2022-09-15 at 13 47 42

lu-c (approved):

connorlbark:
> There's a regression in the checkbox functionality. I can no longer select or unselect checkboxes on `ResearcherInfo_new.js` (the cloud/local computing questions), whereas the `develop` branch can.
> 
> NOTE: There's also a bug where the cloud provider text fields still show up when 'Yes' is selected for the anVIL use question, but that bug existed before this branch, so I'll add a bugfix ticket to address it.
> 
> Kapture 2022-09-15 at 13 47 42 [ Kapture 2022-09-15 at 13 47 42 ](https://user-images.githubusercontent.com/6664778/190475780-fab70a72-251b-4bcb-9ae3-7829b4672b7e.gif) [ ](https://user-images.githubusercontent.com/6664778/190475780-fab70a72-251b-4bcb-9ae3-7829b4672b7e.gif)

Not sure how I missed that! Should be fixed now if you want to check it out, then I will get it merged.

quazi-h:
Took another look and can confirm that the checkboxes are working correctly with changes from the last commit.

PR #1793, opened 2022-09-08
[DUOS][risk=no] Added Translate Page

lu-c:
It's not linked to anywhere within the duos site, you have to type it into the url bar

----

- Verify all tests go green

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -96,7 +96,7 @@ const Routes = (props) => (
       rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/tos_acceptance" component={TermsOfServiceAcceptance} props={props} rolesAllowed={[USER_ROLES.all]} />
     {(props.env === 'local' || props.env === 'dev') && <AuthenticatedRoute path="/data_submission_form" component={DataSubmissionForm} props={props} rolesAllowed={[USER_ROLES.all]} />}
-    <AuthenticatedRoute path="/translate" component={Translator} props={props} rolesAllowed={[USER_ROLES.researcher]}/>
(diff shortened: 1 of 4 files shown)


src/Routes.js
```diff
@@ -95,6 +96,7 @@ const Routes = (props) => (
       rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/tos_acceptance" component={TermsOfServiceAcceptance} props={props} rolesAllowed={[USER_ROLES.all]} />
     {(props.env === 'local' || props.env === 'dev') && <AuthenticatedRoute path="/data_submission_form" component={DataSubmissionForm} props={props} rolesAllowed={[USER_ROLES.all]} />}
+    <AuthenticatedRoute path="/translate" component={Translator} props={props} rolesAllowed={[USER_ROLES.researcher]}/>
```
JVThomas: Since this is an experimental feature, enable it when `env === local`

rushtong: I think Gabe is very hopeful to show this off in conjunction with https://github.com/DataBiosphere/consent-ontology/pull/671 once that is approved

JVThomas: So then should we just limit it to just `dev` and `staging` for now to make it demoable?

rushtong: Seems reasonable to me!

JVThomas (approved):
Outside of the missing env restrictions mentioned above the code looks good. We may need to make adjustments once the ontology endpoint is implemented, but I doubt it'll be anything that'll drastically change this code.

src/Routes.js
lu-c: ok! updated!

PR #1799, opened 2022-09-14
[DUOS][risk=no] translate page x2 - added error handling, added no results block, updated url to be hardcoded

lu-c:
The url is now up and running (despite CORS error), so had to make a few tweaks to this code

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -880,7 +880,7 @@ export const ToS = {
 
 export const Translate = {
   translate: async (body) => {
-    const url = `${await getOntologyUrl()}translate/paragraph`;
+    const url = `${await getOntologyUrl()}/translate/paragraph`;
     const res = await axios.post(url, body, Config.authOpts());
     return res.data;
   },
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -880,7 +880,7 @@ export const ToS = {
 
 export const Translate = {
   translate: async (body) => {
-    const url = `https://consent-ontology.dsde-dev.broadinstitute.org/translate/paragraph`;
+    const url = `${await getOntologyUrl()}translate/paragraph`;
     const res = await axios.post(url, body, Config.authOpts());
     return res.data;
   },
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -880,7 +880,7 @@ export const ToS = {
 
 export const Translate = {
   translate: async (body) => {
-    const url = `${await getOntologyUrl()}/translate/paragraph`;
+    const url = `https://consent-ontology.dsde-dev.broadinstitute.org/translate/paragraph`;
     const res = await axios.post(url, body, Config.authOpts());
     return res.data;
   },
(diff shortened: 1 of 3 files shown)


src/libs/ajax.js
```diff
@@ -880,7 +880,7 @@ export const ToS = {
 
 export const Translate = {
   translate: async (body) => {
-    const url = `${await getOntologyUrl()}/translate/paragraph`;
+    const url = `https://consent-ontology.dsde-dev.broadinstitute.org/translate/paragraph`;
```
lu-c: `await getOntologyUrl()` was returning `https://consent-ontology.dsde-dev.broadinstitute.org/` (with the slash on the end of the string), but other urls in this ajax file looked like `${getOntologyUrl()}/autocomplete`, and when following that standard, the translate URL turned into

`https://consent-ontology.dsde-dev.broadinstitute.org//translate/paragraph` which would error out (because of the double slash)

Not sure if it's safe to remove that slash from the url string, and time is short for this demo, so to rush it, I just threw in the actual url as a string instead of relying on the config

lu-c:
The actual endpoint throws a CORS error - Please correct me I'm wrong, but there's nothing we can do about that, and it might be okay in dev? But if any fixes need to happen, it's in the api?

src/libs/ajax.js
rushtong: Don't we need the config for local proxying?

rushtong:
> The actual endpoint throws a CORS error - Please correct me I'm wrong, but there's nothing we can do about that, and it might be okay in dev? But if any fixes need to happen, it's in the api?

The CORS error is because we're not going through the local proxy.

src/pages/Translator.js
```diff
@@ -133,7 +117,14 @@ export default function Translator() {
             style: { width: '100%', justifyContent: 'center' }
           }, [Spinner]),
 
-          div({ isRendered: !isLoading, className: 'row no-margin' }, [
+          div({ isRendered: error }, [
+            'There was an error running your request'
```
connorlbark: do we want to use error to show something more specific

src/libs/ajax.js
lu-c: Do we? When I throw in this line

`const url = `${await getOntologyUrl()}translate/paragraph``

it errors on `POST https://local.broadinstitute.org:3000/translate/paragraph 404 (Not Found)`

connorlbark (commented):
was able to test it ok, the changes themselves look good, once we figure out the config stuff I will retry + approve.

rushtong (approved):
👍🏽

src/pages/Translator.js
lu-c: i dont want to show the error message if it's something wrong with our servers or our code, only if it's a user error (that they can do something about). For now, I think this is okay, since I'm just trying to rush this code in real quick ^_^;

JVThomas (approved):
UI code is hitting the endpoint and relaying the results to the client.

PR #1803, opened 2022-09-20
[DUOS-2086][risk=no] Update Manage Users Table

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2086

Requires https://github.com/DataBiosphere/consent/pull/1752 to display library cards and institution.

----

- Verify all tests go green


src/components/manage_users_table/ManageUsersTable.js
```diff
@@ -62,21 +64,33 @@ const columnHeaderConfig = {
...
-    cellDataFn: cellData.googleIdCellData,
+    cellDataFn: cellData.emailCellData,
     sortable: false
   },
-  role: {
-    label: 'User Role',
-    cellStyle: { width: styles.cellWidth.role },
-    cellDataFn: cellData.roleCellData,
+  institution: {
+    label: 'Insitution',
```
quazi-h:
```suggestion
    label: 'Institution',
```

quazi-h (approved):
Looks good to me, is there any followup work that is required once the backend ticket ([1752](https://github.com/DataBiosphere/consent/pull/1752) is completed?

src/components/manage_users_table/ManageUsersTableCellData.js
```diff
@@ -17,51 +18,44 @@ export function usernameCellData({displayName, userId, label= 'user-name'}) {
...
     label
   };
 }
 
-//Redirect for admin review page, only used in admin manage dar collections table
-export function roleCellData({userId, roles, label = 'user-role'}) {
+export function permissionsCellData({userId, roles, libraryCards, institution, label = 'permissions'}) {
+  const hasLibraryCard = libraryCards?.map((lc) => lc.institutionId)?.includes(institution?.id);
+  const roleNames = map(roles, 'name').filter((roleName) => roleName !== 'Researcher');
+  const perms = (hasLibraryCard ? roleNames.concat('Library Card') : roleNames);
```
JVThomas: The current search bar filters the table on `displayName`, `email`, and `roles`. Since `Library Card` is a new addition to the permissions column, you'll need to update the search function to account for that as well.

JVThomas (changes requested):
Table is looking good! You just need to make one more adjustment with the search bar

src/libs/utils.js
```diff
@@ -568,13 +568,20 @@ export const getSearchFilterFunctions = () => {
...
 
         const matchable = [displayName, email];
         if (!isNil(roles)) {
           matchable.push(...map((r) => r.name)(roles));
         }
+        if (!isNil(institution)) {
+          matchable.push(institution.name);
+        }
+
+        if (!isNil(libraryCards) && !isEmpty(libraryCards)) {
```
JVThomas: So there's a discrepancy between the logic here and the logic used to set up data for `permissionsCell`. Here you're checking if the user has a library card at all, whereas the permission cell is checking if the user has a library card where `card.institutionId === institution.id`.

The end result is that the filter will let anyone with a library card show up on the filterTerm `library card`, but the permissions cell will state `None` for any researcher who has a library card that doesn't match their current institution. Since users who are only Researchers are allowed to change institutions, the above scenario can happen.

Since this is the admin view, we'd want to show as much information as possible, so I'm inclined to change the permissions logic so that it renders `Library Card` if the user has a Library Card present.  This would also mean that we'd need to list the Library Cards on the individual user page, but that's something to be discussed outside of this ticket.

Screen Shot 2022-09-28 at 9 00 39 AM

JVThomas (dismissed):

src/libs/utils.js
connorlbark: ok - so thoughts on rendering library card like, `Library Card (Broad Institute), Library Card (Other Institution), SigningOfficial`?

connorlbark: I think some backend changes would need to happen for this. It appears the endpoint is returning library cards, but only sometimes returns the institution alongside the library card. For some library cards, it has: 

```
{
...
institutionId: 120,
institution: { ... name: 'TEST', ... }
}
```

for others 
```
{
...
institutionId: 120,
institution: null
}
```
As such, I made the table consistent (always show library card perm if you have ANY library card) but I did not list all library cards that exist. If we wish to do that, I can investigate the backend work needed & get the table updated accordingly.

rushtong: > As such, I made the table consistent (always show library card perm if you have ANY library card) but I did not list all library cards that exist.

This, in my opinion, is the right choice at this stage. If we want more info in the future, we'll address it then.

rushtong:
The heading for this page should be `Manage Users` as opposed to `Edit User`. This is also reflected in the mocks in the ticket. I think it was a good idea to drop the edit button in favor of a link. However, the link could use either a tooltip or title text to explain where it goes.

Screen Shot 2022-09-30 at 11 37 30 AM

rushtong:
Referencing the above screenshot, can we do anything to the permission role names to make them cleaner, i.e. `DataOwner` -> `Data Owner`, etc.?

connorlbark:
@rushtong : used some basic regex to split the perms as described, e.g., `DataOwner` -> 'Data Owner'. Added a tooltip + updated heading to use mock's title & description.

rushtong:
> @rushtong : used some basic regex to split the perms as described, e.g., `DataOwner` -> 'Data Owner'. Added a tooltip + updated heading to use mock's title & description.

All looks good, except I still see the "Edit User" heading. Looking into that now.

rushtong:
@connorlbark - I think we need a change [to this line](https://github.com/DataBiosphere/duos-ui/blob/develop/src/pages/AdminManageUsers.js#L94)
`Edit User` -> `Manage Users`

connorlbark:
> @connorlbark - I think we need a change [to this line](https://github.com/DataBiosphere/duos-ui/blob/develop/src/pages/AdminManageUsers.js#L94) `Edit User` -> `Manage Users`

Oops! You're right. I accidentally changed the wrong line. Should be good now.

src/components/manage_users_table/ManageUsersTableCellData.js
```diff
@@ -17,51 +34,51 @@ export function usernameCellData({displayName, userId, label= 'user-name'}) {
...
+export function permissionsCellData({userId, roles, libraryCards, label = 'permissions'}) {
+  const hasLibraryCard = !isNil(libraryCards) && !isEmpty(libraryCards);
+  const roleNames = map(roles, 'name').filter((roleName) => roleName !== 'Researcher');
+  const perms = (hasLibraryCard ? roleNames.concat('LibraryCard') : roleNames);
+
+  // need to split, e.g., SigningOfficial -> Signing Official
+  // use regex to split by case, then stitch back together with spaces
+  // regex from: https://gist.github.com/JeffJacobson/3841577
+  const splitByWordRegex = /($[a-z])|[A-Z][^A-Z]+/g;
+  const formattedPerms = perms.map((perm) => perm.match(splitByWordRegex).join(' '));
```
rushtong: Nice!

src/components/manage_users_table/ManageUsersTableCellData.js
```diff
@@ -17,51 +34,51 @@ export function usernameCellData({displayName, userId, label= 'user-name'}) {
...
     label,
     id: userId,
   };
 }
 
-export function actionsCellData({userId, user, editUser}) {
+export function institutionCellData({userId, institution, label = 'insitution'}) {
   return {
     isComponent: true,
+    data: institution?.name || 'N/A',
```
rushtong, line 71: TDIL! This is something I [absolutely loved in groovy](https://groovy-lang.org/operators.html#_safe_navigation_operator) and did not know this was added to JS.

rushtong (approved):
👍🏽

src/components/manage_users_table/ManageUsersTableCellData.js
rushtong: I was digging more into this and found a [nice library](https://vocajs.com/#) that does this work as well:

```
const v = require('voca');
const val = v.words('DataSubmitter').join(' ');
console.log(val);
```
Outputs: `Data Submitter`

connorlbark: assuming this is the preference, i added voca and used it to split here!

rushtong: I don't have a preference, either way works!

PR #1804, opened 2022-09-21
[DUOS-1756][risk=no] Step 3 of DAR Application

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1756

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/ResearchPurposeStatement_new.js
+++ b/src/pages/dar_application/ResearchPurposeStatement_new.js
@@ -1,0 +1,206 @@
+import { useState } from 'react';
+import { a, div, fieldset, h2, h3, h4, h, span } from 'react-hyperscript-helpers';
+import isEmpty from 'lodash/fp/isEmpty';
+import './dar_application_new.css';
+import { FormField, FormFieldTypes, FormValidators } from '../../components/forms/forms';
+
+const ResearchPurposeRow = (props) => {
+  const {
+    title,
+    description,
+    id,
+    defaultValue,
+    onChange,
+  } = props;
+
+  return div({
+    className: 'rp-row flex flex-row',
+  }, [
+    div({
+      style: {
+        width: '70%',
+      }
+    }, [
+      div({
+        className: 'rp-row-title',
+      }, [title]),
+      div({
+        className: 'rp-row-description',
+      }, [description]),
+    ]),
+    div({
+      style: {
+        width: '20%',
+      }
+    }, [
+      h(FormField, {
+        type: FormFieldTypes.RADIOGROUP,
+        id: id,
+        validators: [FormValidators.REQUIRED],
+        orientation: 'horizontal',
+        options: [
+          {
+            text: 'Yes',
+            name: 'yes',
+          },
+          {
+            text: 'No',
+            name: 'no',
+          }
+        ],
+        defaultValue: defaultValue ? 'yes' : 'no',
+        onChange: ({key, value, isValid}) => {
+          onChange({key: key, value: (value === 'yes'), isValid: isValid});
+        }
+      }),
+
+    ])
+  ]);
+};
+
+export default function ResearchPurposeStatement(props) {
+
+  const {
+    darCode,
+    formFieldChange,
+    illegalBehavior,
+    notHealth,
+    oneGender,
+    pediatric,
+    psychiatricTraits,
+    sexualDiseases,
+    stigmatizedDiseases,
+    vulnerablePopulation
+  } = props;
+
+  const [gender, setGender] = useState(props.gender);
+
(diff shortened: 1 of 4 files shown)


quazi-h:
Looking good so far. I think two questions might be missing compared to the original DAR form. I don't have access to the wireframe linked in the ticket so please let me know if these were left out! The commercial purpose one was mentioned in the ticket description though.

3.1.1 Will this data be used exclusively or partially for a commercial purpose?
3.1.5 Does the research aim involve the study of alcohol or drug abuse, or abuse of other addictive products?

connorlbark:
> Looking good so far. I think two questions might be missing compared to the original DAR form. I don't have access to the wireframe linked in the ticket so please let me know if these were left out! The commercial purpose one was mentioned in the ticket description though.
> 
> 3.1.1 Will this data be used exclusively or partially for a commercial purpose? 3.1.5 Does the research aim involve the study of alcohol or drug abuse, or abuse of other addictive products?

image

For context, this is the DAR Form Mocks ^ I am not 100% sure if the commercial purpose / drug absue removals were intentional, but I believe they were.

src/pages/dar_application/ResearchPurposeStatement_new.js
```diff
@@ -0,0 +1,206 @@
...
+    notHealth,
+    oneGender,
+    pediatric,
+    psychiatricTraits,
+    sexualDiseases,
+    stigmatizedDiseases,
+    vulnerablePopulation
+  } = props;
+
+  const [gender, setGender] = useState(props.gender);
```
rushtong: In practice, this is not populating the saved gender value of the DAR on the screen. Looking at the object, the gender is being saved and retrieved correctly, just not parsed and/or displayed correctly:

Screen Shot 2022-09-26 at 1 03 17 PM

src/pages/dar_application/dar_application_new.css
```diff
@@ -0,0 +1,36 @@
+.dar-application-row {
+  clear: both;
+  margin-top: 15px;
```
rushtong, line 3: @JVThomas - what's our preference on pixel values in this kind of scenario? Are we preferring rem or %?

rushtong (changes requested):
One functionality bug that needs fixing, otherwise looks good!
Questions:
1. What are we standardizing on for pixels/rems/%
2. The mocks use numbered lists for the questions but that's not in the PR. Are there modifications to the mocks that I'm not aware of?

src/pages/dar_application/ResearchPurposeStatement_new.js
connorlbark: good catch!

connorlbark:
> 2\. The mocks use numbered lists for the questions but that's not in the PR. Are there modifications to the mocks that I'm not aware of?

I think there are multiple mocks running around. I just realized (confusingly) that the ticket links to an outdated mock (I will update this right now). It is my understanding that this is the most recent mock: https://projects.invisionapp.com/d/main#/console/22034746/467411575/preview . In this mock, the questions are not numbered. However, I am happy to number them if preferred.

src/pages/dar_application/ResearchPurposeStatement_new.js
connorlbark: fixed ! i couldn't figure out why the prop wasn't updating but i just removed the state variable entirely so the formData contains the state.

src/pages/dar_application/ResearchPurposeStatement_new.js
```diff
@@ -102,7 +101,7 @@ export default function ResearchPurposeStatement(props) {
             defaultValue: oneGender,
             onChange: ({key, value}) => {
               onChange({key, value});
-              onChange({key: 'gender', value: (value?gender:undefined)});
+              // onChange({key: 'gender', value: (value?gender:undefined)});
```
rushtong: nit: remove the commented out lines - we don't need them. Here and on line 139.

rushtong (approved):
Pending other reviews and minor nits, this looks good from a visual and functional perspective. Tentatively approving 👍🏽

src/pages/dar_application/dar_application_new.css
JVThomas: Since it's the margin-top of a row, I'm guessing it's being used to add space between rows. So the question is do we want the space between rows to grow as the page gets taller or not?

IMO I prefer the spacing to be fixed (or at least not grow beyond a max value), since, in my mind, making the window taller means I want to see more rows at once. Using rem, em, or % here would make the spacing grow if the parent (em, %) or root(rem) also grows with the window.

quazi-h (approved):
Thanks for confirming about the questions, changes here line up with the correct mock and look good to me.

PR #1805, opened 2022-09-22
[DUOS-1758][risk=no] Data Use Agreements - Submission

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1758

Updates styling, removes saving and sending functionality as the new DAR form is still incomplete. 

----

- Verify all tests go green

[DUOS-1758]: https://broadworkbench.atlassian.net/browse/DUOS-1758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/DataAccessRequestApplication_new.js
```diff
@@ -892,13 +892,15 @@ class DataAccessRequestApplicationNew extends Component {
...
-    const step1Invalid = this.step1InvalidResult(this.step1InvalidChecks());
-    const step2Invalid = this.verifyStep2();
-    const step3Invalid = this.step3InvalidResult();
-    const libraryCardInvalid = isEmpty(getOr([], 'libraryCards', this.state.researcher)) && !checkNihDataOnly;
+
+    // VALIDATION CHECKS:
+    // const step1Invalid = this.step1InvalidResult(this.step1InvalidChecks());
+    // const step2Invalid = this.verifyStep2();
+    // const step3Invalid = this.step3InvalidResult();
+    // const libraryCardInvalid = isEmpty(getOr([], 'libraryCards', this.state.researcher)) && !checkNihDataOnly;
```
rushtong: Are we waiting on all other steps to be merged before implementing these? If so, we should add that as a TODO comment here and make sure we have a follow-on story to apply the validation

src/pages/dar_application/DataUseAgreements_new.js
```diff
@@ -0,0 +1,81 @@
+
```
rushtong: nit: rm extra line

src/pages/dar_application/DataUseAgreements_new.js
```diff
@@ -0,0 +1,81 @@
...
+            span({className: 'glyphicon glyphicon-download'}),
+            ' Broad Library Card Agreement'
+          ])
+        ]),
+        div({}, [
+          a({ target: '_blank', href: '', className: 'button button-white' }, [
+            span({className: 'glyphicon glyphicon-download'}),
+            ' NIH Library Card Agreement'
+          ])
+        ])
```
rushtong: Do we need a file for this download?

rushtong (changes requested):
I don't think this story is ready for implementation without the previous components in place and the NIH DAA.

src/pages/dar_application/DataUseAgreements_new.js
connorlbark: yes, I asked pamela about the link and she said we are still waiting on it from NIH

src/pages/DataAccessRequestApplication_new.js
connorlbark: This is correct. I assumed that there would be a ticket for all of the final gluing together of the components, and the actual validation / submission work.

src/pages/dar_application/DataUseAgreements_new.js
rushtong: I think we have this file now in slack.

src/pages/dar_application/DataUseAgreements_new.js
```diff
@@ -0,0 +1,81 @@
...
+    attestAndSend,
+    darCode,
+  } = props;
+
+  return (
+    div({
+      className: 'dar-step-card'
+    }, [
+
+      h2({}, ['Step 4: Attestation and Agreement(s)']),
```
quazi-h: The mock has "Data Use Agreements" as the header for this section, which is not the same as the "Data Use Agreement" section on the lefthand nav bar.
Screen Shot 2022-10-06 at 4 50 37 PM

quazi-h: Another detail I realized when taking this screenshot is that each of the step cards in the mock are separate with white space between each step. Not sure if that change fits into the scope of this PR though cause we don't have that implemented for the previous steps yet either.

connorlbark:
just pushed up: _massive_ cleanup of the old code:

- dar form now a functional component
- validation errors are now strings which we can display to the user. I added little error boxes, but I am happy to change styling or display these errors other ways if preferred.
- cleanup of many of the individual fields 

this still isn't ready for re-review until the tickets in flight for this are done and merged in, but it's pretty close.

connorlbark:
@rushtong @quazi-broad Just merged this in and tested. Was able to save and submit a DAR. I'd also like to see tests made for the DAR form, but I think this ticket was already huge enough... I think a separate ticket focusing in on making tests for the whole form would make sense.

src/components/ConfirmationDialog_new.js
```diff
@@ -0,0 +1,82 @@
+import { mergeAll } from 'lodash/fp';
+import { a, div, h, h2 } from 'react-hyperscript-helpers';
```
rushtong, line 2: Thinking out loud - I've been avoiding hyperscript where possible. New components are great opportunities to ditch this and go straight react.

src/components/ConfirmationDialog_new.js
```diff
@@ -0,0 +1,82 @@
...
+        marginTop: '20px',
+      }, }, [
+        a({
+          id: 'btn_save',
+          className: 'button button-white',
+          style: {
+            marginRight: '2rem',
+          },
+          onClick: props.action.handler(false),
+          disabled: disableNoBtn
```
rushtong, line 69: This doesn't seem to work locally, see screenshot after clicking "No" and the close icon.

Screen Shot 2022-12-12 at 5 56 39 AM

src/libs/ajax.js
```diff
@@ -302,7 +302,7 @@ export const DataSet = {
   },
 
   getDatasetsByIds: async (ids) => {
-    const url = `${await getApiUrl()}/api/dataset/batch?ids=${ids.join('&ids=')}`;
+    const url = `${await getApiUrl()}/api/dataset/batch?ids=${ids.filter((id) => !isNil(id) && isNumber(id)).join('&ids=')}`;
```
rushtong: Curious about what led to this check. Do we have cases where the id being passed in is not in the right format? If so we should file a ticket to look into that case.

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -260,6 +268,26 @@ const DataAccessRequestApplicationNew = (props) => {
     }, 200));
   }, [forcedScroll]);
 
+  //Can't do uploads in parallel since endpoints are post and they both alter attributes in json column
+  //If done in parallel, updated attribute of one document will be overwritten by the outdated value on the other
```
rushtong, line 271: There was supposed to be a ticket where we refactored the DAR upload form to take the entire form, along with files, in one API call and handle this complexity on the back end. Can't find that now, but we should look for it so we can make this more efficient on the front end.

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -290,14 +318,104 @@ const DataAccessRequestApplicationNew = (props) => {
...
+      }
+    }
+    formattedFormData.userId = userId;
+
+    try {
+      //NOTE: the pre-processing saves are adding time to record generation (makes front-end seem slow)
+      //pre-prcessing saves are needed since you can't save documents without a reference id
+      //saves ensure record has a reference id
+      //actual fix would involve generating a blank draft record that is saved on console button click
+      //however that would fall outside the scope of this pr, which is already large enough due to refactored code
```
rushtong: There should always be a reference id, we should never be able to get to this page without one. At one point in time, we could generate a DAR without having a dataset, but we removed that feature and now you can only get to this page with an existing reference id. Let's touch base on this as it should be an error to get here without one.

rushtong (changes requested):
Visually, this is looking good. See comments inline.

src/components/ConfirmationDialog_new.js
rushtong: Side note, clicking "Yes" does work as expected.

src/libs/ajax.js
connorlbark: I honestly don't remember. It's been so long I will try removing it, might have been a debug thing.

src/components/ConfirmationDialog_new.js
connorlbark: Happy to do this. I much prefer JSX so I will convert this.

src/pages/DataAccessRequestApplication_new.js
connorlbark: This comment/code was copied over from the original dar form. I'll look into it, though, to see if there's a better way now.

src/pages/DataAccessRequestApplication_new.js
connorlbark: This was also copied over from the original dar form; I realize now all of these comments seem confusing, but I figured it was best to keep them. I can remove this comment if it's no longer relevant.

src/pages/DataAccessRequestApplication_new.js
connorlbark: I agree with you that it seems there isn't a way to do this. Long term, though, do we want to use FileStorageObjects here? If so this whole code path will change anyway. I also personally think from a "pure REST" perspective, creating multiple resources from one POST call might be unexpected / not desired, since there aren't really any performance concerns that I know of.

src/components/ConfirmationDialog_new.js
connorlbark: Conversion to JSX made tag build & push fail, but I am not sure why. Do you have any insight into this?

rushtong: Here is the [link to the failure](https://github.com/DataBiosphere/duos-ui/actions/runs/3687487278/jobs/6241133647). It looks like a `@mui/system` problem, but I'm not familiar with it. `npm run build` should reproduce the problem locally.

Edit: `npm run build` does not reproduce this locally, it only happens in the docker image build phase.

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -290,139 +317,245 @@ const DataAccessRequestApplicationNew = (props) => {
...
+      setShowDialogSubmit(true);
+    }
+  };
+
+  const submitDARFormData = async () => {
+    const userId = Storage.getCurrentUser().userId;
+    let formattedFormData = cloneDeep(formData);
+
+    for (var key in formattedFormData) {
+      if (formattedFormData[key] === '') {
```
otchet-broad: This will save content with only spaces.  Intended?  Might be better to consider something like noted in this SO post: https://stackoverflow.com/questions/10232366/how-to-check-if-a-variable-is-null-or-empty-string-or-all-whitespace-in-javascri

otchet-broad (commented):
Should form validation run after loading a saved DAR form so that the user can see what needs to be corrected?

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -290,139 +317,245 @@ const DataAccessRequestApplicationNew = (props) => {
...
+      if(!isNil(uploadedIrbDocument) || !isNil(uploadedCollaborationLetter)) {
+        darPartialResponse = await saveDARDocuments(uploadedIrbDocument, uploadedCollaborationLetter, referenceId);
+      }
+      batchFormFieldChange(darPartialResponse);
+      setShowDialogSave(false);
+      setDisableOkButton(false);
+    } catch(error) {
+      setShowDialogSave(false);
+      setDisableOkButton(false);
+      NotyUtil.showError('Error saving partial Data Access Request update');
```
otchet-broad: You might want to include what the user should do at this point?  Try saving the request again?

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -290,139 +317,245 @@ const DataAccessRequestApplicationNew = (props) => {
...
+      }
+      let updatedFormData = assign(formattedFormData, darPartialResponse);
+      await DAR.postDar(updatedFormData);
+      setShowDialogSubmit({
+        showDialogSubmit: false
+      }, Navigation.console(Storage.getCurrentUser(), props.history).response);
+    } catch (error) {
+      setShowDialogSubmit(false);
+      NotyUtil.showError({
+        text: 'Error: Data Access Request submission failed'
```
otchet-broad: What should the user do at this point?

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -290,139 +317,245 @@ const DataAccessRequestApplicationNew = (props) => {
...
-      div({ style: { clear: 'both' } }),
-      form({ name: 'form', 'noValidate': true, className: 'forms-v2' }, [
-        div({ className: 'multi-step-buttons-container' }, [
-          h(Tabs, {
-            value: step,
-            variant: 'scrollable',
-            scrollButtons: 'auto',
-            orientation: 'vertical',
-            TabIndicatorProps: {
+    <div className='container' style={{paddingBottom: '2%'}}>
```
otchet-broad, line 417: Thanks for the conversion away from hyperscript-helpers!

connorlbark:
> Should form validation run after loading a saved DAR form so that the user can see what needs to be corrected?

Personally, I think it would be confusing to open a form to errors. I would be happy to open this conversation up to UX people, though, once Vonne gets back.

src/pages/DataAccessRequestApplication_new.js
connorlbark: Good catch. Updated validation to catch strings with only spaces as well.

rushtong (approved):
Thanks for the changes, they look great 👍🏽

quazi-h (approved):
Changes look good to me, the new Data Use Agreement panel looks like the mock. Tested the save and submit functions on the form and the validation looks good as well.

PR #1814, opened 2022-09-26
[DUOS-2033][risk=no] Update ResearcherProfile to use FormField components

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2033

I also added the ability to show both the suggested and selected institutions when the page loads.

----

- Verify all tests go green


rushtong:
On page load, this is over-writing my existing, selected institution with the suggested values. Checked on dev and it is behaving correctly ... if I have an institution, then that is what is populated.

Screen Shot 2022-09-27 at 4 16 57 PM

connorlbark:
> On page load, this is over-writing my existing, selected institution with the suggested values. Checked on dev and it is behaving correctly ... if I have an institution, then that is what is populated.

I'm not sure what you mean by override. I changed it to where it will show your institution and also what you have previously suggested in the form if both exist, but you still have your current institution. This is because I (personally) found it confusing when I suggested an institution, and the pagre refreshed but nothing changed on the page. Thus, I added this as a way to give some feedback to the user. I am more than happy to revert if the previous approach is preferable.

rushtong:
> I'm not sure what you mean by override. I changed it to where it will show your institution and also what you have previously suggested in the form if both exist, but you still have your current institution. This is because I (personally) found it confusing when I suggested an institution, and the page refreshed but nothing changed on the page. Thus, I added this as a way to give some feedback to the user. I am more than happy to revert if the previous approach is preferable.

What I mean by override is that the institution value I have saved is not what is displayed in the institution field when the page loads. The value that is displayed is from one of the user properties which is incorrect. The two `suggested` fields are only meant to be displayed when real values are not present. For example, if there is no institution "Fred Flintstone University", I can enter that into the the field and it is saved as a user property. Later on, Pamela or Jonathan gets an email saying "Greg Rushton requested 'Fred Flintstone University'". They can then add that institution and set it as my user's institution. From then on, that is my real institution. The current approach as it works now on all environments is the correct approach so yes, let's re-implement that.

connorlbark:
> What I mean by override is that the institution value I have saved is not what is displayed in the institution field when the page loads. The value that is displayed is from one of the user properties which is incorrect. The two `suggested` fields are only meant to be displayed when real values are not present. For example, if there is no institution "Fred Flintstone University", I can enter that into the the field and it is saved as a user property. Later on, Pamela or Jonathan gets an email saying "Greg Rushton requested 'Fred Flintstone University'". They can then add that institution and set it as my user's institution. From then on, that is my real institution. The current approach as it works now on all environments is the correct approach so yes, let's re-implement that.

Ok, I understand. I still feel showing both the real institution and the suggested institution makes more sense to me as a user, since it provides feedback to the user if they suggest something, since, if not show, that value just disappears (I would think, as a user, that nothing happened). Since it was still populating and showing the real institution alongside the suggested, I thought it could be helpful feedback. That being said, I understand just showing only the current value of the institution, it just seems weird to me from a UX perspective that the user both 'selects valid institution' and 'suggests the creation of a new institution' in the same input box, since, they seem like two different operations to me, and effect different fields.

rushtong:
> That being said, I understand just showing only the current value of the institution, it just seems weird to me from a UX perspective that the user both 'selects valid institution' and 'suggests the creation of a new institution' in the same input box, since, they seem like two different operations to me, and effect different fields.

They are two different operations - but for different phases of a user's registration process. This was the intended design that was run through Yvonne/Jonathan/Pamela so if we are going to make changes at this point, we should be going through a design review. I don't have a problem if we choose to do that, but PR reviews are the wrong forum for that discussion.

connorlbark:
> They are two different operations - but for different phases of a user's registration process. This was the intended design that was run through Yvonne/Jonathan/Pamela so if we are going to make changes at this point, we should be going through a design review. I don't have a problem if we choose to do that, but PR reviews are the wrong forum for that discussion.

That is very fair - the original behavior should be returned (but it look slike some cypress errors, will work on those)

connorlbark:
@rushtong I have been investigating this cypress error but I cannot seem to get to the bottom of what is causing it. I can replicate the error locally on my e2e tests; it seems to be an error with e2e background sign in authentication. Very odd since I did not touch that side of the code base. Kind of confused at what's causing it. Rerunning it on github doesn't seem to help, so it's a replicable issue. Just no idea what's causing it. Do you have any ideas?

rushtong:
> I have been investigating this cypress error but I cannot seem to get to the bottom of what is causing it. I can replicate the error locally on my e2e tests; it seems to be an error with e2e background sign in authentication. Very odd since I did not touch that side of the code base. Kind of confused at what's causing it. Rerunning it on github doesn't seem to help, so it's a replicable issue. Just no idea what's causing it. Do you have any ideas?

I think you need to pull in the latest from dev, your branch doesn't have changes from https://github.com/DataBiosphere/duos-ui/pull/1818. That PR moves auth testing to alpha instead of perf.

connorlbark:
> I think you need to pull in the latest from dev, your branch doesn't have changes from #1818. That PR moves auth testing to alpha instead of perf.

ahhhh, that got it! thank you!

src/pages/ResearcherProfile.js
```diff
@@ -216,37 +214,56 @@ export default function ResearcherProfile(props) {
...
-            freetextValue: profile.suggestedInstitution,
             placeholder: 'Search for Institution...',
-            className: 'form-control'
-          })
+            isCreatable: true,
+            defaultValue: {
+              institutionId: selectedInstitution?.institutionId,
+              suggestedInstitution: profile.suggestedInstitution,
+              displayText: (
+                (!isNil(selectedInstitution)
```
JVThomas, line 232: I think you can shorthand this to 
```
displayText: selectedInstitution?.name || profile?.suggestedInstitution || ''
```

rushtong (approved):
👍🏽

src/pages/ResearcherProfile.js
```diff
@@ -379,42 +397,52 @@ export default function ResearcherProfile(props) {
...
-                        };
-                      }),
-                      placeholder: 'Please Select a Signing Official',
-                      searchPlaceholder: 'Search for a Signing Official...',
-                      value: profile.selectedSigningOfficialId,
-                      freetextValue: profile.suggestedSigningOfficial,
-                      className: 'form-control'
-                    })
-                  ]),
+                h(FormField, {
```
JVThomas, line 400: The form field error for `suggestedSigningOfficial` shows up if a user selects a SO from their institution. You can replicate this by selecting an SO from the list, save the changes to your profile, and then either revisit or reload the Researcher Profile Page.

Screen Shot 2022-10-05 at 8 32 47 AM

JVThomas (changes requested):
Updates are looking good! There are a few things that I've left comments on.

src/pages/ResearcherProfile.js
connorlbark: Wasn't quite able to recreate this, but I should have fixed it by only displaying the error if there is no SO id.

JVThomas: The validation error is still there. Are you checking to see if `currentSigningOfficial` is populated?

Kapture 2022-10-06 at 13 59 48

connorlbark: Hmm curious, theoretically they should be the same but I added that check too.

JVThomas (approved):
LGTM

PR #1822, opened 2022-09-29
[DUOS-1966][risk=no] NIH Administrative Data Submission Section

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1966

Note branch is `DUOS-1966-new`; kept Justin's old `DUOS-1966` branch for reference.

~I think I may have done two sections here? The mock had two sections labeled `NIH Administrative Information` but I think the second one should be labeled perhaps `Alternative Data Sharing Plan`~ Turns out these were two different sections, but the mock was wrong and thus only one ticket existed; perhaps it would have made sense to do them in separate PRs. If you all think it's easier I can split these sections into two PRs for separate review.

----

- Verify all tests go green


src/components/forms/formComponents.js
```diff
@@ -435,3 +487,46 @@ export const formInputSlider = (config) => {
...
+      ]),
+      label({
+        for: id,
+        className: 'form-file-label',
+      }, [
+        h(PublishIcon, {}),
+        uploadText,
+      ])
+    ])
+  ]);
```
connorlbark: I am not 100% sure how file uploads work, especially since I don't have a real endpoint to test / figure out how it works on. Is this the right way of creating a generic `File` form field? Does it make sense for the form's value to be `e.target.files`?

src/components/data_submission/NIHAdministrativeInformation.js
```diff
@@ -0,0 +1,339 @@
...
+        defaultValue: initialFormData?.alternativeDataSharingPlanExplanation,
+        onChange,
+        title: 'Explanation for request',
+        validators: [FormValidators.REQUIRED]
+      }),
+      h(FormField, {
+        type: FormFieldTypes.FILE,
+        title: 'Upload your alternative sharing plan.',
+        id: 'alternativeDataSharingPlanFile',
+        onChange,
```
connorlbark: Would like some guidance on how exactly this should function in conjunction with how the file `FormField` should operate. Right now, it adds the `File` object to `formData.alternativeDataSharingPlanFile`

src/components/data_submission/nih_institutions.js
```diff
@@ -0,0 +1,29 @@
+export const nihInstitutions = [
```
rushtong, line 1: This comment is more of an architectural question/suggestion. We have developed a schema to define all of these values already in Consent. It seems duplicative to define these here as well. What I think I would prefer to see is something that queries for the schema, reads it, and uses what's in the schema as opposed to doing all of that work twice. 

Caveat - the one thing that's missing right now is the ability for duos to query for the actual schema. I'm pretty sure we don't have a story for that. I'll get that ball rolling.

src/components/data_submission/NIHAdministrativeInformation.js
```diff
@@ -0,0 +1,134 @@
...
+    h(FormField, {
+      id: 'nihICsSupportingStudy',
+      title: 'NIH ICs Supporting the Study',
+      placeholder: 'Institute/Center Name',
+      onChange,
+      type: FormFieldTypes.SELECT,
+      validators: [FormValidators.REQUIRED],
+      defaultValue: initialFormData?.nihICsSupportingStudy,
+      selectOptions: nihInstitutions,
+    }),
```
rushtong: I'm wondering if this needs to be a multi-select. Can we clarify with Jonathan/Pamela? The schema does NOT support that now, so we'd need to update that if so.

src/components/data_submission/NIHAdministrativeInformation.js
```diff
@@ -0,0 +1,134 @@
...
+    h(FormField, {
+      id: 'nihInstitutionCenterSubmission',
+      title: 'NIH Institute/Center for Submission',
+      placeholder: 'Institute/Center Name',
+      onChange,
+      type: FormFieldTypes.SELECT,
+      validators: [FormValidators.REQUIRED],
+      defaultValue: initialFormData?.nihInstitutionCenterSubmission,
+      selectOptions: nihInstitutions,
+    }),
```
rushtong, line 76: Same question here.

src/components/data_submission/NIHDataManagement.js
```diff
@@ -0,0 +1,221 @@
...
+        onChange,
+        title: 'Explanation for request',
+        validators: [FormValidators.REQUIRED]
+      }),
+      h(FormField, {
+        type: FormFieldTypes.FILE,
+        title: 'Upload your alternative sharing plan.',
+        id: 'alternativeDataSharingPlanFile',
+        onChange,
+      }),
```
rushtong, line 156: This feature doesn't show the uploaded file name after selecting a file. Can we file a separate ticket for that specific aspect of the `FormField` component?

src/components/data_submission/NIHDataManagement.js
```diff
@@ -0,0 +1,221 @@
...
+          id: 'alternativeDataSharingPlanTargetDeliveryDate',
+          style: {
+            width: '45%',
+          },
+          defaultValue: initialFormData?.alternativeDataSharingPlanTargetDeliveryDate,
+          title: 'Target Delivery Date',
+          placeholder: 'Please enter date (YYYY-MM-DD)',
+          validators: [FormValidators.DATE],
+          onChange,
+        }),
```
rushtong, line 203: For date fields, we should have a date selection widget here. I think this can be filed as a separate ticket for updating `FormField`.

rushtong (approved):
Looks good. One question on requirements that would need to be answered in a separate PR and two suggestions for improvements to `FormField`

src/components/data_submission/NIHDataManagement.js
connorlbark: 👍 This is how file uploads look on the DAR form mock:
image
I think this approach makes more sense. I can do this as a separate ticket, or as a part of https://broadworkbench.atlassian.net/browse/DUOS-1755

quazi-h (approved):
The changes look good to me, the new form fields and validators are working well when I tested locally.

src/components/data_submission/NIHDataManagement.js
connorlbark: filed ticket!

src/components/data_submission/NIHDataManagement.js
connorlbark: filed ticket!

src/components/data_submission/NIHAdministrativeInformation.js
connorlbark: Just got back clarification: this should be multiselect, but the other one should stay single-value. I updated the ticket in jira to update json schema to include this change as well.

connorlbark:
@rushtong got feedback from jonathan on the field change + merge - would appreciate a quick rereview then will merge.

rushtong (approved):
👍🏽

PR #1839, opened 2022-10-12
Step 1.4 of new DAR form

quazi-h:
[DUOS-1753](https://broadworkbench.atlassian.net/browse/DUOS-1753)

- Created a new Collaborator component with Row, Summary, Form, and Error components
- CollaboratorList is the entry point, where an existing list of collaborators that are passed in the DAR form are mapped to the CollaboratorRow component.
- The CollaboratorRow component has logic to show the Summary / Form components based on a boolean.
- The CollaboratorForm component uses FormField components for the collaborator field entries.
- CollaboratorErrors does some basic validation on the text fields and validation of email input using a regex check.

Would be nice to add:
- Styling changes to gray out the Add / Save button when the values entered are not valid.
- Hover color for buttons?
- I feel like the styles for my new components are all over the place - some reference index.css, some have the styles explicitly stated - wondering what the best course of action would be to standardize those settings.

----

- Verify all tests go green


src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -174,8 +213,26 @@ export default function ResearcherInfo(props) {
             type: 'button', // default button element type inside a form is "submit".
             className: 'button button-white btn-xs',
             style: { marginTop: 25 },
-            onClick: () => {}
-          }, ['Add Collaborator'])
+            onClick: () => addNewCollaborator(),
+            onChange: ({key: name, value,}) => formFieldChange({name, value}),
```
JVThomas: `onChange` is only applied to DOM elements that can have its value changed. So if you click a radio button its `checked` value changes, or if you type into an input field its value changes whenever the text is updated. But a button has no value assigned to it so its value can never change.

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -174,8 +213,26 @@ export default function ResearcherInfo(props) {
...
+          }, [
+            span({}, ['Add Collaborator'])
+          ]),
+
+          collaboratorsState
+            .map((state, index) => {
+              if (isNil(state)) {
```
JVThomas: There shouldn't be an instance where an `undefined` value exists in any of the staff/collaborator lists. If the remove functionality is working as expected, then the target element is removed altogether from the array, reducing the total number of collaborators by one. 

If the list is initially empty, clicking `Add Collaborator` should add an element to the list with the relevant fields initialized to default values.

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -174,8 +213,26 @@ export default function ResearcherInfo(props) {
...
+                return div({}, []);
+              }
+
+              return h(CollaboratorForm, {
+                key: index,
```
JVThomas: You should pass the state/collaborator data in as well to allow the form to initialize the data with the currently saved values.

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -174,8 +213,26 @@ export default function ResearcherInfo(props) {
...
             style: { marginTop: 25 },
-            onClick: () => {}
-          }, ['Add Collaborator'])
+            onClick: () => addNewCollaborator(),
+            onChange: ({key: name, value,}) => formFieldChange({name, value}),
+          }, [
+            span({}, ['Add Collaborator'])
+          ]),
+
+          collaboratorsState
```
JVThomas: It's cleaner to create a sub-component that will take the list of staff and perform the  template updates inside of it. You can see that approach with the `CollaboratorList` component that we currently use on the DAR Form.

JVThomas: I'm not following what the purpose of `collaboratorsState` is. The list of lab staff is stored in `labCollaborators`, which is being sent in to this component as a prop. Why not use that value directly?

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -174,8 +213,26 @@ export default function ResearcherInfo(props) {
             type: 'button', // default button element type inside a form is "submit".
             className: 'button button-white btn-xs',
             style: { marginTop: 25 },
-            onClick: () => {}
-          }, ['Add Collaborator'])
+            onClick: () => addNewCollaborator(),
+            onChange: ({key: name, value,}) => formFieldChange({name, value}),
+          }, [
+            span({}, ['Add Collaborator'])
```
JVThomas: Text should be updated to say "Add Lab Member"

JVThomas (commented):
There's a couple of things I have to say about how you're managing the list itself on the `ResearcherInfo_new` component. I've outlined some of them, but I don't think those alone  are enough. 

For a reference on how to approach this problem at a high level, look at the `CollaboratorList` sub-component that we currently use on the DAR Form. Take a look at the props and useEffect hooks to understand how the data flows into it and how the component updates based on prop/state changes. If you still have questions, feel free to Slack and/or schedule a video call for further review.

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -170,12 +171,41 @@ export default function ResearcherInfo(props) {
...
+          //   id: 'addNewCollaborator',
+          //   type: 'button', // default button element type inside a form is "submit".
+          //   className: 'button button-white',
+          //   style: { marginTop: 25 },
+          //   onClick: () => {}
+          // }, ['Add Collaborator']),
+          // old style button that adds a new collaborator
+          h(CollaboratorList_new, {
+            formFieldChange,
+            collaborators: labCollaborators,
```
rushtong, line 175: I can't find where `labCollaborators` is being read from the response, and similarly for the other collaborators. Is that intentional? It would be difficult to test how this works with an existing collaborator without that population happening.

rushtong: ~Ignore the above.~ I can see that the page is loading all of the collaborators that exist after restarting/refreshing, but only in the old sections of the page, not this new 1.4 section of the page.

rushtong:
There is a bug where an existing download/approval status is not being rendered in the current form.

### Old:
Screen Shot 2022-11-07 at 3 45 05 PM

### New:
Screen Shot 2022-11-07 at 3 44 54 PM

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -388,29 +391,29 @@ export default function ResearcherInfo(props) {
...
+        //         collaborators: labCollaborators,
+        //         collaboratorKey: 'labCollaborators',
+        //         collaboratorLabel: 'Internal Lab Member',
+        //         showApproval: true,
+        //         disabled: !isEmpty(darCode),
+        //         deleteBoolArray: (new Array(labCollaborators.length).fill(false))
+        //       })
+        //     ])
+        //   ])
+        // ]),
```
rushtong: nit: remove commented code

src/pages/dar_application/collaborator/CollaboratorErrors.js
```diff
@@ -0,0 +1,57 @@
...
+  if (isNil(collaborator.email) || collaborator.email === '') {
+    errors.push('Must specify the email of the collaborator.');
+  } else {
+    var testEmail = /\S+@\S+\.\S+/.test(collaborator.email);
+    if(testEmail === false) errors.push('The email must be a valid email.');
+  }
+
+  if (showApproval) {
+    if (isEmpty(collaborator.approverStatus)) {
+      errors.push('Must specify the approval status.');
```
rushtong: nit: I was initially confused about what approval status was. I suggest we clarify the message here:
```suggestion
      errors.push('Must specify the Designated Download/Approval status.');
```

src/pages/dar_application/collaborator/CollaboratorErrors.js
```diff
@@ -0,0 +1,57 @@
...
+
+  if (isNil(collaborator.title) || collaborator.title === '') {
+    errors.push('Must specify the title of the collaborator');
+  }
+
+  if (isNil(collaborator.email) || collaborator.email === '') {
+    errors.push('Must specify the email of the collaborator.');
+  } else {
+    var testEmail = /\S+@\S+\.\S+/.test(collaborator.email);
+    if(testEmail === false) errors.push('The email must be a valid email.');
```
rushtong: Look at `utils.isEmailAddress` for a standard way of testing the validity of an email address. We should be using this single utility call so we aren't doing the same thing in different ways.

src/pages/dar_application/collaborator/CollaboratorSummary.js
```diff
@@ -0,0 +1,122 @@
...
+    div({
+      id: index,
+
+      style: cardStyle,
+    }, [
+      div({
+        style: {
+          flex: '1 1 100%',
+          marginRight: '1.5rem',
+        }
```
rushtong, line 20: Should we be pulling all of this css into a css file similarly to what we're doing for other parts of [the new dar application](https://github.com/DataBiosphere/duos-ui/tree/develop/src/pages/dar_application)?

rushtong (changes requested):
Overall, it's looking good. Still a couple issues to iron out - see comments inline.

src/pages/dar_application/collaborator/CollaboratorErrors.js
connorlbark: See also `FormValidators.EMAIL.isValid()` for maybe a more flexible way to do this. Right now it just forwards to `isEmailAddress` but it might be a more central/futureproof way of doing minor form input validation.

quazi-h: Updated my error handling to use isEmailAddress. When testing that out, however it was accepting emails with the format: email@domain (emails without a GLTD) so I adjusted the regex slightly.

quazi-h:
Fixed the bug where an existing Download/Approval status was not being displayed. This is because the old form was using booleans to track that option whereas my FormField implementation uses "yes" and "no" radio groups.

https://github.com/DataBiosphere/duos-ui/pull/1839/files#diff-5773e5fffe2ec216dc0218a0730c867178c942863cf2e2a64e506deb82a91e5eR117-R119

src/pages/dar_application/dar_application_new.css
```diff
@@ -49,5 +49,62 @@
   font-weight: 600;
   letter-spacing: 0;
   line-height: 24px;
+}
+
+/*///////////////////////////////////////////////////////COLLABORATOR*/
+
+.collaborator-summary-card {
```
rushtong: Optional feedback. The idea behind separating out CSS this way, as I understood it at the time, was intended to be component-specific. That would suggest this belong in a file called `Collaborator.css` or something similar.

rushtong (approved):
👍🏽 Looks good and is functional. One minor suggestion inline that shouldn't hold up this PR.

connorlbark (approved):
LGTM

PR #1840, opened 2022-10-12
[DUOS-2068][risk=no] Move file type to consent group

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2068

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DataSubmission/consent_group.spec.js
+++ b/cypress/component/DataSubmission/consent_group.spec.js
@@ -38,6 +38,7 @@ describe('Consent Group', function () {
     cy.get('#0_primaryConsent_hmb').check();
     cy.get('#0_col').check();
     cy.get('#0_dataLocation').type('Not Determined{enter}');
+    cy.get('#0_fileType').type('Geno{enter}');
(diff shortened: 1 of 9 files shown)


rushtong:
I'm a little confused on the implementation here. As I understand it, "Filetype" is an object with three properties, File Type, Functional Equivalence, and # of Participants. Looking at this UI, I still see the "Add New Filetype" button which, I believe, should be in the consent group box as well.

Screen Shot 2022-10-13 at 2 06 52 PM

connorlbark:
@rushtong required a decent amount of refactoring, but just got the full file type section moved to the consent group!

quazi-h (approved):
I think fields are in the correct place now - I see the three form fields "File Type*", "Functional Equivalence*", and "# of Participants*" at the bottom of the form that appears with "Add Consent Group". Prior to this change, it was possible to select an existing consent group and also provide the file types but with these changes, adding a file type is only possible when adding a new consent group. Is this intentional?

rushtong (approved):
Changes look good 👍🏽

PR #1842, opened 2022-10-13
[DUOS-1755][risk=no] Step 2 DAR Form

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1755

----

- Verify all tests go green


src/pages/DataAccessRequestApplication_new.js
```diff
@@ -1,22 +1,21 @@
 import { Component } from 'react';
```
connorlbark, line 1: there's still a lot of cleanup to do to get the new DAR form functional + well designed, but I did my best to incorporate the new step 2 section into the old code.

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,331 @@
...
+    formFieldChange,
+    formData,
+    setIrbDocument,
+    setCollaborationLetter,
+    ariaLevel = 2
+  } = props;
+
+  const [datasets, setDatasets] = useState([]);
+  const today = new Date();
+  const irbProtocolExpiration = formData.irbProtocolExpiration || `${today.getFullYear().toString().padStart(4, '0')}-${today.getMonth().toString().padStart(2, '0')}-${today.getDate().toString().padStart(2, '0')}`;
```
connorlbark, line 66: is the irb expiration always one year from the time of upload?

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,330 @@
...
+
+
+  useEffect(() => {
+    fetchAllDatasets(formData.datasetIds).then((datasets) => {
+      setDatasets(datasets);
+    });
+  }, [formData.datasetIds]);
```
connorlbark, line 73: this is _techincally_ inefficient, since it's refetching the datasets every time the user searches something new in the dataset search box. however, in practice, this seems to be robust/fast enough in my testing.

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,330 @@
...
+
+const datasetsContainDataUseFlag = (datasets, flag) => {
+  return datasets.some((ds) => ds?.dataUse[flag] === true);
+
+};
+
+const fetchAllDatasets = async (dsIds) => {
+  let datasets;
+  if (!isNil(dsIds)) {
+    datasets = await Promise.all(dsIds.map((id) => DataSet.getDatasetByIdV2(id)));
```
connorlbark: would it be worth changing the get all datasets endpoint to allow filtering to ids, so we don't have to do many calls? e.g., `/api/v2/dataset?id=1,2,3`

rushtong: I think that's a great idea for a new endpoint. We try not to modify in-use APIs when we can just create a new one and deprecate the old.

connorlbark: okay ! do you think this best be `/api/v3/dataset` or something under `/api/v2/dataset/...`?

rushtong: Not sure about either one ... Justin just did something similar here: https://github.com/DataBiosphere/consent/pull/1564
Maybe `GET /api/datasets?ids=id1,id2,id3` fits the bill?

connorlbark: Oh, I like his idea of `/api/dataset/batch?ids=1,2,3,4`. I will write up a ticket with that (subject to change) and implement it.

src/pages/DataAccessRequestApplication_new.js
connorlbark: I will do the cleanup needed to get this to properly submit & look good (code wise) as part of finishing up https://github.com/DataBiosphere/duos-ui/pull/1805

src/pages/dar_application/DataAccessRequest_new.js
connorlbark: https://github.com/DataBiosphere/consent/pull/1780 <-- once this is merged I can update this GET call.

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,330 @@
...
+  const needsGsoAcknowledgement = useCallback(() => {
+    return datasetsContainDataUseFlag(datasets, 'gsoAcknowledgement');
+  }, [datasets]);
+
+  const needsPubAcknowledgement = useCallback(() => {
+    return datasetsContainDataUseFlag(datasets, 'publicationResults');
+  }, [datasets]);
+
+  const needsDsAcknowledgement = useCallback(() => {
+    return datasetsContainDataUseFlag(datasets, 'diseaseRestrictions');
```
JVThomas: This check is a little different from the rest of the checks since the disease matters.

Let's say there are two datasets with the following DULs ...

Dataset A -> Disease Restrictions: study limited to cancer
Dataset B -> Disease Restrictions: study limited to atrial fibrillation

Both of the datasets have disease restrictions, but the disease that each focuses on makes their DULs unique.

Because of that, a request against these two datasets should trigger the question
`I acknowledge that the dataset can only be used in research consistent with the Data Use Limitations (DULs)...`.  Since you're only checking on whether `diseaseRestrictions` is true, that question will never populate.

In order to figure out if two `diseaseRestrictions` are equivalent to each other you'll need to compare the diseases that each dataset focuses on.

You can test this out by using DUOS-000045 and DUOS-000046 on dev (and local if you're using a copy of dev) when making drafts/requests

src/pages/dar_application/DataAccessRequest_new.js
JVThomas: Good question. @solideoglori, any insights?

src/pages/dar_application/DataAccessRequest_new.js
connorlbark: wait - so is the checkbox rendered if there are any disease restrictions at all (array is not empty) or is there some other condition, (diseases match or do not match dar?)?

JVThomas: The checkbox is rendered if any of the datasets have a Data Use Limitation (DUL) that is different from the other datasets selected. So let's assume...

Dataset A -> For general research use only
Dataset B -> For studies regarding population, origins, and ancestry (POA)

Both of these have different DULs, so the checkbox has to pop up since its a prompt that forces the user to acknowledge that they will follow the restrictions for each dataset rather than just picking the least restrictive DUL (General Use in this case) and applying that across all the selected datasets.

JVThomas: On top of that we need to consider that there is an implicit hierarchy for GRU, HMB, DS, and POA. Here's a diagram of the relationship below...

1-s2 0-S2666979X21000355-gr1_lrg

This comes into play when dealing with older datasets because their data is poorly formatted and may have somewhat nonsensical restrictions listed.

**So as an example**, let's say a dataset has GRU, HMB, and DS listed in their data use restriction. We can see that these options have a nested hierarchy, with DS considered a subset of HMB and HMB a subset of GRU. Since DS is the inner most subset, we evaluate that dataset as having only DS.

**A second example**: let's say that a dataset has both POA and DS selected. Because there's no nested relationship between the two, we can say that the dataset has both POA and DS as limitations. However if a dataset had GRU and POA listed, then we would evaluate it as a POA due to the nested relationship. 

**There's also this edge case to consider**: There exists a dataset with a DS restriction, but there's no diseases listed. While it doesn't make sense to say it has disease based restrictions, we can assume that there is a health/medical/biomedical slant to it, so we can treat it as an HMB.

connorlbark: I think I understand. Here is the new code:

```
  useEffect(() => {
    translateDataUseRestrictionsFromDataUseArray(datasets.map((ds) => ds.dataUse)).then((translations) => {
      setDataUseTranslations(translations);
    });

  }, [datasets]);
```

```
  const needsDsAcknowledgement = useCallback(() => {
    // if any data use translations are different, then this must be displayed.
    return dataUseTranslations.length > 1 && !dataUseTranslations.reduceRight((prevValue, currValue) => isEqual(prevValue, currValue));
  }, [dataUseTranslations]);
```

So this new code will use ontology to translate the data uses properly into the heirarchy / structure that you described and then if there are more than one dataset, the checkbox is rendered if the data uses are not uniform. Is that the correct logic, now?

JVThomas: The high level logic is correct, but using `reduceRight` is messing up your conditional. The way any reducer function works is that it iterates through the list and performs some sort of function that updates an accumulator value. The general structure of the function is 
```
reduceRight((accumulator, current) => {//function body here}, initialValue)
```

So let's say we have three DULs, all of them identical, and we use `reduceRight` as shown above.

First iteration: -> Compares `dataUseTranslations[1]` with `accumulator` with `isEqual`. Since an initialValue isn't provided, `dataUseTranslations[2]` is used by default. Since both are equal, true is returned, making `accumulator = true`

Second iteration -> Same deal as above, except we're now comparing `dataUseTranslations[0]` against `accumulator`. Since accumulator is now `true`, isEqual(object, boolean) will always be false. False is returned, which in turn sets `accumulator = false`.

Loop finishes, `dataUseTranslations.length > 1 && !dataUseTranslations.reduceRight` is essentially `true && false`. False value triggers the checkbox to render when it shouldn't.

You can test this out locally by using three or more datasets with identical DULs. You can DUOS-000005, DUOS-000007, and DUOS-000010 on dev (and local if you're using a copy of dev) to test for proper output.

connorlbark: Oh wow, that makes sense. I see why people like typescript now !! I simplified the logic (I was trying to be fancy but clearly I didn't understand fully how it worked). 

```
  const needsDsAcknowledgement = useCallback(() => {
    // if any data use translations are different, then this must be displayed.
    return dataUseTranslations.length > 1 && !dataUseTranslations.every((translation) => isEqual(dataUseTranslations[0], translation));
  }, [dataUseTranslations]);
```

New logic should be better. `DUOS-000005`, `DUOS-000007`, and `DUOS-000010` now do not render the checkbox.

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,342 @@
...
+  const needsIrbApprovalDocument = useCallback(() => {
+    return datasetsContainDataUseFlag(datasets, 'ethicsApprovalRequired');
+  }, [datasets]);
+
+  const needsCollaborationLetter = useCallback(() => {
+    return datasetsContainDataUseFlag(datasets, 'collaboratorRequired');
+  }, [datasets]);
+
+  const needsGsoAcknowledgement = useCallback(() => {
+    return datasetsContainDataUseFlag(datasets, 'gsoAcknowledgement');
```
JVThomas: This needs to check for `geneticStudiesOnly`

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,342 @@
...
+          title: '2.5 Data Use Acknowledgements',
+          description: 'Please confirm listed acknowledgements and/or document requirements below:',
+          isRendered: needsGsoAcknowledgement() || needsDsAcknowledgement() || needsPubAcknowledgement(),
+        }),
+
+        h(FormField, {
+          id: 'gsoAcknowledgement',
+          type: FormFieldTypes.CHECKBOX,
+          isRendered: needsGsoAcknowledgement(),
+          toggleText: 'I acknowledge that I have selected a dataset limited to use on genetic studies only (SO). I attest that I will respect this data use condition.',
```
JVThomas: The phrasing needs to be `... genetic studies only (GSO)...`

JVThomas:
Another bug found: If a user has a file tied to a DAR, clicks on the file upload button, and then exits out of the file selection window, the application errors out

Kapture 2022-10-27 at 08 02 45

connorlbark:
> Another bug found: If a user has a file tied to a DAR, clicks on the file upload button, and then exits out of the file selection window, the application errors out

Good catch! Simple fix pushed up.

src/pages/dar_application/DataAccessRequest_new.js
connorlbark: added into this PR

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -0,0 +1,338 @@
...
+        }),
+        div({
+          isRendered: needsIrbApprovalDocument(),
+          style: {
+            display: 'flex',
+            justifyContent: 'space-between',
+          }
+        }, [
+          div({}, [
+            h(FormField, {
```
JVThomas, line 308: The FormField.FILE components need to have `defaultValues` as well if the user has already uploaded a file to their draft DAR.

connorlbark: should have default values now!

JVThomas (approved):
Great work! This step's the most complicated out of the group due to the DULs and file processing, but it looks like the core logic is working as expected. Once the rest of the components are up you can check to see if saving and submission works as intended.

quazi-h (approved):
The new step 2 panel looks good and is functioning properly for me when running locally. The  logic for the new components looks good to me.

PR #1844, opened 2022-10-14
[DUOS-2125][risk=no] Add Collapsible DAC Member Vote

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2125

The quick mocks show putting this section in a card, but I think the way I implemented is a bit cleaner in context of the rest of the page. Uses the same vote table as before, but updates language to make it clear that it's the dac member votes and not the chair votes. Adds the vote table to the dac member's vote page. I also added this to the member research purpose vote tab, which wasn't in the spec but seemed to be a logical need.

----

- Verify all tests go green


quazi-h (approved):
These changes are working well for me locally. Checked a few different DAR review pages and observed the collapsible DAC Member Votes component in the Chair and Member slabs, including the RUS slabs. The only nit I have is that the styles do not match up with the mocks on the ticket but I expect that is outside of the scope of this ticket.

JVThomas (changes requested):
Code looks good, but there needs to be an adjustment visually to make the dropdown icon stick out a little more. Making the sub-header and icon bigger should be a quick fix for that

Here's how it currently looks like...
Screen Shot 2022-10-18 at 8 32 04 AM

And here's how scaling up the size can help make the icon more visible...
Screen Shot 2022-10-18 at 8 23 14 AM

connorlbark:
@JVThomas scaled the icon, should be better now! (turns out svg scaling is kind of complicated and doesn't always work as expected... but luckily all i needed was `!important`)

JVThomas (approved):
LGTM

PR #1845, opened 2022-10-17
DUOS-1740[risk=no] Upgrade Cypress to 10.10 and fix failing test

JVThomas:
Addresses [DUOS-2159](https://broadworkbench.atlassian.net/browse/DUOS-2159), but is labelled under DUOS-1740 due to it being a dependency update.

PR upgrades Cypress from 10.9 to 10.10, causing a previously passing test to fail locally. Upon further inspection, the failing tests was missing a stub, indicating that 10.9 failed to correctly evaluate that particular test. Adding the stub corrects the error and causes the test to pass.

There needs to be additional investigations as to why the Github Action also failed to notice the missing stub, but that should be done as a separate task.

- Verify all tests go green

### DIFF
--- a/cypress/component/DataSubmission/data_access_governance.spec.js
+++ b/cypress/component/DataSubmission/data_access_governance.spec.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-undef */
 import React from 'react';
-import { DAC, User } from '../../../src/libs/ajax';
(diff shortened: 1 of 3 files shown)


connorlbark (approved):
👍 works for me locally

quazi-h (approved):
Changes look good to me

PR #1848, opened 2022-10-17
[DUOS-2025][risk=no] Remove IT / SO director fields from applicant information section

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2025

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DarCollectionReview/application_information.spec.js
+++ b/cypress/component/DarCollectionReview/application_information.spec.js
@@ -197,32 +197,32 @@ describe('Application Information', () => {
     requestSpan.contains('Person E, Person F');
   });
 
-  it('renders the signing official and signing official email', () => {
+  it('renders the signing official and signing official', () => {
     const props = {
       signingOfficialEmail: 'test@test.com',
     };
     mount(<ApplicationInformation {...props} />);
 
-    const emailLabel = cy.get('#signing-official-email-label');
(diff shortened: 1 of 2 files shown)


quazi-h (approved):
Changes look good so far, running locally, the SO and IT Director have a single field each that has the name and email for the individuals. One nit I have is that it might be clearer to have those fields labeled as "Signing Official" and "IT Director" without the "Email" in the header since the value of those fields contains both name and (if available) email, but otherwise looks good and we don't have any duplicate data showing up under the Applicant Information slab.

JVThomas:
> Changes look good so far, running locally, the SO and IT Director have a single field each that has the name and email for the individuals. One nit I have is that it might be clearer to have those fields labeled as "Signing Official" and "IT Director" without the "Email" in the header since the value of those fields contains both name and (if available) email, but otherwise looks good and we don't have any duplicate data showing up under the Applicant Information slab.

Agreed. It was included as a note in the ticket itself

```
Note: Remove the email columns for SO and IT Director. These values are essentially the same value, just displayed in two places. 
```

connorlbark:
ohhh good catch! just updated the labels @quazi-broad @JVThomas

JVThomas (approved):
Code looks good to me (tested locally)

PR #1863, opened 2022-10-25
[DUOS-2048][risk=no] Add DAR Dataset Dropdown Table

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2048

----

- Verify all tests go green

### DIFF
--- a/src/components/SimpleTable.js
+++ b/src/components/SimpleTable.js
@@ -3,23 +3,16 @@ import { div, h } from 'react-hyperscript-helpers';
 import { Styles } from '../libs/theme';
 import ReactTooltip from 'react-tooltip';
 import { ArrowDropUp, ArrowDropDown } from '@material-ui/icons';
+import { SpinnerComponent } from '../components/SpinnerComponent';
+import loadingImage from '../images/loading-indicator.svg';
 
-//Component that renders skeleton loader on loading
-const SkeletonLoader = ({columnRow, columnHeaders, baseStyle, tableSize}) => {
-  const rowTemplateArray = (columnRow, columnHeaders) => {
-    const rowsSkeleton = [columnRow];
-    let i = 0;
-    while(i < tableSize) {
-      let row = columnHeaders.map(({cellStyle}, index) => {
-        const style = Object.assign({height: '0.5rem'}, baseStyle, cellStyle);
-        return div({style, className: 'text-placeholder', key: `placeholder-row-${i}-cell-${index}`});
-      });
-      rowsSkeleton.push(div({style: baseStyle, key: `placeholder-row-${i}-container`, className: `placeholder-row-${i}`}, row));
-      i++;
-    }
-    return rowsSkeleton;
-  };
-  return rowTemplateArray(columnRow, columnHeaders);
+// Renders spinning circle while table loading
+const TableLoading = () => {
+  return h(SpinnerComponent, {
+    show: true,
+    name: 'loadingSpinner',
+    loadingImage
+  }, []);
 };
 
 //Simple cell text display
@@ -109,7 +102,6 @@ export default function SimpleTable(props) {
     isLoading,
     styles, //styles -> baseStyle, columnStyle needed to determine sizing and color assignments
     rowWrapper = ({renderedRow}) => renderedRow, // ({rowData, renderedRow}) => ... : allows injecting custom container to row
-    tableSize,
     paginationBar,
     sort,
     onSort
@@ -118,7 +110,7 @@ export default function SimpleTable(props) {
   const {baseStyle, columnStyle, containerOverride} = styles;
   const columnRow = h(ColumnRow, {key: 'column-row-container', columnHeaders, baseStyle, columnStyle, sort, onSort});
   const tableTemplate = [columnRow, h(DataRows, {rowData, baseStyle, columnHeaders, rowWrapper, key: 'table-data-rows'})];
-  const output = isLoading ? h(SkeletonLoader, {columnRow, columnHeaders, baseStyle, tableSize}) : tableTemplate;
+  const output = isLoading ? h(TableLoading, {}) : tableTemplate;
   return div([
     div({className: 'table-data', style: containerOverride || Styles.TABLE.CONTAINER, role: 'table'},
       [output, isNil(paginationBar) ? div() : paginationBar]),
--- a/src/components/SimpleTable.js
+++ b/src/components/SimpleTable.js
@@ -8,11 +8,15 @@ import loadingImage from '../images/loading-indicator.svg';
 
 // Renders spinning circle while table loading
 const TableLoading = () => {
-  return h(SpinnerComponent, {
-    show: true,
-    name: 'loadingSpinner',
-    loadingImage
-  }, []);
+  return div({className: 'table-loading-placeholder'},
+    [
+      h(SpinnerComponent, {
+        show: true,
+        name: 'loadingSpinner',
+        loadingImage
+      }, [])
+    ]
+  );
 };
 
 //Simple cell text display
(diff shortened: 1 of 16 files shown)


src/components/SimpleTable.js
```diff
@@ -90,8 +90,10 @@ const DataRows = ({rowData, baseStyle, columnHeaders}) => {
         }
         return output;
       }));
+
+    return rowWrapper({renderedRow, rowData: row});
```
connorlbark, line 91: I added the table here by adding a function which can inject HTML around a row. I considered a lot of different options, but this seemed the cleanest / least invasive way of achieving adding code inside of the outer DAR collection table itself.

src/components/dar_collection_table/DarCollectionTableCellData.js
```diff
@@ -1,6 +1,7 @@
 import {includes, isEmpty, isNil, toLower} from 'lodash/fp';
 import {formatDate} from '../../libs/utils';
-import {h} from 'react-hyperscript-helpers';
+import {h, div, p} from 'react-hyperscript-helpers';
+import { ArrowDropUp, ArrowDropDown } from '@material-ui/icons';
```
connorlbark: Are there standardized or better drop up / drop down icons ? I used these (as I did somewhere else) but I am not sure they are the most attractive. I couldn't find other usages of drop up/drop down icons in the codebase and struggled to find better ones.

src/components/dar_dataset_table/DarDatasetTableCellData.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export function votesCellData({elections, votes, dataUseGroup, label= 'votes'}) {
+  const dataAccess = elections?.find((election) => election.electionType === 'DataAccess');
+
+  let status = dataAccess?.status || 'N/A';
+
+  if (status === 'Closed') {
+    status = votes?.finalVotes?.some((v) => v.vote === true) ? 'Approved' : 'Denied';
+  }
```
connorlbark: I believe it is, but I just want to check that this is sound logic for the status of a bucket.

rushtong: There could be null vote values for votes that are still in progress. That would be neither approved or denied.

rushtong: If any final vote is true, then it is definitely `Approved`. If any final vote is false, then it is definitely `Denied`

connorlbark: So, would it just be 'In Progress' then?

src/components/dar_collection_table/DarCollectionTableCellData.js
rushtong: I don't think we do have something like that. Perhaps run it by Yvonne for her thoughts.

src/components/dar_dataset_table/DarDatasetTableCellData.js
rushtong: Another possibility is `Canceled` where an election was closed with no final vote values. I think the logic should be something like this:

If any final votes are non-null && true -> `Approved`
If any final votes are non-null && false -> `Denied`
Else - there is no final vote status to report.

rushtong: The above logic is a little too loose in that there's only Approved and Denied, but we do have that third case.

connorlbark: Right now, I am displaying `Canceled`, `Open`, etc. as well since it's deriving the status from the original election status. In this table, do we prefer just showing `Approved` and `Denied`, or do we want to be more specific in the case neither is true, e.g. `Open` `Canceled` etc.

connorlbark: > The above logic is a little too loose in that there's only Approved and Denied, but we do have that third case.

Oh okay, cool. Got it. I replied before this, I think this makes sense.

src/components/dar_collection_table/DarCollectionTableCellData.js
connorlbark: I found one, used in the code + looks like the mock.

src/components/collection_voting_slab/MemberVoteSummary.js
```diff
@@ -28,7 +28,7 @@ export const MemberVoteSummary = (props) => {
       }
     }, [
       title,
-      h((showMemberVotes? ArrowDropUp : ArrowDropDown), {
+      h((showMemberVotes? ExpandLess : ExpandMore), {
```
connorlbark, line 31: Updated the arrow drop up/dropdown that I made in a recent PR to be consistent with the dropup/dropdown added in this PR (looks better in my opinion).

src/components/SimpleTable.js
```diff
@@ -106,6 +108,7 @@ export default function SimpleTable(props) {
     rowData = [], //rowData -> {data, component, style, onClick}
     isLoading,
     styles, //styles -> baseStyle, columnStyle needed to determine sizing and color assignments
+    rowWrapper = ({renderedRow}) => renderedRow, // ({rowData, renderedRow}) => ... : allows injecting custom container to row
```
rushtong, line 108: One thing that's jarring is the styling of the loading phase before the actual content is rendered. Not sure how that's defined, but I think we need to make that more of a spinner than a giant table that goes away within a second.

src/components/collection_voting_slab/MemberVoteSummary.js
rushtong: This looks good to me.

src/components/dar_collection_table/DarCollectionTable.js
```diff
@@ -199,6 +207,38 @@ export const DarCollectionTable = function DarCollectionTable(props) {
...
+  }, [collectionsExpandedState]);
+
+  const collectionIsExpanded = useCallback((id) => {
+    return collectionsExpandedState[id] === true;
+  }, [collectionsExpandedState]);
+
+  const fetchDarCollection = useCallback((darCollectionId) => {
+    if (!isNil(darCollectionCache[darCollectionId])) {
+      return darCollectionCache[darCollectionId];
+    } else {
```
rushtong, line 225: I don't think this is working as intended. When I open/close/open a dar row, the entire collection is re-fetched a second time.

rushtong (commented):
So far, this looks amazing! See comments inline. Still looking at the PR but wanted to get this out to you sooner.

src/components/dar_collection_table/DarCollectionTable.js
rushtong: Follow-on to this, I can see in the console that the collection is stored in the cache so it looks like there's a lookup problem as opposed to a caching problem.

connorlbark: Actually, I think the collection caching is working, or at least it is working for me locally when monitoring local traffic. The issue is that, when bucketing, there are some ontology lookups (looks up disease by DOID). To cache that, we would have to move the bucketing to be outside the table. It would be a bit awkward code-wise but I am happy to do that if preferred.

src/components/SimpleTable.js
connorlbark: I changed the table loading animation to be a spinner in the middle of the screen, and the table is simply empty until the data is fetched. This effects all tables, but I think it looks pretty good, even for the larger tables. If preferred, we could make it a flagged prop if you want the giant table loading animation on the, e.g., admin user page

src/components/dar_collection_table/DarCollectionTable.js
rushtong: Ah, that makes sense - the duplicate calls looked suspicious.

src/components/SimpleTable.js
rushtong: This looks good, thanks for the change!

rushtong (approved):
👍🏽

JVThomas (approved):
The dataset tables look good!

PR #1877, opened 2022-11-03
[DUOS-2158][risk=no] Swap default pages for researcher tab

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2158

The default page for the researcher console is now set to the dataset catalog instead of the user's DARs.

### Old
Screen Shot 2022-11-03 at 5 31 43 AM

### New
Screen Shot 2022-11-03 at 5 31 49 AM

----

- Verify all tests go green

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -124,7 +124,7 @@ export const headerTabsConfig = [
   {
     label: 'Researcher Console',
     link: '/dataset_catalog',
-    search: 'researcher_console',
+    search: 'dataset_catalog',
     children: [
       { label: 'Data Catalog', link: '/dataset_catalog' },
       { label: 'DAR Requests', link: '/researcher_console' }
@@ -181,7 +181,7 @@ const NavigationTabsComponent = (props) => {
               to: {
                 pathname: tab.link,
                 state: {
-                  selectedMenuTab: selectedMenuTab,
+                  selectedMenuTab: tabIndex,
                 },
               },
               component: Link
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -123,11 +123,11 @@ export const headerTabsConfig = [
   },
   {
     label: 'Researcher Console',
-    link: '/researcher_console',
+    link: '/dataset_catalog',
     search: 'researcher_console',
     children: [
-      { label: 'DAR Requests', link: '/researcher_console' },
-      { label: 'Data Catalog', link: '/dataset_catalog' }
+      { label: 'Data Catalog', link: '/dataset_catalog' },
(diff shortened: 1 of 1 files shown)


src/components/DuosHeader.js
```diff
@@ -123,11 +123,11 @@ export const headerTabsConfig = [
...
   {
     label: 'Researcher Console',
-    link: '/researcher_console',
+    link: '/dataset_catalog',
     search: 'researcher_console',
     children: [
-      { label: 'DAR Requests', link: '/researcher_console' },
-      { label: 'Data Catalog', link: '/dataset_catalog' }
+      { label: 'Data Catalog', link: '/dataset_catalog' },
+      { label: 'DAR Requests', link: '/researcher_console' }
```
connorlbark, line 130: is `/researcher_console` an inaccurate url now that it isn't the default? I could see this being a big change though, considering we likely have links pointing to this page in many places, so maybe not worth changing.

connorlbark:
The admin console still has the `DAR Requests` first but when you click on `Researcher` it acts as if it is still under the Admin header, like this:
image
fixing the navbar might be outside of the scope of this ticket, but just something I noticed which this ticket makes look awkward.

connorlbark (approved):
other than the thoughts above (which don't really have any action items for this ticket), this looks good!

quazi-h (commented):
Change looks pretty straightforward herebut I'm noticing the same behavior that Connor pointed out.

Clicking on the Researcher Console tab in the header navigates me to https://local.broadinstitute.org:3000/dataset_catalog but the Header shows one of the other consoles being selected. If already in the DAC Chair Console or DAC Member Console, clicking on the "Researcher Console" switches to one of the "Datasets" tab within the Chair or Member consoles. Otherwise it navigates to the Admin Console / Dataset Catalog and I can not access the Researcher Console / DAR Requests tab. Not sure how to address this.

src/components/DuosHeader.js
rushtong: > is /researcher_console an inaccurate url now that it isn't the default?

Changing that is outside the scope of what I'm trying to do here. The only goal is to arrange the default tabs for the researcher.

rushtong:
> fixing the navbar might be outside of the scope of this ticket, but just something I noticed which this ticket makes look awkward.

There is some weirdness with the selected tab state - I'll dig into that.

rushtong:
@quazi-broad @connorlbark - Connor fixed the bug, PTAL

quazi-h (approved):
The Researcher Console tabs are working correctly with the new changes. Everything looks good!

connorlbark (approved):
👍 updating approval

PR #1881, opened 2022-11-04
[DUOS-2163][risk=no] Add NIH Insitutional Certification Upload to Data Submission Consent Group

connorlbark:
### Addresses

I'm not exactly sure how file uploads will work for the Data Submission Form. As it is, this will expose a file object for each consent group, which can be used in any way once we're ready to send the form data.

https://broadworkbench.atlassian.net/browse/DUOS-2163

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DataSubmission/consent_group.spec.js
+++ b/cypress/component/DataSubmission/consent_group.spec.js
@@ -9,6 +9,9 @@ const props = {
   parentConsentGroup: {},
   saveConsentGroup: () => {},
   deleteConsentGroup: () => {},
+  nihInstitutionalCertificationFile: null,
+  updateNihInstitutionalCertificationFile: () => {},
+
 };
 
 let propCopy = {};
--- a/cypress/component/DataSubmission/data_access_governance.spec.js
+++ b/cypress/component/DataSubmission/data_access_governance.spec.js
@@ -27,21 +27,21 @@ describe('Data Access Governance', function () {
   it('Only renders consent group info if closed access', function () {
     mount(<DataSubmissionForm />);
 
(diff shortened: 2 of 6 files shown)


rushtong:
> I'm not exactly sure how file uploads will work for the Data Submission Form.

You can look at the existing stub API to see this in action. It will take any number of files and validate their name and size per our standard file processing logic. That's all it does right now ... we need to expand the stub API to do the actual work of saving the files appropriately.

connorlbark:
> You can look at the existing stub API to see this in action. It will take any number of files and validate their name and size per our standard file processing logic. That's all it does right now ... we need to expand the stub API to do the actual work of saving the files appropriately.

Ahh yes, I see this now. The only thing here is that the current data submission form does not have a submit button. I am not sure if that is within the scope of this ticket to add.

quazi-h (approved):
New file upload for new Consent Groups under the Data Access Governance on the new data_submission_form looks good and is working well when testing locally. I couldn't find a related mock on Invision to reference but the styling looks good to me.

rushtong:
> > You can look at the existing stub API to see this in action. It will take any number of files and validate their name and size per our standard file processing logic. That's all it does right now ... we need to expand the stub API to do the actual work of saving the files appropriately.
> 
> Ahh yes, I see this now. The only thing here is that the current data submission form does not have a submit button. I am not sure if that is within the scope of this ticket to add.

I think you're right, enabling the form is not in scope here.

rushtong (approved):
👍🏽 Looks good in testing. I have concerns about this approach that need to be hashed out separately so they should not hold up this PR.

PR #1882, opened 2022-11-07
[DUOS-2023][risk=no] Adds download links for COL and IRB provided docs.

otchet-broad:
[DUOS-2023](https://broadworkbench.atlassian.net/browse/DUOS-2023) requests links are added to DARs where the applicant provided a COL or IRB document.
DarCollectionReview was modified to pass in the necessary information to ApplicationInformation to build links to the documents.
ajax was updated to include a download function for enabling document download with the required headers.
applicaiton_information.spec was updated to include tests to ensure the links were rendered correctly.

BEFORE
Screen Shot 2022-11-07 at 1 12 29 PM

AFTER
Screen Shot 2022-11-07 at 1 12 40 PM

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -260,6 +260,18 @@ export const DAR = {
     return res.data;
   },
 
+  downloadDARDocument: async (referenceId, fileType, fileName) => {
+    const authOpts = Object.assign(Config.authOpts(), {responseType: 'blob'});
+    authOpts.headers = Object.assign(authOpts.headers, {
+      'Content-Type': 'application/octet-stream',
+      'Accept': 'application/octet-stream'
+    });
+    const url = `${await getApiUrl()}/api/dar/v2/${referenceId}/${fileType}`;
+    axios.get(url, authOpts).then((response) =>{
+      fileDownload(response.data, fileName);
+    });
+  },
+
   //NOTE: endpoints requires a dar id
   uploadDARDocument: async(file, darId, fileType) => {
     if(isFileEmpty(file)) {
--- a/src/pages/dar_collection_review/DarCollectionReview.js
+++ b/src/pages/dar_collection_review/DarCollectionReview.js
@@ -65,6 +65,7 @@ export default function DarCollectionReview(props) {
   const collectionId = props.match.params.collectionId;
   const [collection, setCollection] = useState({});
   const [darInfo, setDarInfo] = useState({});
+  const [referenceIdForDocuments, setReferenceIdForDocuments] = useState();
   const [isLoading, setIsLoading] = useState(true);
   const [subcomponentLoading, setSubcomponentLoading] = useState(true);
   const [tabs, setTabs] = useState({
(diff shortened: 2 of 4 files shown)


src/pages/dar_collection_review/DarCollectionReview.js
```diff
@@ -82,6 +83,7 @@ export default function DarCollectionReview(props) {
       const collection = await Collections.getCollectionById(collectionId);
       const { dars, datasets } = collection;
       const darInfo = find((d) => !isEmpty(d.data))(collection.dars).data;
+      const referenceIdForDocuments = find((d) => !isEmpty(d.referenceId))(collection.dars).referenceId;
```
otchet-broad, line 86: In speaking with @JVThomas, any one of the collection of DARs should point to the proper referenceId that is needed to obtain the documents from the v2 endpoint.

src/libs/ajax.js
```diff
@@ -260,6 +260,18 @@ export const DAR = {
     return res.data;
   },
 
+  downloadDARDocument: async (referenceId, fileType, fileName) => {
```
rushtong, line 263: The method right above this one (line 252 `getDARDocument`), is what's called when we download the file from the DAR form. Currently, it uses the UUID of the file in GCS as the file name which is pretty poor UX. Can we either refactor that (see usages in `UploadLabelButton`) or file a ticket to do that?

src/pages/dar_collection_review/DarCollectionReview.js
rushtong: That is correct. Alternatively, we can add parallel APIs at the collection level that do the same thing. I'm fine with this approach.

rushtong (approved):
This looks and works great, thank you! One suggestion inline that doesn't have to hold up the PR and can be handled separately.

src/libs/ajax.js
otchet-broad: I'm glad you caught this.  I want to do a bigger refactor here and recommend we make a new ticket for that.  May I do that?

Wherever we're using "createObjectURL" and we're not using revokeObjectURL to remove it, we're potentially leaking memory.  The `fileDownload` utility I've used does that for us automatically.  See: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL


This happens in `UploadLabelButton`, `downloadDataSets`, and in utils `download`

And to your point, I'd also like to make sure we're providing file names whenever possible.

otchet-broad: See [DUOS-2190](https://broadworkbench.atlassian.net/browse/DUOS-2190)

rushtong: > I'm glad you caught this. I want to do a bigger refactor here and recommend we make a new ticket for that. May I do that?

Absolutely. See also https://broadworkbench.atlassian.net/browse/DUOS-2188 where I'd like to do a much better job of handling DAR files.

quazi-h (approved):
The changes look good and I can see IRB and Collaboration Letter download links on DARs that have them. When testing locally, I wasn't able to download anything from DAR-586 but I think that might be because it's a Read-Only for me? After creating a new DAR with IRB/Collaboration files, I was able to download them from the dar_collection_review page.

The console revealed that we were hitting 403 errors when attempting to use those links on a DAR that the current user did not create:
`GET https://local.broadinstitute.org:3000/api/dar/v2/e9340c8c-f4dc-48d4-8f54-627d8a5d1754/irbDocument 403 (Forbidden)`
This is because of the checkAuthorizedUpdateUser in DataAccessRequestResourceVersion2.java. We may need to revise the backend logic to implement a new check that looks to see if the current user is in a list of approved users who should have access to that particular DAR.

otchet-broad:
Great find @quazi-broad!  I'm going to create a PR to address the backend issue.

After speaking with @rushtong, we determined that members, chairs and admins should be able to see the contents.

I'll park this PR until the backend changes are approved.

PR #1895, opened 2022-11-15
[DUOS-1401][risk=low] Updated google authentication

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1401

This PR replaces the existing google authentication scheme with a purely access-token-based approach.
See https://developers.google.com/identity/oauth2/web/guides/migration-to-gis#gis-only for background.

----

- Verify all tests go green

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -22,6 +22,7 @@
   <link rel="icon" type="image/png" href="%PUBLIC_URL%/images/favicon/favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="%PUBLIC_URL%/images/favicon/favicon-16x16.png" sizes="16x16" />
   <link rel="mask-icon" href="%PUBLIC_URL%/images/favicon/safari-pinned-tab.svg">
+  <script src="https://accounts.google.com/gsi/client" async defer id="gsiClient"></script>
   <!--
       Notice the use of %PUBLIC_URL% in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
--- a/cypress/component/SignIn/google_is.spec.js
+++ b/cypress/component/SignIn/google_is.spec.js
@@ -1,0 +1,19 @@
+/* eslint-disable no-undef */
+
+import {GoogleIS} from '../../../src/libs/googleIS';
+
+describe('Google IS Utility', function () {
+  it('GoogleIS.initTokenClient works with a valid client id', function () {
+    cy.viewport(600, 300);
+    cy.readFile('config/alpha.json').then((config) => {
+      const clientId = config.clientId;
+      GoogleIS.initTokenClient(
+        clientId,
+        () => {},
+        () => {})
+        .then(() => {
+          expect(GoogleIS.client).to.not.equal(null);
+        });
+    });
+  });
+});
--- a/cypress/component/SignIn/google_is.spec.js
+++ b/cypress/component/SignIn/google_is.spec.js
@@ -16,4 +16,33 @@ describe('Google IS Utility', function () {
         });
     });
   });
+
+  it('GoogleIS.requestAccessToken: Cannot be tested as it would trigger an OAuth flow', function () {
+    expect(true).to.equal(true);
+  });
+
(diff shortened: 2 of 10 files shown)


otchet-broad:
@rushtong Any idea if other Terra applications are keeping the same button or moving to One Tap?

rushtong:
> @rushtong Any idea if other Terra applications are keeping the same button or moving to One Tap?

Terra has moved to a different oauth library for Azure/B2C support, but not One Tap:
* https://github.com/authts/react-oidc-context
* https://github.com/DataBiosphere/terra-ui/pull/3117

That approach [requires some infrastructure](https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/auth.js#L25) that we don't have in Consent at the moment so I did not pursue it ... yet. It definitely is something we need to look into, but I see that as a moderately larger task that what I'm trying to do in this PR.

connorlbark (approved):
As far as I can tell (not an expert on Google sign in), this looks good and works great. 

Pretty minor, but I was wondering about the signin button beneath Data Access Request Application:

image

Originally, this button was the same as if you were logged in. Since it was changed to being a sign in button, I could see this being a little bit confusing/odd from an end user perspective: would it be possible to change it back, or to add some clarifying text, like `Please sign in to submit a Data Access Request` or something.

otchet-broad:
Nice work.

The code led me down a bit of a reading trail and I ended up on this Google page here: 
[https://developers.google.com/identity/oauth2/web/guides/overview](https://developers.google.com/identity/oauth2/web/guides/overview)

Had you seen this and already considered the note on that page?
 
Screen Shot 2022-11-16 at 1 16 13 PM

If I understood it correctly, it suggests an alternate API because of the limited scope being used to Authenticate.

public/index.html
```diff
@@ -22,6 +22,7 @@
   <link rel="icon" type="image/png" href="%PUBLIC_URL%/images/favicon/favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="%PUBLIC_URL%/images/favicon/favicon-16x16.png" sizes="16x16" />
   <link rel="mask-icon" href="%PUBLIC_URL%/images/favicon/safari-pinned-tab.svg">
+  <script src="https://accounts.google.com/gsi/client" async defer id="gsiClient"></script>
```
otchet-broad, line 25: defer for the win!  well done.

rushtong:
> Originally, this button was the same as if you were logged in. Since it was changed to being a sign in button, I could see this being a little bit confusing/odd from an end user perspective: would it be possible to change it back, or to add some clarifying text, like `Please sign in to submit a Data Access Request` or something.

Great catch, I totally missed that. Will fix it - it should be a button that looks like this:
Screen Shot 2022-11-16 at 1 50 12 PM

rushtong:
> If I understood it correctly, it suggests an alternate API because of the limited scope being used to Authenticate.

I have looked at many of the different use cases. The "Sign In With Google" use case in that link is not exactly what we want. We could make use of it, but then we would only have access to a JWT with basic user information (good). Our use case is to request an access token so we can make back end API calls. If we did use the "Sign In" approach, we would then be required to prompt the user again for an access token since "Sign In" does not grant that. 

As per this content on [their page](https://developers.google.com/identity/oauth2/web/guides/overview):
> Authentication for user sign-in, and authorization to obtain an access token to call Google APIs, now have two separate and distinct user flows; one for [sign-in](https://developers.google.com/identity/gsi/web/guides/overview#how_it_works) and another for [consent](https://developers.google.com/identity/oauth2/web/guides/how-user-authz-works#user_consent) during authorization, with separate user flows to clearly differentiate who you are, from what an app can do.

We really only care about the `consent` case.

rushtong:
@connorlbark - fixed the customized sign in button on the home page under DAR Application, PTAL, thank you!

otchet-broad (commented):
Nice work.  This takes care of the immediate need.  Greg and I are continuing to discuss Authentication outside of this PR.

connorlbark (approved):
New sign in styling looks good to me.

otchet-broad (approved):

PR #1901, opened 2022-11-17
Update Internal and External Collaborator components on the new DAR Form

quazi-h:
[DUOS-1754](https://broadworkbench.atlassian.net/browse/DUOS-1754)
- Implemented the new collaborator components for Internal and External collaborator on the new DAR form:
  - 1.5 Internal Collaborators
  - 1.9 External Collaborators
- Removed the old collaborator components.
- Adjusted styling.

Feedback on style changes would be appreciated, changes look close to the [Invision mock](https://projects.invisionapp.com/d/main#/console/21974870/465993358/preview) but could use some improvement.

----

- Verify all tests go green


src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -182,7 +181,6 @@ export default function ResearcherInfo(props) {
         ]),
 
         div({className: 'dar-application-row'}, [
-          // TODO: DUOS-1754
           h3('1.5 Internal Collaborators'),
           div(
             `Please add Internal Collaborators here Internal Collaborators are defined as individuals
```
otchet-broad:
```suggestion
            `Please list Internal Collaborators here. Internal Collaborators are defined as individuals
```

otchet-broad: Just a suggestion to bring the text in line with the External Collaborators text.

otchet-broad:
Nice work!

Code looks good and runs well.

Some notes:

- We might want to consider is that we now have three buttons with the same text.  It might be worth a quick discussion to consider tailoring the name to each section because some accessibility tools will not provide enough context for a user with a screen reader to know what they are clicking on.  We might want to call the 1.4 button Add Lab Staff, the 1.5 button Add Internal Collaborator and the 1.9 Add External Collaborator.
- 1.4 Heading is Internal Lab Staff, but the form opens and labels the input form "New Internal Lab **Member** Information"
- Maybe we can catch up about input validation and tests in person tomorrow.

quazi-h:
> 
I did notice that discrepancy between the Internal Lab Member/Staff - not sure which one we should default to since the mock/wireframe has them both as well.

I updated the buttons to use the collaboratorLabel so that the text in the buttons includes the specific type of collaborator.

connorlbark:
> * Maybe we can catch up about input validation and tests in person tomorrow.

I agree with adding input validation and tests! As far as validation: My data use agreements panel PR has a pretty big overhaul of validation so it might make sense for that to hold off and merge as-is and then I can go in and add validation stuff. As far as tests, I definitely think that couldn't hurt. I think maybe adding a couple tests to make sure this component functions properly would be great, but as-is I would be happy approving this to get the DAR form rolling.

quazi-h:
Added extensive tests for the new collaborator components! Would appreciate feedback on the changes I made to the cypress commands.js file - was not able to get cypress tests passing without those additions.

connorlbark:
I think you can actaully get rid of the added Cypress commands (dangerous to allow any exception) by stubbing `User.getMe`:

```
const user = { ... };

beforeEach(() => {
  cy.stub(User, 'getMe').returns(user);
});
```

See `cypress/component/DataSubmission/ds_study_information.spec.js`. It looks like all of the exceptions are from the fact that it is trying to hit the backend.

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -196,12 +194,15 @@ export default function ResearcherInfo(props) {
...
-            onClick: () => {}
-          }, ['Add Collaborator'])
+          h(CollaboratorList, {
+            formFieldChange,
+            collaborators: internalCollaborators,
+            collaboratorKey: 'internalCollaborators',
+            collaboratorLabel: 'Internal Collaborator',
+            showApproval: false,
+            disabled: !isEmpty(darCode),
+            deleteBoolArray: (new Array(internalCollaborators.length).fill(false)),
```
connorlbark: if this always stays the same, should it not be a prop? seems like i could be computed inside of the class

src/pages/dar_application/ResearcherInfo_new.js
```diff
@@ -196,12 +194,15 @@ export default function ResearcherInfo(props) {
...
-            style: { marginTop: 25 },
-            onClick: () => {}
-          }, ['Add Collaborator'])
+          h(CollaboratorList, {
+            formFieldChange,
+            collaborators: internalCollaborators,
+            collaboratorKey: 'internalCollaborators',
+            collaboratorLabel: 'Internal Collaborator',
+            showApproval: false,
+            disabled: !isEmpty(darCode),
```
connorlbark: outside of this PR scope, but I think how we determine "is this DAR a draft" is no longer if the darCode is empty... i am not sure how it is though. i think this is ok to merge for now, but this disable condition might change in the future.

src/pages/dar_application/collaborator/CollaboratorList.js
```diff
@@ -4,7 +4,7 @@ import { useState, useEffect} from 'react';
 import { button, div, h } from 'react-hyperscript-helpers';
 import './collaborator.css';
 
-export default function CollaboratorList_new(props) {
```
connorlbark: i vote we keep the file name as `CollaboratorList_new.js` so it is distinct from the old one - at least until we fully delete the old DAR form. mildly confusing as it is

connorlbark (commented):
couple very minor things then I am happy to merge!

cypress/component/DataAccessRequest/researcher_info_new.spec.js
```diff
@@ -65,6 +66,17 @@ const WrappedResearcherInfo = (props) => {
...
 
+const user = {
+  userId: 1,
+  dacUserId: 2,
+  displayName: 'Cindy Crawford',
+  email: 'cc@c.com'
+};
+
+beforeEach(() => {
+  cy.stub(User, 'getMe').returns(Promise.resolve(user));
```
otchet-broad: Can you just returns(user) or does something require a Promise.resolve?

otchet-broad (commented):
Excellent job taking on Cypress!  Thank you!
Two quick things:
  I left you one question about the stub.  Just curious if something needs the Promise or is just looking for the value.
  Check Github for some static code analysis input around quotes.

cypress/component/DataAccessRequest/researcher_info_new.spec.js
connorlbark: we worked through a bug (outside of GH in slack) where it had to return a promise, since it was `await`ing it.

otchet-broad: Do you remember where that was in the code?  I'd like to see that.

It looks like User.getMe returns response.data. In that setup, the request will already have been resolved and the data property of the response is returned.  Could something else be getting in the way?

From MDN (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await):
Non-thenable value: An already-fulfilled Promise is constructed and used.

quazi-h:
Updated the CollaboratorForm and Summary components so that when the user tries to delete a collaborator, they must confirm the delete by pressing another button and having the option to cancel the delete. This is recreating the behavior of the collaborators on the old form. I wasn't sure if the summary delete buttons using glyph icons was intuitive, feedback would be appreciated. You can use this DAR draft with pre-populated collaborators to test the new changes: https://local.broadinstitute.org:3000/dar_application_new/41ea9f9a-22a9-493a-94cd-60161ca2cd26

connorlbark:
> Updated the CollaboratorForm and Summary components so that when the user tries to delete a collaborator, they must confirm the delete by pressing another button and having the option to cancel the delete. This is recreating the behavior of the collaborators on the old form. I wasn't sure if the summary delete buttons using glyph icons was intuitive, feedback would be appreciated. You can use this DAR draft with pre-populated collaborators to test the new changes: https://local.broadinstitute.org:3000/dar_application_new/41ea9f9a-22a9-493a-94cd-60161ca2cd26

I see what you mean about it not being intuitive to delete collaborators. Personally, I think it's better to have than to not have, and as far as being intuitive without text, I think it's as good as it's gonna get. Maybe pinging Vonne about it might be a good idea.

connorlbark (approved):
Pending feedback about confirm / delete (I think it is helpful personally), the code and tests look great. Would love to merge this & get work on finalizing the new DAR form

otchet-broad (approved):
Great accomplishments Quazi!  Nicely done.

PR #1902, opened 2022-11-17
[DUOS-2009][risk=no] DAA Acceptance Modal on SO Console

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2009

----

- Verify all tests go green


src/styles/buttons.css
```diff
@@ -19,3 +19,8 @@
   color: white;
   background-color: #005D9A;
 }
+
+.button-blue:hover {
+  color: white;
+  background-color: #094887;
+}
```
connorlbark, line 26: this fixes the 'Save' button which becomes not readable when you hover over it.

src/components/modals/SigningOfficialDAAModal.js
```diff
@@ -0,0 +1,83 @@
...
+    ]),
+    div({}, [
+      a({ target: '_blank', href: NIHLibraryCardAgreementLink, className: 'button button-white' }, [
+        span({className: 'glyphicon glyphicon-download'}),
+        ' NIH Library Card Agreement'
+      ])
+    ]),
+
+    div({ className: 'flex flex-row', style: { justifyContent: 'flex-end', }, }, [
+      a({
```
otchet-broad: Users should be able to navigate with the keyboard to the "I Agree" button and use the enter key to activate the "click" handler.
Do we need a second option on this modal to log out if they don't agree?

otchet-broad:
Good work Connor!  

Not sure if this is in scope, but you might want to check with the product owner: how can a signing official see these two links again if they need to reference them?

rushtong:
One bit of visual feedback while I review the code. Note how the modal loads behind the tab border:

Screen Shot 2022-11-28 at 3 23 37 PM

rushtong:
Second piece of visual/UX feedback. I have a UX objection to a modal that is not cancelable. We should add a cancel button and if the user has not accepted the terms, then all features should be disabled. Lastly, in this world, we need to somehow give the user the ability to accept them if they haven't already.

src/libs/ajax.js
```diff
@@ -641,6 +641,16 @@ export const User = {
...
+  getAcknowledgements: async () => {
+    const url = `${await getApiUrl()}/api/user/acknowledgements`;
+    const res = await axios.get(url, Config.authOpts());
+    return res.data;
+  },
+  acceptAcknowledgments: async (...keys) => {
+    const url = `${await getApiUrl()}/api/user/acknowledgements`;
+    const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(keys), { method: 'POST' }]));
+    return res.data;
   }
```
rushtong: I don't think we do this anywhere else, but it might be nice to ensure that `keys` isn't empty before trying to make the request.

rushtong (commented):
I'd like to revisit the UX flow for this story and clarify the requirements. Otherwise, code looks good and it functions as expected.

src/components/DuosHeader.css
```diff
@@ -30,6 +30,7 @@
 .navbar-duos .navbar-sub {
   width: 100%;
   padding: 0 20px 0 80px;
+  background: #fff;
```
connorlbark, line 33: fixed the weird styling with the nav bar by filling in the background properly

src/pages/SigningOfficialDataSubmitters.js
```diff
@@ -45,6 +46,7 @@ export default function SigningOfficialConsole() {
       div({style: {}, className: 'signing-official-tabs'}, [
         //NOTE: Links to this custodian table have been removed, we are retaining it with the intention of repurposing it for data submitters
         h(DataCustodianTable, {researchers, signingOfficial, unregisteredResearchers, isLoading}, []),
+        h(SigningOfficialDAAPopup, {}),
```
connorlbark: the logic for determining acceptance status, showing, and accepting is all in the popup class - no logic needed outside to accomplish this, just need to mount the popup component.

src/components/modals/SigningOfficialDAAModal.js
connorlbark: I am down to add this if we want. I'm not sure if it's in scope / the expected behavior of not agreeing though.

src/components/SigningOfficialDAAPopup.js
```diff
@@ -0,0 +1,107 @@
...
+
+  useEffect(() => {
+    if (isLoading) {
+      spinnerService.showAll();
+    } else {
+      spinnerService.hideAll();
+    }
+  }, [isLoading]);
+
+  return h(Modal, {
```
rushtong: This should not be a modal - that comes with extra behaviors that are causing trouble, like clicking outside of the modal area. Since the modal boundary is no longer clearly delineated, clicking anywhere outside of it seems to trigger acceptance.

src/components/SigningOfficialDAAPopup.js
```diff
@@ -0,0 +1,107 @@
...
+    isOpen: hasAccepted === false,
+    onRequestClose: acceptAndClose,
```
rushtong: This is why closing the modal triggers acceptance. We should only call `acceptAndClose` when the "I Agree" button is clicked.

src/pages/SigningOfficialResearchers.js
```diff
@@ -34,7 +37,9 @@ export default function SigningOfficialResearchers() {
   return (
     div({style: Styles.PAGE}, [
       div({style: {}, className: 'signing-official-tabs'}, [
-        h(SigningOfficialTable, {researchers, signingOfficial, isLoading}, [])
+        h(SigningOfficialTable, {researchers, signingOfficial, isLoading}, []),
+        // will automatically pop up if SO has not signed
+        h(SigningOfficialDAAPopup, {})
```
rushtong: In my opinion, we should instead have acceptance be passed into this component from the parent console. Then we can very easily render either one depending on status. Same feedback for the other table components as well.

Second bit of feedback, we should probably rename `SigningOfficialDAAPopup` to something more appropriate, like `SigningOfficialAgreements`.

connorlbark: Ok, I can do that. My only "ick" with that is that there is duplicate logic in three different console pages, but it does clean up the idea of the modal and such. I will go with this approach

connorlbark: I figured out how to do this in a way that I think is much better than before: I wrapped the consoles in a wrapper component, which queries to see agreement status, and conditionally renders its children or the agreement information. This means that the signing official consoles require no extra logic while still being able to replace their content effectively. Let me know if you think this is the right approach

connorlbark:
@rushtong @otchet-broad this should be ready for a re-review! no longer a modal

src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -0,0 +1,101 @@
...
+      ]),
+      div({}, [
+        a({ target: '_blank', href: NIHLibraryCardAgreementLink, className: 'button button-white' }, [
+          span({className: 'glyphicon glyphicon-download'}),
+          ' NIH Library Card Agreement'
+        ])
+      ]),
+
+      div({ className: 'flex flex-row', style: { justifyContent: 'flex-end', }, }, [
+        a({
```
otchet-broad: Is this the correct control to be using?  I still can't tab to it and press the enter key and see the onClick fire.

src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -0,0 +1,101 @@
...
+        setIsLoading(true);
+        setHasAccepted(await hasSOAcceptedDAAs());
+        setIsLoading(false);
+      } catch(error) {
+        Notifications.showError({text: 'Error: Unable to retrieve current user from server'});
+        setIsLoading(false);
+      }
+    };
+    init();
+  }, []);
```
rushtong: One issue in pushing this down to the component is that other API calls are still made while this one loads which is unnecessary API traffic. Not a huge deal, but it is visible on the page. 
Screen Shot 2022-12-06 at 12 28 01 PM

otchet-broad: I was thinking about that too!  I'm wondering if we should flex that "Storage" muscle a bit more often.  This might be a good candidate for that.

rushtong: > I was thinking about that too! I'm wondering if we should flex that "Storage" muscle a bit more often. This might be a good candidate for that.

I agree, that's a decent approach to this problem. If the parent component made the call and set this in storage, then we could completely skip rendering unnecessary components.

otchet-broad: @rushtong @connorlbark - I pushed an example into the branch DUOS-2009-2 of what I was thinking around this.

src/components/SigningOfficialDaaAgreementWrapper.js
rushtong: Conceptually, it should be a button object, not a link. Both div and span work for styling purposes, but I don't know how to make the tabbing work.`tabindex` doesn't seem to have any effect here.

rushtong: Add `tabIndex: 0` to each of the tab-able elements, then change the a to `button`. That allows tabbing and enter to activate the click handler

rushtong: Even without the tabindex, just using a button works for tab/select.

src/pages/SigningOfficialDarRequests.js
```diff
@@ -7,6 +7,7 @@ import {Collections} from '../libs/ajax';
 import { USER_ROLES } from '../libs/utils';
 import { DarCollectionTableColumnOptions, DarCollectionTable } from '../components/dar_collection_table/DarCollectionTable';
 import { consoleTypes } from '../components/dar_collection_table/DarCollectionTableCellData';
+import SigningOfficialDaaAgreementWrapper from '../components/SigningOfficialDaaAgreementWrapper';
```
rushtong: I was very confused by what's happening in the SO console here with the need to wrap the sub each sub tab. I had expected there to be a parent component like we do for all other console pages so we could render individual nav content conditionally. Instead, we have three separate components that all look like a parent console but then behave as a sub nav. I think we need to review this choice as it forces some common logic to be repeated for each tab.

connorlbark: I completely agree with you. The issue is the subnav logic is very complex... basically I'm not sure there's a way to say "don't render subnavs, go to this page, then after that render subnavs" - at least without significant refactor of the NavBar code (unless there is some approach that I am not aware of). After talking to Elliot briefly over slack, I thought perhaps it would make sense to remove this entirely from the console pages, and add basically a component wrapper which can be used in the routes. Like so: 

```
    <AuthenticatedRoute path="/signing_official_console/researchers" component={ensureSoHasDaaAcknowledgement(SigningOfficialResearchers)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
    <AuthenticatedRoute path="/signing_official_console/dar_requests" component={ensureSoHasDaaAcknowledgement(SigningOfficialDarRequests)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
    <AuthenticatedRoute path="/signing_official_console/data_submitters" component={ensureSoHasDaaAcknowledgement(SigningOfficialDataSubmitters)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
```

This injects code that will show the DAA acknowledgement page, so there's no logic in the page itself. I also incorporated Elliot's code for storage and caching... which makes it snappy. Let me know what you think of this approach.

src/components/SigningOfficialDaaAgreementWrapper.js
connorlbark: I like this - I made the logic generic, too, so it works with all future acknowledgements.

src/components/SigningOfficialDaaAgreementWrapper.js
connorlbark: Now using Elliot's branch's button, which works with tabbing

src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+// Wraps component and ensures that SO agrees to the
+// Broad and NIH agreements before proceeding to the given
+// component.
+export const ensureSoHasDaaAcknowledgement = (component) => {
+  return (props) => h(SigningOfficialDaaAgreementWrapper, {}, [
+    h(component, props),
+  ]);
+};
```
rushtong, line 100: This is a nice solution to the lack of a parent console page!

src/libs/ajax.js
```diff
@@ -641,6 +641,20 @@ export const User = {
...
+    const res = await axios.get(url, Config.authOpts());
+    return res.data;
+  },
+  acceptAcknowledgments: async (...keys) => {
+    if (keys.length === 0) {
+      return {};
+    }
+
+    const url = `${await getApiUrl()}/api/user/acknowledgements`;
+    const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(keys), { method: 'POST' }]));
```
rushtong: I'm not sure if we've documented this anywhere, but we did start moving toward using `axios` over `fetch`. Try `axios.post` instead:

```suggestion
    const res = await axios.post(url, Config.jsonBody(keys), Config.authOpts());
```

rushtong (approved):
I'm good with this moving forward as is. Feel free to implement suggestions as you see fit. What I would like to see is a follow on story to refactor the three pages so that we have a parent console that is more consistent with how we're doing all of our other console pages.

otchet-broad (approved):
Nice work!

PR #1908, opened 2022-11-22
[DUOS-1615][risk=no] Show data use abbreviations in dataset table.

otchet-broad:
[DUOS-1615](https://broadworkbench.atlassian.net/browse/DUOS-1615) provides a screenshot of the Structured Data Use Limitations with a list of data abbreviations.

This PR:
Hooks the enrichment of data use abbreviations to table changes so the column enriches the value as it is needed for display.  This approach lazily enriches each table row with the data abbreviations because a data abbreviation's computation may require one or more requests to be issued to the ontology service.  Performing these all at once was causing a 12 second delay as fetches were performed. 

Known issues:
These abbreviations are not provided in the download datasets link.
Data Use abbreviations can only be searched if the abbreviation has been rendered.
The existing page has a large state machine.  We may want to refactor this into something simpler moving forward.

BEFORE:
Screen Shot 2022-11-22 at 12 33 31 PM

AFTER:
Screen Shot 2022-11-22 at 12 34 50 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetCatalog.js
+++ b/src/pages/DatasetCatalog.js
@@ -13,6 +13,8 @@ import {Storage} from '../libs/storage';
 import {Theme} from '../libs/theme';
 import datasetIcon from '../images/icon_dataset_.png';
 import {getBooleanFromEventHtmlDataValue, USER_ROLES} from '../libs/utils';
+import {DataUseTranslation} from '../libs/dataUseTranslation';
+import {spinnerService} from '../libs/spinner-service';
 import { ArrowDropUp, ArrowDropDown } from '@material-ui/icons';
 
 const tableBody = {
@@ -28,6 +30,7 @@ export default function DatasetCatalog(props) {
   const [sort, setSort] = useState({ field: 'alias', dir: 1 });
   const [numDatasetsSelected, setNumDatasetsSelected] = useState(0);
   const [selectedDatasets, setSelectedDatasets] = useState([]);
+  const [datasetsOnPage, setDatasetsOnPage] = useState([]);
 
   // Selection States
   const [currentPageAllDatasets, setCurrentPageAllDatasets] = useState(1);
@@ -60,7 +63,6 @@ export default function DatasetCatalog(props) {
       row.ix = index;
       row.dbGapLink =
         getOr('')('propertyValue')(find({propertyName: 'dbGAP'})(row.properties));
-
       // Extracting these fields to make sorting easier
       row['Dataset Id'] = row.alias;
       row['Data Access Committee'] = findDacName(localDacs, row);
@@ -68,7 +70,6 @@ export default function DatasetCatalog(props) {
       row['Principal Investigator (PI)'] = findPropertyValue(row, 'Principal Investigator(PI)');
       row['# of Participants'] = findPropertyValue(row, '# of participants');
       row['Data Custodian'] = findPropertyValue(row, 'Data Depositor');
-
       return row;
     })(datasets);
     applyDatasetSort(sort, datasets);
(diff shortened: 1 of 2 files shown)


src/pages/DatasetCatalog.js
```diff
@@ -101,6 +102,40 @@ export default function DatasetCatalog(props) {
     init();
   }, [getDatasets]);
 
+  useEffect( () => {
+    const doEnrichment = async() => {
+      spinnerService.showAll();
+      const theCurrentPage = filterToOnlySelected ? currentPageOnlySelected : currentPageAllDatasets;
```
otchet-broad, line 108: There's a complex state machine in this code.  
I believe pulling this into a hook is the right approach, but the immediate impact is the need for managing all of the elements that can change state within the hook, hence why there's some seemingly duplicate code.  If I were to refactor this page, I would get more granular with state because the current approach isn't actually saving enough in certain conditions (for example, searchTable was invoked multiple times).

src/pages/DatasetCatalog.js
```diff
@@ -101,6 +102,40 @@ export default function DatasetCatalog(props) {
...
+          return text.toLowerCase().includes(query.toLowerCase());
+        }
+        return true;
+      };
+      const visibleDatasets = filterToOnlySelected ? selectedDatasets : datasetList;
+      const results = visibleDatasets.filter(searchTable(searchDulText)).slice((theCurrentPage - 1) * pageSize, theCurrentPage * pageSize);
+      await Promise.all(results.map(async(dataset) => {
+        if (isNil(dataset.codeList)) {
+          if (!dataset.dataUse || isEmpty(dataset.dataUse)) {
+            dataset.codeList = 'None';
```
quazi-h, line 121: I remember when you were demoing this you mentioned you were providing "None" when there are no dataUse restrictions. I was flipping through a dataset catalog and noticed a dataset that had an empty string displayed. Digging deeper I found that it had two restrictions listed as false, so there was still a "dataUse" property for that dataset. Should datasets such as these be displaying "None"?
**dataSetId: 441**
`    "dataUse": {
      "generalUse": false,
      "other": "false"
    },`
Screen Shot 2022-11-23 at 11 10 06 AM

quazi-h (approved):
Changes look good to me, I am able to see the data use abbreviations in the column on the dataset catalog on the different consoles. Abbreviations look good. I agree that at some point this page should be refactored to address the complexity of states being handled here but I'm not sure on the best way to approach that. I left one comment regarding a dataset with a blank value in the column but otherwise this looks good to me.

src/pages/DatasetCatalog.js
connorlbark: agreed. below this, i think it would just be:

```
            dataset.codeList = isEmpty(codes) ? 'None' : codes.join(', ');
```

connorlbark (approved):
Agree with quazi on adding `None` if no codes; other than that, looks good! I imagine these code paths would all be cleaner once we figure out a solution for the repeat ontologies calls.

src/pages/DatasetCatalog.js
otchet-broad: This one is interesting.  Excellent find @quazi-broad 
It seems this is a case where our code was changed, but existing values in dataUse were not.

For DUOS-000078, the dataUse object is: `dataUse:{generalUse: false, other: 'false'}`
When I generate a new entry similar to it (see DUOS-000612 on dev), the dataUse object is: `dataUse:{otherRestrictions: true, other: 'What should the data use abbreviation be for this case?'}`

otherRestrictions is key here.  It causes a proper abbreviation to be generated where our existing code doesn't for DUOS-000078.

I'm inclined to leave this code as is because the absence is highlighting a bug.  Is that okay with you @connorlbark ?  I'm going to see if I can fix inside of dataUseTranslation.

otchet-broad: BEFORE:

Screen Shot 2022-11-23 at 5 18 02 PM

AFTER:
Screen Shot 2022-11-23 at 5 18 40 PM

rushtong: > I'm inclined to leave this code as is because the absence is highlighting a bug.

I agree with this. In theory, there should never be a completely empty data use. That would be an error case in the data and possibly a validation error when processing a new DAR.

rushtong (approved):
Looks good 👍🏽

quazi-h (approved):
New changes look good - the dataset I was referencing with blank use codes now has "OTH1". I see no more instances of datasets with a blank value for that field. Thanks for the changes!

PR #1914, opened 2022-11-28
[DUOS-2148][risk=no] Update custom error page

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2148

## Changes
* Point 400s to the app so so we can control the display.
* `Election404` is unused.
* `k8s.yaml` is unused. When we deployed to our own project via circleci, we did use this. It was missed in the migration to terra projects.
* Refactor the `NotFound` component to be a pure function.
* Made `Dockerfile` a little easier to work with when testing nginx config changes - docker builds are easier since we can use the cached app image between reruns.

## Testing
To test, you need to do a full docker build and point a local browser to some invalid location such as `https://local.broadinstitute.org/static/`. In those cases, nginx will redirect to root so that the app can handle the error case appropriately.

----

- Verify all tests go green


otchet-broad:
Impact of the server_tokens off setting in response headers:

BEFORE: 
Screen Shot 2022-11-28 at 3 37 52 PM

AFTER:
Screen Shot 2022-11-28 at 3 35 24 PM

Thanks for all of the improvements, especially the cache!

otchet-broad (approved):
Look, ma!  No hyperscript!

connorlbark (approved):
👍

PR #1926, opened 2022-12-06
[DUOS-1758][risk=no] Data Use Agreements - Add Validation to DAR Form

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1758

Converts DAR Form to functional component, fixes some bugs, and adds validation. There is no submit functionality. This is the largest part of the work needed for submit: not sure how to make smaller, since 90% of the work was refactoring so that it would easier to keep track of state for validation.

----

- Verify all tests go green


src/pages/DataAccessRequestApplication_new.js
```diff
@@ -221,763 +205,235 @@ class DataAccessRequestApplicationNew extends Component {
...
+                  style: {
+                    marginRight: '2rem',
+                  }
+                }, ['Save']),
+                a({
+                  id: 'btn_submit', isRendered: isNil(formData.darCode), onClick: () => mockSubmit(),
+                  className: 'button button-blue '
+                }, ['Attest and Send']),
+              ])
+            ]),
```
connorlbark: these are here just in place of the new DataUseAgreements panel. clicking save does nothing, clicking submit runs the validation code.

rushtong:
The location of the profile validation message is confusing. In the existing view, it's between sections 1.2 and 1.3. In the new, it's in section 1.9

### Old:
Screen Shot 2022-12-07 at 11 01 14 AM

### New:
Screen Shot 2022-12-07 at 10 59 31 AM

rushtong:
Section 1.2 should be marked as optional if 1.2.3 is selected:

### Old
Screen Shot 2022-12-07 at 11 04 33 AM

### New
Screen Shot 2022-12-07 at 11 04 47 AM

rushtong:
Strange bug report - selecting a different dataset in a DAR clears out the page with form utils error:
```
Uncaught TypeError: Cannot read properties of null (reading 'geneticStudiesOnly')
    at darFormUtils.js:9:1
    at Array.some (<anonymous>)
    at datasetsContainDataUseFlag (darFormUtils.js:9:1)
    at needsGsoAcknowledgement (darFormUtils.js:21:1)
    at DataAccessRequest (DataAccessRequest_new.js:273:1)
    at renderWithHooks (react-dom.development.js:16305:1)
    at updateFunctionComponent (react-dom.development.js:19588:1)
    at beginWork (react-dom.development.js:21601:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:4164:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:1)
```

connorlbark:
> Strange bug report - selecting a different dataset in a DAR clears out the page with form utils error:

do you know which dataset you added? looking at the report, I am pretty sure the dataset added must have had a null dataUse field. if so, this should be good now.

rushtong:
> do you know which dataset you added? looking at the report, I am pretty sure the dataset added must have had a null dataUse field. if so, this should be good now.

I was searching on Herman, so there are two that fit that bill, both triggered it.

src/pages/dar_application/DataAccessRequest_new.js
```diff
@@ -38,71 +59,58 @@ const formatSearchDataset = (ds) => {
...
-  const needsGsoAcknowledgement = useCallback(() => {
-    return datasetsContainDataUseFlag(datasets, 'geneticStudiesOnly');
-  }, [datasets]);
-
-  const needsPubAcknowledgement = useCallback(() => {
-    return datasetsContainDataUseFlag(datasets, 'publicationResults');
-  }, [datasets]);
+  const primaryChange = ({key, value}) => {
+    if (key === 'diseases' && value === true) {
+      // in this case, reset all fields.
```
rushtong: nit, language
```suggestion
      // in this case, reset all primary data use fields.
```

rushtong (commented):
In general, this looks good. See comments and issues inline.

rushtong:
I think the layout of the error messages is a little confusing. These are disconnected from where the error actually is:

Screen Shot 2022-12-09 at 1 51 35 PM

And for this one, cloud computing is below the fold:
Screen Shot 2022-12-09 at 1 53 08 PM

This is certainly not the final revision, so I don't consider these blockers. I think we could use some UX guidance/discussion on how to treat errors on this page. Maybe something in the left nav that indicates error states in each step? Maybe all errors should be grouped together or placed directly above or below the field they are associated to? Let's have that conversation.

rushtong (approved):
I'm in favor of moving this forward even with the outstanding UX questions. Let's get this in front of Jonathan/Yvonne and sort out enhancements.

quazi-h (approved):
Refactor to organize the validation code is looking great.

The only nits I have were similar to Greg's comments, re: placement of the error notifications.
Most of them seem to be grouped above the corresponding step, which makes sense to me. I think it might be more helpful to have the specific error right above the actual component where the issue needs to be addressed.

When clicking "Attest and Send", the UI scrolls up to the topmost error most of the time but I noticed one instance where I left the following two fields blank:
Is the primary purpose of this research to investigate a specific disease(s)?
2.4 Non-Technical Summary*
And when clicking Submit, the UI scrolls up to the second error, not the first one.

I also agree with Greg that these are minor and could benefit from UX guidance and overall the validation is working great and the code is much more organized with this PR so this should get merged since I'm sure we will have another iteration of improvements once we have more feedback. Great work!

connorlbark:
> And when clicking Submit, the UI scrolls up to the second error, not the first one.

I think that's actually  more than likely an issue with the gotostep function, it's behavior can be erratic sometimes. I'll go ahead and merge this and make sure to note that in the ticket so we can investigate gotostep

connorlbark:
> And when clicking Submit, the UI scrolls up to the second error, not the first one.

I think that's actually  more than likely an issue with the gotostep function, it's behavior can be erratic sometimes. I'll go ahead and merge this and make sure to note that in the ticket so we can investigate gotostep

PR #1929, opened 2022-12-07
[DUOS-1929] New DAC Datasets Table for Chair and Data Submitter Consoles

quazi-h:
[DUOS-1929](https://broadworkbench.atlassian.net/browse/DUOS-1929)
The new tab should display datasets belonging or submitted to the user's DAC(s).
A DAC chair should be able to approve and deny datasets as well as edit Accepted datasets.
Approval/Rejection actions should send out confirmation emails to the Data Submitter.
The new table should be searchable and paginated, with hyperlinks to the dataset statistics and data submission form.

**This PR:**
- Added a new DAC Datasets sub-nav tab to the Chair and DS consoles.
- Pulls datasets via User.getUserRelevantDatasets, pulls dacApproval status by calling Dataset.getDataSetsByDatasetId (this could be optimized)
- Gets specific dataset properties within the CellData code.
- Added a new DAC ajax method that hits the updateApprovalStatus endpoint.
- Added conditional logic for the status column for DAC datasets to render a "YES" / "NO" button if a dataset is not yet approved/rejected and then an "ACCEPTED" / "REJECTED" status to reflect the DAC's decision on approval.

**Still working on:**
- Need to add hover actions:
  Hover over the Dataset ID shall display a hover over that says View Dataset Statistics
  Hover over the hyperlinked Data Use abbreviations shall display a hover over that displays all data use terms for the dataset
- Status column YES/NO buttons need to trigger confirmation emails
- Search appears to only be working for the DUOS ID (alias).
- The new page is loading data inconsistently (custodian, submitter, dataUse) and not showing updates when an action is made - need to refactor to optimize for those issues.
- Need to update the isRendered/disabled logic for the status column items
- Need to add tests

**Need clarification on:**
- Should the redirect to the add dataset form be the old form or new one? (dataset_registration vs data_submission_form)
- Clicking the hyperlinked Dataset Name -> read only view of data_submission_form, how do I render a read-only version of a page?
- Unclear on the difference between the Chair Console and DS Console implementations of this new tab.
- Is Data Submitter == 'Data Depositor'? or create user?
- Is Data Custodian == 'Data Depositor'?

----

- Verify all tests go green

[DUOS-1929]: https://broadworkbench.atlassian.net/browse/DUOS-1929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rushtong:
Here's a stab at some of the open questions:

> Should the redirect to the add dataset form be the old form or new one? (dataset_registration vs data_submission_form)

We can only point to the existing page now. The new page is not functional, so we cannot direct users there.

> Clicking the hyperlinked Dataset Name -> read only view of data_submission_form, how do I render a read-only version of a page?

We don't officially have a read-only view. This will need to be a new ticket and we'll need to sort out what that looks like there. We do have a non-linked page for dataset statistics which is what we used to link to from the catalog page. We disabled that due to a bug that we haven't addressed yet. 

> Unclear on the difference between the Chair Console and DS Console implementations of this new tab.

This story covers the Chair console, not the data submitter console. We should not be doing the DS version until we hit that story.

> Is Data Submitter == 'Data Depositor'? or create user?

Data Submitter is the create user for the dataset.

> Is Data Custodian == 'Data Depositor'?

We need clarification from @solideoglori on this question. We technically don't have either role in the system yet. We only have `DATA_OWNER` which could be the custodian if we decide it to be. Also, I believe there can be multiple `DATA_OWNER` users for a dataset so this needs to be followed up in a different story. For now, I suggest we leave that column empty and file a ticket for implementation/discussion.

connorlbark:
I'm concerned about this from a UX perspective.

Some thoughts:
- The status column is confusing. What does `Yes` do? What does `No` do? I couldn't figure out what they did until I read about it / clicked on it
- What is the difference between clicking on the Dataset Name and the DUOS ID? How would the user know which one to click on? Should there be an Edit button and a Statistics button which are more clearly labeled?
- The `Enter Search Terms` bar is i think styled differently than other pages
- Usually, `Add` buttons are placed next to the searchbar as well.
- Should be able to search on `Accepted` and only see accepted results.
- Maybe should also be able to sort by accepted status as well

src/Routes.js
```diff
@@ -92,6 +93,7 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/dataset_catalog" component={DatasetCatalog} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.all]} />
     <AuthenticatedRoute path="/dataset_statistics/:datasetId" component={DatasetStatistics} props={props}
       rolesAllowed={[USER_ROLES.all]} />
+    <AuthenticatedRoute path="/dac_datasets" component={DACDatasets} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.all]} />
```
connorlbark: should the rolesAllowed just be the dac chair?

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+  const dacAccepted = div({
+    style: {color: '#1ea371', fontWeight: 'bold'}
+  },[
+    span('ACCEPTED'),
+    h(Link,{
+      style: {marginLeft: '15px'},
+      id: `${dataset.dataSetId}_edit`,
+      className: 'glyphicon glyphicon-pencil',
+      isRendered: true,
+      to: `dataset_registration/${dataset.dataSetId}`
```
otchet-broad, line 89: Clicking the pencil icon didn't take me to the intended page. You might need to adjust the destination to`../dataset_registration....`, or we might have a bug in the Link component.

src/utils/DatasetUtils.js
```diff
@@ -0,0 +1,27 @@
+import {find, getOr, isNil} from 'lodash/fp';
+import {isEmpty} from 'lodash/fp';
+import {DataUseTranslation} from '../libs/dataUseTranslation';
+
+export const findPropertyValue = (dataSet, propName, defaultVal) => {
+  const defaultValue = isNil(defaultVal) ? '' : defaultVal;
+  return getOr(defaultValue)('propertyValue')(find({ propertyName: propName })(dataSet.properties));
+};
+
+export const getDataUseCodes = async (dataset) => {
```
connorlbark, line 10: I think this should return the data use codes instead of modifying dataset. I think it would confuse me / would worry me about side effects if I was another developer and was using this function. this could be kept as react state on the component level and sent down to everything that needs it.

src/components/dac_dataset_table/DACDatasetTable.js
```diff
@@ -0,0 +1,173 @@
...
+
+  //Helper function to update page
+  const goToPage = useCallback(
+    (value) => {
+      updatePage(value, pageCount, setCurrentPage);
+    },
+    [pageCount]
+  );
+
+  return h(SimpleTable, {
```
otchet-broad: It looks like this table might benefit from some style modifications:
1) When the table is narrow, cell data is spilling out of the "cell" and overlaps contents from another cell.  Ideally, we should have a margin between cells.
2) We might want to consider using ellipses for long cell content.

src/pages/DACDatasets.js
```diff
@@ -0,0 +1,120 @@
...
+    setFilteredList
+  ), [datasets, filterFn]);
+
+  const getDatasets = useCallback(async () => {
+    const userDatasets = await User.getUserRelevantDatasets();
+    userDatasets.map(async dataset => {
+      const datasetObject = await DataSet.getDataSetsByDatasetId(dataset.dataSetId);
+      dataset.dacApproval = datasetObject.dacApproval;
+    });
+    return userDatasets;
```
connorlbark: Ah, ok. I see the issue. `/me/dac/datasets` uses DatasetDTO objects (which are deprecated and don't include all fields. i think we should update this endpoint if we're going to still need it to return regular Dataset objects, then you shouldn't have to do this map operation

src/pages/DACDatasets.js
```diff
@@ -0,0 +1,120 @@
...
+import {img} from 'react-hyperscript-helpers';
+import lockIcon from '../images/lock-icon.png';
+import SearchBar from '../components/SearchBar';
+import {DACDatasetTable, DACDatasetTableColumnOptions} from '../components/dac_dataset_table/DACDatasetTable';
+import {getSearchFilterFunctions, Notifications, searchOnFilteredList} from '../libs/utils';
+import {useRef} from 'react';
+import { consoleTypes } from '../components/dac_dataset_table/DACDatasetTableCellData';
+import '../components/dac_dataset_table/dac_dataset_table.css';
+
+export default function DACDatasets(props) {
```
connorlbark, line 21: should there be a `DataSubmitterDatasets` component as well? would they show different thigns?

src/components/dac_dataset_table/DACDatasetTable.js
```diff
@@ -0,0 +1,173 @@
...
+  STATUS: 'status',
+};
+
+const columnHeaderConfig = {
+  duosId: {label: 'DUOS ID', cellStyle: {width: styles.cellWidths.duosId}, cellDataFn: cellData.duosIdCellData, sortable: true},
+  dataSubmitter: {label: 'Data Submitter', cellStyle: {width: styles.cellWidths.dataSubmitter}, cellDataFn: cellData.dataSubmitterCellData, sortable: true},
+  datasetName: {label: 'Dataset Name', cellStyle: {width: styles.cellWidths.datasetName}, cellDataFn: cellData.datasetNameCellData, sortable: true},
+  dataCustodian: {label: 'Data Custodian', cellStyle: {width: styles.cellWidths.dataCustodian}, cellDataFn: cellData.dataCustodianCellData, sortable: true},
+  dataUse: {label: 'Data Use', cellStyle: {width: styles.cellWidths.dataUse}, cellDataFn: cellData.dataUseCellData, sortable: false},
+  status: {label: 'Status', cellStyle: {width: styles.cellWidths.status}, cellDataFn: cellData.statusCellData, sortable: false},
```
otchet-broad: Why turn off sorting for any of these?  Funny enough, this was the first column I wanted to sort on.  :-)

src/components/DuosHeader.js
```diff
@@ -114,10 +115,11 @@ export const headerTabsConfig = [
...
   {
     label: 'DS Console',
-    link: '/data_submission_form',
-    search: 'data_submission_form',
+    link: '/dac_datasets',
+    search: 'dac_datasets',
     children: [
-      { label: 'Datasets', link: '/data_submission_form' }
+      { label: 'Datasets', link: '/dac_datasets' },
+      // { label: 'Datasets', link: '/data_submission_form' }
```
connorlbark: We should still keep this... but maybe label it something different.

src/components/dac_dataset_table/DACDatasetTable.js
otchet-broad: As @connorlbark might have mentioned, I think we need a better term to describe this column, or even consider status (accepted, pending) separate from actions.  I haven't looked to see what Yes/No is all about, but it isn't clear from the UI what I'd be acting on.

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+    })
+  ]);
+
+  const dacRejected = div({
+    style: {color: '#000000', fontWeight: 'bold'}
+  },[span('REJECTED')]);
+
+  const approvalActions = div({}, [
+    button({
```
otchet-broad, line 98: We have standard coloring on voting buttons (green/red).  Should that be applied to these buttons as well?

src/pages/DACDatasets.js
```diff
@@ -0,0 +1,120 @@
...
+      dataset.dacApproval = datasetObject.dacApproval;
+    });
+    return userDatasets;
+  }, []);
+
+  useEffect(() => {
+    const init = async() => {
+      try {
+        setIsLoading(true);
+        // const listOfDatasets = await User.getUserRelevantDatasets();
```
otchet-broad: Dead code.

src/libs/utils.js
```diff
@@ -590,7 +591,20 @@ export const getSearchFilterFunctions = () => {
...
-    }
+    },
+    datasets: (term, targetList) => filter(dataset => {
+      const duosId = dataset.alias;
+      const dataSubmitter = findPropertyValue(dataset, 'Data Depositor');
+      const datasetName = findPropertyValue(dataset, 'Dataset Name');
+      const dataCustodian = findPropertyValue(dataset, 'Data Depositor');
+      getDataUseCodes(dataset);
+      const dataUse = dataset.codeList;
+      return includes(term, toLower(duosId)) ||
```
otchet-broad, line 475: Why not allow search on status?

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+      onClick: () => updateApprovalStatus(false),
+      className: 'btn-primary-dac-datasets'
+    }, ['NO'])
+  ]);
+
+  const status = (!isNil(dataset?.dacApproval))
+    ? dataset.dacApproval
+      ? dacAccepted
+      : dacRejected
+    : approvalActions; // todo: add buttons and actions
```
otchet-broad, line 118: Is that todo still correct?

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+  },[span('REJECTED')]);
+
+  const approvalActions = div({}, [
+    button({
+      id: 'btn_approveDataset',
+      isRendered: true,
+      disabled: false,
+      onClick: () => updateApprovalStatus(true),
+      className: 'btn-primary-dac-datasets'
+    }, ['YES']),
```
otchet-broad, line 104: Maybe Accept as the name?

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+    button({
+      id: 'btn_rejectDataset',
+      isRendered: true,
+      disabled: false,
+      onClick: () => updateApprovalStatus(false),
+      className: 'btn-primary-dac-datasets'
+    }, ['NO'])
```
otchet-broad, line 111: Maybe Reject?

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+  return {
+    data: div({
+      style: {
+        display: 'flex',
+        alignItems: 'center',
+      }
+    }, [status]),
+    value: status,
+    id: dataset.dataSetId,
+    cellStyle: { width: styles.cellWidths.status },
```
otchet-broad, line 130: Should we use the DUOS Red/Green colors for accept and reject?  Looks like we're only using green for accept.

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+
+export function datasetNameCellData({dataset, label='datasetNameCellData'}) {
+  const datasetName = findPropertyValue(dataset, 'Dataset Name');
+  return {
+    data: div({
+      style: {
+        display: 'flex',
+        alignItems: 'center',
+      }
+    }, [h(Link, {to: `dataset_registration/${dataset.dataSetId}`}, [datasetName])]),
```
otchet-broad, line 43: None of the links seem to be working correctly for me.

src/pages/DACDatasets.js
```diff
@@ -0,0 +1,120 @@
...
+          margin: '0 3% 0 0',
+          borderRadius: '100px',
+        },
+      }),
+    ]),
+    div([
+      button({
+        id: 'btn_addDataset',
+        className: 'btn-primary-dac-datasets',
+        onClick: () => history.push({ pathname: 'data_submission_form' }),
```
otchet-broad, line 135: This doesn't seem to be working as expected for me.

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+      isRendered: true,
+      disabled: false,
+      onClick: () => updateApprovalStatus(true),
+      className: 'btn-primary-dac-datasets'
+    }, ['YES']),
+    button({
+      id: 'btn_rejectDataset',
+      isRendered: true,
+      disabled: false,
+      onClick: () => updateApprovalStatus(false),
```
otchet-broad, line 109: Consider what the grid looks like after the user clicks Yes or No.
When I clicked "No" I would have expected the status in that cell to update after the spinner was done, but it didn't.  Reloading the page caused the page to display the correct value, but that's a hard interruption of my workflow to see the outcome.

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+    label
+  };
+}
+
+export function statusCellData({dataset, label='statusCellData'}) {
+  const updateApprovalStatus = async (status) => {
+    await DAC.updateApprovalStatus(dataset.dacId, dataset.dataSetId, status);
+  };
+
+  // todo: update isRendered/ disabldisabledfor each
```
otchet-broad: Still the correct comment/todo?

src/components/dac_dataset_table/DACDatasetTableCellData.js
```diff
@@ -0,0 +1,143 @@
...
+  return {
+    data: dataCustodian,
+    id: dataset.dataSetId,
+    cellStyle: { width: styles.cellWidths.dataCustodian },
+    label
+  };
+}
+
+export function dataUseCellData({dataset, label='dataUseCellData'}) {
+  getDataUseCodes(dataset);
```
otchet-broad, line 64: You probably want to lazily enrich the getDataUseCodes based on the record that are visible because this can lead to many async calls, depending on the dataset configuration.  The behavior that was accepted on the Researcher Dataset table was users won't be able to search all of the values until every row is visible, but they won't have to wait for all of the calls to complete upfront.  @rushtong and I have talked about improving this so that the backend object might include the necessary information without the frontend making all of the calls.

src/libs/utils.js
```diff
@@ -590,7 +591,20 @@ export const getSearchFilterFunctions = () => {
...
         return !isNil(match);
       })(targetList);
-    }
+    },
+    datasets: (term, targetList) => filter(dataset => {
+      const duosId = dataset.alias;
+      const dataSubmitter = findPropertyValue(dataset, 'Data Depositor');
+      const datasetName = findPropertyValue(dataset, 'Dataset Name');
+      const dataCustodian = findPropertyValue(dataset, 'Data Depositor');
+      getDataUseCodes(dataset);
```
otchet-broad, line 473: See my comment about lazily enriching.  This will not be performant if this collection includes many datasets that need to call ontology.  The underlying code in getDataUseCodes may need to make multiple requests of ontology to produce the correct set of codes.

otchet-broad (changes requested):
Good start, @quazi-broad!  I've left feedback and one overall suggestion is to add some test coverage as you continue to move through the work.

rushtong:
Closing in favor of https://github.com/DataBiosphere/duos-ui/pull/1999

PR #1938, opened 2022-12-09
[DUOS-2116][risk=no] Filter objects and values to only include relevant datasets as return…

otchet-broad:
[DUOS-2116](https://broadworkbench.atlassian.net/browse/DUOS-2216) identified two issues.  One issue is related to UX and workflow.  It _is_ functioning properly and as designed, but is reveals complexities behind the state machine.  New tickets were created to address those concerns.

The second issue related to the expandable list of datasets embedded in rows when the list of DARs are displayed to DAC members and chairs.
This PR code adds logic to filter the DAR Collection object to include only the datasets and elements needed for the datasets they have access to.  In deciding on the approach to implement, this was the least invasive as it left the core logic of bucket building as-is and added input filtering to existing functions.  Other (admin) code elements relied on the bucketing logic as-is.

An entirely different approach to consider may be to perform this filtering at the consent tier so that resource "collection" responses don't include the DARs, datasets, and elections that aren't relevant for the DAC Chair or Member.

----

- Verify all tests go green

### DIFF
--- a/src/components/dar_dataset_table/DarDatasetTable.js
+++ b/src/components/dar_dataset_table/DarDatasetTable.js
@@ -6,12 +6,13 @@ import PaginationBar from '../PaginationBar';
 import { recalculateVisibleTable, goToPage as updatePage } from '../../libs/utils';
 import SimpleTable from '../SimpleTable';
 import cellData from './DarDatasetTableCellData';
-import {flow, isNil} from 'lodash/fp';
+import {flow, isNil, intersection} from 'lodash/fp';
 import {
   generatePreProcessedBucketData,
   processDataUseBuckets,
 } from '../../utils/DarCollectionUtils';
 import {Notifications} from '../../libs/utils';
+import { isEmpty } from 'lodash';
 
 
 const storageDarDatasetSort = 'storageDarDatasetSort';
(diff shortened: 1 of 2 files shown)


src/components/dar_dataset_table/DarDatasetTable.js
```diff
@@ -151,14 +152,32 @@ export const DarDatasetTable = (props) => {
...
+      let { dars, datasets } = cloneDeep(collection);
+      datasets = datasets ? datasets.filter(dataset => summary.datasetIds.includes(dataset.dataSetId)) : null;
+      const darKeys = Object.keys(dars);
+      darKeys.forEach(darKey => {
+        if(intersection(summary.datasetIds, dars[darKey].datasetIds).length > 0) {
+          const electionKeys = Object.keys(dars[darKey].elections);
+          electionKeys.forEach(electionKey => {
+            if (!summary.datasetIds.includes(dars[darKey].elections[electionKey].datasetId)) {
+              delete dars[darKey].elections[electionKey];
+            }
```
rushtong, line 175: This looks like an unusual edge case ... writing this out so I'm clear on what's happening.

1. We're starting with a list of all datasets on the DAR
2. We then filter that on the dataset ids accessible to the chair/member as calculated by the back end
3. We then look at all elections on the DAR
4. We further restrict the visible datasets so that we are only seeing datasets for which we have elections.

We need step 4 because there could be datasets for which no elections have yet been created (not sure if this would also happen with canceled elections). I'd like to run this by Jonathan as it means we're purposefully NOT showing the user a dataset that the DAR is requesting only because no election has been created for it yet.

otchet-broad: Step 2 (line 167) filters the datasets to those included in the summary response.
Step 3 loops the individual DARs in the collection.  The existing code refers to these as DARs, so I kept the naming the same, but the history is apparent in this code.  :-) 
I then start to loop the individual DARs in the collection at 169.  The line at 173 is satisfied only when an election is found that **doesn't** pertain the the list of datasets we're filtering down to.  At that point, I'm removing that **election** information (#174) (not the dataset).
If the DAR does not pertain to one of the filtered datasets, it is removed. (Line 178, below)

rushtong: Gotcha. Will rephrase my current understanding. There are two tasks here. 

1.  Filter the datasets to those that are visible to the chair/member.
2. Since we are displaying vote totals, we also have to filter elections that are not associated to the datasets in question.

We do have a method that should do the same thing, at least in theory, (`utils.filterCollectionElectionsByUser`) but I can see the utility here. I also think we could do a better job of packing up collection specific functions for common use.

rushtong (approved):
Looks good!

src/components/dar_dataset_table/DarDatasetTable.js
otchet-broad: Great rephrasing, Greg!

Agreed on the collection specific function packaging.  I left it as-is for two reasons: 1) It's only current use is this function, 2) I wasn't sure if this should eventually make it into src/libs/utils.js **or** src/utils/DarCollectionUtils.js

Do you think we need to apply this logic elsewhere?

rushtong: > Great rephrasing, Greg!
> 
> Agreed on the collection specific function packaging.  I left it as-is for two reasons: 1) It's only current use is this function, 2) I wasn't sure if this should eventually make it into src/libs/utils.js **or** src/utils/DarCollectionUtils.js
> 
> Do you think we need to apply this logic elsewhere?

I think it's fine here, I don't see any other needs for this right now. Thank you!

connorlbark (approved):
👍

PR #1955, opened 2022-12-20
[DUOS-1611][risk=no] Custom highlighting for Narrative RUS

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1611

Initially, I tried to use the old code in utils for highlighting. However, it used raw regex and HTML tags, which gave me concerns of safety, since we would have to sanitize user input while still forcing raw html to render. Instead, I created a new, custom highlighting component which allows multiple colors and multiple banks of words and does not use regex or any raw html, so user input is guaranteed sanitized. I looked into packages that could do this, but I couldn't find a major highlighting package which allowed multiple different colors.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/highlight_text.spec.js
+++ b/cypress/component/highlight_text.spec.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-undef */
 import HighlightText from '../../src/components/HighlightText';
 import { mount } from 'cypress/react';
+import React from 'react';
 
 describe('HighlightText - Tests', function() {
   it('Renders text with no highlighting if no matches', function() {
--- a/cypress/component/highlight_text.spec.js
+++ b/cypress/component/highlight_text.spec.js
@@ -27,7 +27,7 @@ describe('HighlightText - Tests', function() {
 
   it('Highlights with the correct color, irrelevant of casing.', function() {
     const text = (
-      'The quick brown fox jumps over the lazy log.'
+      'ThE quick BroWn fox jUMpS over tHe laZy log.'
     );
 
     mount(
--- a/cypress/component/highlight_text.spec.js
+++ b/cypress/component/highlight_text.spec.js
@@ -1,0 +1,172 @@
+import HighlightText from '../../src/components/HighlightText';
+import { mount } from 'cypress/react';
+
+describe('HighlightText - Tests', function() {
+  it('Renders text with no highlighting if no matches', function() {
+    const lorem = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.';
+
+    mount(
+      <HighlightText
+        highlight={[
+          {
+            bgColor: 'black',
+            textColor: 'black',
+            words: ['unrelated', 'unused', 'words'],
+          }
+        ]}
+        text={lorem}
+      />
+    );
+    // first highlight chunk is entire text
+    cy.get('[data-cy="highlight-0"]').contains(lorem);
+    // no further chunks
+    cy.get('[data-cy="highlight-1"]').should('not.exist');
+  });
+
+  it('Highlights with the correct color, irrelevant of casing.', function() {
+    const text = (
+      'The quick brown fox jumps over the lazy log.'
+    );
+
+    mount(
+      <HighlightText
+        highlight={[
+          {
+            bgColor: 'rgb(0, 0, 0)',
+            textColor: 'rgb(255, 255, 255)',
+            words: ['the', 'jumps'],
+          }
+        ]}
+        text={text}
+      />
+    );
+    // first highlight chunk; regex allows us to ensure there is no extra text
+    cy.get('[data-cy="highlight-0"]').contains(/^The$/);
+    // highlighted CSS props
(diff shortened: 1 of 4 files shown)


src/components/HighlightText.js
```diff
@@ -0,0 +1,153 @@
...
+        </span>
+      );
+    };
+
+
+    // keeps track of last unhighlighted text
+    let endOfLastHighlight = 0;
+
+    // scan through text by character
+    for (let idx = 0; idx < text.length; idx++) {
```
otchet-broad, line 106: Why not search with regular expressions here?

I spent a good time thinking about this one.  The best example around what I was thinking about that I could find was in this repo here: https://github.com/bvaughn/highlight-words-core/blob/master/src/utils.js

cypress/component/highlight_text.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    cy.get('[data-cy="highlight-2"]').should('have.css', 'color', 'rgb(255, 255, 255)');
+
+
+    cy.get('[data-cy="highlight-3"]').contains(/^. ahhhhmultiple words$/);
+    // default, unchanged CSS props
+    cy.get('[data-cy="highlight-3"]').should('have.css', 'background-color', 'rgba(0, 0, 0, 0)');
+    cy.get('[data-cy="highlight-3"]').should('have.css', 'color', 'rgb(0, 0, 0)');
+
+  });
+});
```
otchet-broad, line 174: Should words that are substrings of other words match?

For example if:
`words: ['doctor','cancer']`

and

`text={'The doctors found many types of cancers growing'}`

Would 'doctor' be highlighted in 'doctors' and would 'cancer' be highlighted in 'cancers'?

otchet-broad:
Nice work Connor!  I really like how you were able to add support for multiple colors.

I left a couple of small questions/notes for you to consider.  One might need UX or Johnathan's feedback (should we highlight words found in words).

src/components/HighlightText.js
connorlbark: I assumed it would need each word to be unique / not a subword. For example “Origin” is in the word list, and we wouldn’t want to highlight origin in “original”. I assumed, if we wanted to include plurals such as diseases, we could just add it to the word list, but maybe we should check in with Jonathan first. The big issue with regex for me was that a lot of browsers do not support positive lookbehind, which is the only real way to ensure that a word is surrounded by whitespace to ensure that it is its own word. It’s still possible to use regex, but we would have to manually filter the results if we wanted to make sure that sub-words are not highlighted, which seemed clunky to me. However, if we wanted to highlight subwords, then regex is much easier.

cypress/component/highlight_text.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    );
+    // first highlight chunk is entire text
+    cy.get('[data-cy="highlight-0"]').contains(lorem);
+    // no further chunks
+    cy.get('[data-cy="highlight-1"]').should('not.exist');
+  });
+
+  it('Highlights with the correct color, irrelevant of casing.', function() {
+    const text = (
+      'The quick brown fox jumps over the lazy log.'
```
rushtong: Suggestion - add a variety of casing here, something like: 
```suggestion
      'ThE quick BroWn fox jUMpS over tHe laZy log.'
```

src/components/HighlightText.js
rushtong: > lot of browsers do not support positive lookbehind

Huh - I had to look this up: https://www.lambdatest.com/web-technologies/js-regexp-lookbehind
Looks like this is true. For the cases we really care about (Chrome, newest FF), it is fully supported. On a policy note, we [officially support Chrome](https://support.terra.bio/hc/en-us/articles/360028235911-How-to-register-for-a-Terra-account). I'm fine with the char-by-char approach but would love to see some performance metrics for long RUS statements.

rushtong: Following up on my question, I used [this example RUS](https://www.ncbi.nlm.nih.gov/projects/gap/cgi-bin/GetPdf.cgi?id=phd001030) in DAR-601 (dev). The timing for a rather large RUS seems pretty reasonable. Highlighting takes the most time in rendering that element, which makes sense.
 
Screen Shot 2023-01-09 at 7 34 31 AM

src/components/collection_voting_slab/ResearchProposalVoteSlab.js
```diff
@@ -91,6 +92,27 @@ const animationAttributes = {
...
+      'Algorithm',
+      'Population',
+      'Origin',
+      'Ancestry',
+      'Controls',
+      'Commercial',
+      'Profit'
+    ]
+  }
+];
```
rushtong, line 114: I'm torn on where to define this. I like having the color options local to the component, but I think the words themselves would be more of a fit in a service-level utility. For this PR, I think this is fine. If we add this feature to other components (I could see it in the Application Info component too), then we should pull this up to a higher level of abstraction.

rushtong (approved):
Looks great 👍🏽  See optional comments inline, but I think this is good as is.

src/components/HighlightText.js
connorlbark: I am more than happy to look into making it more performant, maybe using some hashmaps or mixing regex and manual search. But if it's good enough as is, then I am happy with merging too.

rushtong: > But if it's good enough as is, then I am happy with merging too.

I do, thank you!

rushtong:
@connorlbark - Looks like there is a failing test here: https://github.com/DataBiosphere/duos-ui/actions/runs/3878993653/jobs/6615727023#step:5:2125

connorlbark:
> @connorlbark - Looks like there is a failing test here: https://github.com/DataBiosphere/duos-ui/actions/runs/3878993653/jobs/6615727023#step:5:2125

Oops yep, trivial fix.  Thanks!

quazi-h (approved):
The new highlighting component works well for me locally. I am able to observe the correct terms being highlighted in the RUS for members and chairs. Tests look good!

PR #1963, opened 2023-01-03
[DUOS-2077][risk=no] Button to Send Reminder to DAC Member to Vote

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2077

I tried out multiple ways of presenting this to the user. The ticket specifically suggests using a bell icon with a tooltip. However, upon implementation, I found it confusing and felt that an unfamiliar user would have no idea what the bell does until they hover over it (if they ever do). I changed it to be like the original, old table, which just had some clickable text which says "Send Reminder" if the user is a chair and the member has not voted.

----

- Verify all tests go green

[DUOS-2077]: https://broadworkbench.atlassian.net/browse/DUOS-2077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/component/MultiDatasetVoteTab/multi_dataset_vote_slab.spec.js
+++ b/cypress/component/MultiDatasetVoteTab/multi_dataset_vote_slab.spec.js
@@ -427,7 +427,29 @@ describe('MultiDatasetVoteSlab - Tests', function() {
     cy.get('.row-data-3').should('contain.text', 'Matt').should('contain.text', 'Yes');
   });
 
+
   it('Renders filler text when some fields of vote are empty', function() {
+    mount(
+      <MultiDatasetVoteSlab
+        title={'GROUP 1'}
+        bucket={{
+          elections: [openElection2],
+          votes: [votesForOpenElection2]
+        }}
+        dacDatasetIds={[20]}
+        isChair={false}
+      />
+    );
+    cy.stub(Storage, 'getCurrentUser').returns({userId: 100});
+
+    cy.get('.table-data').should('not.exist');
+    cy.get('#show-member-vote-dropdown').click();
+    cy.get('.table-data').should('exist');
+    cy.get('.row-data-0').should('contain.text', 'Joe').should('contain.text', '- -');
+    cy.get('.row-data-2').should('contain.text', 'Matt').should('contain.text', '- -');
+  });
+
+  it('Renders send reminder button when user is chair and no vote', function() {
     mount(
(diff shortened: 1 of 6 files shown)


rushtong (commented):
This works as intended. I think we need some feedback to indicate that a reminder has been sent to a user. In the original version, we would use Noty to show a brief onscreen notification. See utils `Notifications.showSuccess()` for examples.

connorlbark:
Good point @rushtong, adding that. I also wanted to ask, since we're saving the emails that we're sending now, do we want to make sure only one email is sent if the user clicks more than once? I could open up a quick separate backend PR for that.

rushtong:
> I also wanted to ask, since we're saving the emails that we're sending now, do we want to make sure only one email is sent if the user clicks more than once? I could open up a quick separate backend PR for that.

@connorlbark - We do need something, let's put it in a ticket. The idea is that we want the API to return an error response (of some sort) if the user has hit it within some time period, to be defined. We could also do something smart on the front end too so that we change the text to "Reminder email sent" once clicked. I don't think we need too much state retention here, so if the user refreshes, this goes away, so let's keep it simple.

connorlbark:
> @connorlbark - We do need something, let's put it in a ticket. The idea is that we want the API to return an error response (of some sort) if the user has hit it within some time period, to be defined. We could also do something smart on the front end too so that we change the text to "Reminder email sent" once clicked. I don't think we need too much state retention here, so if the user refreshes, this goes away, so let's keep it simple.

What do you think of just keeping this in the frontend? Perhaps when they click send reminder and it's successful, it changes to green text saying `Reminder Sent`. That way they can send multiple if they really really want to

quazi-h (approved):
Changes look good to me, I see the Send Reminder button under the VOTE column for the DAC member votes. (Note: the mock on the ticket shows the link under the "Rationale" column instead of the "Vote" column but I think it's good the way implemented it). Clicking the button sends an email and I see a "Successfully sent reminder" notification.

rushtong:
> What do you think of just keeping this in the frontend? Perhaps when they click send reminder and it's successful, it changes to green text saying `Reminder Sent`. That way they can send multiple if they really really want to

@connorlbark - Sorry for the late reply on this, I think your suggestion is perfect, thank you.

rushtong (approved):
Thanks for the changes!

PR #1972, opened 2023-01-06
[DUOS-2138][risk=no] Update NPOA logic

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2138

Minor update to the current dataset edit/create page to show the **NPOA** option correctly:

When editing an existing dataset:
* If dataset's `populationOriginsAncestry` is true (or null), we do not show **NPOA**
* If dataset's `populationOriginsAncestry` is false, we show **NPOA** as selected

When creating a new dataset (only visible when selecting **GRU**):
* If **NPOA** is unchecked, we do not set the dataset's `populationOriginsAncestry`
* If **NPOA** is checked, we set the dataset's `populationOriginsAncestry` as `false`

----

- Verify all tests go green


quazi-h (approved):
New logic looks sound to me, if the dataset POA is false, then the result should be NOT POA.

connorlbark (approved):
👍

PR #1977, opened 2023-01-09
[DUOS-1740][risk=no] Add separate build action

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1740

Currently, the only action that runs a docker build is run for non-dependabot PRs. So, when a dependabot update breaks the docker build ([for example](https://github.com/DataBiosphere/duos-ui/pull/1957)), we only find out about it if we run that locally or merge it to dev. This action will catch those kinds of errors so we know about them beforehand.

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,10 @@
+on: [pull_request]
+name: docker build
+jobs:
+  scan:
+    name: docker build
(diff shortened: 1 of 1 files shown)


connorlbark (approved):
👍 seems reasonable to me

PR #1993, opened 2023-01-20
[DUOS-2232][risk=no] Add email to DAC edit screens

kaitlynoflaherty:
## Addresses

https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2232

Updates all add/edit screens to allow for the email field.

image

----

- Verify all tests go green


src/libs/ajax.js
```diff
@@ -102,9 +102,9 @@ export const DAC = {
     return res.json();
   },
 
-  update: async (dacId, name, description) => {
+  update: async (dacId, name, description, email) => {
     const url = `${await getApiUrl()}/api/dac`;
-    const dac = { 'dacId': dacId, 'name': name, 'description': description };
+    const dac = { 'dacId': dacId, 'name': name, 'description': description, 'email': email };
```
nmalfroy: General note about newer ES syntax, because the variables match, something like:
```const dac = { 'dacId': dacId, 'name': name, 'description': description, 'email': email };```
can be written in shorthand as:
```const dac = { dacId, name, description, email };```

...I'll leave it up to you if you have a preference - just a neat syntax hack :)

File: src/pages/manage_dac/AddDacModal.js [hunk 2]
```diff
@@ -294,6 +303,25 @@ export const AddDacModal = hh(class AddDacModal extends Component {
             ])
           ]),
 
+          div({ className: 'form-group first-form-group' }, [
+            label({
+              id: 'lbl_dacEmail',
+              className: 'col-lg-3 col-md-3 col-sm-3 col-xs-4 control-label common-color'
+            }, ['Email']),
+            div({ className: 'col-lg-9 col-md-9 col-sm-9 col-xs-8' }, [
```
nmalfroy, line 286: I get that you just followed convention here but eventually, I think it might be better to use a dedicated `Grid` component rather than the classes since those could change I think.

that's a much bigger change and I think you did the right thing in this PR to follow convention.  Just noting for posterity

src/pages/manage_dac/AddDacModal.js
```diff
@@ -191,6 +191,15 @@ export const AddDacModal = hh(class AddDacModal extends Component {
           return prev;
         });
         break;
+      case 'email':
```
nmalfroy: seems like this whole switch statement could be simplified to something like:
```
this.setState(prev => {
  let newDac = Object.assign({}, prev.dac);
  newDac[name] = value;
  prev.dac = newDac;
  prev.dirtyFlag = true;
  return prev;
});
```

nmalfroy (approved):
this looks good!  Not sure how to get to this in the UI to get a sense of what the behavior is that it's modifying. I put some comments in around different code approaches but those grow the scope of the PR so I don't know if we want to tackle those here

src/components/TableIconButton.js
```diff
@@ -43,7 +43,8 @@ export default function TableIconButton(props) {
   } = props;
   const Icon = props.icon;
   const classes = useStyles();
-  const appliedStyle = setStyle(disabled, style, 'color');
+  //const appliedStyle = setStyle(disabled, style, 'color');
```
nmalfroy: looks like you've got commented code here

src/libs/ajax.js
```diff
@@ -95,16 +95,16 @@ export const DAC = {
     return res.json();
   },
 
-  create: async (name, description) => {
+  create: async (name, description, email) => {
     const url = `${await getApiUrl()}/api/dac`;
-    const dac = { 'name': name, 'description': description };
+    const dac = { 'name': name, 'description': description, 'email': email };
```
nmalfroy: you can do the same thing as below here.  E.g.
```suggestion
    const dac = { name, description, email };
```

src/libs/ajax.js
rushtong: That's nice, didn't know that either.

src/pages/manage_dac/AddDacModal.js
rushtong: I agree Muscles - we do need to migrate a lot of form components. We have started that process, but this is one of the earlier things I did when we started full time on DUOS. You can see in the PR screenshot how we have some alignment issues and I'm sure that these bootstrap classes are the cause.

src/libs/ajax.js
```diff
@@ -95,16 +95,16 @@ export const DAC = {
     return res.json();
   },
 
-  create: async (name, description) => {
+  create: async (name, description, email) => {
     const url = `${await getApiUrl()}/api/dac`;
-    const dac = { 'name': name, 'description': description };
+    const dac = { name, description, email };
```
rushtong, line 100: The back end considers email optional - not sure if we've tested that case (I haven't). Might be worth some logic around the construction of the `dac` object.

rushtong (approved):
👍🏽 Thank you!

PR #2001, opened 2023-01-26
[DUOS=2244][risk=low] UX Followons for info page

nmalfroy:
- Update images for size and color
- Fix some text
- Make pages responsive
- Add alt text to images

Check out the behavior for the various sections.  At some point, upgrading to Material UI 5 (e.g. MUI) makes this easier

Top of page
top-resize

Middle sections
middle-resize

Bottom bullet points
bottom-resize

----

- Verify all tests go green


src/pages/DMSPolicyInfo.css
```diff
@@ -49,19 +59,43 @@
 .subtitle-section .section-text {
   padding-top: 10px;
 }
+/*
+  This means that when the window is less than 80px wide (e.g. grid breakpoint = l) then the laptop image will be hidden
```
fboulnois:
```suggestion
  This means that when the window is less than 800px wide (e.g. grid breakpoint = l) then the laptop image will be hidden
```

src/pages/DMSPolicyInfo.css
```diff
@@ -49,19 +59,43 @@
...
 .store-data-section {
   background-color: #ececec; /* d8d8d8d with 0.5 opacity over white */
   padding: 50px 0;
 }
 .store-data-section img {
   width: 100%;
-  padding-left: 40px;
 }
-
+/* remove padding in mobile mode */
```
fboulnois:
```suggestion
/* add padding in desktop view */
```

I would either put `/* no padding in mobile view */` on the `.store-data-section` without media queries, or I would recommend using my suggestion instead. As is it seems a bit confusing to have the comment above the media query. Ditto for the other cases.

fboulnois (commented):
Some minor suggestions. Also do we use `svg`s anywhere in the UI? If so the duos chart seems like a prime candidate, especially since it means we can losslessly resize it on different viewports.

rushtong (approved):
Looked at it in mobile mode, looks great 👍🏽

src/pages/DMSPolicyInfo.css
nmalfroy: gah, nice catch

src/pages/DMSPolicyInfo.css
nmalfroy: ah yeah, that's a fair point.  Will update

nmalfroy:
Yeah, there are SVGs but I just used what I got from Lou for this one

fboulnois (approved):
lgtm 😎

PR #2004, opened 2023-01-27
[DUOS-2270][risk=no] Getting started guide for DUOS

fboulnois:
Adds a `getting-started.md` guide which helps new software engineers onboard onto the DUOS team.

[View the rendered markdown.](https://github.com/DataBiosphere/duos-ui/blob/407aee8507a4960990671d23e4f2b214516227b6/docs/getting-started.md)

The philosophy is to make the onboarding guide as straightforward as possible:

- Align our guide with the TDR and TDC ones where it makes sense
- Automate steps when possible to lower the barrier to setup and to prevent developers from making mistakes
- Explain a minimum number of unfamiliar concepts to new developers so they can understand why a step is necessary

### DIFF
--- a/docs/Brewfile
+++ b/docs/Brewfile
@@ -1,0 +1,10 @@
+brew "git"
+brew "git-secrets"
+brew "jq"
+brew "vault", restart_service: true
+brew "volta"
+cask "docker"
+cask "google-cloud-sdk"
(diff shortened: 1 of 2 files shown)


rushtong (approved):
Thank you for this! 👍🏽

kaitlynoflaherty (approved):
Looks great! Thank you

PR #2030, opened 2023-02-03
[DUOS-2260][risk=no] Fix missing key console errors

rushtong:
This PR adds keys to all of the navigation tabs

----

- Verify all tests go green

### DIFF
--- a/src/components/DuosHeader.js
+++ b/src/components/DuosHeader.js
@@ -72,13 +72,13 @@ export const headerTabsConfig = [
     label: 'Admin Console',
     link: '/admin_manage_dar_collections',
     children: [
-      { label: 'DAR Requests', link: '/admin_manage_dar_collections' },
-      { label: 'Dataset Catalog', link: '/dataset_catalog' },
-      { label: 'DACs', link: '/manage_dac' },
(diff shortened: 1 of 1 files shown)


kaitlynoflaherty (commented):
I tested this locally and ran into an issue. When I log out and log back in, I can navigate the tabs without any errors. When I reload the page, the same warning pops up. I'm not sure what could be causing it to only appear after reloading the page.
image

fboulnois (approved):
curious about Kaite's error, otherwise lgtm

rushtong:
> I tested this locally and ran into an issue. When I log out and log back in, I can navigate the tabs without any errors. When I reload the page, the same warning pops up. I'm not sure what could be causing it to only appear after reloading the page.

Thanks @kaitlynoflaherty ... gonna dig in some more, not sure I saw that but could have missed something

kaitlynoflaherty (approved):
Does the same error occur on your local environment? I think its good otherwise

rushtong:
> Does the same error occur on your local environment? I think its good otherwise

It does. I'm going to close this and we can address it separately.

PR #2032, opened 2023-02-03
[DUOS-2294][DUOS-2295][risk=no] Update notifications for minimized navigation panel

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2294
https://broadworkbench.atlassian.net/browse/DUOS-2295

This PR updates the minimized drawer navigation panel so that it doesn't try to render the notifications a second time. It also reduces the spacing over the notification. Lastly, it addresses some react console warnings about missing keys in the navigation menu.

### New
Screenshot 2023-02-06 at 10 02 21 AM

### Old
Screenshot 2023-02-03 at 3 02 29 PM

----

- Verify all tests go green


src/pages/DMSPolicyInfo.css
```diff
@@ -126,10 +126,6 @@
   margin-bottom: auto !important;
 }
 
-p {
-  margin-top: 10px;
-}
```
rushtong: This global style update interfered with other `p` tags across the site.

fboulnois (approved):
looks great!

src/pages/DMSPolicyInfo.css
nmalfroy: Gah sorry about that. Rather than deleted, is it possible to scope it?  I think this might mess up the DMS page

rushtong: I think we have to scope it on the DMS page. The `p` tags where I'm seeing issues is when we convert text to markdown in the notifications. Let me take another stab at it ... might be able to make something work.

nmalfroy: I think you should be able to just do:
```
.base-style p {
  margin-top: 10px;
}
```

(probably need a more dms-page specific base class)

I also noticed the same thing right below for the `b` tag

rushtong: OK, I think I have a fix for it - just have to override what's there.

rushtong: I ended up reverting this change so we can address both global styles separately

nmalfroy: Ah gotcha

nmalfroy (approved):
looks good!

PR #2035, opened 2023-02-06
[DUOS-1442][risk=no] Migrate to get datasets v2 call

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1442 but does not at all complete that story.

This PR just updates the UI calls for two features, "Dataset Catalog" and "View DAC Datasets", to use the real `Dataset` object from Consent as opposed to what they're using now which is a `DatasetDTO`. The `DatasetDTO` is deprecated (we've been migrating for quite some time now). The only difference between the two is that the DTO treats the `alias` field as the long form identifier where we have a newer field for long form identifier called `datasetIdentifier`.

----

- Verify all tests go green


fboulnois (approved):
lgtm

kaitlynoflaherty (approved):
Looks good!

PR #2036, opened 2023-02-07
[DUOS-2239][risk=no] Add Submit Button to Data Submission Form

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2239

This PR includes a lot of small / minor changes that are needed to get this form cworking.
- Some field names / enum values were wrong, and had to change to match backend
- Updated `Number` field type to be properly functional, and return to the form values of a number type 
- Revamped how file state is maintained to be easier to send to the backend

Functionally / visually, nothing should change except minor edits and the addition of a submit button.

Right now, the submit button works, but validation does not. As such, you will get a server error if there are any validation errors. In the future, UI-side validation will prevent this popup from occurring.

----

- Verify all tests go green


src/pages/DataSubmissionForm.js
```diff
@@ -41,50 +41,88 @@ export const DataSubmissionForm = () => {
...
+    const files = {};
+
+    if (!isEmpty(registration.alternativeDataSharingPlanFile)) {
+      files.alternativeDataSharingPlan = registration.alternativeDataSharingPlanFile;
+      delete registration.alternativeDataSharingPlanFile;
+    }
+
+    const consentGroups = registration['consentGroups'];
+
+    for (let i = 0; i < consentGroups?.length; i++) {
```
rushtong: nit: prefer map with index over for loops.

src/pages/DataSubmissionForm.js
```diff
@@ -41,50 +41,88 @@ export const DataSubmissionForm = () => {
...
+      if (!isEmpty(consentGroups[i].nihInstitutionalCertificationFile)) {
+        files['consentGroups['+i+'].nihInstitutionalCertification'] = registration.nihInstitutionalCertificationFile;
+        delete consentGroups[i].nihInstitutionalCertificationFile;
+      }
```
rushtong: Trying to grok this and not quite getting why we need to do it. I can see the desired need to refrain from uploading duplicate files, but not quite sure we're doing that right. The main form doesn't have an `nihInstitutionalCertificationFile` form field - the only place we have this form element is on the consent group panel. Additionally, what is the value of `nihInstitutionalCertificationFile`? Is it a file name? If so, can it be true that we have two files with the same name (but from different directories) on the user's drive?

rushtong (commented):
Still reviewing, a couple of comments so far.

src/pages/DataSubmissionForm.js
rushtong: Thinking more on this. I believe that restricting the user from uploading the same file for multiple consent groups is technically incorrect. A user should be able to use the same file multiple times. We should revisit this on the API side as well as I think we're doing something similar there.

connorlbark: These are `File` objects, so they contain information about the filename, etc. All this function does is copy it from the `consentGroups` objects where they are populated by the subcomponents of the form, into their own object so they can be uploaded in a separate form data body part. There is no issue with duplicate files, you can upload the same file, multiple files with the same name, etc. as many times as you want, and that's all good.

connorlbark: I updated the code paths so that files have their own `formData` type object, so hopefully it's all a bit clearer. I also uploaded the same file in 3 different places and it all uploaded without a hitch.

src/pages/DataSubmissionForm.js
fboulnois: I had this thought too. Too bad Java doesn't have a built-in in `Enumerate` method like Python!

src/components/forms/formComponents.js
```diff
@@ -74,14 +74,29 @@ export const formInputGeneric = (config) => {
...
       placeholder: placeholder || title,
-      value: formValue,
+      value: formValue.toString(),
       readOnly: readOnly,
       style: { ...styles.inputStyle, ...inputStyle },
       disabled: disabled,
-      onChange: (event) => onFormInputChange(config, event.target.value),
+      onChange: (event) => {
+        const value = event.target.value;
+        if (type?.inputType === 'number') {
```
fboulnois: This parsing looks pretty complicated, especially since `onFormInputChange` calls `validateFormInput`. Perhaps the functionality could be moved into a `parseFormInput` function which is called after `validateFormInput`?

fboulnois (approved):
One suggestion:

rushtong (approved):
Looks good to me 👍🏽

PR #2040, opened 2023-02-07
[DUOS-2259][risk=no]SO Console - Data Submitter Tab content update

kaitlynoflaherty:
## Adresses 
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2259
Makes changes to only Data Submitters tab on the SO Console:

- Changes “Library Card” to “Data Submitter” in the heading and sub-heading
- Changes “Broad Library Card Agreement” to “DUOS Data Submitter Agreement” in the button
- Replaces the downloadable file with https://docs.google.com/document/d/1UTPProxotG4xuPd291NS71VEIatVQqRO/edit?usp=sharing&ouid=105186821507829169555&rtpof=true&sd=true 
- Removes the second button with “NIH Library Card Agreement”
image

----

- Verify all tests go green


src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -60,23 +68,27 @@ export const SigningOfficialDaaAgreementWrapper = (props) => {
...
       h2({}, [
-        'Agree to Library Card Terms',
+        'Agree to ' + replacementText + ' Terms'
       ]),
 
       p({
         style: {marginBottom: '20px'}
       }, [
-        'To begin issuing Library Cards to researchers from your institution, please review the terms of both data access agreements below and click \'I agree\' when finished.'
+        'To begin issuing ' + replacementText + 's to researchers from your institution, please review the terms of the data access agreement(s) below and click \'I agree\' when finished.'
```
fboulnois: I feel like "to begin issuing Data Submitter to researchers..." sounds strange, even though that's the recommendation.

src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -2,21 +2,29 @@ import { h, h2, a, p, div, span, button } from 'react-hyperscript-helpers';
...
-    children
+    children,
   } = props;
 
+  const {
+    props: {
+      dataSubmitter,
+    },
+  } = children;
+
```
fboulnois: I wonder if there's a better way to do this, the repeated 'props' element causes issues.

fboulnois (commented):
A few minor comments:

src/components/SigningOfficialDaaAgreementWrapper.js
kaitlynoflaherty: I think "to begin issuing Data Submitter privileges to researches" sounds better. What do you think?

rushtong: As a side note, language choices are often required for various reasons. I think in this case, we can make some common-sense changes. In other places, especially around data use questions and phrasing, we need to defer to Jonathan since they often have roots in GA4GH or Broad legal requirements.

src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -60,23 +68,27 @@ export const SigningOfficialDaaAgreementWrapper = (props) => {
...
+          span({className: 'glyphicon glyphicon-download'}),
+          ' DUOS Data Submitter Agreement'
+        ]) :
         a({ target: '_blank', href: BroadLibraryCardAgreementLink, className: 'button button-white', }, [
           span({className: 'glyphicon glyphicon-download'}),
           ' Broad Library Card Agreement'
         ])
       ]),
-      div({}, [
+      div({}, [dataSubmitter === true ? isNull :
```
rushtong: I think this is fine as is. In the past, we've seen bugs around this kind of thing where the field was truth-y (i.e. 'true', "true", "TRUE", etc., but not a boolean primitive.

rushtong (approved):
👍🏽

kaitlynoflaherty (commented):
I took out the 'replacementText variable' and put the condition straight inline because the text wasn't easily reusable. I'm not sure if this is the best approach, so let me know if I should rework it. Also, figured out the props issue!

rushtong (approved):
This approach is definitely easier to grok 👍🏽

src/Routes.js
```diff
@@ -88,7 +88,7 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/profile" component={ResearcherProfile} props={props} rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/signing_official_console/researchers" component={ensureSoHasDaaAcknowledgement(SigningOfficialResearchers)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
     <AuthenticatedRoute path="/signing_official_console/dar_requests" component={ensureSoHasDaaAcknowledgement(SigningOfficialDarRequests)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
-    <AuthenticatedRoute path="/signing_official_console/data_submitters" component={ensureSoHasDaaAcknowledgement(SigningOfficialDataSubmitters)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
+    <AuthenticatedRoute path="/signing_official_console/data_submitters" component={ensureSoHasDaaAcknowledgement(SigningOfficialDataSubmitters, true)} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.signingOfficial]} />
```
fboulnois, line 91: This is definitely cleaner.

src/components/SigningOfficialDaaAgreementWrapper.js
```diff
@@ -93,8 +98,8 @@ export const SigningOfficialDaaAgreementWrapper = (props) => {
 // Wraps component and ensures that SO agrees to the
 // Broad and NIH agreements before proceeding to the given
 // component.
-export const ensureSoHasDaaAcknowledgement = (component) => {
-  return (props) => h(SigningOfficialDaaAgreementWrapper, {}, [
+export const ensureSoHasDaaAcknowledgement = (component, isDataSubmitter=false) => {
```
fboulnois:
```suggestion
export const ensureSoHasDaaAcknowledgement = (component, isDataSubmitterTab=false) => {
```

One minor suggestion, switch `isDataSubmitter` to `isDataSubmitterTab` across the board since the user role isn't what is actually changing.

fboulnois (commented):
One suggestion:

fboulnois (approved):
looks great!

PR #2041, opened 2023-02-07
[DUOS-2264][risk=no]Expand collapsible table width on all DAR tables 

kaitlynoflaherty:
## Adresses 
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2264

- Displays the collapsible DAR ‘Data Use Group’ table with the same width as its parent row on all Console DAR Tables
- Decreases the spacing between the column titles and the first row in both the primary and secondary tables by 50%

**I'm  a little confused about the following instruction, but I think it looks good. Let me know if you think it needs updates: 
- The increased row width should be ~70% allocated to increasing the width of the Datasets column, and ~30% allocated to the Data Use Group column. 

image

----

- Verify all tests go green

### DIFF
--- a/src/libs/theme.js
+++ b/src/libs/theme.js
@@ -187,10 +187,9 @@ export const Styles = {
       fontFamily: 'Montserrat',
     },
     CARDCONTAINER: {
-      margin: '2rem -12.5%',
+      margin: '-.75rem -12.5%',
       border: '1px solid #979797',
       backgroundColor: 'rgb(184,205,211,0.08)',
-      borderRadius: '5px',
       padding: '1rem 4rem',
       color: '#7B7B7B',
       fontFamily: 'Montserrat',
--- a/src/libs/theme.js
+++ b/src/libs/theme.js
@@ -187,11 +187,11 @@ export const Styles = {
       fontFamily: 'Montserrat',
     },
     CARDCONTAINER: {
-      marginTop: '2rem',
+      margin: '2rem -12.5%',
       border: '1px solid #979797',
       backgroundColor: 'rgb(184,205,211,0.08)',
       borderRadius: '5px',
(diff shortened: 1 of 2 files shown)


src/libs/theme.js
```diff
@@ -187,11 +187,11 @@ export const Styles = {
       fontFamily: 'Montserrat',
     },
     CARDCONTAINER: {
-      marginTop: '2rem',
+      margin: '2rem -12.5%',
```
rushtong: Can we tighten up the border between the top of the opened table with the row above it? Locally, I was able to get those to line up with `margin: '-.75rem -12.5%',`. However, `rem` is somewhat flexible and could change in different browsers. Might be nice to modify the border radius on the top of the table and bottom of the summary row to make them look even more tightly coupled.

## Original
Screenshot 2023-02-09 at 7 17 04 AM

## Updated
Screenshot 2023-02-09 at 7 22 17 AM

rushtong (approved):
See optional suggestions inline, implement as you see fit, but as is, I think this is a nice improvement 👍🏽

src/libs/theme.js
```diff
@@ -187,11 +187,10 @@ export const Styles = {
       fontFamily: 'Montserrat',
     },
     CARDCONTAINER: {
-      marginTop: '2rem',
+      margin: '-.75rem -12.5%',
```
fboulnois:
```suggestion
      margin: '-.75rem -12.5% 2rem',
```

I think it's still worth visually separating the opened table from the row **below** it. So I'd recommend something like this.

fboulnois (changes requested):
One recommendation:

fboulnois (approved):

PR #2046, opened 2023-02-13
[DUOS-2338][risk=no] Conditionally display the SO Data Submitters sub nav

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2338

This PR was a little more complicated than I had hoped it would be. We currently have render logic (based on the user) at the first tab level, but we had no logic at the subtab level. This PR does several things:

1. Extract all conditional route logic in `Routes.js` to a new utility.
2. Add the current application environment to local storage so we can access it throughout the app without having to await a promise from `Config`.
3. Add optional and flexible rendering logic at the sub nav level
4. Apply that to the SO Data Submitters sub-nav

### Testing

To test, manipulate the `env` value in local storage, then switch to the "SO Console" tab. You have to switch out and back again between top level tabs to see your changes to the `env` value. The "Data Submitters" sub tab should only show up when the env is `local`, `dev` or `alpha`. Any other value will hide the rendering of the sub tab.

Screenshot 2023-02-14 at 10 38 22 AM

Screenshot 2023-02-14 at 10 39 23 AM

Screenshot 2023-02-14 at 10 40 40 AM

----

- Verify all tests go green


src/libs/storage.js
```diff
@@ -65,5 +66,13 @@ export const Storage = {
...
   removeData: key => {
     sessionStorage.removeItem(key);
+  },
+
+  setEnv: (value) => {
+    sessionStorage.setItem(ENV, value);
+  },
+
+  getEnv: () => {
+    return sessionStorage.getItem(ENV);
```
connorlbark, line 76: nice. i ran into an issue like this before and didn't how to handle it best - i like this solution.

connorlbark (approved):
👍 works great for me locally, loving envrionmentutils addition

src/components/DuosHeader.js
```diff
@@ -322,7 +332,7 @@ const NavigationTabsComponent = (props) => {
             },
           },
           component: Link
-        });
+        }) : <div/>;
```
fboulnois: Should we return an empty div or no element if the condition is false?

src/utils/EnvironmentUtils.js
```diff
@@ -0,0 +1,41 @@
...
+ *  * prod
+ *    * staging
+ *      * alpha
+ *      * dev
+ *      * local
+ * @type {{NON_STAGING: string[], NON_PROD: string[]}}
+ */
+export const envGroups = {
+  NON_PROD: ['local', 'dev', 'alpha', 'staging'],
+  NON_STAGING: ['local', 'dev', 'alpha']
```
fboulnois, line 17: I imagine these environments will be pretty stable, but I wonder if it would be useful to be extensible somehow:

```
const ALL_ENVS = [ 'local', 'dev', 'alpha', 'staging', 'prod' ];
const NON_PROD = _.without(ALL_ENVS, 'prod');
const NON_STAGING = _.without(ALL_ENVS, 'staging', 'prod');

export const envGroups = {
    ALL_ENVS: ALL_ENVS,
    NON_PROD: NON_PROD,
    NON_STAGING: NON_STAGING
};
```

Perhaps this is too much unnecessary complexity though.

fboulnois (approved):
lgtm, a few comments:

src/utils/EnvironmentUtils.js
rushtong: I think this would be a good idea if we needed something else that used `ALL_ENVS`. I think what I have now is extensible enough for future development and yet is the minimum needed to accomplish the goal.

PR #2058, opened 2023-02-16
[DUOS-2261][risk=no] Rewrite bucketing logic

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2261

Rewrite of the bucketing logic so that buckets can be distinct in each case where there is an `Other` or undefined restriction. This PR tries to arrange the bucket construction in a more stepwise fashion that should be easier to extend when we need to make more changes for Progress Reports and Closeouts. It tries to be a drop-in replacement for the original bucket objects. 

The only logic change involves filtering buckets by user/dac. Previous logic filtered by user by examining votes to ensure that the user had a vote for the dar-dataset combination. This PR changes that to instead look at the user's DACs. If the dataset exists in one of the user's DACs, the bucketing logic will optionally exclude those datasets (as well as votes and elections) from the bucket which has the same effect. One possible error case that might exist (will need further examination and testing) is the case where a user is a member of a DAC, but for some reason, doesn't have a vote for the dar-dataset combination. This error case should not exist, but we should look into flagging/exposing it.

----

- Verify all tests go green


src/components/collection_voting_slab/DataUseAlertBox.js
```diff
@@ -28,8 +28,8 @@ const styles = {
 const dataUseDescriptions = (translatedDataUse) => {
   return flatMap(key => {
     const dataUses = translatedDataUse[key];
-    return map(dataUse => {
-      return div({key: dataUse.code}, [dataUse.description]);
+    return map.convert({cap: false})((dataUse, index) => {
+      return div({key: dataUse.code + '-' + index}, [dataUse.description]);
     })(manuallyReviewedDataUses(dataUses));
```
rushtong, line 33: In this new world, we can have duplicate keys with slightly different data uses so we need to ensure key uniqueness here

src/components/dar_dataset_table/DarDatasetTable.js
```diff
@@ -182,14 +181,12 @@ export const DarDatasetTable = (props) => {
         setBuckets([]);
         return;
       }
-      const processedBuckets = await flow([
-        generatePreProcessedBucketData,
-        processDataUseBuckets,
-      ])({ dars, datasets });
-      const filteredDataUseBuckets = processedBuckets.filter(
+      const binnedBuckets = await binCollectionToBuckets(collection);
+      const filteredBinnedBuckets = binnedBuckets.filter(
```
rushtong: There's a bug directly above this block. Lines 165-179 effectively remove dars/datasets from the buckets if the user doesn't have any votes in them. I need to add the same filtering logic in the voting pages to this page as well.

src/utils/BucketUtils.js
```diff
@@ -0,0 +1,251 @@
...
+ * Step 4: Coalesce the algorithm decision per bucket
+ * Step 5: Pull all votes up to a top level bucket field for easier iteration
+ * Step 6: Prepend an RP Vote bucket for the DAC to vote on the research purpose
+ *
+ * @param collection The full Data Access Request Collection
+ * @param dacIds An optional array of dac ids. If provided, bucket contents will be filtered to datasets matching
+ *        the provided dac ids. This will extend to elections and votes as well.
+ * @returns {Promise<*[Bucket]>}
+ */
+export const binCollectionToBuckets = async (collection, dacIds = []) => {
```
connorlbark, line 25: would it be possible to break this function up into smaller pieces? it took me a long time to grok what was happening here, and it mostly looks good, but it's hard to tell / keep it all in my mind at once

src/utils/BucketUtils.js
```diff
@@ -0,0 +1,251 @@
...
+    // All the same match value? Choose the first one
+    if (matchVals.length === 1) {
+      const match = bucket.matchResults[0];
+      const { createDate, failureReasons, id } = match;
+      return {
+        result: processMatchData(match),
+        createDate,
+        failureReasons,
+        id
+      };
```
connorlbark: nit: could you collapse these into one branch? it seems like the `uniq` call would cover the case where there's only 1 match as well as >1

src/utils/BucketUtils.js
```diff
@@ -0,0 +1,251 @@
+import {any, compact, findIndex, filter, flatMap, flow, forEach, get, getOr, includes, isEmpty, isEqual, isUndefined, join, map, toLower, uniq, values} from 'lodash/fp';
```
connorlbark: looking at all of these at a high level, i almost feel like these bucketing functions should exist in ontology or consent. it feels like the frontend is doing too much of the nitty gritty logic. it seems like it would be easier / more stable to let the backend handle that. i think for now, to get these bug fixes out, this is great cleanup, but long term i feel like it would be good to push this back.

src/utils/BucketUtils.js
```diff
@@ -0,0 +1,251 @@
...
+  // Find all match results for this collection. This will be placed into each
+  // bucket based on the dataset that the match applies to in step 1.a
+  const referenceIds = flow(
+    map(d => d.referenceId),
+    uniq
+  )(collection.dars);
+  const matchData = referenceIds.length > 0 ? await Match.findMatchBatch(referenceIds) : [];
+  const datasets = isEmpty(dacIds) ?
+    get('datasets')(collection) :
+    filter(d => includes(d.dacId)(dacIds))(get('datasets')(collection));
```
connorlbark: nit: personally, i find `lodash/fp` a bit awkward, but in particular i find the nested `lodash/fp` calls here confusing, because it doesn't flow left->right. maybe break into multiple lines?

src/utils/BucketUtils.js
```diff
@@ -0,0 +1,251 @@
...
+  // Find all translated data uses for all datasets. `translateDataUseRestrictionsFromDataUseArray` creates a parallel,
+  // ordered array in the same order as rawDataUses, so we can associate them by index. Unfortunately, it also creates
+  // empty elements per translation (one for any missing potential translation), so we need to filter those out.
+  const rawDataUses = compact(map(d => d.dataUse)(datasets));
+  const translatedDataUses = await translateDataUseRestrictionsFromDataUseArray(rawDataUses);
+  const flatTranslatedDataUses = map(t => compact(t))(translatedDataUses);
+
+  // Step 1: Create buckets for unique dataset groups
+  map(d => {
```
connorlbark: nit: i feel like `datasets.map(dataset => {` would be cleaner for the long, multiline code that's being executed, as it's hard to tell what data type to expect when the resolution of the map call is at the end, so you have to go back and forth. even just renaming `d` to `dataset` would help, i think.

src/utils/BucketUtils.js
```diff
@@ -0,0 +1,251 @@
...
+    b.votes.push(processVotesForBucket(b.elections));
+  })(buckets);
+
+  // Step 6: Populate RUS Vote bucket with RP votes
+  const rpVotes = createRpVoteStructureFromBuckets(buckets);
+  buckets.unshift({
+    isRP: true,
+    key: 'RUS Vote',
+    votes: rpVotes
+  });
```
connorlbark, line 130: suggestion: structurally, i find the `isRP` confusing, since multiple types are intermixed in one array. what if the returned object was of the structure `{ rpBucket: {}, dataUseBuckets: [{}, {}] }`? i am ok keeping it the same way if it's preferred.

connorlbark (commented):
looks like a great cleanup! some initial thoughts! mostly, i think it would be good to break up `binCollectionToBuckets` into smaller and more easily testable steps, and test those components more individually.

src/utils/BucketUtils.js
rushtong: I can break it up some more ... I tried to break out the larger parts into smaller methods already. I can certainly make it more granular

src/utils/BucketUtils.js
rushtong: Nice idea, will do.

src/utils/BucketUtils.js
rushtong: I'll try to make this more legible.

src/utils/BucketUtils.js
rushtong: Actually, I don't need a map at all ... this should be a `forEach` since there is no need for a return value.

src/utils/BucketUtils.js
rushtong: TBH, I found it confusing too. However, that's the field that the original version defined and is used in several other functions/components. My goal here was to make as much of a drop-in replacement as possible.

src/utils/BucketUtils.js
fboulnois: +1, this is the core logic change of this PR and I had to reread it a few times.

fboulnois (approved):
I think this approach is definitely more extensible than the previous approach. The logic is still somewhat challenging to follow, I also think `binCollectionToBuckets` could be broken apart more. Doing so would be clearer and easier to unit test to give us confidence that the logic works as expected.

src/utils/BucketUtils.js
rushtong: Part of the problem is that the buckets are different depending on the view. As an admin, buckets are unfiltered. As a chair, buckets have non-dac datasets/votes/elections filtered out. This is purely a UI preference since there is no systemic reason why a dac chair cannot know about elections in other DACs. I can see the appeal of moving it to the back end, but we're going to change our minds on a somewhat regular basis about what a view/voting screen looks like. Having a stable backend API that provides the correct data will let us be flexible in how we display it.

connorlbark (approved):
👍 much cleaner than before! i still think it could likely be broken up more into subfunctions, but i think it's much clearer now, and i can piece together the logic much more naturally.

fboulnois (approved):
lgtm

PR #2059, opened 2023-02-16
[DUOS-2304][risk=no] PART 2 - Add JSONSchema validation to Data Submission Form 

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2304

## PART 1

See https://github.com/DataBiosphere/duos-ui/pull/2068

## PART 2 (this ticket)
- Added a JSON Schema utility function which is able to validate the form according to a JSON schema and then convert it into the above validation state object
  - e.g., uses JSON schema from server to convert `{ someField: "asdf", someOtherField: "not a url" }` -> `{ someField: { valid: true }, someOtherField: { valid: false, failed: ['uri'] } }`
- Data submission thus: 
  - downloads the schema from the backend when the page laods
  - keeps validation state in a top level object
  - updates this object as the user fills out the form
  - when the user hits submit, runs the new JSON schema validation code path
    - if invalid, then the validation state is set to the JSON schema results
    - if valid, sends submission to backend.

## Notes / Other Changes
- Had to allow `/schemas` through the backend proxy

## Known issues
- The frontend and backend don't match perfectly on validation; frontend marks some non-required fields as required, and vice versa.
- Backend schema _allows_ required array fields which have no elements, so, e.g., even though `consentGroups` is required, `consentGroups: []` is considered valid.

----

- Verify all tests go green


rushtong:
> Backend schema allows required array fields which have no elements, so, e.g., even though consentGroups is required, consentGroups: [] is considered valid.

We should revisit this with the next round of schema fixes.

package.json
```diff
@@ -8,6 +8,7 @@
     "@mui/material": "5.11.0",
     "@mui/icons-material": "5.11.0",
     "@mui/styles": "5.11.0",
+    "ajv": "8.12.0",
```
rushtong, line 10: Thanks for pinning an exact version here!

rushtong:
Not sure if validation is working as intended - quite a few fields are marked as invalid when they are, in fact, valid

Screenshot 2023-02-28 at 1 56 21 PM

connorlbark:
> Not sure if validation is working as intended - quite a few fields are marked as invalid when they are, in fact, valid

Now testing, this is a strange regression - I don't know what changed, it was working fine before the merge. I have a feeling I should make some major changes to how the form is being saved anyway (convert it to react state), so perhaps I should do that now... I'm gonna make this into a draft and investigate.

connorlbark:
Changes turned out to be simpler than I thought - should be much more stable now that it's proper react state.

rushtong:
Separate from code touched in this PR, the anvil component is validating strangely:
```
      h(FormField, {
        id: 'sequencingCenter',
        title: 'Sequencing Center',
        placeholder: 'Email',
        validators: [FormValidators.REQUIRED, FormValidators.EMAIL],
        onChange,
        validation: validation.sequencingCenter,
        onValidationChange,
      }),
```
I'm pretty sure that should not be an email validator.

Screenshot 2023-03-02 at 12 45 59 PM

rushtong:
On save, the data custodian email is not validating correctly:

Screenshot 2023-03-02 at 12 54 00 PM

connorlbark:
> Separate from code touched in this PR, the anvil component is validating strangely:
> 
> ```
>       h(FormField, {
>         id: 'sequencingCenter',
>         title: 'Sequencing Center',
>         placeholder: 'Email',
>         validators: [FormValidators.REQUIRED, FormValidators.EMAIL],
>         onChange,
>         validation: validation.sequencingCenter,
>         onValidationChange,
>       }),
> ```
> 
> I'm pretty sure that should not be an email validator.
> 
> Screenshot 2023-03-02 at 12 45 59 PM

Yeah, there were a few like that I noticed. Not sure how those happened. I can change this here.

connorlbark:
> On save, the data custodian email is not validating correctly:

Ok, so I think it's working properly as-is, but the multi-text component is confusing. Inputting text is not enough - you have to hit enter or hit the plus button to actually "save" it.

Maybe if the user unfocuses, it could error and say "Please save the value by pressing the plus button" or something? Could be another quick ticket.

connorlbark:
Something I was thinking about - we don't have to do it here, but it could be possible to automatically, according to the JSON Schema, use the correct validators. So, e.g., we would never have the Sequencing Center falsely have an email validator and the backend does not. Obviously out of scope here, but could be a cleanup ticket post-release.

rushtong:
More fixes that are separate from this PR (I think) ... the participant number validator needs an update:

Screenshot 2023-03-03 at 12 21 26 PM

rushtong:
Here's a case where I'm seeing the lack of a validation message. I can't be sure if this is triggering the validation error message, but it seems like it should be highlighted:

Screenshot 2023-03-03 at 12 26 08 PM

rushtong:
There is a hidden error preventing a form from being valid:

Screenshot 2023-03-03 at 12 35 29 PM

connorlbark:
> There is a hidden error preventing a form from being valid:

That is because it doesn't match the backend's enum values, since they changed. Theoretically, this would never happen once the frontend is updated, but I can add some checks to give it a more descriptive error (though it shouldn't be user facing I think)

connorlbark:
> Here's a case where I'm seeing the lack of a validation message. I can't be sure if this is triggering the validation error message, but it seems like it should be highlighted:

The schema doesn't have it as conditionally required. Apparently it can be done: https://stackoverflow.com/questions/38717933/jsonschema-attribute-conditionally-required , but it's an issue with the schema, not the frontend.

rushtong (approved):
Thanks for the changes 👍🏽

fboulnois (approved):
lgtm!

PR #2085, opened 2023-03-01
[DUOS-2324][risk=no] Handle open access in consent group

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2324

### Summary

Move open access selection to consent group and update logic

----

- Verify all tests go green

### DIFF
--- a/src/components/data_submission/DataAccessGovernance.js
+++ b/src/components/data_submission/DataAccessGovernance.js
@@ -1,27 +1,10 @@
 import ConsentGroupForm from './consent_group/ConsentGroupForm';
 import { useEffect, useState, useCallback } from 'react';
 import { isNil, every, cloneDeep } from 'lodash/fp';
-import { div, h, h2, h3, a, span } from 'react-hyperscript-helpers';
+import { div, h, h2, a, span } from 'react-hyperscript-helpers';
 import { DAC } from '../../libs/ajax';
-import { FormFieldTypes, FormField } from '../forms/forms';
 
 import './ds_common.css';
-
-const OPEN_ACCESS = 'Open Access';
-const CONTROLLED_ACCESS = 'Controlled Access';
-
-const openControlledRadioOptions = [
-  {
-    id: 'open_access',
-    name: OPEN_ACCESS,
-    text: 'Open Access'
-  },
-  {
-    id: 'controlled_access',
-    name: CONTROLLED_ACCESS,
-    text: 'Controlled Access',
-  }
-];
 
 export const DataAccessGovernance = (props) => {
   const {
@@ -30,7 +13,6 @@ export const DataAccessGovernance = (props) => {
 
   const [consentGroupsState, setConsentGroupsState] = useState([]);
   const [dacs, setDacs] = useState([]);
-  const [isControlledAccess, setIsControlledAccess] = useState(false);
 
   useEffect(() => {
     addNewConsentGroup();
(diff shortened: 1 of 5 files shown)


rushtong:
I think we're missing the data access committee selection menu in this view.

rushtong:
> I think we're missing the data access committee selection menu in this view.

Aha - I do see it now, it only shows up after selecting a data use term. I think it should always be visible.

rushtong:
The ticket mentions this consequence of selecting no restrictions:

> Add a ‘No Restrictions (Open Access Data)’ radio button option after ‘Populations, Origins, Ancestry Use’ in the 'Primary Data Use Terms' section
> If selected, disable the ‘Data Access Committee’ & ‘Secondary Data Use Terms’ fields

In this PR, those are hidden, not visible and disabled. I think this is a valid approach, but we should confirm with Jonathan if that's what was intended as opposed to visible and disabled.

rushtong:
I don't think the combination of "No Restrictions" is valid with "Other". No Restrictions should clear out all other primary and secondary selections.

Screenshot 2023-03-02 at 2 25 35 PM

fboulnois:
> I don't think the combination of "No Restrictions" is valid with "Other". No Restrictions should clear out all other primary and secondary selections.
> 
> Screenshot 2023-03-02 at 2 25 35 PM

This is currently how the form works. You can select GRU + Other or POA + Other: https://duos-k8s.dsde-dev.broadinstitute.org/data_submission_form

fboulnois:
> The ticket mentions this consequence of selecting no restrictions:
> 
> > Add a ‘No Restrictions (Open Access Data)’ radio button option after ‘Populations, Origins, Ancestry Use’ in the 'Primary Data Use Terms' section
> > If selected, disable the ‘Data Access Committee’ & ‘Secondary Data Use Terms’ fields
> 
> In this PR, those are hidden, not visible and disabled. I think this is a valid approach, but we should confirm with Jonathan if that's what was intended as opposed to visible and disabled.

How about hiding them if and only if No Restrictions is selected? This would be easier than setting the disabled state on each secondary use item.

connorlbark:
> > I don't think the combination of "No Restrictions" is valid with "Other". No Restrictions should clear out all other primary and secondary selections.
> > Screenshot 2023-03-02 at 2 25 35 PM
> 
> This is currently how the form works. You can select GRU + Other or POA + Other: https://duos-k8s.dsde-dev.broadinstitute.org/data_submission_form

I think that is a bug; my understanding is that selecting other should clear the form.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -70,6 +52,10 @@ export const DataAccessGovernance = (props) => {
     });
   }, []);
 
+  useEffect(() => {
```
connorlbark, line 56: just curious cause i've never seen something like this, what does this use effect do?

rushtong:
> How about hiding them if and only if No Restrictions is selected? This would be easier than setting the disabled state on each secondary use item.

I think that would work fine!

rushtong:
> This is currently how the form works. You can select GRU + Other or POA + Other: https://duos-k8s.dsde-dev.broadinstitute.org/data_submission_form

Thank you - great point, please ignore this feedback. If we need to change it, we'll do so as a separate ticket after reviewing it as a group.

src/components/data_submission/consent_group/EditConsentGroup.js
```diff
@@ -105,297 +109,332 @@ export const EditConsentGroup = (props) => {
...
+        }),
+      ]),
 
-      // location
+      // data access committee
+      div({
+        isRendered: consentGroup.openAccess !== true,
+      }, [
+        h(FormField, {
+          id: 'dataAccessCommitteeId',
```
connorlbark: should be `idx + '_dataAccessComitteeId'` - that way there aren't multiple fields with the same id if there are multiple consent groups.

src/components/data_submission/DataAccessGovernance.js
connorlbark: oh, is this to make a new consent group on initialize?

src/components/data_submission/consent_group/EditConsentGroup.js
```diff
@@ -105,297 +109,332 @@ export const EditConsentGroup = (props) => {
...
-      }),
+        h(FormField, {
+          type: FormFieldTypes.RADIOBUTTON,
+          id: idx + '_primaryConsent_otherPrimary',
+          name: 'primaryConsent',
+          value: 'otherPrimary',
+          toggleText: 'Other',
+          defaultValue: selectedPrimaryGroup(consentGroup),
+          onChange: ({ value }) => {
+            onPrimaryChange({ key: value, value: otherPrimaryText });
```
connorlbark, line 231: so i think the reason it's glitchy is because `otherPrimaryText` starts out as undefined; either here or where it's defined we need to make sure it's at least an empty string, and if so, then I think it should fix the glitch where it can be double selected.

src/components/data_submission/consent_group/EditConsentGroup.js
```diff
@@ -105,297 +109,332 @@ export const EditConsentGroup = (props) => {
...
+          id: idx + '_primaryConsent_openAccess',
+          name: 'primaryConsent',
+          value: 'openAccess',
+          toggleText: 'No Restrictions (Open Access Data)',
+          defaultValue: selectedPrimaryGroup(consentGroup),
+          onChange: ({ value }) => {
+            onPrimaryChange({ key: value, value: true });
+          },
+          validation: validation.openAccess,
+        }),
```
connorlbark, line 221: should we have a tooltip explaining why fields go away and why it doesn't need a DAC?

connorlbark (commented):
looks great! couple minor things on my first pass of review, but overall great

src/components/data_submission/DataAccessGovernance.js
fboulnois: Yes, that's correct.

connorlbark: Makes sense to me, cool use of useEffect. maybe a quick comment would be helpful to clarify what it does but that might just be my novice-ness to react.

src/components/data_submission/consent_group/EditConsentGroup.js
fboulnois: I'll make another ticket for this, the fix is probably straightforward but I tried a few things and wasn't able to resolve the double selection.

fboulnois: Actually I was able to resolve it 👍

src/components/data_submission/consent_group/EditConsentGroup.js
fboulnois: I'll create a separate ticket for this, it might need a little UX work and I'm having trouble making the tooltip inline.

fboulnois: I created https://broadworkbench.atlassian.net/browse/DUOS-2372 for this.

connorlbark (approved):
thank you for the changes! looks great.

kaitlynoflaherty (approved):
Looks great!

PR #2088, opened 2023-03-07
[DUOS-2367][risk=no] Use external validation in DAR form

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2367

As a side effect, fixes: https://broadworkbench.atlassian.net/browse/DUOS-2360

This PR is good if:
- Validation works as expected for all fields - no change from develop functionality
- Clicking submit correctly highlights all invalid fields - new functionality, replaces old validation messages
- Clicking submit correctly scrolls the UI to the section which includes the first invalid field (e.g., if only project title invalid, scroll to section 2) - no change from develop functionality

----

- Verify all tests go green


cypress/component/DataAccessRequest/data_access_request_validation.spec.js
```diff
@@ -0,0 +1,225 @@
...
+    cy.get('#pediatric').should('have.class', 'errored');
+    cy.get('#vulnerablePopulation').should('have.class', 'errored');
+    cy.get('#illegalBehavior').should('have.class', 'errored');
+    cy.get('#sexualDiseases').should('have.class', 'errored');
+    cy.get('#psychiatricTraits').should('have.class', 'errored');
+    cy.get('#notHealth').should('have.class', 'errored');
+    cy.get('#stigmatizedDiseases').should('have.class', 'errored');
+  });
+
+  it('Internal / external / lab collaborators error properly', () => {
```
connorlbark, line 179: one note: this test was wildly inconsistent (50/50 shot of passing), and I couldn't figure out why; it was failing upon calling a stubbed endpoint. then I changed something (still, am not sure what exactly) and it didn't fail after running it ~20 times. this is a potentially flaky test

rushtong:
I'm not sure what's causing it, but there is something that triggers re-rendering many times after trying to submit an invalid form that jumps to the PI field. I'm seeing lots and lots of re-renders that looks up the disease terms, so it seems like the dataset widget might be getting called too often.

Screenshot 2023-03-09 at 6 50 26 AM


Screenshot 2023-03-09 at 6 52 29 AM

rushtong:
Scrolling through a page with validation errors essentially re-calls every API call the page needs infinitely.

Screenshot 2023-03-09 at 6 55 29 AM

src/pages/DataAccessRequestApplication_new.js
```diff
@@ -190,6 +192,47 @@ const DataAccessRequestApplicationNew = (props) => {
     });
   }, [datasets]);
 
+  const resetForcedScrollDebounce = useCallback(() => {
```
rushtong: Once validation is triggered, this method is called every time the user scrolls the window. As a consequence of that, the `init` callback is also called which triggers infinite reloading when the user pages up and down the form.

connorlbark: good catch - i will investigate this, thank you

fboulnois (approved):
lgtm, contingent on @rushtong 's suggestion

src/pages/DataAccessRequestApplication_new.js
connorlbark: this should be fixed! i am not sure why this wasn't happening before, to be honest. but i separated this out into two different use effects and confirmed that the expected behavior (init once, remove onScroll once) was happening.

rushtong: I'm still seeing a similar behavior after a fresh pull and build. After trying to submit an incomplete form, the validation is triggered. Scrolling anywhere on the page triggers all of the API calls that populate the page content. Here's a screen-cap of the requests made after scrolling the page once. Let's follow up in person so I can show you a repeatable case.

Screenshot 2023-03-15 at 6 22 25 AM

connorlbark: I figured it out! No idea why my changes made this worse, but basically because: `init` depends on `onScroll` which depends on `forcedScroll` - which controls debounce on forcing scroll, to ensure scroll doesn't occur too quickly. since submitting scrolls to one of the errors, it triggers a debounce update / sets debounce timer. every time the debounce timer function is updated (see: often), then it would trigger onScroll update... then init update... downloading everything again. removing the debounce mechanism seems to make the scroll more stable and does not trigger init update anymore.

rushtong (approved):
Definitely working better, thanks for the changes! 👍🏽

PR #2122, opened 2023-03-28
[DUOS-2402][risk=no] Cleanup Consent Group Validation

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2402

Consent groups had a strange way of validation with top-level messages. This removes these top level messages and replaces them with a much cleaner usage of external validation state, which will highlight fields if they are invalid on save, mirroring the rest of the form.

----

- Verify all tests go green

### DIFF
--- a/src/components/forms/forms.js
+++ b/src/components/forms/forms.js
@@ -302,16 +302,15 @@ export const FormField = (config) => {
 */
 export const FormTable = (config) => {
   const {
-    id, name, formFields,
+    id, formFields,
     enableAddingRow, addRowLabel,
     disabled, onChange, minLength,
-    validation, defaultValue
+    validation, onValidationChange, defaultValue
   } = config;
 
   const [formValue, setFormValue] = useState(defaultValue || [{}]);
 
-
-  const key = name || id;
+  const key = getKey(config);
 
   return div({ id, className: `formField-table formField-${id}` }, [
     // generate columns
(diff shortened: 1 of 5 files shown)


src/components/forms/forms.js
```diff
@@ -302,16 +302,15 @@ export const FormField = (config) => {
 */
 export const FormTable = (config) => {
```
connorlbark, line 303: there are some weird bugs in formtable that doing this ticket made me notice.

the only one i was fully unable to fix can be recreated by:

1. create >1 file types
2. fill out the first one with a value for number of participants, but not the others
3. delete the first one

curiously... the value stays, but if you try to submit, it will error out as if it wasn't present. i have realized this is because:
- the `defaultValue` of the original field went from `undefined` -> `something`, so the value being passed down "took control" of the form field and replaced the internal state
- when the 0th field gets delete called on it, the value being passed down does from `something` -> `undefined`
- since there isn't a key here, the same component gets reused - and since that component's default value is now `undefined`, it looks to its internal state variable - which is `something`, the old value from the deleted row.
- until there is a change which triggers the external `defaultValue` to get become not undefined... the value is the same as the old value, since the form component isn't being controlled by `defaultValue`.

The only solution I could think of was to implement unique keys for each form value in the list, which would be consistent across deletions and creating new rows... but this turned out to be really tricky.

Anyway it's a pretty minor bug that the user can bypass pretty intuitively by just changing the value. Just a strange consequence of how we keep values in form components.

rushtong (approved):
👍🏽

kaitlynoflaherty (approved):

PR #2123, opened 2023-03-28
[DUOS-2377][risk=no] Change hardcoded DS form JSON schema version

connorlbark:
Fixes fallout from - 
https://broadworkbench.atlassian.net/browse/DUOS-2377

The backend schema version changed, which broke frontend.

----

- Verify all tests go green

### DIFF
--- a/src/utils/JsonSchemaUtils.js
+++ b/src/utils/JsonSchemaUtils.js
@@ -1,4 +1,4 @@
-import Ajv2019 from 'ajv/dist/2019';
+import {default as Ajv} from 'ajv/dist/2019';
 import {
   urlValidator,
   dateValidator,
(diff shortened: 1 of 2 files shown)


src/utils/JsonSchemaUtils.js
```diff
@@ -1,4 +1,4 @@
-import Ajv2020 from 'ajv/dist/2020';
+import Ajv2019 from 'ajv/dist/2019';
```
fboulnois:
```suggestion
import Ajv2019 as Ajv from 'ajv/dist/2019';
```

Could we do this instead to make it easy to switch out?

fboulnois (approved):
One suggestion:

src/utils/JsonSchemaUtils.js
connorlbark: oh true - I did `Ajv = Ajv2019` down below but this is probably easier

kaitlynoflaherty (approved):

rushtong (approved):
👍🏽

PR #2134, opened 2023-04-03
[DUOS-2415][risk=no] Precompile DS Form JSONSchema

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2415

----

- Verify all tests go green


src/utils/schemas/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+For the Dataset Registration schema, the call looks like:
+
+```
+ajv compile -s consent/src/main/resources/dataset-registration-schema_v1.json -o duos-ui/src/utils/schemas/DatasetRegistrationSchemaValidation.js --spec draft2019 -c ajv-formats --strict false --all-errors
+``` 
+
+`-s` refers to the input file. `-o` refers to the output file. `--spec` specifies the specification version. `-c` adds formats such as "date", "uri", etc. `--strict false` is needed as our schema has some unknown metadata fields. `--all-errors` is needed because, by default, the validate function only returns one error at a time.
+
+Once generated, the file needs to have the `/* eslint-disable */` flag on the first line, or it will create many linting errors.
```
fboulnois: I am wondering if we should create a tiny `sh` script to do this instead.

fboulnois (approved):
lgtm, a small suggestion:

src/utils/schemas/README.md
rushtong: I agree with an alternate approach here. I'm also thinking that the home for any generated validation code is in `src/assets` as opposed to here in `src/utils/schemas`.

rushtong (commented):
Looks good aside from schema generation 👍🏽

src/utils/schemas/README.md
connorlbark: When should this be generated? Let's say you're running it locally - would it generate it on `npm start`? there would need to be code there for vscode to consider that function as existing - so would there be a 'placeholder' old version present that gets overridden?

rushtong: Great question. This is a new pattern for us, so we don't have a precedent. This sounds like a pre-deploy+build but maybe a post-dependency kind of thing? NPM docs (https://docs.npmjs.com/cli/v8/commands/npm-install) show a `postinstall` phase that might be helpful here.

connorlbark: ~What are your thoughts on `postinstall` to initially generate and also generating on `poststart` so it keeps relatively synced? And it grabs it from the backend configured in `public/config.json`? or just always from dev?~

Edit: Modeling after the code-generation process in one of my classes, I made the files a) tracked by github (so we're confident everyone has the same copy, and the dev is alerted when it changes) and b) regenerated on `prestart` according to the backend that the user has selected - so it is up-to-date every time a user runs their code.

One question I have - I am not exactly sure how production build processes work. Does it use `npm start` / would it trigger `pre-start`? I assume it would do a build and then just serve those files, so npm start would never be ran. This might be for the best because it won't stray from what is on git. But I am not sure.

scripts/compile-jsonschema.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
fboulnois, line 1: Run `shellcheck` on this script after installing it: `brew install shellcheck`

scripts/compile-jsonschema.sh
```diff
@@ -0,0 +1,12 @@
...
+APIURL=https://consent.dsde-dev.broadinstitute.org/
+if test -f "./public/config.json"; then
+  APIURL=$(jq '.apiUrl' ./public/config.json -r)
```
fboulnois: Check that both `jq` and `ajv` are installed. Here's an example for `jq`:

```sh
# ensure that jq is installed
check_jq_installed() {
    if ! jq --version 1>/dev/null 2>&1; then
        abort "jq v1.6 or above is required; install jq to continue"
    fi
}
```

scripts/compile-jsonschema.sh
```diff
@@ -0,0 +1,12 @@
...
+fi
+
+
+curl -s -S -X 'GET'  "$APIURL/schemas/dataset-registration/v1"  -H 'accept: application/json' -o ./src/assets/schemas/DataRegistrationV1.json
+
+ajv compile -s ./src/assets/schemas/DataRegistrationV1.json -o ./src/assets/schemas/DataRegistrationV1Validation.js --spec draft2019 -c ajv-formats --strict false --all-errors
+
+echo "/* eslint-disable */ $(cat ./src/assets/schemas/DataRegistrationV1Validation.js)" > ./src/assets/schemas/DataRegistrationV1Validation.js
```
fboulnois: Wrap this in a function and call it to make the script a little cleaner.

scripts/compile-jsonschema.sh
fboulnois, line 1: For stricter error checking set these in your script:

```sh
set -eu
set -o pipefail
```

fboulnois, line 1: Add a description of what the script does underneath the shebang with usage. Example:

```sh
#
# Deletes old Google App Engine deployments of Terra UI in an environment. You
# MUST have jq installed to be able to use this script.
#
# USAGE: ./delete-old-app-engine-versions.sh ENV
#   ENV must be one of dev, alpha, or staging
#
```

fboulnois (changes requested):
Suggestions for improving the script:

src/utils/schemas/README.md
connorlbark: I ended up removing it from pre-start - there are so many contexts in which `start` is ran that I felt like trying to track down all of them was difficult. In particular, integration tests were failing and I wasn't sure why. Possibly, we could target the build phase - but I feel like it might be more stable to just have documentation that schemas need to be regenerated + an easy npm script to do so.

rushtong: > One question I have - I am not exactly sure how production build processes work. Does it use `npm start` / would it trigger `pre-start`? I assume it would do a build and then just serve those files, so npm start would never be ran. This might be for the best because it won't stray from what is on git. But I am not sure.

All deployed versions are built using `Dockerfile`. There, you'll see there that we have a build phase and the copy the compiled code over to `/usr/share/nginx/html` and mount that in the base nginx image. That compiled image is pushed to GCR where it is referenced by the k8s pods with env-specific `config.json` files. So, there are two ways that we generally start the UI, locally (`npm start`) and via the compiled image. The piece that I've been concerned about is having the code in `assets/schemas` checked in via this PR since it is not tied to this repo. Ideally, that code should be generated during a build step.

scripts/compile-jsonschema.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+# Download the data submission schema
+curl -s -S -X 'GET'  "$APIURL/schemas/dataset-registration/v1"  -H 'accept: application/json' -o ./src/assets/schemas/DataRegistrationV1.json
+
+# Generate the data submission schema validation code
+ajv compile -s ./src/assets/schemas/DataRegistrationV1.json -o ./src/assets/schemas/DataRegistrationV1Validation.js --spec draft2019 -c ajv-formats --strict false --all-errors
+
+# Prevent linting errors (minified file)
+prepend_text "/* eslint-disable */" ./src/assets/schemas/DataRegistrationV1Validation.js
```
fboulnois: It's probably worth wrapping this in a function and calling it.

fboulnois (approved):
One small suggestion, otherwise lgtm:

scripts/compile-jsonschema.sh
connorlbark: I think i am confused on what you mean, I don't do a lot of bash scripting. What should be wrapped in a function? I added the `prepend_text` function because I thought that was what you meant before, but I think I misunderstood.

src/utils/schemas/README.md
connorlbark: I guess I feel like there are two solutions:

1. have the code checked out in the repo & easily manually updatable (this PR)
2. update the code as part of build

I feel like updating the code as part of build is a) unstable because it relies on hitting the backend which could fail - if the process just used local files then I feel like it would be stable - and b) hard to debug if it goes wrong. Keeping it a part of the repo and easily updatable via a script ensures everyone has the same code & doesn't add any complexity to the build process that could go wrong.

scripts/compile-jsonschema.sh
fboulnois: Something like:

```sh
prepare_schema() {
  # Download the data submission schema
  curl -s -S -X 'GET'  "$APIURL/schemas/dataset-registration/v1"  -H 'accept: application/json' -o ./src/assets/schemas/DataRegistrationV1.json

  # Generate the data submission schema validation code
  ajv compile -s ./src/assets/schemas/DataRegistrationV1.json -o ./src/assets/schemas/DataRegistrationV1Validation.js --spec draft2019 -c ajv-formats --strict false --all-errors

  # Prevent linting errors (minified file)
  prepend_text "/* eslint-disable */" ./src/assets/schemas/DataRegistrationV1Validation.js
}

prepare_schema
```

rushtong (approved):
👍🏽 I'm good with this approach. If we discover any issues, we can address them then.

PR #2148, opened 2023-04-10
[DUOS-2352][risk=no] Release new DAR form

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2352

Step 1 of a 2 step process for releasing the new DAR form. This PR sets up the new form as the destination for the `dar_application` path for non-prod, non-staging envs. Also maintains the `_new` paths pointing to the new component and adds `_old` paths that all point to the old component. This allows for manual testing on staging/prod using the `_new` urls as we've been doing in dev. 

Step 2 will be to remove the env checks so the new component serves the default url.

Possible step 3 will be to clean up all the old paths and remove unused components/code.

### Testing
You can test by setting up your local config with different envs. I suggest just changing the `env` var which is what the router uses to determine which component to direct the various urls to.

----

- Verify all tests go green


fboulnois (approved):
👍

kaitlynoflaherty (approved):
👍

PR #2176, opened 2023-04-28
[DUOS-2455][risk=no] Enable calls to add/remove data submitter role

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2455

* Enables the issue/remove links for data submitters on the SO Console Data Submitters page
* Moves SO pages around so they are centrally located and more easily found.

----

- Verify all tests go green


fboulnois (approved):
👍 , I think at some point we should switch away from relative paths, but that is completely separate PR.

src/pages/signing_official_console/DataCustodianTable.js
```diff
@@ -90,12 +92,12 @@ const IssueDataCustodianButton = (props) => {
...
 const researcherFilterFunction = getSearchFilterFunctions().signingOfficialResearchers;
 
-const CustodianCell = ({
+const SubmitterCell = ({
   researcher,
   showConfirmationModal
 }) => {
   const id = researcher.userId || researcher.email;
-  const button = researcher.isCustodian
+  const button = researcher.isDataSubmitter
```
connorlbark: nit: would it be better to use a function, `isDataSubmitter(user)` here? I found the `setDataSubmitterProp` a bit unintuitive at first until I fully understood how it worked/interplayed with the rest of the system. i could also see someone in the future forgetting to call it, and not understanding why this button isn't showing up. totally understand if you'd rather just get it through, though.

connorlbark (approved):
This part looks good - although, when we issue the data submitter role, the text on the right column does not update with the added `DataSubmitter` until there is a refresh. Maybe this should be a separate ticket?

src/pages/signing_official_console/DataCustodianTable.js
rushtong: Yeah ... not proud of this bit of code. I was shooting for a quicker fix, but this is definitely a code smell. Will refactor that to be more consistent with how we're looking at other user roles.

fboulnois (approved):
🎉

PR #2232, opened 2023-06-08
[DUOS-2522][risk=no] Split out initializations into separate useEffects

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2522

### Summary
I'm not sure why this happened, but for some users, not all initialization variables were being stored in local storage. Specifically, the `env` var was not stored for some users, but was stored fine for other users. I theorize that having all initializations in the same useEffect may have caused this. After splitting them out into separate use effects, I was able to get past this error and the problematic user was no longer problematic after the update.

### Testing
I could only reproduce this with one user who had an SO and DO role. If you can reproduce this, look at local storage after logging in and not navigating anywhere. If you can replicate this, then you'll see that there is a missing `env` var in session storage. After this PR, I was able to consistently populate `env`

----

- Verify all tests go green

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -6,7 +6,7 @@ import './App.css';
 import {Config} from './libs/config';
 import DuosFooter from './components/DuosFooter';
 import DuosHeader from './components/DuosHeader';
-import {useHistory} from 'react-router-dom';
+import {useHistory, useLocation} from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
lgtm

kaitlynoflaherty (approved):
👍

PR #2240, opened 2023-06-16
[DUOS-2187][risk=no] Add form to edit datasets

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2187

### Summary

Add form to edit datasets

Screenshot 2023-06-16 at 14-00-55 Broad Data Use Oversight System

----

- Verify all tests go green

### DIFF
--- a/src/components/data_update/DatasetUpdate.js
+++ b/src/components/data_update/DatasetUpdate.js
@@ -1,0 +1,394 @@
+import { useCallback, useState, useEffect } from 'react';
+import { button, h, h2, div } from 'react-hyperscript-helpers';
+import { find, isArray, isNil } from 'lodash/fp';
+
+import { FormFieldTypes, FormField, FormValidators } from '../forms/forms';
+import { DAC, DAR, DataSet } from '../../libs/ajax';
+import { Notifications } from '../../libs/utils';
+
+export const DatasetUpdate = (props) => {
+  const { dataset } = props;
+
+  const [formData, setFormData] = useState({ dac: {}, dataUse: {}, properties: {} });
+
+  const searchOntologies = (query, callback) => {
+    let options = [];
+    DAR.getAutoCompleteOT(query).then(
+      items => {
+        options = items.map(function (item) {
+          return item.label;
+        });
+        callback(options);
+      });
+  };
+
+  const getDiseaseLabels = async (ontologyIds) => {
+    let labels = [];
+    if (!isNil(ontologyIds)) {
+      const idList = ontologyIds.join(',');
+      const result = await DAR.searchOntologyIdList(idList);
+      labels = result.map(r => r.label);
+    }
+    return labels;
+  };
+
+  const extract = useCallback((propertyName) => {
+    const property = find({ propertyName })(dataset.properties);
+    return property?.propertyValue;
+  }, [dataset]);
+
+  const asProperty = (propertyName, propertyValue) => {
+    return {
+      propertyName,
+      propertyValue
+    };
+  };
+
+  const normalizeDataUse = useCallback(async (dataUse) => {
+    let du = dataUse;
+    if (!isNil(dataUse.diseaseRestrictions)) {
+      du.hasDiseaseRestrictions = true;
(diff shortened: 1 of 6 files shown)


src/components/data_update/DatasetUpdate.js
```diff
@@ -0,0 +1,392 @@
...
+
+    const formElement = event.target.form;
+    const submitData = new FormData(formElement);
+    const nihFileData = submitData.get('nihInstitutionalCertificationFile');
+    const consentGroups = [{ nihInstitutionalCertificationFile: nihFileData }];
+
+    const newDataset = {
+      name: dataset.name,
+      dacId: formData.dac.dacId,
+      properties: [
```
fboulnois, line 84: Although the OpenAPI schema for DatasetUpdate says it requires a `datasetProperties` object, the `DatasetUpdate` object in Consent actually requires a `properties` object instead.

src/components/data_update/DatasetUpdate.js
```diff
@@ -0,0 +1,392 @@
...
+        asProperty('Data Depositor', formData.properties.dataDepositor),
+        asProperty('Principal Investigator(PI)', formData.properties.principalInvestigator),
+      ]
+    };
+
+    const multiPartFormData = new FormData();
+    multiPartFormData.append('dataset', JSON.stringify(newDataset));
+    multiPartFormData.append('consentGroups', consentGroups);
+
+    DataSet.updateDatasetV3(dataset.dataSetId, multiPartFormData).catch(() => {
```
fboulnois: Additionally, although the response returns 200, the underlying dataset does not actually get updated. I think there's probably an issue with the `updateDataset` endpoint.

kaitlynoflaherty (approved):
The form looks great! When testing locally, everything works as expected besides the submit button. It seems like some required fields have been excluded from the endpoint, so I agree that the API should be updated.

src/components/data_update/DatasetUpdate.js
rushtong: Nice catch, we should file a bug fix for that.

src/components/data_update/DatasetUpdate.js
rushtong: I'm seeing some updates when using the submit button. For example, the dataset name updates, but the species doesn't seem to update. I think the properties are not saving correctly. What I think is missing is some UI indication of success.

rushtong (commented):
The form looks good, but I think we need an indicator of success somewhere on the page. We use the noty library now for that kind of thing.

The back end does seem to not be saving the properties correctly. This did work in testing so we need to sort out what's happening there.

src/components/data_update/DatasetUpdate.js
```diff
@@ -0,0 +1,392 @@
...
+    event.preventDefault();
+    event.stopPropagation();
+
+    const formElement = event.target.form;
+    const submitData = new FormData(formElement);
+    const nihFileData = submitData.get('nihInstitutionalCertificationFile');
+    const consentGroups = [{ nihInstitutionalCertificationFile: nihFileData }];
+
+    const newDataset = {
+      name: dataset.name,
```
rushtong: I think this should align with the same value in `formData.properties.datasetName`. As it stands now, there are two different names being sent over in the payload. This is a bit of a holdover from previous technical debt where the name was not a top level field, but just a prop.

fboulnois: Thanks, good to know, I wasn't sure which of these I should use.

src/components/data_update/DatasetUpdate.js
fboulnois: https://broadworkbench.atlassian.net/browse/DUOS-2579

src/components/data_update/DatasetUpdate.js
fboulnois: Just like the submission form, I added a success notification and a redirect to the catalog page.

rushtong (approved):
👍🏽

PR #2260, opened 2023-06-27
[DUOS-2456][risk=no] Do not normalize each data use group

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2456

### Summary

This is one potential way to avoid fractional votes.

For example, `DAR-632` has a Data Use Group with 2 datasets. One dataset has 10 DAC members and the other dataset has 5 DAC members. As a result, 15/2 = 7.5, and this ends up with fractional votes.

Avoiding the normalization on the dataset size fixes this issue, but I am not sure it makes sense to display the numbers this way.

----

- Verify all tests go green


rushtong (approved):
👍🏽

kaitlynoflaherty (approved):
👍

PR #2261, opened 2023-06-28
[DUOS-2527] Create an Edit Study page

kaitlynoflaherty:
### Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2527

### Summary
Note:
1. Some of the data is corrupted in the previous studies. It is best to use a new study with all the proper info when testing

----

- Verify all tests go green


src/libs/ajax.js
```diff
@@ -393,7 +393,18 @@ export const DataSet = {
...
+
+  getStudyById: async studyId => {
+    const url = `${await getApiUrl()}/api/dataset/study/${studyId}`;
+    const res = await fetchOk(url, Config.authOpts());
+    return await res.json();
+  },
+
+  updateStudy: async (studyId, studyObject) => {
+    const url = `${await getApiUrl()}/api/dataset/study/${studyId}`;
+    return await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(studyObject), {method: 'PUT'}]));
```
rushtong: Suggestion - we should try to re-use the register pattern. This should work:
```
await axios.put(url, Config.jsonBody(studyObject), Config.multiPartOpts());
```

src/pages/StudyUpdateForm.js
```diff
@@ -0,0 +1,258 @@
...
+        const me = await User.getMe();
+        setUser(me);
+        setStudy(await DataSet.getStudyById(studyId));
+        setFailedInit(false);
+      } catch (error) {
+        Notifications.showError({ text: 'Failed to load study' });
+      }
+    };
+    init();
+  }, [studyId]);
```
rushtong: You can get the user from local storage instead of an API call to streamline things. 
Also, I'm not sure if this is populating the data submitter values, but the page is showing me my own info, which is not correct in this case. We might need to pull back more study data if this isn't there already.

Screenshot 2023-08-01 at 11 53 31 AM

src/components/data_submission/consent_group/ConsentGroupForm.js
```diff
@@ -12,38 +12,46 @@ export const ConsentGroupForm = (props) => {
...
-    otherSecondary: null, // string
+    nmds: consentGroupsState[idx].consentGroup.nmds || false, // No Methods Development or validation studies
+    gso: consentGroupsState[idx].consentGroup.gso || false, // genetic studies only
+    pub: consentGroupsState[idx].consentGroup.pub || false, // publication required
+    col: consentGroupsState[idx].consentGroup.col || false, // collaboration required
+    irb: consentGroupsState[idx].consentGroup.irb || false, // irb approval required
+    gs: consentGroupsState[idx].consentGroup.gs || null, // string: geographic restriction
+    mor: consentGroupsState[idx].consentGroup.mor || undefined, // date (string): publication moratorium
+    npu: consentGroupsState[idx].consentGroup.npu || false, // non profit only
+    otherSecondary: consentGroupsState[idx].consentGroup.otherSecondary || null, // string
```
rushtong: The checkboxes are still hover-able even if not changeable. We should try to disable the pointer icon in this case. This might be a simple fix with the checkbox form component so might not need to be a part of this PR.

Screenshot 2023-08-01 at 11 56 48 AM

src/components/data_submission/consent_group/ConsentGroupForm.js
```diff
@@ -12,38 +12,46 @@ export const ConsentGroupForm = (props) => {
...
     // "AnVIL Workspace", "Terra Workspace",
     // "TDR Location", "Not Determined"
-    dataLocation: null,
+    dataLocation: consentGroupsState[idx].consentGroup.dataLocation || null,
 
-    url: '',
-    numberOfParticipants: undefined, // numeric
-    fileTypes: [{}],
+    url: consentGroupsState[idx].consentGroup.url || '',
+    numberOfParticipants: consentGroupsState[idx].consentGroup.numberOfParticipants || undefined, // numeric
```
rushtong: Study ID 8 on dev has participants, but it isn't displayed:
```
        {
          "propertyId": 8992,
          "dataSetId": 1426,
          "propertyName": "File Types",
          "propertyValue": [
            {
              "numberOfParticipants": 3
            }
          ],
          "schemaProperty": "fileTypes",
          "propertyType": "Json"
        }
```
Screenshot 2023-08-01 at 12 03 54 PM

rushtong (commented):
Just a few bits of feedback in the interim. Still reviewing!

src/pages/StudyUpdateForm.js
kaitlynoflaherty: Easy fix! I'll correct this now

src/components/data_submission/consent_group/ConsentGroupForm.js
kaitlynoflaherty: The data here is weird because in this case the number of participants is in the fileTypes properties where in a normal study the number of participants would be it's own property, which is how I extracted it. The property value for fileTypes in this case is corrupt I believe

rushtong: Ah, Ok, got it.

rushtong:
We need to improve the efficiency here. I see 6 calls to get all dacs, and similarly 6 calls to get dac id 4

Screenshot 2023-08-01 at 2 55 40 PM

src/components/data_submission/consent_group/ConsentGroupForm.js
```diff
@@ -12,38 +12,46 @@ export const ConsentGroupForm = (props) => {
...
     deleteConsentGroup,
     disableDelete,
+    consentGroupsState,
+    studyEditMode,
+    dacs,
   } = props;
 
   const [consentGroup, setConsentGroup] = useState({
-    consentGroupName: '',
+    // this will need to change to studyEditMode ? <consentGroupsState property> : <default>
```
fboulnois:
```suggestion
```

nit: remove these comments

src/components/data_submission/consent_group/ConsentGroupForm.js
```diff
@@ -12,38 +12,46 @@ export const ConsentGroupForm = (props) => {
...
 
     // primary: one of these needs to be filled
-    generalResearchUse: undefined,
-    hmb: undefined,
-    diseaseSpecificUse: undefined, // string
-    poa: undefined,
-    openAccess: undefined,
-    otherPrimary: undefined, // string
+    // if in edit study mode, give each of these the value of the consent group state at the index
+    // otherwise give them the default value
```
fboulnois:
```suggestion
```

nit: same here

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -1,34 +1,88 @@
...
+      labels = result.map(r => r.label);
+    }
+    return labels;
+  };
+
+  const normalizeDataUse = useCallback(async (dataUse) => {
+    let du = dataUse;
+    if (!isNil(dataUse.diseaseRestrictions)) {
+      du.hasDiseaseRestrictions = true;
+      try {
```
fboulnois, line 39: I would add a comment somewhere here that the reason we `try... catch` and do nothing is that sometimes the disease restrictions don't cleanly map to disease labels.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -1,34 +1,88 @@
...
+
+    property?.propertyValue.forEach((propValue)=> {
+      fileTypesName === 'fileType' ?
+        fileTypesArr.push(propValue?.fileType)
+        : fileTypesArr.push(propValue?.functionalEquivalence);
+    });
+    return fileTypesArr;
+  }, []);
+
+  // method to extract consent group data from datasets
```
fboulnois:
```suggestion
```

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
+        disableFields: false,
       });
 
       return newConsentGroupsState;
     });
   }, []);
 
+  // extract consent groups from datasets
+  const prefillConsentGroups = useCallback(async () => {
+    // from datasets array to consent group array (needs to match the existing consent group structure)
```
fboulnois:
```suggestion
```

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
+    var consentGroups = await Promise.all(datasets?.map(async (dataset, idx) => {
+      const dataUse = await normalizeDataUse(dataset?.dataUse);
+      // DAC is required if openAcess is false
+      const openAccess = extract('Open Access', dataset);
+      const dac = openAccess ? undefined : await DAC.get(dataset?.dacId);
+      const dacs = openAccess ? undefined : await DAC.list();
+
+      // extract the consent group data from the dataset
```
fboulnois:
```suggestion
```

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
       return newConsentGroupsState;
     });
   }, []);
 
+  // extract consent groups from datasets
+  const prefillConsentGroups = useCallback(async () => {
+    // from datasets array to consent group array (needs to match the existing consent group structure)
+    var consentGroups = await Promise.all(datasets?.map(async (dataset, idx) => {
+      const dataUse = await normalizeDataUse(dataset?.dataUse);
+      // DAC is required if openAcess is false
```
fboulnois:
```suggestion
      // DAC is required if openAccess is false
```

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
+      setConsentGroupsState(consentGroups);
+    }
+    return consentGroupsState;
+  }, [consentGroupsState, datasets, normalizeDataUse, extract, extractFileTypes]);
+
   // pre-populate the page with a consent group
   useEffect(() => {
-    addNewConsentGroup();
-  }, [addNewConsentGroup]);
+    // when in edit study mode, pre-populate the page with the study's consent groups otherwise add new
```
fboulnois:
```suggestion
```

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
+        disableFields: true,
+      };
+    }));
+    if (isEmpty(consentGroupsState)) {
+      setConsentGroupsState(consentGroups);
+    }
+    return consentGroupsState;
+  }, [consentGroupsState, datasets, normalizeDataUse, extract, extractFileTypes]);
+
   // pre-populate the page with a consent group
```
fboulnois:
```suggestion
  // when in edit study mode, pre-populate the page with the study's consent groups otherwise add new
```

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
   const addNewConsentGroup = useCallback(() => {
     setConsentGroupsState((consentGroupsState) => {
       const newConsentGroupsState = cloneDeep(consentGroupsState);
-      newConsentGroupsState.push({
+      newConsentGroupsState?.push({
```
fboulnois:
```suggestion
      newConsentGroupsState.push({
```

Since on L176 you check if consentGroupState is empty, this should never be null or undefined.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -1,34 +1,88 @@
...
 
-    const groups = filteredConsentGroupsState.map(state => state.consentGroup);
-    const valid = every(filteredConsentGroupsState.map(state => (!state.editMode) && state.valid));
+    const groups = filteredConsentGroupsState?.map(state => state.consentGroup);
+    const valid = every(filteredConsentGroupsState?.map(state => (!state.editMode) && state.valid));
 
     onChange({ key: 'consentGroups', value: groups, isValid: valid });
 
-    filteredConsentGroupsState.forEach((cgState, idx) => {
+    filteredConsentGroupsState?.forEach((cgState, idx) => {
```
fboulnois:
```suggestion
    filteredConsentGroupsState.forEach((cgState, idx) => {
```

Since on L176 you check if consentGroupState is empty, this should never be null or undefined.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -1,34 +1,88 @@
...
   }, [setDacs]);
 
   useEffect(() => {
-    const filteredConsentGroupsState = consentGroupsState.filter(state => !isNil(state));
+    const filteredConsentGroupsState = consentGroupsState?.filter(state => !isNil(state));
 
-    const groups = filteredConsentGroupsState.map(state => state.consentGroup);
-    const valid = every(filteredConsentGroupsState.map(state => (!state.editMode) && state.valid));
+    const groups = filteredConsentGroupsState?.map(state => state.consentGroup);
+    const valid = every(filteredConsentGroupsState?.map(state => (!state.editMode) && state.valid));
```
fboulnois:
```suggestion
    const groups = filteredConsentGroupsState.map(state => state.consentGroup);
    const valid = every(filteredConsentGroupsState.map(state => (!state.editMode) && state.valid));
```

Since on L176 you check if consentGroupState is empty, this should never be null or undefined.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -1,34 +1,88 @@
...
+
   useEffect(() => {
     DAC.list(false).then((dacList) => {
       setDacs(dacList);
     });
   }, [setDacs]);
 
   useEffect(() => {
-    const filteredConsentGroupsState = consentGroupsState.filter(state => !isNil(state));
+    const filteredConsentGroupsState = consentGroupsState?.filter(state => !isNil(state));
```
fboulnois:
```suggestion
    const filteredConsentGroupsState = consentGroupsState.filter(state => !isNil(state));
```

Since on L176 you check if consentGroupState is empty, this should never be null or undefined.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -144,7 +270,7 @@ export const DataAccessGovernance = (props) => {
           a({
             id: 'btn_addConsentGroup',
             className: 'btn-primary btn-add common-background',
-            onClick: () => addNewConsentGroup(),
+            onClick: () => { addNewConsentGroup(); }
```
fboulnois:
```suggestion
            onClick: () => addNewConsentGroup(),
```

src/components/data_submission/NIHAdministrativeInformation.js
```diff
@@ -22,13 +23,16 @@ export const NIHAdministrativeInformation = (props) => {
...
       id: id,
     };
   };
 
   return div({
-    isRendered: [YES_NHGRI_YES_PHS_ID, YES_NHGRI_NO_PHS_ID, NO_NHGRI_YES_ANVIL].includes(formData.nihAnvilUse),
+    isRendered:
+      (studyEditMode ?
+        ['yes_nhgri_yes_phs_id', 'yes_nhgri_no_phs_id', 'no_nhgri_yes_anvil'].includes(formData.properties.nihAnvilUse)
+        : [YES_NHGRI_YES_PHS_ID, YES_NHGRI_NO_PHS_ID, NO_NHGRI_YES_ANVIL].includes(formData.nihAnvilUse)),
```
fboulnois:
```suggestion
        ['yes_nhgri_yes_phs_id', 'yes_nhgri_no_phs_id', 'no_nhgri_yes_anvil'].includes(formData.properties.nihAnvilUse)
        : [YES_NHGRI_YES_PHS_ID, YES_NHGRI_NO_PHS_ID, NO_NHGRI_YES_ANVIL].includes(formData.nihAnvilUse)),
```

I believe there's four possible states for the `nihAnvilUse`, is this expected?

src/components/data_submission/NIHAdministrativeInformation.js
```diff
@@ -53,7 +60,7 @@ export const NIHAdministrativeInformation = (props) => {
       title: 'NIH Grant or Contract Number',
       validators: [FormValidators.REQUIRED],
       onChange,
-      defaultValue: formData?.nihGrantContractNumber,
+      defaultValue: studyEditMode ? formData?.properties.nihGrantContractNumber : formData?.nihGrantContractNumber,
```
fboulnois: Can formData ever be null or undefined? If not, the `?` should be removed for this one and all the other properties.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -96,7 +220,7 @@ export const DataAccessGovernance = (props) => {
   }, []);
 
   useEffect(() => {
-    setAllConsentGroupsSaved(consentGroupsState.every((state) => state.editMode === false));
+    setAllConsentGroupsSaved(consentGroupsState?.every((state) => state.editMode === false));
```
fboulnois: I don't think this should ever be null or undefined.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
+              functionalEquivalence: extractFileTypes('File Types', 'functionalEquivalence', dataset)
+            }
+          ],
+          dataAccessCommitteeId: { ...dac, dacs },
+        },
+        key: `prefilledConsentGroups[${idx}]`,
+        nihInstitutionalCertificationFile: null,
+        editMode: false,
+        valid: false,
+        disableFields: true,
```
fboulnois, line 155: If I'm understanding correctly, when you are prefilling the consent groups (i.e. `studyEditMode = true`), you set `disableFields = true` as well, and then you use this variable to determine whether fields should be disabled.

Is there ever a case where `disableFields = false` and `studyEditMode = true` or vice versa? If this is not the case, then the value of `studyEditMode` will always be equivalent to `disableFields`, which means you can simply do  `disabled: studyEditMode` in `ConsentGroupForm.js`. This avoids having to carry extra state around that is duplicated on each index.

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -40,22 +94,92 @@ export const DataAccessGovernance = (props) => {
...
+            {
+              fileType: extractFileTypes('File Types', 'fileType', dataset),
+              functionalEquivalence: extractFileTypes('File Types', 'functionalEquivalence', dataset)
+            }
+          ],
+          dataAccessCommitteeId: { ...dac, dacs },
+        },
+        key: `prefilledConsentGroups[${idx}]`,
+        nihInstitutionalCertificationFile: null,
+        editMode: false,
```
fboulnois, line 153: Same question with `editMode`, except the situation would be `const [editMode, setEditMode] = useState(!studyEditMode);`

src/components/data_submission/ds_study_information.js
```diff
@@ -42,7 +48,9 @@ export default function DataSubmissionStudyInformation(props) {
       validators: [FormValidators.REQUIRED],
       validation: validation.studyName,
       onChange,
-      onValidationChange
+      onValidationChange,
+      defaultValue: studyEditMode ? formData?.studyName : undefined,
+      disabled: studyEditMode ? true : false,
```
fboulnois:
```suggestion
      disabled: studyEditMode,
```

src/pages/DataSubmissionForm.js
```diff
@@ -172,11 +172,11 @@ export const DataSubmissionForm = (props) => {
...
-      <DataAccessGovernance onChange={onChange} onFileChange={onFileChange} validation={formValidation} onValidationChange={onValidationChange} setAllConsentGroupsSaved={setAllConsentGroupsSaved} />
+      <DataSubmissionStudyInformation onChange={onChange} validation={formValidation} onValidationChange={onValidationChange} studyEditMode={studyEditMode}/>
+      <NihAnvilUse onChange={onChange} formData={formData} validation={formValidation} onValidationChange={onValidationChange} studyEditMode={studyEditMode}/>
+      <NIHAdministrativeInformation formData={formData} onChange={onChange} institutions={institutions} validation={formValidation} onValidationChange={onValidationChange} studyEditMode={studyEditMode}/>
+      <NIHDataManagement formData={formData} onChange={onChange} onFileChange={onFileChange} validation={formValidation} onValidationChange={onValidationChange} studyEditMode={studyEditMode}/>
+      <DataAccessGovernance onChange={onChange} onFileChange={onFileChange} validation={formValidation} onValidationChange={onValidationChange} setAllConsentGroupsSaved={setAllConsentGroupsSaved} studyEditMode={studyEditMode}/>
```
fboulnois, line 179: nit: it might be clearer to set `studyEditMode=false` on all of these so it's clear at a glance what the value of `studyEditMode` is.

src/pages/StudyUpdateForm.js
```diff
@@ -0,0 +1,264 @@
...
+          return 'no_nhgri_yes_anvil';
+        case 'i am not nhgri funded and do not plan to store data in anvil':
+          return 'no_nhgri_no_anvil';
+        default:
+          return undefined;
+      }
+    }
+  };
+
+  // convert property to date
```
fboulnois:
```suggestion
```

src/pages/StudyUpdateForm.js
```diff
@@ -0,0 +1,264 @@
...
+    properties['nihAnvilUse'] = radioSelectionToLabels(properties['nihAnvilUse']);
+    properties['alternativeDataSharingPlanTargetDeliveryDate'] = toYYYYMMDD(properties['alternativeDataSharingPlanTargetDeliveryDate']);
+    properties['alternativeDataSharingPlanTargetPublicReleaseDate'] = toYYYYMMDD(properties['alternativeDataSharingPlanTargetPublicReleaseDate']);
+    properties['embargoReleaseDate'] = toYYYYMMDD(properties['embargoReleaseDate']);
+
+    return properties;
+  }, [study]);
+
+  const prefillFormData = useCallback(async (study) => {
+    // get dataSubmitter information
```
fboulnois:
```suggestion
```

fboulnois (commented):
A few suggestions:

src/components/data_submission/NIHAdministrativeInformation.js
kaitlynoflaherty: This section is only rendered from 3/4 states.

src/components/data_submission/DataAccessGovernance.js
kaitlynoflaherty: I think this state is necessary because when `studyEditMode = true`, it is possible that you will be adding a new consent group in which `editMode = true` to edit and `disableFields = false` because the name, dac, etc should not be disabled when adding new but should be disabled when editing a preexisting dataset. I think your logic would work if you are only editing the preexisting consent groups

src/pages/DataSubmissionForm.js
kaitlynoflaherty: I could be doing this wrong, but I'm getting this error 
image

src/components/data_submission/DataAccessGovernance.js
fboulnois: Perhaps instead you could have a `isPrefilled` variable with value set to `true` and `false` otherwise, and then set `disabled: isPrefilled` and `editMode: !isPrefilled`, that way you only need to carry one boolean around for each consent group and it's a little more clear why and when the boolean is set.

src/components/data_submission/consent_group/ConsentGroupForm.js
```diff
@@ -12,38 +12,42 @@ export const ConsentGroupForm = (props) => {
     updateNihInstitutionalCertificationFile,
     deleteConsentGroup,
     disableDelete,
+    consentGroupsState,
+    studyEditMode
   } = props;
 
   const [consentGroup, setConsentGroup] = useState({
```
fboulnois, line 20:
```suggestion
  const curConsentGroup = consentGroupsState[idx]?.consentGroup; // not sure if ? is needed.

  const [consentGroup, setConsentGroup] = useState({
```

It might be useful to pull this out beforehand so you can use it for all the properties below.

fboulnois (commented):
Another small suggestion:

src/components/data_submission/NIHAdministrativeInformation.js
kaitlynoflaherty: I think that it can be null

src/components/data_submission/DataAccessGovernance.js
kaitlynoflaherty: That's a good idea, thanks!

kaitlynoflaherty: The only issue with this logic is that you can't use `isPrefilled` to determine the state of editMode. `editMode` can be true when prefilled or not. With this, I'm not sure that having `isPrefilled` rather than `disableFields` is much more efficient. I think with both `disableFields` and `isPrefilled`  they are the same in controlling the `disabled:` field on the form components, so it is not necessary to have a condition with `studyEditMode` and I will update that now. I will definitely make the change from `disableFields` to `isPrefilled` if you think it is still worthwhile! I think the main difference would be declaring `isPrefilled` as a variable rather than a state on `consentGroupsState`. Please let me know your thoughts!

src/components/data_submission/consent_group/ConsentGroupForm.js
kaitlynoflaherty: I think that the checkbox component needs to be edited so I made a quick ticket in the backlog here: https://broadworkbench.atlassian.net/browse/DUOS-2635

src/components/data_submission/DataAccessGovernance.js
```diff
@@ -1,18 +1,73 @@
...
+
+  const getDiseaseLabels = async (ontologyIds) => {
+    let labels = [];
+    if (!isEmpty(ontologyIds)) {
+      const idList = ontologyIds.join(',');
+      const result = await DAR.searchOntologyIdList(idList);
+      labels = result.map(r => r.label);
+    }
+    return labels;
+  };
```
rushtong, line 32: This looks like a copy of `DatasetUpdate.getDiseaseLabels`. We should try to extract common functions like this. It might be helpful to include the try/catch in the normalize function below to make the function more resilient.

rushtong:
Clicking on the first option in `NIH and AnVIL Use`:

Screenshot 2023-08-07 at 9 40 09 AM

Should trigger the expanded view with additional questions for that section as well as `NIH Administrative Information` and `NIH Data Management & Sharing Policy Details` sections.

Update: It looks like the first three options are not showing their expanded states correctly, just the last one.

rushtong (approved):
👍🏽  I think we should merge this and then do a bug session once https://github.com/DataBiosphere/consent/pull/2119 is merged

src/utils/JsonSchemaUtils.js
```diff
@@ -70,7 +70,7 @@ const updateValidation = (existingValidation, validationError) => {
   }
 
   // if the field is required and empty, we shouldn't also error on, e.g., that it isn't a date format
-  if (existingValidation.failed.includes('required') || validationError === 'required') {
+  if (existingValidation.failed?.includes('required') || validationError === 'required') {
```
fboulnois: Is this still required?

src/pages/StudyUpdateForm.js
```diff
@@ -0,0 +1,261 @@
...
+        const me = Storage.getCurrentUser();
+        setUser(me);
+        setStudy(await DataSet.getStudyById(studyId));
+        setFailedInit(false);
+      } catch (error) {
+        Notifications.showError({ text: 'Failed to load study' });
+      }
+    };
+    init();
+  }, [studyId]);
```
fboulnois: I think these two init functions should be combined into one.

fboulnois (commented):
A few final questions / recommendations:

src/pages/StudyUpdateForm.js
```diff
@@ -0,0 +1,255 @@
...
+        });
+      }
+      try {
+        const me = Storage.getCurrentUser();
+        setUser(me);
+        setStudy(await DataSet.getStudyById(studyId));
+        setFailedInit(false);
+      } catch (error) {
+        Notifications.showError({ text: 'Failed to load study' });
+      }
```
fboulnois:
```suggestion
      try {
        const me = Storage.getCurrentUser();
        setUser(me);
        setStudy(await DataSet.getStudyById(studyId));
      } catch (error) {
        setFailedInit(true);
        Notifications.showError({ text: 'Failed to load study' });
      }
```

fboulnois: Since the default for `failedInit` is `false`, if the previous try/catch passes it will continue to be `false`, so the statement has no effect. I believe you want to check whether the init fails in the error handler instead.

fboulnois (commented):
I think I caught another small issue:

fboulnois (approved):
👍 looks good, thanks for your diligence on this ticket!

PR #2283, opened 2023-07-11
DUOS-2602 User Profile Page Update

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2603

### Summary
- created a new researcher profile page
- added a new route for new page (test_user_profile)
- removed unnecessary sections for new page

Screenshot 2023-07-11 at 2 10 00 PM


src/Routes.js
```diff
@@ -62,6 +63,7 @@ const Routes = (props) => (
     <Route path="/data_sharing_language_tool" component={DataSharingLanguageTool} />
     <Route path="/nih_dms_policy" component={NIHDMSPolicyInfo} />
     <Route path="/anvil_dms_policy" component={AnVILDMSPolicyInfo} />
+    <AuthenticatedRoute path="/test_user_profile" component={UserProfile} props={props} rolesAllowed={[USER_ROLES.all]} />
```
rushtong: nit: I think we can simplify this to `user_profile` which won't require a path change once this is the official version of the page.

src/pages/UserProfile.js
```diff
@@ -0,0 +1,397 @@
...
+    institutionId: undefined,
+    suggestedInstitution: undefined,
+    selectedSigningOfficialId: undefined,
+    suggestedSigningOfficial: undefined,
+    eraCommonsId: undefined,
+    email: undefined,
+    id: undefined
+  });
+
+  const possibleSupportRequests = [
```
rushtong: I think this is a candidate for pruning since it is only for a component that isn't on this page any more.

src/pages/UserProfile.js
```diff
@@ -0,0 +1,397 @@
...
+        setSigningOfficialList((institution.signingOfficials ? institution.signingOfficials : []));
+      });
+    } else {
+      setSelectedInstitution(null);
+    }
+  }, [profile.institutionId]);
+
+  useEffect(() => {
+    setCurrentSigningOfficial(findSigningOfficial(profile.selectedSigningOfficialId));
+  }, [profile.selectedSigningOfficialId, signingOfficialList, findSigningOfficial]);
```
rushtong: If we remove `currentSigningOfficial`, we can remove this too.

src/pages/UserProfile.js
```diff
@@ -0,0 +1,397 @@
...
+
+  const processSupportRequests = () => {
+    const filteredRequests = possibleSupportRequests.filter(request => supportRequests[request.key]);
+    return [
+      filteredRequests.length > 0,
+      filteredRequests
+        .map(x => `- ${x.label}`)
+        .join('\n')
+    ];
+  };
```
rushtong: This should be pruned

src/pages/UserProfile.js
```diff
@@ -0,0 +1,397 @@
...
+      Notifications.showSuccess(
+        {text: 'Sent Requests Successfully', layout: 'topRight', timeout: 1500}
+      );
+    } else {
+      Notifications.showError({
+        text: `ERROR ${response.status} : Unable To Send Requests`,
+        layout: 'topRight',
+      });
+    }
+  };
```
rushtong: Prune this too.

rushtong (changes requested):
I think we should prune as much as possible from this page and stick with just the name/email as the mocks show. The institution element should be in a separate component for "Affiliation & Roles"

From my understanding of the mocks, user name and email are read-only so we can probably pull them out of form elements too.

hams7504:
- removed all sections except for name and email
Screenshot 2023-07-11 at 3 15 29 PM

hams7504:
- removed unnecessary functions

src/pages/UserProfile.js
```diff
@@ -0,0 +1,80 @@
...
+        await getResearcherProfile();
+
+        setNotificationData(await NotificationService.getBannerObjectById('eRACommonsOutage'));
+      } catch (error) {
+        Notifications.showError({text: 'Error: Unable to retrieve user data from server'});
+      }
+    };
+
+    init();
+  }, [props.history]);
```
rushtong: I don't think you need this effect to be dependent on props.history - there's nothing in the code now that modifies it. We may need it in future iterations, but let's do that when we need it.

src/pages/UserProfile.js
```diff
@@ -0,0 +1,80 @@
+import {useState, useEffect} from 'react';
+import {div, h, hr} from 'react-hyperscript-helpers';
+import {PageHeading} from '../components/PageHeading';
+import {Notification} from '../components/Notification';
+import {FormField, FormFieldTypes, FormValidators} from '../components/forms/forms';
+import {User} from '../libs/ajax';
+import {NotificationService} from '../libs/notificationService';
+import { Notifications} from '../libs/utils';
+
+export default function ResearcherProfile(props) {
```
rushtong: Should be `UserProfile(props)`

On a code organization level, we are definitely going to be adding profile-specific components so let's put this in a `user_profile` directory where all of the relevant components can live. If we decide to re-use any of them, we can move them to the shared component directory.

src/pages/UserProfile.js
```diff
@@ -0,0 +1,80 @@
...
+          PageHeading({
+            id: 'researcherProfile',
+            color: 'common',
+            title: 'Your Profile',
+            description: 'Please complete the form below to start using DUOS.'
+          }),
+          hr({ className: 'section-separator' })
+        ]),
+        div({ className: 'col-md-10 col-md-offset-1 col-xs-12 no-padding' }, [
+          h(FormField, {
```
rushtong: Let's touch base with Jonathan today - I'm not convinced we need actual form elements here since I don't think they're modifiable in the new mocks. The input form details are just visual there.

rushtong (changes requested):
Looking much better! See comments inline.

rushtong (approved):
Thank you for the changes, looks great!

fboulnois (approved):
🎉

PR #2290, opened 2023-07-20
Duos UI 2602 affiliation and roles

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2613

### Summary
- Created affiliation and roles component
- Added new component to main profile page
- Removed react-bootstrap class names used for styling
- Users should be able to edit their institution as long as they are not a signing official with an existing institution

Screenshot 2023-07-20 at 1 16 15 PM


kaitlynoflaherty (approved):
Works great! Good job 😃

package.json
```diff
@@ -26,7 +26,7 @@
     "react-dropzone": "14.2.3",
     "react-ga": "3.3.1",
     "react-google-charts": "4.0.0",
-    "react-hyperscript-helpers": "2.0.0",
+    "react-hyperscript-helpers": "^2.0.0",
```
fboulnois: Do we need to update `react-hyperscript-helpers`?

fboulnois (commented):
One question:

package.json
hams7504: Just tested it out on my end and the resulting page is the same in both cases. I am not sure why npm updated it for me, but here's something I found about this:
"The main difference between "react-hyperscript-helper 2.0.0" and "^2.0.0" is the specificity of the version being requested. The former will install exactly version 2.0.0, while the latter will install any version that is 2.x.x but not versions that may break compatibility. It's generally a good practice to use version constraints like "^2.0.0" when you want to allow for patch and minor updates without risking major breaking changes."
Do you think it would be a good idea to include it or should I remove those changes for this branch?

fboulnois: I think the changes to `package.json` and `package-lock.json` should be reverted, since they're unrelated to the functionality of the ticket. Otherwise, the changes look good to me.

hams7504: sounds good, will do!

hams7504: @fboulnois Done!

fboulnois (approved):
I added a commit that ensures that the line endings in the `package-lock.json` file remain `LF` instead of `CRLF`, otherwise you'll see +30k, -30k changes made to the lockfile.

Otherwise, lgtm 👍

PR #2305, opened 2023-08-08
[DUOS-2452] DAC Dataset Acceptance => Public Visibility

kaitlynoflaherty:
### Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2452

### Summary
A dataset should only be visible if Public Visibility is YES and it is ACCEPTED by the DAC, 
unless it is an open access or external access dataset in which case it only needs Public Visibility to be set to YES.
Unless/until we migrate legacy datasets to have studies we rely on the Active dataset prop

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetCatalog.js
+++ b/src/pages/DatasetCatalog.js
@@ -27,14 +27,16 @@ const canApplyForDataset = (dataset) => {
   return dataset.active && !isNil(dataset.dacId);
 };
 
-
-// add condition based on dac vote decision:
-// if vote === true -> disabled = publicVisibility
 const isVisible = (dataset) => {
-  // need to get study? need to get correct study info to get public visibility
-  //const publicVisibility = dataset.study?.publicVisibility;
-  // return dataset.dacApproval ? dataset.study?.publicVisibility : false;
-  return dataset.dacApproval && dataset.active;
+  if(dataset.dacApproval) {
+    if(!isNil(dataset.study.publicVisibility)){
+      return dataset.study?.publicVisibility && dataset.active;
+    } else {
+      return dataset.active;
+    }
+  } else {
+    return false;
+  }
 };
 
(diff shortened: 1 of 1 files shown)


src/pages/DatasetCatalog.js
```diff
@@ -27,6 +27,21 @@ const canApplyForDataset = (dataset) => {
...
+const extractDatasetProp = (propertyName, dataset) => {
+  const property = find({ propertyName })(dataset.properties);
+  return property?.propertyValue;
+};
+
+const isVisible = (dataset) => {
+  const openAccess = extractDatasetProp('Open Access', dataset);
+  if(!isNil(openAccess)){
+    // if open Access is false, dac approval required
+    return openAccess ? dataset.study.publicVisibility : (dataset.dacApproval && dataset.study.publicVisibility);
```
rushtong: `study` can be null, do we need null-safe traversal here?

rushtong (approved):
One suggestion inline, otherwise 👍🏽

src/pages/DatasetCatalog.js
kaitlynoflaherty: My thinking was since the openAccess is a property on a study,  if `openAccess` is not null then the study must exist.

rushtong: I think that's true today, but the code doesn't enforce that currently and could be incorrect in the near future.

kaitlynoflaherty: Good point. Do you think that a good approach would be to just add `dataset.study?.publicVisibility` or do more thorough handling here?

rushtong: I think that's fine for this case.

fboulnois (approved):
Glad we're simplifying this logic 👍

PR #2315, opened 2023-08-21
[DUOS-2612] Controlled Access Grants Component

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2612

### Summary
- Created a sortable table component
- Used table component to create the controlled access grants section of the user profile

----

- Verify all tests go green


src/components/SortableTable.js
```diff
@@ -0,0 +1,292 @@
...
+  order: PropTypes.oneOf(['asc', 'desc']).isRequired,
+  orderBy: PropTypes.string.isRequired,
+};
+
+export default function SortableTable(props) {
+
+  const {
+    rows,
+    headCells
+  } = props;
```
rushtong: I think this component would be more useful if a predefined `theme` were passed in as a prop from the caller.
As a side note, we do have another `theme.js` class which can seem confusing. I'm wondering if `SortableTable` could be it's own package level (i.e. directory) component with `EnhancedTableHead` and `Themes` as separate components in that package.

src/components/SortableTable.js
```diff
@@ -0,0 +1,292 @@
...
+                  const labelId = `enhanced-table-checkbox-${index}`;
+
+                  return (
+                    <TableRow
+                      hover
+                      onClick={(event) => handleClick(event, row.name)}
+                      tabIndex={-1}
+                      key={row.datasetIdentifier}
+                      selected={isItemSelected}
+                    >
```
rushtong: minor nit: collapse the closing tag to the previous line. Here and in other places as well.

rushtong (approved):
Some minor comments inline, but otherwise, this looks great! I can see some potential improvements such as sorting on different columns, but we should only tackle that if it becomes a requirement.

hams7504:
> Some minor comments inline, but otherwise, this looks great! I can see some potential improvements such as sorting on different columns, but we should only tackle that if it becomes a requirement.

Could you explain what you mean by "sorting on different columns"? Currently, the table should be able to sort based on all columns passed in through the headCells variable.

src/components/SortableTable.js
hams7504: I can implement it that way! I think the SortableTable file is way too crowded.

fboulnois (commented):
Once the Codacy issues have been resolved, let's rebase and flatten this PR to make it easier to review.

rushtong:
> > Some minor comments inline, but otherwise, this looks great! I can see some potential improvements such as sorting on different columns, but we should only tackle that if it becomes a requirement.
> 
> Could you explain what you mean by "sorting on different columns"? Currently, the table should be able to sort based on all columns passed in through the headCells variable.

You are correct, I didn't see the additional sorts, sorry about that!

PR #2320, opened 2023-08-28
[DUOS-1442][DUOS-2641][risk=no] Remove unused ajax calls.

rushtong:
### Addresses
Partial front end support for:
* https://broadworkbench.atlassian.net/browse/DUOS-1442
* https://broadworkbench.atlassian.net/browse/DUOS-2641

### Summary
This PR removes several unused ajax calls
* `getAutoCompleteDS`: This calls an API that returns a deprecated model object.
* `getDARDocument`: This was used in the old DAR form for documents associated to a DAR. We don't use that any longer so we can remove this call.
* `disableDataset`: This was used to set the active property on a dataset to true/false which used to control visibility. We no longer use that feature for visibility so we can remove this call.

----

- Verify all tests go green


fboulnois (approved):
Always good to see cleanup 👍

hams7504 (approved):
looks good! 👍

PR #2324, opened 2023-08-30
[DUOS-2616] Finalizing User Profile Page

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2616

### Summary

----

- Verify all tests go green


src/pages/user_profile/UserProfile.js
```diff
@@ -54,68 +55,114 @@ export default function UserProfile(props) {
...
+      />
+    </h>
+    <div style={{ marginTop: '10px' }} />
+    <h>
+      <FormField
+        type={FormFieldTypes.TEXT}
+        id='profileEmail'
+        defaultValue={profile.email}
+        readOnly={true}
+      />
```
rushtong: For these fields, when they are read-only, have a normal selectable cursor. The institution selection, on the other hand, has the cursor that prevents selection. We should try to have the same non-selection cursor in this case.

rushtong: My second question (and maybe we've addressed this before) is why we have a form field if it is always read-only? I think maybe this decision was about styling according to the mocks. I would suggest that we reconsider this decision (and the mock) if that is the case ... if it's never editable, then why make it look like it could be?

src/pages/user_profile/AffiliationAndRoles.js
```diff
@@ -142,67 +141,67 @@ export default function AffiliationAndRole(props) {
...
-      div({ className: '', style: { 'marginTop': '83px' } }, []),
-      p({}, ['Role']),
-    ]),
-    p({}, [profile.roles]),
-  ]);
+        }}
+        disabled={!formIsValid()}
+      >
+        Save
+      </button>
```
rushtong: I think we could use some tool-tip text here to indicate why the save button is disabled. This is not in the stated requirements, but it would help reduce reliance on customer support.

src/pages/user_profile/ResearcherStatus.js
```diff
@@ -79,7 +79,7 @@ export default function ResearcherStatus(props) {
       location: pageProps.location,
       header: false
     })}
-    <div style={{ marginTop: '10px' }} />
+    <div style={{ marginTop: '20px' }} />
```
rushtong, line 81: On line 76, we set the destination to `profile`: 

```
    {eRACommons({
      destination: 'profile',
      onNihStatusUpdate: () => { },
      location: pageProps.location,
      header: false
    })}
```

That is the url fragment the eRA auth login flow uses to return the user back to after authenticating. In this case, it returns you back to `https://local.broadinstitute.org:3000/profile` instead of `https://local.broadinstitute.org:3000/user_profile`. If you change that to the following:

```
    {eRACommons({
      destination: 'user_profile',
      onNihStatusUpdate: () => { },
      location: pageProps.location,
      header: false
    })}
```

It should redirect you back to this new user profile page. This is the essential trick to using this component anywhere in the app ... as long as you give it a valid url fragment, it will reload after authenticating.

rushtong (approved):
Please see comments inline. Please address as you see fit. This is looking great, love the new styling and simplification.

src/pages/user_profile/ResearcherStatus.js
hams7504: I switched Researcher Profile with User Profile in Routes.js so that the User Profile page loads for the `/profile` url instead of Researcher Profile. That should fix this!

fboulnois (approved):
I like @rushtong 's suggestions, but otherwise looks good 👍

src/pages/user_profile/UserProfile.js
hams7504: Got it! I added the same non-selection cursor.

fboulnois (approved):
👍

rushtong:
Side note - as a result of this PR, `ResearcherProfile` becomes dead code. We tend to add clean up tickets to handle that kind of work, so let's do that here too.

src/components/sortable_table/SortableTable.js
```diff
@@ -11,6 +11,34 @@ import { ThemeProvider } from '@mui/material/styles';
...
+//   },
+//   {
+//     id1: 'id1_value2',
+//     id2: 'id2_value2'
+//   },
+//   {
+//     id1: 'id1_value3',
+//     id2: 'id2_value3'
+//   }
+// ];
```
rushtong: nit: rm commented lines

rushtong:
Minor visual nit. When sorting results in one column being wider than it was before the sort, the headers shift size as well which is a bit jarring. Consider this an optional request to see if it's possible to fix the column sizes making it less dynamic, but more stable when sorting.

Compare the two screens noting where the sort icon is for dataset identifier:

Screenshot 2023-09-07 at 2 30 30 PM

Screenshot 2023-09-07 at 2 30 27 PM

src/pages/user_profile/AffiliationAndRoles.js
```diff
@@ -1,13 +1,40 @@
...
+        styleOverrides: {
+          tooltip: {
+            fontSize: '14px',
+            color: 'white',
+            fontFamily: 'Montserrat'
+          }
+        }
+      },
+      MuiButton: {
+        // styleOverrides: {
```
rushtong: nit: rm commented line(s)

rushtong:
I have another feature request that should be in a separate ticket. The `Accepted Terms & Policies` currently looks for user acknowledgements that we store locally (i.e. PDF agreements the user has clicked on and read). However, there is another Terms item that all DUOS users have to agree to and that's Terms of Service

Screenshot 2023-09-07 at 2 44 02 PM

Can we add at the top of the list a note describing that the user has agreed to the DUOS/Terra ToS? We know that for two reasons, one is that the user can see this page. The second is that it's part of the response of the `/api/sam/register/self/diagnostics` API call (not made from this page now, but could be added):

```
https://consent.dsde-dev.broadinstitute.org/api/sam/register/self/diagnostics
...
{
  "adminEnabled": true,
  "enabled": true,
  "inAllUsersGroup": true,
  "inGoogleProxyGroup": true,
  "tosAccepted": true
}
```

rushtong (approved):
Looks good! 👍🏽

src/components/sortable_table/SortableTable.js
hams7504: Sure! My original intention for including it was to have a guide for people who want to use SortableTable in the future. But I can remove those lines if it's better.

hams7504:
> I have another feature request that should be in a separate ticket. The `Accepted Terms & Policies` currently looks for user acknowledgements that we store locally (i.e. PDF agreements the user has clicked on and read). However, there is another Terms item that all DUOS users have to agree to and that's Terms of Service
> 
> Screenshot 2023-09-07 at 2 44 02 PM
> 
> Can we add at the top of the list a note describing that the user has agreed to the DUOS/Terra ToS? We know that for two reasons, one is that the user can see this page. The second is that it's part of the response of the `/api/sam/register/self/diagnostics` API call (not made from this page now, but could be added):
> 
> ```
> https://consent.dsde-dev.broadinstitute.org/api/sam/register/self/diagnostics
> ...
> {
>   "adminEnabled": true,
>   "enabled": true,
>   "inAllUsersGroup": true,
>   "inGoogleProxyGroup": true,
>   "tosAccepted": true
> }
> ```

Of course, I can add that.
Is the time and date stored when the user agrees to the DUOS/Terra Terms of Service? If not, should I leave that part blank?

src/components/sortable_table/SortableTable.js
rushtong: Ah, I didn't get that. There is a standard for JS doc ... it would be good to use that at the top of the component. Look at `BucketUtils.js` for an example.

rushtong:
> Is the time and date stored when the user agrees to the DUOS/Terra Terms of Service? If not, should I leave that part blank?

Correct, that date is not something that is exposed from Sam, AFAIK.

rushtong (approved):
👍🏽

PR #2329, opened 2023-09-11
[DUOS-2635] Implemented disabled checkboxes for data submission form

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2635

### Summary
- Added a non-selection cursor to the checkboxes in the Data Submission Form when the consent group is prefilled (when the checkboxes are disabled)

----

- Verify all tests go green

### DIFF
--- a/src/components/forms/formComponents.js
+++ b/src/components/forms/formComponents.js
@@ -485,6 +485,7 @@ export const FormInputCheckbox = (config) => {
     label({
       className: `regular-checkbox ${!isValid(validation) ? 'errored' : ''}`,
       htmlFor: `${id}`,
+      style: disabled ? { cursor: 'not-allowed' } : null
     }, [toggleText])
   ]);
 };
--- a/src/components/data_submission/consent_group/EditConsentGroup.js
+++ b/src/components/data_submission/consent_group/EditConsentGroup.js
@@ -305,6 +305,7 @@ export const EditConsentGroup = (props) => {
           name: 'gso',
           type: FormFieldTypes.CHECKBOX,
           toggleText: 'Genetic studies only (GSO)',
+          disabled: disableFields,
           defaultValue: consentGroup.gso,
           onChange,
           validation: validation.gso,


src/components/forms/formComponents.js
```diff
@@ -485,6 +485,7 @@ export const FormInputCheckbox = (config) => {
     label({
       className: `regular-checkbox ${!isValid(validation) ? 'errored' : ''}`,
       htmlFor: `${id}`,
+      style: disabled ? { cursor: 'not-allowed' } : null
     }, [toggleText])
   ]);
```
rushtong, line 490: I'm seeing two things unrelated to the ticket that shouldn't hold up this PR, maybe we can address separately.
1. I had the question of _Why is the input not inside the label?_ and needed to look that up. I learned HTML in the 90's, but apparently it can be [before, around, or after](https://www.w3.org/TR/html401/interact/forms.html#edef-LABEL)
2. We're still using hyperscript here ... would be really nice to get rid of that and stick with jsx.

rushtong (approved):
Looks good! 👍🏽

src/components/forms/formComponents.js
hams7504: 1. Yes! And I also don't understand why the input tag is not directly creating the checkbox.
2. Definitely agree!

src/components/forms/formComponents.js
```diff
@@ -485,6 +485,7 @@ export const FormInputCheckbox = (config) => {
     label({
       className: `regular-checkbox ${!isValid(validation) ? 'errored' : ''}`,
       htmlFor: `${id}`,
+      style: disabled ? { cursor: 'not-allowed' } : null
```
nmalfroy, line 488: I'm assuming that this will default to `pointer` in the null case?

nmalfroy (approved):
Looks good!  Might just be worth double checking the behavior if it's enabled (I'm assuming it's fine)

src/components/forms/formComponents.js
hams7504: Yes, that's right!

PR #2354, opened 2023-10-13
[DUOS-2720] Fix date and uri validation

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2720

### Summary
There are some issues with trying to pass custom validator functions into the Ajv-generated validator code. In our case, we have custom functions in `JsonSchemaUtils.js` but it doesn't look like we're actually passing them into the script that does code generation (in a separate script). Therefore, they're never actually used.

As a workaround, this PR fixes the date validation and re-implements the URI validation that was previously commented out due to this same problem. These were being pulled from `ajv-formats` but seem to be intermittently undefined, i.e., the validation function would exist for one field, but not for another later on in the validation process.

----

- Verify all tests go green


src/assets/schemas/DataRegistrationV1Validation.js
```diff
@@ -1,17 +1,681 @@
...
+      "label": "Does the data need to be managed under Controlled or Open Access?",
+      "description": "Does the data need to be managed under Controlled or Open Access?",
+      "enum": ["Controlled Access", "Open Access"]
+    }
+  }
+};
 const func3 = require("ajv/dist/runtime/ucs2length").default;
-const formats0 = { "_items": ["require(\"ajv-formats/dist/formats\").", { "str": "fullFormats" }, ""] }.date;
+const dateRegex = /^\d{4}-(02-(0[1-9]|[12][0-9])|(0[469]|11)-(0[1-9]|[12][0-9]|30)|(0[13578]|1[02])-(0[1-9]|[12][0-9]|3[01]))$/;
+const dateValid = (value) => dateRegex.test(value);
```
rushtong, line 537: This is the meat of the fix - this date regex comes from `/src/components/forms/formValidation.js`

fboulnois: I think this validation file is automatically generated from `ajv`, if this change is implemented manually will it be overwritten the next time we run `ajv`?

fboulnois (commented):
One question:

hams7504 (approved):
lgtm! 👍

src/assets/schemas/DataRegistrationV1Validation.js
rushtong: Yes, there is a script that we have to run that re-writes this file. Most of the changes here are from that process. This is not a sustainable process and we need to move away from it which is what I'm looking into now.

fboulnois (approved):
lgtm 👍

PR #2357, opened 2023-10-16
[DUOS-2721] Inherit all data libraries from base url

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2721

### Summary

Refactors data libraries so that are all available at `/datalibrary/<name>`. 

Advantages of this approach include:

- Not having to create a new route in code every time
- All data libraries inherit from a standardized url scheme
- Easier to add more data libraries in the future
- Falls back to instant data libraries if the path doesn't exist

----

- Verify all tests go green


rushtong (approved):
I like this approach, very clean 👍🏽

hams7504 (approved):
looks good!

PR #2378, opened 2023-11-02
[DUOS-2773] Truncate Table Cells

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2773

### Summary

----

- Verify all tests go green


src/components/data_search/DatasetSearchTable.js
```diff
@@ -115,30 +115,39 @@ export const DatasetSearchTable = (props) => {
           data: [
             {
               value: entry[0].study.studyName,
+              truncate: true,
```
rushtong, line 117: It would be nicer if there were a default value set to `false` and only specified `true` when necessary.

src/components/CollapsibleTable.js
```diff
@@ -145,9 +164,16 @@ const CollapsibleRow = (props) => {
           </IconButton>
         </StyledTableCell>
         {row.data.map((cell, i) => (
-          <StyledTableCell key={i}>
-            {cell.value}
-          </StyledTableCell>
+          cell.truncate ?
```
rushtong: Would optional chaining work here and allow for fewer changes in `DatasetSearchTable.js`?
```suggestion
          cell?.truncate ?
```

src/components/CollapsibleTable.js
```diff
@@ -114,6 +117,22 @@ const StyledTableCell = styled(TableCell)(() => ({
   },
 }));
 
+const TruncatedTableCell = styled(StyledTableCell)(() => ({
```
fboulnois, line 120: Instead of expanding on hover, can we make the ellipsis a button that can be clicked to reveal more of the text?

fboulnois (approved):
I agree with @rushtong 's suggestion to make the property optional. Also, another idea:

src/components/CollapsibleTable.js
```diff
@@ -144,11 +163,20 @@ const CollapsibleRow = (props) => {
...
         </StyledTableCell>
-        {row.data.map((cell, i) => (
-          <StyledTableCell key={i}>
-            {cell.value}
-          </StyledTableCell>
-        ))}
+        {row.data.map((cell, i) => {
+          if (!('truncate' in cell)) {
+            cell.truncate = false;
+          }
```
fboulnois: Is this check needed? I think the `cell?.truncate` should take care of the condition.

fboulnois (commented):
One comment:

src/components/CollapsibleTable.js
hams7504: Thanks for the suggestion! I discussed this with @solideoglori and he said he might prefer the hover option for now.

src/components/CollapsibleTable.js
rushtong: Same feedback that @fboulnois has here, I don't think this is necessary.

rushtong (approved):
👍🏽

PR #2385, opened 2023-11-08
Revert "[DUOS-1740-dependabot]: Bump the npm-dependencies group with 6 updates"

rushtong:
Reverts DataBiosphere/duos-ui#2380

This causes a strange react error on the dataset catalog page. Reverting this and will look into updating each library separately.
```
Uncaught ReferenceError: Cannot access 'K' before initialization
    at cI (DatasetCatalog.js:107:7)
    at ki (react-dom.production.min.js:167:137)
    at kl (react-dom.production.min.js:290:337)
    at wc (react-dom.production.min.js:280:389)
    at yc (react-dom.production.min.js:280:320)
    at vc (react-dom.production.min.js:280:180)
    at ic (react-dom.production.min.js:271:88)
    at ac (react-dom.production.min.js:268:429)
    at S (scheduler.production.min.js:13:203)
    at MessagePort._ (scheduler.production.min.js:14:128)
```

**Ancillary change:** The authentication integration test is failing. It works locally, I suspect something has changed in the test environment with the service account. Disabling in favor of a different approach where we test authentication against a BEE environment.


hams7504 (approved):
👍

fboulnois (approved):

PR #2390, opened 2023-11-08
[DUOS-2690] Institution Not Updated Correctly

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2690

### Summary

----

- Verify all tests go green


src/pages/user_profile/AffiliationAndRoles.js
```diff
@@ -115,6 +116,8 @@ export default function AffiliationAndRole(props) {
     };
 
     let updatedUser = await User.updateSelf(payload);
+    const user = await User.getMe();
+    setUserRoleStatuses(user, Storage);
```
rushtong: `updatedUser` should be the same object as `user` in this context so you probably don't have to make the call to `User.getMe()`. Can you verify this and maybe simplify?

fboulnois (approved):
Agreed with @rushtong , otherwise lgtm.

rushtong (approved):
Thanks for the changes!

PR #2400, opened 2023-11-15
[DUOS-2757] Increase first column width in study and dataset tables

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2757

### Summary
Screenshot 2023-11-15 at 2 42 24 PM

----

- Verify all tests go green

### DIFF
--- a/src/components/CollapsibleTable.js
+++ b/src/components/CollapsibleTable.js
@@ -131,6 +131,15 @@ const TruncatedTableCell = styled(StyledTableCell)(() => ({
       wordBreak: 'break-all'
     },
   },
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍🏽

fboulnois (dismissed):
The `CollapsibleTable` component should be general purpose, so this fixes it for the data library at the cost of breaking other potential `CollapsibleTable`s. I would add a new `style` property in the data and then switch the size based on the value of that property.

rjohanek (approved):
Dataset table on the SO console looks good and has wider name/id column!

PR #2412, opened 2023-11-27
[DUOS-2774] Update Library Styling

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2774

### Summary

----

- Verify all tests go green


rushtong (approved):
I think this looks good so far and takes us in the right direction. There are some other tweaks that need to be made but can be done in separate tickets.

---
Note the thicker border that stops just past the inner table (upper circle). I think it has to do with the border definitions around the inner table. We should try and get rid of the doubled line and also extend the top of the NEXT row all the way across the screen (see lower circle).
Screenshot 2023-11-27 at 9 28 41 AM

---
The mocks call for some space between the rows. This can also be a separate ticket but should be addressed:

Screenshot 2023-11-27 at 9 29 53 AM

hams7504:
> I think this looks good so far and takes us in the right direction. There are some other tweaks that need to be made but can be done in separate tickets.
> 
> Note the thicker border that stops just past the inner table (upper circle). I think it has to do with the border definitions around the inner table. We should try and get rid of the doubled line and also extend the top of the NEXT row all the way across the screen (see lower circle). Screenshot 2023-11-27 at 9 28 41 AM
> 
> The mocks call for some space between the rows. This can also be a separate ticket but should be addressed:
> 
> Screenshot 2023-11-27 at 9 29 53 AM

Sounds good! I can make a separate ticket for those changes. I didn't end up implementing the space between rows since the ticket did not mention that, but I can make include that in the new ticket as well! Thank you.

rjohanek (approved):
Approve with some comments/changes - Mainly, the subtable now has headers that wrap to the second line causing a thick header. I think it looked better when the titles could fit all on one line. Maybe you can make the subtable wider? Additionally, minor thing, the mock has a thinner header for both the larger table and the sub table, if you're trying to match the mock I think this should be thinner. Finally, like Greg said, if the thick line could be extended for the entire table instead of stopping around participants I think it would look neater.

hams7504:
> Approve with some comments/changes - Mainly, the subtable now has headers that wrap to the second line causing a thick header. I think it looked better when the titles could fit all on one line. Maybe you can make the subtable wider? Additionally, minor thing, the mock has a thinner header for both the larger table and the sub table, if you're trying to match the mock I think this should be thinner. Finally, like Greg said, if the thick line could be extended for the entire table instead of stopping around participants I think it would look neater.

Thank you! I'll add these suggestions to the new styling ticket

hams7504:
Created new ticket to address additional styling updates: https://broadworkbench.atlassian.net/browse/DUOS-2820

PR #2414, opened 2023-11-30
[DUOS-2797] Create Request LC page

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2797

### Summary
Screenshot 2023-11-30 at 10 58 07 AM

----

- Verify all tests go green

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -15,7 +15,7 @@ import NIHICWebform from './pages/NIHicWebform';
 import PrivacyPolicy from './pages/PrivacyPolicy';
 import ResearcherConsole from './pages/researcher_console/ResearcherConsole';
 import UserProfile from './pages/user_profile/UserProfile';
-import RequestRole from './pages/user_profile/RequestRole';
+import RequestForm from './pages/user_profile/RequestForm';
 import SigningOfficialResearchers from './pages/signing_official_console/SigningOfficialResearchers';
 import SigningOfficialDarRequests from './pages/signing_official_console/SigningOfficialDarRequests';
 import SigningOfficialDataSubmitters from './pages/signing_official_console/SigningOfficialDataSubmitters';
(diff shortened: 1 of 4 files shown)


src/pages/user_profile/RequestForm.js
```diff
@@ -31,14 +31,31 @@ export default function RequestRole(props) {
     }
   ];
 
-  const [hasSupportRequests, setHasSupportRequests] = useState(false);
-  const [supportRequests, setSupportRequests] = useState({
+  const possibleLCRequests = [
```
Shakespeared: another way you could structure this would be to keep the single const `possibleSupportRequests` and set the keys and labels in the logic on line 42 where it checks which type of request is coming through (easier in a cond/switch case block than a ternary). That would be a little more streamlined, but maybe less obvious which request was triggered.

src/Routes.js
```diff
@@ -66,7 +66,8 @@ const Routes = (props) => (
     <Route path="/nih_dms_policy" component={NIHDMSPolicyInfo} />
     <Route path="/anvil_dms_policy" component={AnVILDMSPolicyInfo} />
     <AuthenticatedRoute path="/profile" component={UserProfile} props={props} rolesAllowed={[USER_ROLES.all]} />
-    <AuthenticatedRoute path="/request_role" component={RequestRole} props={props} rolesAllowed={[USER_ROLES.all]} />
+    <AuthenticatedRoute path="/request_role" component={RequestForm} props={Object.assign({isRequestRolePage: true}, props)} rolesAllowed={[USER_ROLES.all]} />
+    <AuthenticatedRoute path="/request_lc" component={RequestForm} props={Object.assign({isRequestLCPage: true}, props)} rolesAllowed={[USER_ROLES.all]} />
```
rushtong: When using Object.assign, the safe thing to do is to use an empty object to mutate and then merge in others as required. For example, if there is a `isRequestLCPage` in props, then it will be overwritten with the value in props instead of `true` here. 
```suggestion
    <AuthenticatedRoute path="/request_role" component={RequestForm} props={Object.assign({}, props, {isRequestRolePage: true})} rolesAllowed={[USER_ROLES.all]} />
    <AuthenticatedRoute path="/request_lc" component={RequestForm} props={Object.assign({}, props, {isRequestLCPage: true})} rolesAllowed={[USER_ROLES.all]} />
```

hams7504: That makes sense, thank you!

rushtong (approved):
Thank you for the changes 👍🏽

fboulnois (approved):
👍

PR #2416, opened 2023-12-04
[DUOS-2825] Replace institution id with name in /datalibrary/myinstitution page

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2825

### Summary
Screenshot 2023-12-04 at 1 55 28 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetSearch.js
+++ b/src/pages/DatasetSearch.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useEffect, useState } from 'react';
 import { Notifications } from '../libs/utils';
-import { DataSet } from '../libs/ajax';
+import { DataSet, Institution } from '../libs/ajax';
 import DatasetSearchTable from '../components/data_search/DatasetSearchTable';
 import broadIcon from '../logo.svg';
 import duosIcon from '../images/duos-network-logo.svg';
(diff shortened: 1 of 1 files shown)


src/pages/DatasetSearch.js
```diff
@@ -180,6 +181,10 @@ export const DatasetSearch = (props) => {
         props.history.push('/profile');
         return;
       }
+      else {
+        const res = await Institution.getById(institution);
+        setInstitutionName(res.name);
```
rushtong: It would be nice to have some error handling here ... if `res` comes back as not found, we could see an undefined error on `res.name`

rushtong: Another question ... I think we have this value on the user already as part of the `/api/user/me` API call which should also be stored in `Storage.getCurrentUser()`. Why do we need to query for this value again here?

fboulnois: I don't think we need to `useState` either. The bug can be fixed by setting `const institution = user.institution` and then extracting the `institution?.name` and `institution?.id`.

fboulnois (changes requested):
See comment below:

fboulnois (approved):
👍

src/pages/DatasetSearch.js
```diff
@@ -48,6 +48,7 @@ export const DatasetSearch = (props) => {
   const [loading, setLoading] = useState(true);
   const user = Storage.getCurrentUser();
   const institution = user.institution?.id;
```
fboulnois: Maybe switch this to `institutionId` for consistency and correctness

rushtong (approved):
👍🏽

PR #2422, opened 2023-12-13
[DUOS-2845][risk=no] Fix dac datasets modal bug

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2845

### Summary
Refactor the dac datasets modal to be a newer functional component and fix the error when using the read more component.

Also note that we're removing Consent ID and Sample Collection ID from the display ... those are no longer supported in the new Dataset/Study model.

### Why
Using the `ReadMore` component as part of the translated data use caused a problem with too many `useEffect` hooks in combination with the parent components. The "fix" here is to use a shortened (or ellipses concatenated string) along with using the full translation as the title text of the span it lives in and remove the use of the `ReadMore` component altogether.

----

- Verify all tests go green

### DIFF
--- a/src/components/modals/DacDatasetsModal.js
+++ b/src/components/modals/DacDatasetsModal.js
@@ -1,120 +1,125 @@
 import find from 'lodash/find';
 import get from 'lodash/get';
 import isEmpty from 'lodash/isEmpty';
-import { Component } from 'react';
-import { a, div, hh, span, table, tbody, td, th, thead, tr } from 'react-hyperscript-helpers';
-import { BaseModal } from '../BaseModal';
-import { ReadMore } from '../ReadMore';
-import { DataUseTranslation } from '../../libs/dataUseTranslation';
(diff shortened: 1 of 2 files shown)


hams7504 (approved):
lgtm! 👍

Ghost-in-a-Jar (approved):
I am not really very familiar with the DUOS UI but this seems reasonable. 

I think it would maybe be worth converting DacDatasetsModal to ts, the ts-ignore comments peppered around worry me a bit, but I didnt want to block fixing a bug.

rushtong:
> think it would maybe be worth converting DacDatasetsModal to ts, the ts-ignore comments peppered around worry me a bit, but I didnt want to block fixing a bug.

Thank you @Ghost-in-a-Jar - yes, I was looking into that too. The TS conversion is non-trivial at the moment since we are just starting that journey. I do intend to do that, but was hoping to get past the bug first.

PR #2423, opened 2023-12-14
[DUOS-2805][risk=no] Turn off TS checking for old javascript

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2805

### Summary
Minor config update so that TypeScript errors are not displayed in JS files

#### Old

Screenshot 2023-12-14 at 5 05 26 PM

#### New

Screenshot 2023-12-14 at 5 05 39 PM

----

- Verify all tests go green

### DIFF
--- a/src/components/modals/DacDatasetsModal.js
+++ b/src/components/modals/DacDatasetsModal.js
@@ -5,7 +5,6 @@ import React, {useEffect, useState} from 'react';
 import {BaseModal} from '../BaseModal';
 import {DataUseTranslation} from '../../libs/dataUseTranslation';
 
-// @ts-ignore
 const DacDatasetsModal = (props) => {
 
   const {showModal, onCloseRequest, datasets, dac} = props;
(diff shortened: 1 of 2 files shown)


hams7504 (approved):
looks good! 👍

fboulnois (approved):
Approved, although I see the following risks:

- Without these chunks being flagged, there may be no / low desire to move over the older Javascript to Typescript. What is our long term approach and commitment to migrating to Typescript?
- Not checking Javascript alongside with Typescript may obscure bugs we already have but are not triggering or bugs that are triggering but are not noticing.

rushtong:
> * Without these chunks being flagged, there may be no / low desire to move over the older Javascript to Typescript. What is our long term approach and commitment to migrating to Typescript?

I agree with this, we need to commit to moving to TS as a team. Will bring this up in team meetings.

> * Not checking Javascript alongside with Typescript may obscure bugs we already have but are not triggering or bugs that are triggering but are not noticing.

I don't think I agree with this ... we ARE checking our js with other tools (eslint, cypress) in our current CI process. This is intended to reduce the current noise where valid JS code is not syntactically valid TS code.

PR #2427, opened 2024-01-04
[DUOS-2861] Return undefined instead of error when no data custodian

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2861

### Summary

Return undefined instead of error when no data custodian

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTable.js
+++ b/src/components/data_search/DatasetSearchTable.js
@@ -148,7 +148,7 @@ export const DatasetSearchTable = (props) => {
               truncate: true,
             },
             {
-              value: entry[0].study.dataCustodianEmail.join(', '),
+              value: entry[0].study.dataCustodianEmail?.join(', '),
               truncate: true,
             },
           ],


hams7504 (approved):
looks great!

src/components/data_search/DatasetSearchTable.js
```diff
@@ -148,7 +148,7 @@ export const DatasetSearchTable = (props) => {
               truncate: true,
             },
             {
-              value: entry[0].study.dataCustodianEmail.join(', '),
+              value: entry[0].study?.dataCustodianEmail.join(', '),
```
rushtong: IIUC, this will break on an undefined `study`, not an undefined `dataCustodianEmail`. If we move the `?` to `study.dataCustodianEmail?.join`, we'll still get an `undefined` passed into the `join` method which will fail.

fboulnois: Whoops I put it on the wrong side. According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining , it will short circuit and not run the join if the email is undefined.

src/components/data_search/DatasetSearchTable.js
```diff
@@ -148,7 +148,7 @@ export const DatasetSearchTable = (props) => {
               truncate: true,
             },
             {
-              value: entry[0].study.dataCustodianEmail.join(', '),
+              value: entry[0].study.dataCustodianEmail?.join(', '),
```
rushtong, line 151: I think this will still error out. This will pass `undefined` to `join` which will throw. I think we need a conditional here to avoid trying to send anything invalid to `join`

fboulnois: See previous comment, according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining , it will short circuit and not run the join if the email is undefined.

fboulnois: Minimal example:

```
a = { b: [ 'c', 'd' ], e: undefined };
a.b.join(', ') // "c, d"
a.e.join(', ') // undefined
```

fboulnois: Minimal example:

```
a = { b: [ 'c', 'd' ], e: undefined };
a.b.join(',')  // "c,d"
a.e?.join(',') // undefined
```

rushtong: Oops, missed that comment, thank you!

rushtong (approved):
👍🏽

PR #2432, opened 2024-01-09
[DUOS-2853][risk=no] Cache doid responses in local storage

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2853

### Summary
Cache DOID api calls so we don't make unnecessary API calls.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ontologyService.js
+++ b/src/libs/ontologyService.js
@@ -1,6 +1,7 @@
 import axios from 'axios';
-import { Notifications } from './utils';
-import { getOntologyUrl } from './ajax';
+import {Notifications} from './utils';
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
looks good, how much does this improve performance in the typical case?

rushtong:
> looks good, how much does this improve performance in the typical case?

The performance increase isn't that much, and depends greatly on the number of datasets you have access to. For someone like Jonathan, it will probably boost performance by 20% or so. For others, like me, very little.

aarohinadkarni (approved):

PR #2439, opened 2024-01-18
[DUOS-2823] Add delete modal dialog for accepted DAC datasets

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2823

### Summary

Add delete modal dialog for accepted DAC datasets

----

- Verify all tests go green

### DIFF
--- a/src/components/dac_dataset_table/DACDatasetApprovalStatus.js
+++ b/src/components/dac_dataset_table/DACDatasetApprovalStatus.js
@@ -1,5 +1,5 @@
 import React, {useState} from 'react';
-import {DAC} from '../../libs/ajax';
+import {DAC, DataSet} from '../../libs/ajax';
 import {Link} from 'react-router-dom';
 import {isNil} from 'lodash/fp';
 import Button from '@mui/material/Button';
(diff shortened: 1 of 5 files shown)


src/components/dac_dataset_table/DACDatasetApprovalStatus.js
```diff
@@ -1,14 +1,33 @@
 import React, {useState} from 'react';
-import {DAC} from '../../libs/ajax';
+import {DAC, DataSet} from '../../libs/ajax';
 import {Link} from 'react-router-dom';
 import {isNil} from 'lodash/fp';
 import Button from '@mui/material/Button';
 import ReactTooltip from 'react-tooltip';
 import style from '../../pages/DACDatasets.module.css';
+import { ConfirmationDialog } from '../modals/ConfirmationDialog';
+import IconButton from '@mui/material/IconButton';
```
aarohinadkarni: Can we delete the imports on lines 9 & 10 if they aren't being used?

src/components/dac_dataset_table/DACDatasetApprovalStatus.js
```diff
@@ -1,14 +1,33 @@
...
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  const handleAction = (datasetId) => {
+    setOpen(false);
+    DataSet.deleteDataset(datasetId).then(() => {
+      window.location.reload();
+    });
+  }
```
rushtong: This function should only be called if the dataset is deletable. In the case that it isn't, the user will get an error:
```
{"message":"Dataset is in use and cannot be deleted.","code":400}
```
Look for the  `"deletable": false` property on the dataset to check this condition.

rushtong (changes requested):
Aside from the conditional check needed, the code looks good!

src/components/dac_dataset_table/DACDatasetApprovalStatus.js
fboulnois: I was not aware of this property! Good catch.

aarohinadkarni (approved):
👍

rushtong (approved):
Thank you for the changes 👍🏽

PR #2440, opened 2024-01-18
[DUOS-2882][risk=no] Bug fix for component test

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2882

### Summary
Minor fix for click items that clearly exist, but are for some reason inaccessible without `{force: true}`

----

- Verify all tests go green


aarohinadkarni (approved):
👍

fboulnois (changes requested):
Although it solves the problem, I'm not sure if it's the right solution. According to the documentation, `force` skips the actionable checks Cypress would otherwise apply. In particular, it skips the following checks: https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Actionability

This suggests to me that the button is either hidden, disabled, detached, readonly, animating, or covered by something else and there is a deeper issue going on.

PR #2442, opened 2024-01-19
[DUOS-2811][risk=no] Enable data submitter dataset deletion via the UI on data submissions

aarohinadkarni:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2811

### Summary

Adds a delete confirmation modal for the deletion of data submissions not yet accepted on the Data Submissions tab.

----

- Verify all tests go green


src/pages/researcher_console/DatasetSubmissionsTable.jsx
```diff
@@ -65,11 +68,49 @@ export default function DatasetSubmissionsTable(props) {
...
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  const removeDataset = async (termToDelete) => {
+    const termName = termToDelete.datasetName;
+    const termId = termToDelete.datasetId;
+    let updatedTerm = await DataSet.deleteDataset(termId).then(() => {
+        window.location.reload();
+    });
```
rushtong: This causes a refresh while the dialog is open which is slightly confusing as a user - I wasn't clear on what was happening.

rushtong: In the error case, this doesn't close the dialog - it remains open. The error case happens when a dataset is in use. Look for the `deletable` property on the dataset. On the back end, we query the DAR-Dataset join table to see if the dataset has been requested. If it is in use, we mark this field as false.

src/pages/researcher_console/DatasetSubmissionsTable.jsx
```diff
@@ -89,6 +130,24 @@ export default function DatasetSubmissionsTable(props) {
...
+            className={'glyphicon glyphicon-trash'}
+            onClick={() => handleClick(term)}
+            to={`#`}
+          />
+          <ConfirmationDialog 
+            title="Delete dataset" 
+            openState={open} 
+            close={handleClose}
+            action={() => removeDataset(selectedTerm)}
+            description={`Are you sure you want to delete the dataset named '${selectedTerm.datasetName}'?`}
```
rushtong: Nit: I suggest using the dataset identifier here - names in production are extremely long and will affect the modal display here.

rushtong (commented):
See comments inline. Let's figure out which approach we want to take and consolidate our approaches.

src/components/modals/ConfirmationDialog.js
```diff
@@ -30,4 +30,4 @@ export const ConfirmationDialog = (props) => {
             </DialogActions>
         </Dialog>
     );
-}
+}
```
rushtong: nit: revert this to minimize unnecessary changes

rushtong (approved):
Thanks for this, it looks good! One minor suggestion, then 👍🏽

fboulnois (approved):
looks good 👍

PR #2450, opened 2024-01-29
[DUOS-2807] Add export to terra in DUOS libraries

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2807

### Summary

See #2441 for the original PR. Add export to terra in DUOS libraries. This version is rebased on our latest develop. You will only see the button IF you have access to the dataset in TDR.

Testing this PR requires multiple steps:

- You must use the **same** user for TDR and DUOS
- You must have the `terraUrl` and `tdrApiUrl` variables defined in your configs, for example:

```
  "terraUrl": "https://bvdp-saturn-dev.appspot.com",
  "tdrApiUrl": "https://jade.datarepo-dev.broadinstitute.org",
```

- You must create a snapshot and have `action: share_policy::reader` on it
- You must assign the DUOS ID to the snapshot in the TDR UI

Once these steps are completed, you should be able to click the export to Terra button. After it is clicked, you will need to wait a few seconds until it is replaced with the export icon. Clicking that will start the import process.

If you get the error `Invalid import request. Unable to load snapshot.`, then you need to clear your cookies **for that specific Terra instance**. There's a bug in Terra with storing a ton of OIDC stuff.

----

- Verify all tests go green

### DIFF
--- a/public/config-example.json
+++ b/public/config-example.json
@@ -4,6 +4,7 @@
   "hash": "dev",
   "apiUrl": "http://localhost:8180/api",
   "ontologyApiUrl": "https://ontologyURL.org/",
+  "tdrApiUrl": "https://tdrApiUrl.org/",
   "clientId": "111111111111-11111111111111111111111111111111.apps.googleusercontent.com",
   "errorApiKey": "example",
   "gaId": "",
--- a/public/config-example.json
+++ b/public/config-example.json
@@ -4,6 +4,7 @@
   "hash": "dev",
   "apiUrl": "http://localhost:8180/api",
   "ontologyApiUrl": "https://ontologyURL.org/",
+  "terraUrl": "https://terraURL.org/",
   "tdrApiUrl": "https://tdrApiUrl.org/",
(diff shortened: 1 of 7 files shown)


public/config-example.json
```diff
@@ -4,6 +4,8 @@
   "hash": "dev",
   "apiUrl": "http://localhost:8180/api",
   "ontologyApiUrl": "https://ontologyURL.org/",
+  "terraUrl": "https://terraURL.org/",
+  "tdrApiUrl": "https://tdrApiUrl.org/",
```
rushtong, line 8: We're going to need changes to https://github.com/broadinstitute/terra-helmfile/blob/master/charts/duos/templates/_config.json.tpl for this to work on deployment. Those need to be in before this gets to dev.

rushtong (approved):
I was able to test this successfully 👍🏽 I can envision some additional enhancements like info panels describing which snapshot the export button points to, etc., but this is a great start.

aarohinadkarni (approved):
Looks great! I agree with Greg's comment above, but this seems to be an awesome start.

PR #2451, opened 2024-02-02
[DUOS-2903][risk=no] Update dataset selection in DAR form

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2903

### Summary
* Convert from hyperscript to jsx
* Use new dataset autocomplete API instead of slower search API
  * Note that the new API returns data in a slightly different form since it is not a full dataset object. For that reason, in several places in the code, we look for one of two property names like `id` and `dataSetId` or `identifier` and `datasetIdentifier` due to the different models.
* Update the population and selection of dataset objects when populating the form
* Ensure that we're not querying for invalid/null dataset ids

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/DataAccessRequest.js
+++ b/src/pages/dar_application/DataAccessRequest.js
@@ -52,7 +52,8 @@ const searchDatasets = (query, callback, currentDatasets) => {
         identifier: ds.identifier || ds.datasetIdentifier,
         datasetIdentifier: ds.identifier || ds.datasetIdentifier,
         datasetName: ds.name || ds.datasetName,
-        name: ds.name || ds.datasetName
+        name: ds.name || ds.datasetName,
+        ... ds
       };
     });
     let options = processedDatasets.filter((ds) => !currentDatasetIds.includes(ds.dataSetId)).map(function (item) {
--- a/src/pages/dar_application/DataAccessRequest.js
+++ b/src/pages/dar_application/DataAccessRequest.js
@@ -38,10 +38,11 @@ const autocompleteOntologies = (query, callback) => {
 };
 
 const searchDatasets = (query, callback, currentDatasets) => {
-  const currentDatasetIds = currentDatasets.map((ds) => ds.id || ds.dataSetId);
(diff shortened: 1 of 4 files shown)


aarohinadkarni (approved):
👍

src/pages/dar_application/DataAccessRequest.js
```diff
@@ -50,12 +50,13 @@ const searchDatasets = (query, callback, currentDatasets) => {
...
+    key: ds.id || ds.datasetId,
+    value: ds.id || ds.datasetId,
     dataset: ds,
-    displayText: ds.datasetIdentifier,
-    label: span({}, [
-      span({style: {fontWeight: 'bold'}}, [ds.datasetIdentifier]), ' | ', ds.name]),
+    displayText: ds.identifier || ds.datasetIdentifier,
+    label: <span>
+      <span style={{fontWeight: 'bold'}}>{ds.identifier || ds.datasetIdentifier}</span> | {ds.name || ds.datasetName}
+    </span>
```
fboulnois: The fact that we have `.id`, `.datasetId`, `.identifier`, `.datasetIdentifier`, and also `.dataSetId` is confusing. I think the key and value needs to be `ds.dataSetId`.

fboulnois (approved):
See below:

src/pages/dar_application/DataAccessRequest.js
rushtong: They are different things depending on the object. I agree, this is confusing, but does reflect reality. To make them all the same, we need to modify the API to make that happen, which is also confusing. Another possibility might be to try and pre-process the objects and conform them to a common standard.

rushtong: Mapping the autocomplete response so they look like the old version of a dataset should simplify these conditions.

rushtong: Looking into the broken component test.

fboulnois (approved):
neat idea, 👍

PR #2456, opened 2024-02-07
[DUOS-2901] Add search bar to libraries

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2901

### Summary
Adds search bars to our two data libraries: Researcher Console, Data Library and SO Console, My Datasets.

----

- Verify all tests go green


rushtong:
One detail I'm seeing is that the search bar and the access type filters are over-writing the results without changing the filter values. For example, enter a value in the search bar and it will filter records. Then check one of the access type checkboxes, and the search bar filter is no longer in effect, yet the value still shows in the search bar. A similar effect happens in reverse. I think we need to take into account BOTH the search bar value and the access type checkboxes in the search handler function.

src/components/data_search/DatasetSearchTable.js
```diff
@@ -102,6 +105,13 @@ export const DatasetSearchTable = (props) => {
     history.push(`/dar_application/${darDraft.referenceId}`);
   };
 
+  const handleSearchChange = useCallback((searchTerms) => searchOnFilteredList(
+    searchTerms,
+    datasets,
+    getSearchFilterFunctions().datasetTerms,
+    setFiltered
+  ), [datasets]);
```
rushtong: I think `getSearchFilterFunctions().datasetTerms` needs to be integrated with the `filterHandler` function instead of being handled separately here. This way, the same function that implements the faceted search also implements the free text filter.

fboulnois: I suspect that it should be possible to show / hide search items purely based on the filtered data list, without touching the datasets at all either.

rushtong (approved):
I'm very impressed with this implementation, thank you! 👍🏽

src/components/data_search/DatasetSearchTable.js
```diff
@@ -253,17 +342,60 @@ export const DatasetSearchTable = (props) => {
...
+              </Box>
+              :
+              isEmpty(filtered) ?
+              <Box sx={{
+                display: 'flex',
+                flexDirection: 'column', alignItems: 'center', justifyContent: 'center', height: '100%' }}>
+                <h1>There are no datasets that fit these criteria.</h1>
+              </Box>
+              :
               <CollapsibleTable
```
fboulnois: I think a series of `if` `else` statements here might be more clear.

src/components/data_search/DatasetSearchTable.js
```diff
@@ -253,17 +342,60 @@ export const DatasetSearchTable = (props) => {
...
+              fontSize: '1.5rem'
+            }}
+            onChange={() => filterHandler(null, datasets, '', searchRef.current.value)}
+            ref={searchRef}
+          />
+          <div/>
+          <Box sx={{ display: 'flex', flexDirection: 'row', justifyContent: 'flex-end', paddingLeft: '1em' }}>
+            <Button variant="contained" onClick={clearSearchRef}>
+                Clear Search
+            </Button>
```
fboulnois: Perhaps we could expand this button horizontally a bit (transform by 1.5)

src/components/data_search/DatasetSearchTable.js
```diff
@@ -40,40 +43,121 @@ export const DatasetSearchTable = (props) => {
...
-        }
-        return false;
-      });
-    }
-    setFiltered(newFiltered);
+    const fullQuery = assembleFullQuery(searchTerm, newFilters);
+    const search = async () => {
+      try {
+        await DataSet.searchDatasetIndex(fullQuery).then((filteredDatasets) => {
+          setFiltered(datasets.filter(value => filteredDatasets.some(item => isEqual(item, value))));
```
fboulnois:
```suggestion
          var newFiltered = datasets.filter(value => filteredDatasets.some(item => isEqual(item, value)));
          setFiltered(newFiltered);
```

nit: we can inspect the results a bit more easily this way.

src/components/data_search/DatasetSearchTable.js
```diff
@@ -40,40 +43,121 @@ export const DatasetSearchTable = (props) => {
...
+          'field': 'study'
+        }
+      }
+    ];
+  
+    // do not apply search modifier if there is no searchTerm
+    if (searchTerm !== '') {
+      const searchModifier = [
+        {
+          "multi_match": {
```
fboulnois, line 67: Right now this matches only as a prefix, but this is a good start, so perhaps we can add a ticket to the backlog to investigate how to search for whole phrases.

fboulnois (commented):
Overall looks good, a few minor suggestions:

src/components/data_search/DatasetSearchTable.js
aarohinadkarni: Ticket created: https://broadworkbench.atlassian.net/browse/DUOS-2978

fboulnois (approved):
looks good 👍

PR #2468, opened 2024-02-26
[DUOS-2926] Allow user to change their display name

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2926

### Summary
Allows user to change their display name and seeing the effect without having to logout & login again.

----

- Verify all tests go green

### DIFF
--- a/src/pages/user_profile/UserProfile.js
+++ b/src/pages/user_profile/UserProfile.js
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import { FormField, FormFieldTypes } from '../../components/forms/forms';
 import { PageHeading } from '../../components/PageHeading';
 import { Notification } from '../../components/Notification';
-import { Institution } from '../../libs/ajax';
+import { Institution, User } from '../../libs/ajax';
 import { Storage } from '../../libs/storage';
 import { NotificationService } from '../../libs/notificationService';
 import { Notifications, getPropertyValuesFromUser } from '../../libs/utils';
@@ -12,12 +12,14 @@ import AcceptedAcknowledgements from './AcceptedAcknowledgements';
 import ControlledAccessGrants from './ControlledAccessGrants';
 import ga4ghLogo from '../../images/ga4gh-logo.png';
 import userProfileIcon from '../../images/user-profile.png';
+import {cloneDeep} from 'lodash/fp';
 
 export default function UserProfile(props) {
 
   const [user, setUser] = useState({});
   const [userProps, setUserProps] = useState({});
   const [institutions, setInstitutions] = useState([]);
+  const [name, setName] = useState('')
 
   const [profile, setProfile] = useState({
     profileName: '',
@@ -27,10 +29,26 @@ export default function UserProfile(props) {
 
   const [notificationData, setNotificationData] = useState({});
 
+  const onChange = ({key, value}) => {
+    console.log('key', key);
+    console.log('value', value);
+    setName(value);
+    let newUser = cloneDeep(user);
+    newUser.displayName = value;
+
+    User.update(newUser,newUser.userId).then(() => {
+      Notifications.showSuccess({ text: 'Name updated successfully!' });
+    }, () => {
+      Notifications.showError({ text: 'Some errors occurred, the user\'s name was not updated.' });
+    });
+  };
+
+
   useEffect(() => {
     const init = async () => {
       try {
         const user = Storage.getCurrentUser();
+        console.log('user', user);
         setUser(user);
         setUserProps(getPropertyValuesFromUser(user));
         setProfile({
@@ -38,6 +56,9 @@ export default function UserProfile(props) {
           email: user.email,
           id: user.userId
         });
+        setName(user.displayName)
+        console.log('name', name);
+        console.log('profile', profile);
         const institutions = await Institution.list();
         setInstitutions(institutions);
         setNotificationData(await NotificationService.getBannerObjectById('eRACommonsOutage'));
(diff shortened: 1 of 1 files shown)


src/pages/user_profile/UserProfile.js
```diff
@@ -27,6 +30,24 @@ export default function UserProfile(props) {
 
   const [notificationData, setNotificationData] = useState({});
 
+  const updateRef = ({key, value}) => {
+    setName(value);
+    let newUser = cloneDeep(user);
+    newUser.displayName = value;
+    setUser(newUser);
```
rushtong: When this is called, it changes the user in state. That triggers a re-render of ResearcherStatus. That triggers a refresh to `User.getSOsForCurrentUser()` in useEffect. The end result of this is that every keystroke triggers an unnecessary call to get signing officials for the current user. Digging in to see if I can find a cleaner approach.

src/pages/user_profile/UserProfile.js
```diff
@@ -27,6 +30,24 @@ export default function UserProfile(props) {
...
+
+  const updateName = () => {
+    User.update(user,user.userId).then(() => {
+      Notifications.showSuccess({ text: 'Name updated successfully!' });
+    }, () => {
+      Notifications.showError({ text: 'Some errors occurred, the user\'s name was not updated.' });
+    });
+
+    setUserRoleStatuses(user, Storage);
+  }
```
rushtong: nit: 
```suggestion
  };
```

src/pages/user_profile/UserProfile.js
```diff
@@ -124,12 +146,25 @@ export default function UserProfile(props) {
...
-      defaultValue={profile.profileName}
-      disabled={true}
-    />
+    <div style={{ display: 'flex', alignItems: 'center' }}>
+      <FormField
+        type={FormFieldTypes.TEXT}
+        id='profileName'
+        defaultValue={name}
+        onChange={updateRef}
+        style={{ width: '90%', marginTop: '10px',
```
rushtong: nit: extra space at end of line
```suggestion
        style={{ width: '90%', marginTop: '10px',
```

src/pages/user_profile/UserProfile.js
```diff
@@ -38,6 +59,7 @@ export default function UserProfile(props) {
           email: user.email,
           id: user.userId
         });
+        setName(user.displayName)
```
rushtong: nit: missing semicolon
```suggestion
        setName(user.displayName);
```

src/pages/user_profile/UserProfile.js
rushtong: OK, here's my solution to this problem.
1. Create a new state variable that will not trigger the researcher status re-render: `const [updatedUser, setUpdatedUser] = useState({});`
2. Update that object instead of the user object when calling `updateRef`: 
```
  const updateRef = ({key, value}) => {
    setName(value);
    let newUser = cloneDeep(user);
    newUser.displayName = value;
    setUpdatedUser(newUser);
  };
```
3. In `updateName`, use that variable to make the API call. Also, only call the storage if the response is a sucess:
```
  const updateName = () => {
    User.update(updatedUser, updatedUser.userId).then((response) => {
      setUserRoleStatuses(response, Storage);
      Notifications.showSuccess({ text: 'Name updated successfully!' });
    }, () => {
      Notifications.showError({ text: 'Some errors occurred, the user\'s name was not updated.' });
    });
  };
```

There might be some additional work to ensure that the save button works in the case that the user hasn't changed their name, but this approach at least solves the nested component re-render problem.

aarohinadkarni: I added some logic to ensure we aren't getting an undefined API call if the user hasn't changed their name here.

src/pages/user_profile/UserProfile.js
```diff
@@ -27,6 +31,27 @@ export default function UserProfile(props) {
...
+  const updateRef = ({key, value}) => {
+    setName(value);
+    let newUser = cloneDeep(user);
+    newUser.displayName = value;
+    setUpdatedUser(newUser);
+  };
+
+  const updateName = () => {
+    if (updatedUser.displayName) {
+      User.update(updatedUser, updatedUser.userId).then((response) => {
```
rushtong: I think we're using the wrong API here. This one calls https://github.com/DataBiosphere/consent/blob/38ea7d8e7273fa724c145a74c9a556fefa751954/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java#L208 which is limited to Admin users. I think we want to call https://github.com/DataBiosphere/consent/blob/38ea7d8e7273fa724c145a74c9a556fefa751954/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java#L229 instead.

rushtong: Try this:
```suggestion
      User.updateSelf(updatedUser).then((response) => {
```

src/pages/user_profile/UserProfile.js
```diff
@@ -27,6 +31,31 @@ export default function UserProfile(props) {
 
   const [notificationData, setNotificationData] = useState({});
 
+  const updateRef = ({key, value}) => {
+    setName(value);
+    // let newUser = cloneDeep(user);
+    // newUser.displayName = value;
```
rushtong: nit: rm commented lines

rushtong (approved):
Looks good after comment removal 👍🏽

rushtong (approved):
👍🏽

fboulnois (approved):
👍

PR #2476, opened 2024-03-02
[DUOS-2968][risk=no] ERA Commons Component Refactor

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2968

### Summary
* Convert to a functional component from a class based component
* Refactor the external call to the profile server to parse the token and instead, parse it ourselves
* Note that github (not git) cannot distinguish the changed file from `eRACommons.js` -> `ERACommons.jsx`.

### Testing
This will function normally from:
* Account linking on the user profile page
* Account linking on the data access request page

### Notes
The "Full DAR" tab on a voting page also uses this component (in `read-only` view) but has an existing bug so that a user with a commons id is not displayed. This is proving challenging to fix currently, so will file a new ticket that can be addressed separately from this update.

----

- Verify all tests go green


aarohinadkarni (approved):
looks good! 👍

cypress/component/ERACommons/era_commons.spec.js
```diff
@@ -0,0 +1,109 @@
...
+describe('ERA Commons Utility', function () {
+  it('decodeNihToken works with a valid base64 encoded token', async function () {
+    const token = {'nih-username-token': encodedToken};
+    const decoded = await decodeNihToken(token);
+    expect(decoded).to.not.equal(null);
+    expect(decoded.eraCommonsUsername).to.equal('testing');
+    expect(decoded.iat).to.equal(1709561276);
+    expect(decoded.exp).to.equal(1709564876);
+  });
+  it ('decodeNihToken returns null if the token is invalid', async function () {
```
fboulnois:
```suggestion
  it('decodeNihToken returns null if the token is invalid', async function () {
```

nit: remove extra space

src/components/ERACommons.jsx
```diff
@@ -0,0 +1,187 @@
...
+      if (props.readOnly && props.researcherProfile) {
+        setResearcherProfile(props.researcherProfile);
+        const propsData = researcherProfile.properties;
+        setCommonResearcherPropertyState(propsData, researcherProfile.eraCommonsId);
+      } else {
+        await getUserInfo();
+      }
+    };
+    fetchData();
+  }, []);
```
fboulnois: Is it necessary to have a separate `useEffect` for this functionality?

src/components/ERACommons.jsx
```diff
@@ -0,0 +1,187 @@
...
+    setCommonResearcherPropertyState(updatedUserProps, decodedToken.eraCommonsUsername);
+    document.getElementById('era-commons-id').scrollIntoView(
+      {block: 'start', inline: 'nearest', behavior: 'smooth'}
+    );
+  };
+
+  const redirectToNihLogin = async () => {
+    const destination = window.location.origin + '/' + props.destination + '?nih-username-token=<token>';
+    const nihUrl = `${ await Config.getNihUrl() }?${qs.stringify({ 'return-url': destination })}`;
+    window.location.href = nihUrl;
```
fboulnois, line 98:
```suggestion
    props.history.push(nihUrl);
```

Should we use the React mechanism for redirecting instead?

fboulnois (approved):
A few questions, but otherwise looks good:

src/components/ERACommons.jsx
rushtong: This only redirects internally, it doesn't redirect externally ... i.e., it tacks on the `nihUrl` to the base url which ends up failing. I don't see another mechanism for doing this, but would love to see something other than the `window.location.href` approach

src/components/ERACommons.jsx
rushtong: Great point, I had a ton of `useEffect` problems when testing (infinite reloads, etc), but it turns out we can combine them.

fboulnois (approved):
looks good 👍

PR #2487, opened 2024-03-14
[DUOS-2988] Edit text on DAR form 1.5 Internal Collaborators description

PintoGideon:
### Addresses

- https://broadworkbench.atlassian.net/browse/DUOS-2988  

    image-20240313-040124
  
- https://broadworkbench.atlassian.net/browse/DUOS-2498

  Before:
  
   Screenshot from 2024-03-14 18-00-00

  After:
 
  Screenshot from 2024-03-14 17-50-04
  
  

-   An alternative design I tried for DUOS-2498 was the one shared in the mockup attached to the issue, but opted for the former.

    Screenshot from 2024-03-14 17-47-06

  
  

   
   

----

- Verify all tests go green


src/pages/dar_application/ResearcherInfo.js
```diff
@@ -175,11 +175,7 @@ export default function ResearcherInfo(props) {
             laboratory) who assists with the PI's research project involving controlled-access data 
             subject to the NIH GDS Policy. Internal collaborators are employees of the Requesting 
             PI's institution and work at the same location/campus as the PI. Internal Collaborators 
-            must be at the PI or equivalent level and are required to have a Library Card in order 
-            to access data through this request. Internal Collaborators will have Data 
-            Downloader/Approver status so that they may add their own relevant Internal Lab Staff. 
-            Internal Collaborators will not be required to submit an independent DAR to collaborate 
-            on this project.`
+            must be at the PI or equivalent level and are required to have a Library Card and submit their own data access.`
```
rushtong, line 178: The ticket looks like there should be a link around the text: `and submit their own data access.` but the link is not specified in the ticket, so we should confirm with @solideoglori about what is required here.

rushtong (commented):
We should revert the changes to `package-lock.json`. I have one question about the text change (see comment inline) but otherwise looks good.

src/pages/dar_application/ResearcherInfo.js
solideoglori: Thanks @rushtong - I don't think we should link to 'submit a DAR' here as this is an Signing Official page, and we would rarely if ever expect an SO to submit their own DAR. I think we can proceed as is, but feel free to call out anything I'm missing.

rushtong:
@PintoGideon - I think you need to either rebase this PR onto the latest develop or re-apply your changes to a new copy of develop and start from scratch. There are lots of commits from previously merged PRs here that are confusing the overall PR

PR #2498, opened 2024-03-18
[DUOS-3014] Replace 'Request a Library Card' button with text guidance

PintoGideon:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-3014

### Summary

If the user does not have any library cards, replace the button to request a card with some textual guidance.

----

- Verify all tests go green

### DIFF
--- a/src/Routes.jsx
+++ b/src/Routes.jsx
@@ -65,7 +65,7 @@ const Routes = (props) => (
     <Route path="/nih_dms_policy" component={NIHDMSPolicyInfo} />
     <Route path="/anvil_dms_policy" component={AnVILDMSPolicyInfo} />
     <AuthenticatedRoute path="/profile" component={UserProfile} props={props} rolesAllowed={[USER_ROLES.all]} />
-    <AuthenticatedRoute path="/request_role" component={RequestForm} props={Object.assign({}, props, {isRequestRolePage: true})} rolesAllowed={[USER_ROLES.all]} />
+    <AuthenticatedRoute path="/request_role" component={RequestForm} rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/admin_review_collection/:collectionId" component={DarCollectionReview} props={Object.assign({adminPage: true}, props)} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/admin_manage_users" component={AdminManageUsers} props={props} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/admin_edit_user/:userId" component={AdminEditUser} props={props} rolesAllowed={[USER_ROLES.admin]} />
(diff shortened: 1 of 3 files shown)


src/pages/user_profile/ResearcherStatus.js
```diff
@@ -99,15 +99,10 @@ export default function ResearcherStatus(props) {
...
-          style={{
-            marginTop: '10px',
-            marginBottom: '50px'
-          }} >
-          Request a Library Card
-        </button>
+        <p style={{
+          marginTop: '10px',
+          marginBottom:'50px'
+        }}>You must have a Library Card to submit a data access request. To obtain one, your Institutional Signing Official must register in DUOS, request and receive Signing Official permissions, and issue you a Library Card.</p>
```
rushtong: This leaves the `goToRequestRole` method unused. 
That takes the user to `/request_lc` which now, no longer needs a route in `Routes.js`.
That takes us to `RequestForm` which now no longer uses the prop `isRequestLCPage`
This is essentially the "Request Library Card" feature so this code change makes that code path unreachable now. 
We should potentially do two things here
1. Confirm with @solideoglori that we no longer have a need for this feature
2. If so, remove all of the unreachable code

solideoglori: 1. Yes, we no longer need.
2. Agreed!

aarohinadkarni (approved):
looks good! 👍

src/pages/user_profile/RequestForm.js
```diff
@@ -44,19 +43,6 @@ export default function SupportRequestsPage(props) {
       extraRequest: undefined
     };
   }
```
rushtong: The previous if condition here (`if (props.isRequestRolePage)`) is always true so we don't need that any longer. It was only used to distinguish between the LC request vs Role request.

src/Routes.jsx
```diff
@@ -66,7 +66,6 @@ const Routes = (props) => (
     <Route path="/anvil_dms_policy" component={AnVILDMSPolicyInfo} />
     <AuthenticatedRoute path="/profile" component={UserProfile} props={props} rolesAllowed={[USER_ROLES.all]} />
     <AuthenticatedRoute path="/request_role" component={RequestForm} props={Object.assign({}, props, {isRequestRolePage: true})} rolesAllowed={[USER_ROLES.all]} />
```
rushtong: See related comment in `RequestForm` ... we don't need the `isRequestRolePage` property any more - it is always true so we can simplify both this props assignment and the `RequestForm` component by removing it.

src/pages/user_profile/ResearcherStatus.js
```diff
@@ -10,20 +10,13 @@ export default function ResearcherStatus(props) {
...
   const [hasCard, setHasCard] = useState(true);
 
-  const goToRequestRole = () => {
-    pageProps.history.push({
-      pathname: '/request_lc',
-      state: { data: profile }
-    });
-  };
-
+
```
rushtong: There is a stray space here that eslint is reporting on

rushtong (commented):
See comments inline, I think we can do a little more cleanup here to keep the logic easier to grok.

rushtong (approved):
Looks good, thank you 👍🏽

PR #2508, opened 2024-03-25
[DUOS-3028][risk=no] Address `key` is not a prop warning

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-3028

### Summary
* Minor property rename to address the following warning:
```
Warning: DataUsePill: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://reactjs.org/link/special-props)
    at DataUsePill (https://local.broadinstitute.org:3000/static/js/bundle.js:10234:5)
    at div
    at div
    at DataUseSummary (https://local.broadinstitute.org:3000/static/js/bundle.js:11221:5)
```

----

- Verify all tests go green

### DIFF
--- a/src/components/collection_voting_slab/DataUsePill.jsx
+++ b/src/components/collection_voting_slab/DataUsePill.jsx
@@ -32,10 +32,10 @@ const styles = {
 };
 
 export const DataUsePill = (props) => {
-  const { dataUse, key } = props;
+  const { dataUse, index } = props;
 
   return (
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
good catch 👍

PintoGideon (approved):
Looks good !!

PR #2512, opened 2024-03-27
[DUOS-2462][DUOS-2850] B2C OAuth Integration

sjkobori:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2462
https://broadworkbench.atlassian.net/browse/DUOS-2850

## B2C
 - Add `OidcBroker.ts` for wrapping calls to the `oidc-client-ts` library
 - Adds `Auth.ts` for managing DUOS UI specific session logic and communicating to the OidcBroker
 - Adds `OAuth2.ts` for making api calls to get our oauth config
 - Adds `appLoader.ts`, `RedirectFromOAuth.ts` to handle redirect after signing in through popup window
 - Removes `GoogleIS` and related code
 
## Other
 - As per request, moved the SignIn button from the `Home.jsx` to `DuosHeader.jsx`
 - Sign out calls now delegate through `Auth.signOut`
 - Typescript conversion for files affected by these changes (except for large files requiring their own ticket to convert)


package.json
```diff
@@ -23,6 +23,7 @@
     "mixin-deep": "2.0.1",
     "moment": "2.30.1",
     "noty": "3.2.0-beta",
+    "oidc-client-ts": "^3.0.1",
```
rushtong: nit: please pin this explicitly, we've hit bugs using flexible versioning
```suggestion
    "oidc-client-ts": "3.0.1",
```

src/libs/auth/oidcBroker.ts
```diff
@@ -0,0 +1,93 @@
...
+let userManagerSettings: UserManagerSettings | null = null;
+let userManager: UserManager | null = null;
+
+const generateOidcUserManagerSettings = async (
+  config: OAuthConfig
+): Promise<UserManagerSettings> => {
+  const metadata: Partial<OidcMetadata> = {
+    authorization_endpoint: `${await Config.getApiUrl()}/oauth2/authorize`,
+    token_endpoint: `${await Config.getApiUrl()}/oauth2/token`,
+  };
```
dvoet, line 31: I don't know the right answer here, but I wonder if adding this metadata is required. Does `oidc-client-ts` know how to use the authority config value to lookup the `.well-known/openid-configuration`? If so, we probably don't need to use the custom authorization and token endpoints.

Note that the custom endpoints were ported by me from TDR. I didn't fully understand why it has many of the custom bits. I now realize that it was mostly to be able to support switching back and forth between B2C and Google. All of that is over with and we are B2C all the way. So the custom endpoints don't really provide any value. They are used by the swagger ui but I have figured out a better way that does not require them. 

Bottom line: I think we should work towards removing these endpoints.

dvoet: https://authts.github.io/oidc-client-ts/#md:required-settings

dvoet: looks like our [config url](https://terradevb2c.b2clogin.com/terradevb2c.onmicrosoft.com/v2.0/.well-known/openid-configuration?p=b2c_1a_signup_signin_dev) (in dev) does not have CORS headers set. We could look into setting those or add a new DUOS api to serve the config. It actually already gets and caches the config, there is jut no api.

rtitle: I'm all for removing the oauth2 proxy endpoints if they are not needed anymore. I think another thing the custom endpoints do is inject the clientId as a scope and add `?prompt=login` to the `authorize` request. See:

* https://github.com/broadinstitute/terra-helmfile/blob/fa1dabdd92833cb4e736e7203d7ef6d0e315b3a3/charts/consent/templates/_consent.yaml.tpl#L69-L73
* https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/OidcService.java#L33-L36

There is probably a way to set these in the `oidc-client-ts` frontend instead.

src/libs/auth/oidcBroker.ts
```diff
@@ -0,0 +1,93 @@
...
+  return {
+    authority: config.authorityEndpoint,
+    client_id: config.clientId,
+    popup_redirect_uri: `${window.origin}/redirect-from-oauth`,
+    silent_redirect_uri: `${window.origin}/redirect-from-oauth-silent`,
+    metadata,
+    prompt: 'consent login',
+    scope: 'openid email profile',
```
rtitle, line 39: Re my above comment: I see `prompt=login` is already set here. We'd need to figure out how to set the clientId as a scope if not using the custom oauth2 endpoints. (Or validate that this is still a requirement in B2C?)

src/libs/auth/oidcBroker.ts
```diff
@@ -0,0 +1,93 @@
...
+    return oidcStorage !== null ? User.fromStorageString(oidcStorage) : null;
+  },
+  
+  signIn: async (popup: boolean, args?: SigninPopupArgs | SigninSilentArgs): Promise<User | null> => {
+    const um: UserManager = OidcBroker.getUserManager();
+    return popup ? await um.signinPopup(args) : await um.signinSilent(args);
+  },
+  
+  signOut: async (): Promise<void> => {
+    await OidcBroker.getUserManager().removeUser();
```
rtitle: Just wondering if we have to clear any storage upon signout, or does `removeUser` take care of it?

(Just remembering that terra-ui had to clear several things upon signOut, but that might have been Terra UI's custom state store.)

rtitle (approved):
Overall it looks good to me!

I do think it's an interesting exploration whether the backend oauth2 proxy endpoints are truly needed anymore..

src/libs/auth/oidcBroker.ts
sjkobori: It is probably worth throwing in a `clearStaleState` (https://authts.github.io/oidc-client-ts/classes/UserManager.html#clearStaleState) here just to clean out anything else that may be in the store. The specific `duos-ui` state is cleared out in `Auth.signOut`. Good callout 👍

src/components/SignInButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+          description={errorInfo.description}
+        />
+      </div>
+    );
+  };
+
+  return <div>{!showError ? signInButton() : errorAlert(errorInfo)}</div>;
+};
+
+export default SignInButton;
```
Ghost-in-a-Jar, line 137: Just for future pr's it is nice for these sorts of js -> ts conversions to have inline pr comments on the new file for reviewers of what you changed relative to the js file since there is no direct diff shown in github. I just opened up both side by side and did the comparison for this pr.

Ghost-in-a-Jar (approved):
This looks good overall, I would definitely smoke test this thoroughly and expect there to maybe be some unforeseen behavior/bugs given the changes. In particular make sure that the background token expiration/refresh works as expected etc. I think you set yourself up well with logging and metrics collection.

src/Routes.jsx
```diff
@@ -50,12 +49,6 @@ const Routes = (props) => (
     <Route exact path="/home" render={(routeProps) => <Home {...routeProps} {...props} />} />
     <Route exact path="/status" render={(routeProps) => <Status {...routeProps} {...props} />} />
     <Route exact path="/liveness" component={HealthCheck} />
-    <Route exact path="/backgroundsignin" render={
-      (routeProps) =>
-        checkEnv(envGroups.NON_STAGING)
-          ? <BackgroundSignIn {...routeProps} />
-          : <NotFound />
-    } />
```
rushtong: This page is what we use for E2E tests so we didn't have to manually trigger an auth dance. Currently, they're not run on PRs or in GHA, but we can run it locally. We either need a B2C version of this, or bring it back so we can test authenticated pages.

src/pages/TermsOfServiceAcceptance.jsx
```diff
@@ -19,6 +20,9 @@ export default function TermsOfServiceAcceptance(props) {
 
   const acceptToS = useCallback(async () => {
     await TosService.acceptTos();
+    // TODO: Determine if this is the correct user flow
+    // Can the user only accept ToS once the have signed in?
+    // What state is the user in after signIn, but before accepting ToS?
```
rushtong, line 25: This is the correct flow. Yes, the user can only accept ToS after signing in - otherwise we don't know who they are and could not register their ToS acceptance in Sam. If they sign in, but have not yet accepted ToS, they are directed to this page where their only choice is to accept or reject.

src/libs/auth/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+      //TODO: Add metrics for user loaded
+    });
+    um.events.addAccessTokenExpiring((): void => {
+      //TODO: Add an alert that session will expire soon
+      console.log('accessTokenExpiring');
+    });
+    um.events.addAccessTokenExpired((): void => {
+      Auth.signOut();
+      //TODO: Add an alert that session has expired
+    });
```
rushtong: Let's either resolve these todos or add tickets for them

src/libs/auth/oidcBroker.ts
dvoet: Extra parameters can be set with [the extraQueryParams config setting](https://authts.github.io/oidc-client-ts/interfaces/OidcClientSettings.html#extraQueryParams).

`prompt=login` is good but I don't know if the client id is required. In my experimentation with the swagger UI's oidc integration it doesn't seem to be.

src/libs/ajax.js
```diff
@@ -1,15 +1,15 @@
...
 import {StackdriverReporter} from './stackdriverReporter';
-import {Storage} from './storage';
 import axios from 'axios';
 
 //define axios interceptor
 //to log out user and redirect to home when response has 401 status
 //return responses with statuses in the 200s and reject the rest
 const redirectOnLogout = () => {
-  Storage.clearStorage();
+  Auth.signOut();
```
rushtong, line 12: I like how the storage piece is moved into `Auth.signOut()`

src/components/SignInButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export const SignInButton = (props: SignInButtonProps) => {
+  const [errorInfo, setErrorInfo] = useState<ErrorInfo>({});
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [showError, setShowError] = useState<boolean>(false);
+  const { customStyle } = props;
+  const history = useHistory();
+
+  // Utility function called in the normal success case and in the undocumented 409 case
```
rushtong: We should look more into this - I think with the update to v2 ToS APIs, the 409 case is no longer a thing.

.eslintrc
```diff
@@ -51,6 +51,16 @@
...
+  },
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx"],
+      "rules": {
+        "no-undef": "off"
+      }
+    }
+  ]
 }
```
rushtong: Side note - I'd really like this kind of thing to be in a separate PR so we can benefit from it sooner.

cypress/component/SignIn/sign_in.spec.js
```diff
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import { mount } from 'cypress/react';
-import SignIn from '../../../src/components/SignIn';
+import SignInButton from '../../../src/components/SignInButton';
 import { Config } from '../../../src/libs/config';
 
 const signInText = 'Sign-in';
```
rushtong: This text is no longer accurate - the first test should pass with 
```suggestion
const signInText = 'Sign In';
```

cypress/component/SignIn/sign_in.spec.js
```diff
@@ -17,14 +17,14 @@ describe('Sign In Component', function() {
...
       cy.contains(signInText).should('exist');
     });
   });
   it('Spinner loads when client id is empty', function () {
     cy.viewport(600, 300);
     cy.stub(Config, 'getGoogleClientId').returns('');
-    mount(<SignIn />);
+    mount(<SignInButton />);
     cy.contains(signInText).should('not.exist');
   });
```
rushtong: I don't think we need this test any longer - there is longer a spinner that loads in place of the sign in text during the google client id loading time.

.eslintrc
```diff
@@ -51,6 +51,16 @@
     "@babel/implicit-arrow-linebreak": "off",
     "@babel/array-bracket-newline": "off",
     "@babel/brace-style": "off",
-    "@babel/semi": "warn"
-  }
+    "@babel/semi": "warn",
+    // `any` is useful for incremental type improvements during the transition from JS to TS.
+    "@typescript-eslint/no-explicit-any": "off",
```
nawatts: Depending on how often an explicit `any` is desired, consider disabling this for specific lines or files instead of in global ESLint config. That way, it increases the friction to adding `any` and makes it easier to ratchet up type safety while incrementally removing `any` later.

package.json
nawatts: For what it's worth, even with the range selector, an exact version would be pinned in `package-lock.json`.

src/components/DuosHeader.jsx
```diff
@@ -226,6 +228,13 @@ const NavigationTabsComponent = (props) => {
           };
         </div>
         {/* Navbar right side */}
+        {!isLogged && (
+          <div
+            style={{ display: 'flex', alignItems: 'center', flexDirection: orientation === 'vertical' ? 'column' : 'row' }}
+          >
+            <SignInButton customStyle={{ whiteSpace: 'nowrap',  }}></SignInButton>
```
nawatts: Nit:

```suggestion
            <SignInButton customStyle={{ whiteSpace: 'nowrap',  }} />
```

src/components/SignInButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+import loadingImage from '../images/loading-indicator.svg';
+import { Auth } from '../libs/auth/auth';
+import eventList from '../libs/events';
+import { StackdriverReporter } from '../libs/stackdriverReporter';
+import CSS from 'csstype';
+import { useHistory } from 'react-router';
+import { OidcUser } from 'src/libs/auth/oidcBroker';
+
+interface SignInButtonProps {
+  customStyle: CSS.Properties | undefined;
```
nawatts: Nit: suggest naming this prop `style` so it's more intuitive.

```suggestion
  style: CSS.Properties | undefined;
```

src/components/SignInButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+    const user: DuosUser = await User.getMe();
+    if (!user.roles) {
+      await StackdriverReporter.report(
+        'roles not found for user: ' + user.email
+      );
+    }
+    setUserRoleStatuses(user, Storage);
+    const userStatus = await ToS.getStatus();
+    const { tosAccepted } = userStatus;
+    if (!isEmpty(userStatus) && !tosAccepted) {
```
nawatts: If `userStatus` was empty, wouldn't the destructuring in `const { tosAccepted } = userStatus;` throw an error?

When would we expect `userStatus` to be empty?

src/components/SignInButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+    }
+  };
+
+  const getRedirectTo = (): string => {
+    const queryParams = new URLSearchParams(window.location.search);
+    return queryParams.get('redirectTo') || window.location.pathname;
+  };
+
+  const shouldRedirectTo = (page: string): boolean =>
+    page !== '/' && page !== '/home';
```
nawatts: Since they don't depend on state, etc, these functions look like they could be moved out of the component.

Consider doing that wherever possible to simplify the component and to avoid recreating functions on every render.

src/components/SignInButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+          id='dialog'
+          type='danger'
+          title={errorInfo.title}
+          description={errorInfo.description}
+        />
+      </div>
+    );
+  };
+
+  return <div>{!showError ? signInButton() : errorAlert(errorInfo)}</div>;
```
nawatts: Nit: flipping the branches allows removing the `!`, reducing the amount of syntax a reader has to parse.

```suggestion
  return <div>{showError ? errorAlert(errorInfo) : signInButton()}</div>;
```

src/libs/ajax/OAuth2.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Config } from "../config";
+
+export interface OAuthConfig {
+  clientId: string;
+  authorityEndpoint: string;
+}
+
+export const OAuth2 = {
+  getConfig: async (): Promise<OAuthConfig> => getConfig(),
+};
```
nawatts, line 11: For what it's worth, as an alternative to namespace objects, modules could export individual methods and consumers could:

```ts
import * as OAuth2 from 'path/to/OAuth2';
```

if they really want the namespace.

src/libs/auth/RedirectFromOAuth.ts
```diff
@@ -0,0 +1,22 @@
...
+  const url = window.location.href;
+  const isSilent = window.location.pathname.startsWith(
+    '/redirect-from-oauth-silent'
+  );
+
+  if (isSilent) {
+    userManager.signinSilentCallback(url);
+  } else {
+    userManager.signinPopupCallback(url);
+  }
```
nawatts: This looks like it belongs in a `useEffect` hook instead of in `render`.

But does this need to be in a React component at all? It looks like `oauth-redirect-loader` could call the `UserManager` functions directly (and render a Spinner if desired).

src/libs/auth/RedirectFromOAuth.ts
```diff
@@ -0,0 +1,22 @@
...
+  return Spinner;
```
nawatts:
```suggestion
  return <Spinner />;
```

src/libs/auth/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+    } else {
+      await Auth.signOut();
+    }
+  },
+
+  signIn: async (
+    popup: boolean,
+    onSuccess?: (response: OidcUser) => Promise<void> | void,
+    onFailure?: (response: any) => Promise<void> | void
+  ): Promise<void> => {
```
nawatts: Since this is already returning a Promise, consider leaning into that pattern instead of mixing it with callbacks.

Consumers could handle success or failure with `signIn(...).then(onSuccess, onFailure)`.

src/libs/auth/oidcBroker.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export interface OidcUser extends User {
+  profile: B2cIdTokenClaims;
+}
+
+type OidcUserManager = UserManager;
+
+let config: OAuthConfig | null = null;
+let userManagerSettings: UserManagerSettings | null = null;
+let userManager: UserManager | null = null;
```
nawatts, line 23: A possible alternative to global state could be to have an OidcBroker class and have this module export a singleton instance of it?

src/components/SignInButton.tsx
rushtong: > When would we expect `userStatus` to be empty?

I believe `userStatus` is one of the flavors of the Sam user info object. We collect that on every back end API call as part of the DUOS user to determine ToS acceptance.

package.json
rushtong: > For what it's worth, even with the range selector, an exact version would be pinned in `package-lock.json`.

What we've seen is that when depdendabot PRs regenerate `package-lock.json` it can pull in newer versions.

rushtong (changes requested):
See comments inline - we need at least one test fix and the background sign-in feature replaced. Outside of that, this works as expected. I've been able to test standard OAuth with google users and OAuth with an azure account.

src/components/SignInButton.tsx
sjkobori: Yeah, I definitely agree it is hard to tell the difference easily. In hindsight, I should have done this TS conversion in a separate PR and then made the b2c changes in this one.

src/libs/auth/auth.ts
sjkobori: New tickets:

Metrics:
https://broadworkbench.atlassian.net/browse/DUOS-3072
Session Alerts:
https://broadworkbench.atlassian.net/browse/DUOS-3082

cypress/component/SignIn/sign_in.spec.js
sjkobori: I'll update this test with the new `SignInButton` functionality.

src/libs/auth/RedirectFromOAuth.ts
sjkobori: I don't think this needs to be in a React component since this should only happen once each time the user signs in through the popup window, and then I believe that app instance ends once that window closes.

src/components/SignInButton.tsx
sjkobori: That makes sense, I did not consider that pure functions in a React component would get recreated. I guess that would be a reason to make a React class component instead, but I do prefer moving these to their own module.

src/libs/auth/RedirectFromOAuth.ts
sjkobori: Seems like spinner was not a proper React component. Since it was only used in one other place, I converted it to one in order to make this change.

src/pages/TermsOfServiceAcceptance.jsx
sjkobori: That makes sense. I was trying to determine if the ToS accept button should set the `setUserIsLogged` to true, since by signing in, userIsLogged should already be set to true. With this context, I believe the ToS accept button should not change the userIsLogged state.

rushtong: Correct - `userIsLogged` state is separate and distinct from ToS acceptance.

src/libs/auth/RedirectFromOAuth.ts
nawatts: Oh. I assumed it was based on the name. But if it was an element instead, that change will likely break other places it was used.

fboulnois (approved):
I was able to run the PR after correctly making certain dependencies explicit. After limited testing, I was able to log in and out of my Google and Azure accounts and verify that the sign in button was in the login bar.

sjkobori:
https://github.com/DataBiosphere/duos-ui/pull/2611 Will continue this work as this PR too large to easily track the changes.

PR #2514, opened 2024-03-27
[DUOS-2986] Allow DUOS-000000 in URLs for datasets

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2986

### Summary

Allow DUOS-000000 in URLs for datasets. BONUS: Enables a dataset short link which lives at duos.org/duos-000000

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -359,8 +359,8 @@ export const DataSet = {
 
   getDatasetByDatasetIdentifier: async datasetIdentifier => {
     const url = `${await getApiUrl()}/api/tdr/${datasetIdentifier}`;
-    const res = await fetchOk(url, Config.authOpts());
-    return await res.json();
+    const res = await axios.get(url, Config.authOpts());
+    return res.data;
   }
 };
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -356,6 +356,12 @@ export const DataSet = {
     const url = `${await getApiUrl()}/api/dataset/study/${studyId}`;
     return await axios.put(url, studyObject, Config.multiPartOpts());
   },
+
+  getDatasetByDatasetIdentifier: async datasetIdentifier => {
+    const url = `${await getApiUrl()}/api/tdr/${datasetIdentifier}`;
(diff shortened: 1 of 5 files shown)


src/libs/ajax.js
```diff
@@ -356,6 +356,12 @@ export const DataSet = {
     const url = `${await getApiUrl()}/api/dataset/study/${studyId}`;
     return await axios.put(url, studyObject, Config.multiPartOpts());
   },
+
+  getDatasetByDatasetIdentifier: async datasetIdentifier => {
+    const url = `${await getApiUrl()}/api/tdr/${datasetIdentifier}`;
+    const res = await fetchOk(url, Config.authOpts());
```
rushtong: The fetchOK method will only return if the response is successful. In the case of a 404, for instance, it errors out and the page doesn't display anything. Can we add some logic in `DatasetStatistics` to capture that? You could call `fetchAny` instead and manually handle the non-success case in the UI. Separate from that, we've been trying to migrate to axios (one of our many transitions) so it might be cleaner using that approach.

src/pages/DatasetStatistics.jsx
```diff
@@ -9,15 +9,25 @@ import { formatDate } from '../libs/utils';
...
+  const datasetIdentifier = extractIdentifier(props.match.params);
   const [dataset, setDataset] = useState();
   const [dars, setDars] = useState();
   const [isLoading, setIsLoading] = useState(true);
 
   useEffect(() => {
-    setData(datasetId);
-  }, [datasetId]);
+    DataSet.getDatasetByDatasetIdentifier(datasetIdentifier).then((dataset) => {
+      setData(dataset.dataSetId);
```
PintoGideon, line 28: Fails silently for incorrect dataset identifiers. We could add a notification for an error here for consistency, though I doubt this error state would be triggered as the user navigates to this from the dataset catalog.

PintoGideon (approved):
Manually tested. Works as expected apart from a minor issue of silent errors.

rushtong (approved):
Thank you for the changes 👍🏽

PR #2522, opened 2024-04-03
[DUOS-3006] Improve Data Use in the Data Library

PintoGideon:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-3006

### Summary

- Include both primary and secondary data use terms in the Data Library.
- Refactor the TranslatedDulModal to display terms from both the Data Library and Dataset Catalog.
- Tweak the Dataset Catalog to ensure the data use prop is consistent for both components.

----

- Verify all tests go green


rushtong (approved):
This looks reasonable to me and works as expected 👍🏽

src/components/data_search/DatasetSearchTable.js
```diff
@@ -351,81 +370,91 @@ export const DatasetSearchTable = (props) => {
...
       </Box>
-    </Box>
+      {
+        showTranslatedDULModal &&
+        <TranslatedDulModal
+          showModal={showTranslatedDULModal}
+          dataUse={dataUse}
+          onCloseRequest={()=>setShowTranslatedDULModal(false)}
+        />
+      }
```
fboulnois, line 456: Since we are using material UI on the newer pages, I think we should switch this modal to follow material UI or create a new material UI modal for the translated DUL.

fboulnois (commented):
See below:

src/components/data_search/DatasetSearchTable.js
PintoGideon: Sure @fboulnois , I can update the code to use material ui.

PintoGideon: @fboulnois, I have added an UpdatedBaseModal component using the modal from @material-ui. I have kept the appearance of the modal consistent with the older design.

aarohinadkarni (approved):
looks good! 👍

src/components/data_search/DatasetSearchTable.js
fboulnois: I think we want to keep the styling consistent between the data library and the modals. For example, see `modals/ConfirmationDialog.js` for how this could be implement using only Material UI. I have a PR which uses that modal here: https://github.com/DataBiosphere/duos-ui/pull/2439/files

One problem with the older modals is they do not respect z-ordering, and are thus often inaccessible or unreachable if another modals or the header is at the top of the screen.

PintoGideon: @fboulnois  I made some updates. Let me know what you think.

fboulnois: Looks great, thank you! 👍

fboulnois (approved):
looks great 👍

PR #2511, opened 2024-03-26
[DUOS-2974] Refactor the Duos Header Component

PintoGideon:
### Addresses

https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?selectedIssue=DUOS-2974

### Summary

- Remove the usage of react-hyperscript-helpers
- Convert from class component in DuosHeader to function component
- Rename the file extension from js to jsx
----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2523, opened 2024-04-03
[DUOS-3021][risk=no] Handle null roles better

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-3021

### Summary
* Minor fix for handling the case where an existing user does not have any roles assigned.

See also this back end PR: https://github.com/DataBiosphere/consent/pull/2279

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2524, opened 2024-04-04
[DUOS-3038][risk=no]Added divKey to pass key to notification component

sjkobori:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-3038

### Summary
This react warning was introduced in https://github.com/DataBiosphere/duos-ui/pull/2511 . This PR fixes that warning and passes the `key` value to the child components through `divKey`.
image

### DIFF
--- a/src/components/DuosHeader.jsx
+++ b/src/components/DuosHeader.jsx
@@ -365,7 +365,7 @@ const DuosHeader = (props) => {
   };
 
   const makeNotifications = () => {
-    return state.notificationData.map((d, index) => <Notification notificationData={d} key={index} />);
+    return state.notificationData.map((d, index) => <Notification notificationData={d} key={index} divKey={index} />);
   };
 
   const toggleDrawer = (boolVal) => {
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
Thank you for catching this 👍

rushtong (approved):
Thank you - I also saw this and had a similar fix in https://github.com/DataBiosphere/duos-ui/pull/2523.

src/components/DuosHeader.jsx
```diff
@@ -365,7 +365,7 @@ const DuosHeader = (props) => {
   };
 
   const makeNotifications = () => {
-    return state.notificationData.map((d, index) => <Notification notificationData={d} key={index} />);
+    return state.notificationData.map((d, index) => <Notification notificationData={d} key={index} divKey={index} />);
```
fboulnois, line 368:
```suggestion
    return state.notificationData.map((d, index) => <Notification notificationData={d} key={index} index={index} />);
```

I like Greg's approach in the other PR.

fboulnois (commented):
See below

src/components/DuosHeader.jsx
sjkobori: Ah, I just saw this after merging :/. Will update

PR #2538, opened 2024-04-12
[DUOS-3065][risk=no] Split up ajax and ran linter

sjkobori:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-3065

### Summary
In order to make it easier to add to and maintain the ajax functions, each ajax function namespace has been put into their own module.


src/libs/ajax/DatasetMetrics..js
```diff
@@ -0,0 +1,11 @@
+import { Config } from '../config';
+import { getApiUrl, fetchOk } from '../ajax';
+
+
+export const DatasetMetrics = {
```
sjkobori: This was originally `Metrics` but that term was overloaded with the API calls to Bard, so I renamed it. If there is a more accurate name, please let me know.

src/libs/ajax/DatasetMetrics..js
okotsopoulos, line 1: Is this filename correct? `DatasetMetrics..js`

sjkobori: Thank you.. not sure why the import resolver didnt catch that

fboulnois:
Which linter did you run? I see there are some discrepancies with our current code style. Probably not something we need to fix in this ticket, but it may be worth a future ticket.

okotsopoulos (dismissed):
You made the right call splitting this out into its own PR and ticket.  Thanks for taking this on!

Maybe OOS: that's a lot of old lint that you gleaned.  It's great to have it removed but when we glean lint periodically rather than within the relevant changes we can lose some helpful `git blame` context. I'm of the mind that we should automate this as a PR check, here's how TDR UI does it: https://github.com/DataBiosphere/jade-data-repo-ui/actions/workflows/lint.yml

src/libs/ajax/DAC.js
```diff
@@ -0,0 +1,82 @@
+import * as fp from 'lodash/fp';
```
fboulnois, line 1: Now that these are all in separate modules, it should make it easier to test. Maybe we can add a ticket in the backlog to test these?

okotsopoulos (commented):
Ah, given @fboulnois ' comment above around possible linter discrepancy, I'm going to hold back on my approval until that gets hashed out.

src/pages/Status.jsx
```diff
@@ -64,6 +64,6 @@ class Status extends Component {
     );
   }
 
-};
+}
```
fboulnois: This is one of those rules that's linter specific. I'm curious, which linter did you use? I don't have a strong opinion either way, as long as it's consistent. Probably not worth doing in this ticket, but we can add a ticket to the backlog.

fboulnois (approved):
Looks good, some comments:

src/pages/Status.jsx
sjkobori: I had ran this `npx eslint . --ext .js,.jsx,.ts,.tsx --ignore-pattern \"/cypress/*\" --fix` so I believe it is running the config from duos-ui's `.eslintrc`

src/libs/ajax/DAC.js
sjkobori: What kind of tests are you thinking? Would these be unit tests to verify that the requests are in the shape we are expecting (since I believe that is/should be the only logic in these ajax functions)? Or are you thinking of something else?

src/pages/Status.jsx
rushtong: I use the eslint tool in Intelli-j which also enforces this rule. We do have linting rules - I'm surprised that different tools honor those differently

rushtong:
Similar to some of the linting comments, I suggest reverting changes that are unrelated such as the white space changes to `dataset_utils.spec.js` and the quote change to `dac_dataset_table.spec.js`, for example.

src/libs/ajax/DAC.js
rushtong: In our Java apps, we use MockServer to mock the potential responses. I'm not familiar with a JS equivalent, but we should look into it.

rushtong: Another backlog ticket related to this effort is to finish migrating the `fetch` calls to `axios`.

aarohinadkarni (approved):
looks good, thanks for this! 👍

rushtong (approved):
Thank you for the changes! 👍🏽

sjkobori:
@okotsopoulos To your comment about running the linter as part of a workflow, heavy +1 from me.

PR #2550, opened 2024-04-23
[DUOS-3051][DCJ-94][risk=no] Fix bucket data use logic

rushtong:
### Addresses
* https://broadworkbench.atlassian.net/browse/DUOS-3051
* https://broadworkbench.atlassian.net/browse/DCJ-94

### Summary
Fixes a bug where the bucketing logic would choose an incorrect data use for a bucket when using lodash fp's `findIndex` method. For example, in the production scenario, these are all of the data uses:
```
[
  {"hmbResearch": true, "other": "Samples and information may not be sold for profit."},
  {"generalUse": true},
  {"hmbResearch": true},
  {"generalUse": true},
  {"hmbResearch": true}
]
```
When looking for `{"hmbResearch": true}`, the index returned was `0` which is incorrect. I believe the `findIndex` method was only looking at the first element of the the term instead of fully matching the term against the array elements. The fix here is to use the native `Array.findIndex` method with a full object equality check which returns the correct index, `2`.

Additionally, this PR removes the outdated `otherRestrictions` data use condition which we no longer use.

----

- Verify all tests go green

### DIFF
--- a/src/utils/BucketUtils.js
+++ b/src/utils/BucketUtils.js
@@ -272,10 +272,9 @@ const processV3Abstain = (matchResults) => {
  * @returns boolean
  */
 const isOther = (dataUse) => {
-  const otherRestrictions = getOr(false)('otherRestrictions')(dataUse);
   const primaryOther = !isEmpty(getOr('')('other')(dataUse));
   const secondaryOther = !isEmpty(getOr('')('secondaryOther')(dataUse));
-  return otherRestrictions || primaryOther || secondaryOther;
+  return primaryOther || secondaryOther;
 };
 
 /**
--- a/src/utils/BucketUtils.js
+++ b/src/utils/BucketUtils.js
@@ -2,7 +2,6 @@ import {
   any,
   compact,
   filter,
-  findIndex,
   flatMap,
   flow,
   forEach,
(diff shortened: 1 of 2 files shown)


src/utils/BucketUtils.js
```diff
@@ -99,7 +98,7 @@ export const binCollectionToBuckets = async (collection, dacIds = []) => {
 
     // Step 1.b: Populate translated dataUses
     if (dataset.dataUse) {
-      const index = findIndex(dataset.dataUse)(rawDataUses);
+      const index = rawDataUses.findIndex(du => du === dataset.dataUse);
```
fboulnois, line 101: I don't know how you feel about this generally, but my preference is to use the ES6+ built-ins instead of `lodash` instead precisely for these reasons.

fboulnois (approved):
good catch 👍

src/utils/BucketUtils.js
rushtong: > I don't know how you feel about this generally, but my preference is to use the ES6+ built-ins instead of `lodash` instead precisely for these reasons.

I'm starting to lean in that same direction - it's making more sense.

fboulnois (approved):
👍

aarohinadkarni (approved):
nice catch 👍

PR #2557, opened 2024-04-29
[DCJ-260][risk=no] Converted User.js -> User.ts

sjkobori:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-260 Typescript Conversion
~~https://broadworkbench.atlassian.net/browse/DCJ-332 Unit tests~~ Moved to https://github.com/DataBiosphere/duos-ui/pull/2610
### Summary
- Converts User.js to TypeScript as well as their corresponding types.
- ~~Adds Tests for User methods.~~


src/libs/ajax/User.ts
```diff
@@ -3,31 +3,41 @@ import { cloneDeep, flow, unset } from 'lodash/fp';
...
     try {
-      const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), Config.jsonBody(user), { method: 'POST' }]));
+      const res = await fetchOk(
+        url,
+        fp.mergeAll([
+          Config.authOpts(),
+          Config.jsonBody(request),
+          { method: 'POST' },
+        ])
+      );
```
fboulnois, line 39: We should create a ticket in the backlog to use Axios for all the calls.

src/types/model.ts
```diff
@@ -0,0 +1,187 @@
+import { DuosUserResponse } from './responseTypes';
```
fboulnois, line 1: I wonder if there's a tool we can use to generate these automatically. Future investigation?

fboulnois (commented):
Looks reasonable minus tests.

src/types/model.ts
sjkobori: I would also imagine there are tools to generate the swagger doc from Springboot as well, so it might be a way we can make it consistent across the board.

src/libs/ajax/User.ts
sjkobori: https://broadworkbench.atlassian.net/browse/DCJ-278 this is for this file specifically

package.json
```diff
@@ -80,6 +80,7 @@
     "react-scripts": "5.0.1",
     "source-map-explorer": "2.5.3",
     "typescript": "4.9.5",
+    "typescript-eslint": "^7.8.0",
```
rushtong: nit: please pin the version explicitly
```suggestion
    "typescript-eslint": "7.8.0",
```

src/libs/ajax/User.ts
```diff
@@ -3,31 +3,41 @@ import { cloneDeep, flow, unset } from 'lodash/fp';
...
 
-  getById: async (id) => {
+  getById: async (id: number): Promise<DuosUserResponse> => {
     const url = `${await getApiUrl()}/api/user/${id}`;
     const res = await axios.get(url, Config.authOpts());
     return res.data;
   },
 
-  list: async (roleName) => {
+  //TODO: Determine what are legal roleNames (swagger says only these two)
```
rushtong: This is correct: https://github.com/DataBiosphere/consent/blob/9b64b3eab319c694a200f89ae2f6a62faa1efc0f/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java#L75

src/types/model.ts
rushtong: I think you're thinking of https://swagger.io/tools/swagger-codegen/
We are using Dropwizard on the back end, so no spring in this particular stack.

src/types/model.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export type UserRoleName =
+  | 'Admin'
+  | 'Chairperson'
+  | 'Member'
+  | 'Researcher'
+  | 'Alumni'
+  | 'SigningOfficial'
+  | 'DataSubmitter'
+  | 'All';
```
rushtong, line 11: Technically, we don't have a back end role for `'All'`, but it might be helpful in this context if all means 
> Any role is allowed to access this resource

src/types/model.ts
```diff
@@ -0,0 +1,187 @@
...
+  study: Study;
+  active: boolean;
+  needsApproval: boolean;
+  alias: string;
+  datasetIdentifier: string;
+  objectId: string;
+  /**
+   * TODO: Determine what type this is
+   */
+  dataUse: unknown;
```
rushtong: This is here: https://consent-ontology.dsde-dev.broadinstitute.org/#/Data%20Use/get_schemas_data_use_v4

src/libs/ajax/User.ts
```diff
@@ -68,46 +92,63 @@ export const User = {
...
-    const res = await fetchOk(url, fp.mergeAll([Config.authOpts(), { method: 'GET' }]));
+    const res = await fetchOk(
+      url,
+      fp.mergeAll([Config.authOpts(), { method: 'GET' }])
+    );
     return res.json();
   },
 
-  getUnassignedUsers: async () => {
+  getUnassignedUsers: async (): Promise<DuosUserResponse> => {
```
aarohinadkarni: Should this be returning a list?
```suggestion
  getUnassignedUsers: async (): Promise<DuosUserResponse[]> => {
```

src/libs/ajax/User.ts
```diff
@@ -116,9 +157,10 @@ export const User = {
     const res = await axios.post(url, keys, Config.authOpts());
     return res.data;
   },
-  getApprovedDatasets: async () => {
+
+  getApprovedDatasets: async (): Promise<ApprovedDataset> => {
```
aarohinadkarni: Should this also be returning a list?
```suggestion
  getApprovedDatasets: async (): Promise<ApprovedDataset[]> => {
```

aarohinadkarni (commented):
looks good besides tests & two possible changes 👍

src/libs/ajax/User.ts
sjkobori: yes, good catch 👍

src/libs/ajax/User.ts
sjkobori: Yep, I missed this one too 👍

src/libs/ajax/User.ts
```diff
@@ -0,0 +1,168 @@
...
+    // This below code does not seem to work at all and
+    // does not seem appropriate for this request anyway.
+    // The UpdateDuosUserRequestV2 is not the same shape as a DuosUser
+    // like this flow suggests.
+    const filteredUser = flow(
+      cloneDeep,
+      unset('updatedUser.createDate'),
+      unset('updatedUser.institution'),
+      unset('updatedUser.libraryCards')
+    )(user);
```
sjkobori, line 80: This seems to be a relic of an older implementation of this API. I don't think it is within the scope of this ticket to update/remove this logic, however, I did want to check with the team about this before spinning it off into its own ticket

src/libs/ajax/User.ts
```diff
@@ -0,0 +1,168 @@
...
+        url,
+        fp.mergeAll([
+          Config.authOpts(),
+          Config.jsonBody(request),
+          { method: 'POST' },
+        ])
+      );
+      if (res.ok) {
+        return res.json();
+      }
```
sjkobori, line 42: Wanted to make a callout here. This previously read `return res.json` without the parentheses and Typescript helped me catch that. There is only one place this method is used and it is being used as a truthy value so I don't believe this should change any behavior.

Regardless, I just wanted to make reviewers aware since it is not clear at all from the diff.

src/libs/ajax/User.ts
sjkobori: To be more specific (and I can add this to the comment directly if desired), these unset calls do not in fact unset those properties but the payload being used in the update request does not have those fields anyway.

rushtong: This is an artifact of the older API. The latest version accepts a limited number of fields that can be updated as per the API docs. I think it is appropriate to leave this logic as is and we can address it in a future ticket.
* [UserUpdateFields](https://github.com/DataBiosphere/consent/blob/8305d4748f6c6589fc68723de759c871c9a2283e/src/main/java/org/broadinstitute/consent/http/models/UserUpdateFields.java#L15)
* [Current API Docs](https://github.com/DataBiosphere/consent/blob/8305d4748f6c6589fc68723de759c871c9a2283e/src/main/resources/assets/paths/userByIdv2.yaml#L4)

src/libs/ajax/User.ts
rushtong: I think this is fine, but it might be redundant if `fetchOk` fails on a non-OK response. There could be descendant code that also handles the non-OK case in a particular manner, for example, showing a custom error response depending on what `res.json()` does or does not contain.

rushtong: Just to clarify - I don't see any issues with moving from `res.json` -> `res.json()`

src/libs/ajax/User.ts
```diff
@@ -0,0 +1,168 @@
...
+    } catch (err) {
+      return false;
+    }
+  },
+
+  updateSelf: async (payload: UpdateDuosUserRequestV1): Promise<UpdateDuosUserResponse> => {
+    const url = `${await getApiUrl()}/api/user`;
+    // We should not be updating the user's create date, associated institution, or library cards
```
rushtong, line 50: Note that this is currently broken in dev: https://broadworkbench.atlassian.net/browse/DCJ-362

cypress/component/unit/ajax/User.spec.ts
```diff
@@ -0,0 +1,795 @@
...
+        try {
+          await User.list('Admin');
+        } catch (err) {
+          assert(true);
+        }
+      });
+    });
+  });
+
+  //TODO
```
rushtong: Is this still relevant?

cypress/component/unit/ajax/User.spec.ts
```diff
@@ -0,0 +1,795 @@
...
+        expect(returnedUsers[1].displayName).to.eq(testUser2.displayName);
+        expect(returnedUsers[1].userId).to.eq(testUser2.userId);
+      });
+    });
+    describe('on rejection', () => {
+      it('does not handle the rejection', async () => {
+        cy.intercept('GET', 'api/user/role/*', { statusCode: 404 }).as('list');
+        try {
+          await User.list('Admin');
+        } catch (err) {
```
rushtong: Do we need anything before the catch to ensure that we actually did hit the assert in the catch block? This pattern does test as expected, so I'm fine with this as is.

rushtong (approved):
I tested this locally, no regressions. Code looks good, test coverage also looks good 👍🏽 Consider inline comments optional.

src/libs/ajax/User.ts
sjkobori: I think ultimately, we should change all these `fetchOk`s to use axios instead. I think it would be wise to have a consistent way of making these requests and if we decide to do any handling of errors, they should happen outside of these ajax calls

cypress/component/unit/ajax/User.spec.ts
sjkobori: Nope, missed that one 👍

cypress/component/unit/ajax/User.spec.ts
sjkobori: That's a good point, I'll put some `assert(false)`s after the ajax calls before the catch block to make sure it fails if it does not throw for some reason

fboulnois (approved):
Approved after tests pass.

PR #2579, opened 2024-05-10
[DCJ-58] Update DAR application to indicate all required DAAs

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-58

### Summary
Shows all required DAAs for datasets in a DAR in the Data Use Agreement section of the DAR application instead of the "default" Broad & NHGRI LCAs. In the case that a DAR only consists of datasets that are "Open Access" or "External to DUOS," we are currently showing the DUOS Code of Conduct/DAA. This may change after conversations with Jonathan, but this is what was decided for now.

----

- Verify all tests go green


src/pages/dar_application/DataUseAgreements.jsx
```diff
@@ -14,11 +13,77 @@ export default function DataUseAgreements(props) {
...
+      );
+    });
+    if (fileNames.size === 0) {
+      return (
+        <div></div>
+      );
+    } else {
+      return (
+        <div>
+          <h3>By submitting this data access request and in accordance with your Institution’s issuance of Library Cards to you for the agreement(s) below,</h3>
```
rushtong: Grammar nit: I think this should end in a period
```suggestion
          <h3>By submitting this data access request and in accordance with your Institution’s issuance of Library Cards to you for the agreement(s) below.</h3>
```

src/pages/dar_application/DataUseAgreements.jsx
```diff
@@ -14,11 +13,77 @@ export default function DataUseAgreements(props) {
...
     isAttested,
     cancelAttest,
+    datasets
   } = props;
+  const [daas, setDaas] = useState([]);
+
+  useEffect(() => {
+    const init = async () => {
+      try {
+        const daaList = await DAA.getDaas();
```
rushtong: This means we're getting ALL data access agreements every time we load this page. I think a more sustainable approach would be to get a list of DAAs for a distinct set of dataset ids OR a list of DAAs for a DAR reference id (I like this approach better) and that way we can move the filtering logic to the server which should allow the UI to be more streamlined. This would be a new API/ticket and an incremental improvement.

A related issue we have here is that there is the case where there are no DAAs for a particular dataset. We should clarify what the requirements should be in that case which can leave a DAR in a strange state.

src/pages/dar_application/DataUseAgreements.jsx
rushtong, line 1: Can we rename this to `DataAccessAgreements.jsx`?

src/pages/dar_application/DataUseAgreements.jsx
```diff
@@ -14,11 +13,77 @@ export default function DataUseAgreements(props) {
...
+      } catch (error) {
+        Notifications.showError({
+          text: 'Error: Unable to retrieve DAAs from server',
+        });
+      }
+    };
+    init();
+  }, []);
+
+  const RequiredDAAs = () => {
```
rushtong: This could be a separate file/component

src/pages/dar_application/DataUseAgreements.jsx
```diff
@@ -14,11 +13,77 @@ export default function DataUseAgreements(props) {
...
+      fileNames.add(fileName);
+      return (
+        <div key={id}>
+          {DAADownload(id, fileName)}
+        </div>
+      );
+    });
+    if (fileNames.size === 0) {
+      return (
+        <div></div>
```
rushtong: In the empty case, should we default to the original `BroadLibraryCardAgreementLink` and `NhgriLibraryCardAgreementLink` components? I think we should clarify the requirements here.

rushtong (commented):
This works as expected. I have some questions inline, and I think we need to clarify some edge cases.

src/pages/dar_application/DataUseAgreements.jsx
aarohinadkarni: I think getting a list of DAAs for a DAR reference id makes sense, the ticket to add that functionality is linked here: https://broadworkbench.atlassian.net/browse/DCJ-364

I don't think there will be a case where there are no DAAs for a particular dataset. If a dataset is associated with a DAC, it will have a DAA associated with it. In the case that a dataset is Open Access or External to DUOS, we should not be creating a DAR for it. The ticket for addressing this existing bug allowing us to create DARs that are Open Access or External to DUOS is linked here: https://broadworkbench.atlassian.net/browse/DCJ-359

src/pages/dar_application/DataUseAgreements.jsx
aarohinadkarni: Ideally, we should not be running into this empty case, because we should only be opening DARs for datasets that are associated with DACS (not Open Access or External to DUOS), and thus that have DAAs. I think currently until we have a fix in for that in, in the empty case we should just be showing the DUOS "DAA"/Code of Conduct that is displayed in text here. What do you think?

rushtong: > Ideally, we should not be running into this empty case, because we should only be opening DARs for datasets that are associated with DACS (not Open Access or External to DUOS), and thus that have DAAs. I think currently until we have a fix in for that in, in the empty case we should just be showing the DUOS "DAA"/Code of Conduct that is displayed in text here. What do you think?

That is the ideal case. I think we still need a way to deal with the edge case which is when a DAC does not have a DAA associated to it. In testing on dev, I was able to create a DAR against an open access dataset. We should be protecting against those cases, but they inevitably slip through.

aarohinadkarni: Update: We have decided that an empty div and the displaying of the DUOS Code of Conduct makes sense to show in these cases where a DAR is created against an open access or external dataset.

rushtong:
I have a functional question here. Even if my user does **NOT** have a LC-DAA association, I still see the linked data access agreement in my DAR. Is that correct? I'm not sure we've thought this use case through yet.

Screenshot 2024-05-17 at 9 07 04 AM

src/pages/dar_application/RequiredDAAs.jsx
```diff
@@ -0,0 +1,38 @@
...
+    return (
+      <div key={id}>
+        {daaDownload(id, fileName)}
+      </div>
+    );
+  });
+  if (fileNames.size === 0) {
+    return (
+      <div></div>
+    );
```
rushtong, line 27: If we release this to prod, all DARs will no longer show any data access agreements. I want to confirm that this is intended behavior. The SO DAA screens are currently flagged on the environment, but this is not so we are in an interim state until the SO DAA screens are in production.

rushtong (commented):
See comments inline - I think we need to address a release strategy before merging this.

fboulnois (commented):
Holding off on review until we resolve Greg's concern.

src/pages/dar_application/RequiredDAAs.jsx
aarohinadkarni: That's a great point, I don't think we would want no DAAs showing up if we release this to prod but I'm not positive. I propose that we wait until the SO DAA screens are in production before we release this to prod to ensure all parts necessary for this change to make sense are released together.

aarohinadkarni:
> I have a functional question here. Even if my user does **NOT** have a LC-DAA association, I still see the linked data access agreement in my DAR. Is that correct? I'm not sure we've thought this use case through yet.

@rushtong I believe that future work ([DCJ-49](https://broadworkbench.atlassian.net/browse/DCJ-49) and [DCJ-47](https://broadworkbench.atlassian.net/browse/DCJ-47)) should ensure that if you don't have the LC-DAA association for the DAA that governs a dataset, you should not be able to get to this screen where you would be opening a DAR against that dataset. 

[DCJ-49]: https://broadworkbench.atlassian.net/browse/DCJ-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DCJ-47]: https://broadworkbench.atlassian.net/browse/DCJ-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

rushtong:
> @rushtong I believe that future work ([DCJ-49](https://broadworkbench.atlassian.net/browse/DCJ-49) and [DCJ-47](https://broadworkbench.atlassian.net/browse/DCJ-47)) should ensure that if you don't have the LC-DAA association for the DAA that governs a dataset, you should not be able to get to this screen where you would be opening a DAR against that dataset.

Thank you, that's helpful. This means that during the time between SO's creating LC-DAA associations, users will see the DAAs for datasets even though they don't have any and their SOs will not have created any. It also means that if an SO purposefully does NOT create an LC-DAA association for a user, the user will still see it. 

As part of the release strategy, I think we need to work on some kind of data population process that creates these for users with open DARs.

[DCJ-49]: https://broadworkbench.atlassian.net/browse/DCJ-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DCJ-47]: https://broadworkbench.atlassian.net/browse/DCJ-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

aarohinadkarni:
> As part of the release strategy, I think we need to work on some kind of data population process that creates these for users with open DARs.

This makes sense— let's discuss with the team.

src/pages/dar_application/RequiredDAAs.jsx
rushtong: My concern here has been resolved with the conditional render. This will only be displayed when we fully release this to higher environments.

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -606,13 +608,23 @@ const DataAccessRequestApplication = (props) => {
...
               {!props.readOnlyMode ?
                 <div className='step-container'>
-                  <DataUseAgreements
-                    darCode={formData.darCode}
-                    cancelAttest={() => setIsAttested(false)}
-                    isAttested={isAttested}
-                    attest={attemptSubmit}
-                    save={() => setShowDialogSave(true)}
-                  />
+                  {checkEnv(envGroups.DEV) ?
```
rushtong, line 611: Tested locally with `checkEnv('staging')` which works as expected.

fboulnois (approved):
lgtm 👍

rushtong (approved):
👍🏽  Thank you for the updates!

PR #2581, opened 2024-05-11
[DCJ-224] Refactor NIHicWebform.js to JSX

PintoGideon:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-224

### Summary

1. Remove the use of react-hyperscript-helpers
2. Update the file extension from js to jsx

----

- Verify all tests go green


src/pages/NIHicWebform.js
```diff
@@ -218,1667 +217,1192 @@ export default function NIHICWebform() {
...
+      <div className='row no-margin'>
+        <PageHeading
+          imgSrc={addDatasetIcon}
+          color='common'
+          descriptionStyle={{ color: Theme.palette.primary, fontSize: '19px' }}
+          title='Extramural Genomic Data Sharing Plan & Institutional Certification'
+          description='This integrated GDSP & IC form combines duplicate fields, allows for digital tracking and statistics, and assigns machine-readable GA4GH Data Use Ontology terms to the datasets upon completion!'
+        />
+      </div>
+      <hr className='section-seperator' />
```
okotsopoulos: I see in the diff that this was previously `section-separator`, I think this is a typo.

src/pages/NIHicWebform.js
```diff
@@ -218,1667 +217,1192 @@ export default function NIHICWebform() {
...
+      <form name='form' noValidate={true}>
+
+        <div id='form-views'>
+          <div className='col-lg-10 col-lg-offset-1 col-xs-12'>
+            <fieldset>
+              <h3 className='rp-form-title common-color'>Administrative Information</h3>
+
+              <div className='form-group'>
+                <div className='col-xs-12 rp-group'>
+                  <label className='control-label rp-title-question common-color'>Principal Investigator Name</label>
```
okotsopoulos: In the original, there was a space after the label value.  Are those no longer needed? (Ditto for the label values below, too.)

rushtong (approved):
👍🏽

okotsopoulos (commented):
This looks good as far as I can see, I left a few comments.

I am still new to the DUOS world, so more seasoned engineers may come in and give an approval without the following.  No objections to you merging if so.  Is there a way that you could deploy this to a BEE to make it easier for reviewers to compare old and new forms, along with providing instructions on how to get to the form in the UI?

src/pages/NIHicWebform.js
PintoGideon: Ah, thanks for catching that !!

src/pages/NIHicWebform.js
PintoGideon: @okotsopoulos  That's a great suggestion. I don't have access to Beehive. I've always assumed that the other team members had a local instance of the UI that they tested against, or they tested against the production instance of DUOS. More than happy to learn how to set up  a BEE instance for testing if that is helpful and the way to go for ui pr's.

fboulnois (approved):
looks good, thank you 👍

PR #2615, opened 2024-07-01
DCJ-367: Link to data library instead of old catalog

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-367

### Summary
Replace all links to the old catalog page to the new data library page
This PR does not remove the older custom dataset catalog functionality. There are no in-app links to that feature but it has been shared with customers. This should be addressed in a separate ticket as a redirect feature.

----

- Verify all tests go green

### DIFF
--- a/src/Routes.jsx
+++ b/src/Routes.jsx
@@ -6,7 +6,6 @@ import ManageDac from './pages/manage_dac/ManageDac';
 import ManageEditDac from './pages/manage_dac/ManageEditDac';
 import AdminManageUsers from './pages/AdminManageUsers';
 import DataAccessRequestApplication from './pages/dar_application/DataAccessRequestApplication';
-import DatasetCatalog from './pages/DatasetCatalog';
 import DACDatasets from './pages/DACDatasets';
 import DatasetRegistration from './pages/DatasetRegistration';
 import Home from './pages/Home';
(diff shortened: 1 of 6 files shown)


src/components/DuosHeader.jsx
```diff
@@ -82,7 +82,7 @@ export const headerTabsConfig = [
     link: '/admin_manage_dar_collections',
     children: [
       { label: 'DAR Requests', link: '/admin_manage_dar_collections' },
-      { label: 'Dataset Catalog', link: '/dataset_catalog' },
+      { label: 'Data Library', link: '/datalibrary', search: 'datalibrary' },
```
okotsopoulos, line 85: What does this `search` field represent?

src/Routes.jsx
```diff
@@ -98,7 +97,6 @@ const Routes = (props) => (
     <AuthenticatedRoute path="/admin_manage_lc/" component={AdminManageLC} props={props} rolesAllowed={[USER_ROLES.admin]} />
     <AuthenticatedRoute path="/admin_manage_dar_collections/" component={AdminManageDarCollections} props={props} rolesAllowed={[USER_ROLES.admin]} />
     {checkEnv(envGroups.NON_STAGING) && <AuthenticatedRoute path="/dataset_catalog/:variant" component={CustomDatasetCatalog} props={props} rolesAllowed={[USER_ROLES.researcher]}/>}
```
okotsopoulos, line 99: Is this a place that also needs to be updated?

okotsopoulos (approved):
Looks good as far as I can tell -- I agree with your choice to limit the scope to updating links only.  Will you file that follow-on ticket for full removal of the `DatasetCatalog` component and related code?

src/Routes.jsx
rushtong: This is the `custom dataset catalog functionality` in the PR description that I did not want to remove yet. Jonathan has shared out links to this so we need to figure out what to do with it.

src/components/DuosHeader.jsx
rushtong: The `search` value is used in a couple places in this file. I don't remember exactly how it works, but I'm pretty sure it has to do with displaying the correct tab based on your role and where you have navigated to/from. For example, if you have multiple roles, each one with access to the same sub-tab (i.e. data library), then this feature helps restore the correct sub-tab when the user navigates between higher level role tabs.

rushtong:
> Will you file that follow-on ticket for full removal of the `DatasetCatalog` component and related code?

Yes, thank you, I will get that ticket into the backlog.

aarohinadkarni (approved):
looks good to me! 👍

PR #2636, opened 2024-07-25
[DCJ-47] UI: Enable DAA Restriction in Data Library

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-47

### Summary
Adds a modal that shows which of the selected datasets in the Data Library that the user creating a DAR for has approval to request access to before proceeding to the DAR. This is only enabled in dev & staging, and the original user journey remains in place in prod.

----

- Verify all tests go green


rushtong:
Moving this back into draft today, we need to address one more design comment:

> what needs to be added to the modal PR for if the user has access to none of the datasets selected:
>> Hi - Modal will display the message that they do not have access to those dataset and need to send request to SO. The example of the model with two columns would have the text you need. (message from Lou if you have any questions)

PR #2646, opened 2024-08-14
DCJ-606: Use ES to query for My DAC Datasets

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-606

### Summary
This PR changes the source of truth for where we get datasets from and how we display them. Originally, we would find all datasets for each DAC the user was a member/chair for. Then, we'd iterate through them and reprocess the Data Use through custom UI logic.

The new and improved method is to query the indexed dataset terms from Elastic Search. These have already been pre-processed to show the correct data use. This accomplishes several goals:

1. Speeds up the process orders of magnitude. The dataset APIs that query the db are very slow due to the large amount of processing. The ES index is very fast.
2. It narrows down the code paths we have that translate Data Use objects into more friendly UI-consumable entities. Existing UI code replicates some of this logic inefficiently whereas the ES indexing does it faster and more consistently.
3. It narrows down the code paths that query for datasets and studies. Many UI elements are now querying a single source of truth instead of multiple. This allows us to better debug inconsistencies across pages.

See also: https://github.com/DataBiosphere/consent/pull/2374
This PR takes advantage of the newly added `phsId` term added to the ES index.

### Screen
The new view should look and function exactly like the original, just faster and with more accurate Data Use information.

Screenshot 2024-08-15 at 7 04 37 AM

Screenshot 2024-08-15 at 7 42 51 AM

----

- Verify all tests go green


src/utils/DatasetService.js
```diff
@@ -1,57 +0,0 @@
...
-  },
-
-  getDataUseTranslations: async (dataset) => {
-    if (!dataset.dataUse || isEmpty(dataset.dataUse)) {
-      return {
-        codeList: '- -',
-        translations: []
-      };
-    } else {
-      const translations = await DataUseTranslation.translateDataUseRestrictions(dataset.dataUse);
```
rushtong, line 48: Reviewers note: this is a particularly expensive operation that we want to try and prune the usages of. The only way around this currently is to use the ES index as the primary source of truth for translated data use objects.

fboulnois (approved):
improvements look reasonable to me 👍

okotsopoulos (approved):
Looks good, thank you!

PR #2651, opened 2024-08-19
DCJ-619: Simplify/fix useEffect infinite reload problem

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-619

### Summary
This PR does a few of things that hopefully clarify and simplify the intended functionality behind this page.

1. Move all initialization logic into a single method
2. Load DACs based on the user role. Here, we check state to see if the user is Admin before loading all DACs on the page. This prevents a flash of seeing all DACs before they get filtered by role and instead, only become visible once state has loaded and we know which role the user is accessing DACs for.
3. Clean up usages of lodash to use native javascript filter/map/includes functionality
4. Reduce the `useEffect` dependencies to an empty array

----

- Verify all tests go green

### DIFF
--- a/src/pages/manage_dac/ManageDac.jsx
+++ b/src/pages/manage_dac/ManageDac.jsx
@@ -4,7 +4,6 @@ import {useState, useEffect, useCallback} from 'react';
 import { Styles } from '../../libs/theme';
 import lockIcon from '../../images/lock-icon.png';
 import { DAC } from '../../libs/ajax/DAC';
-import {contains, filter, map} from 'lodash/fp';
 import {Storage} from '../../libs/storage';
 import {Notifications} from '../../libs/utils';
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
good catch 👍

pshapiro4broad (approved):
The change looks OK but client side filtering based on role always makes me nervous. I assume that it's OK that a user can see all DACs even if they can't operate on them? So the filtering isn't for security, but to hide elements in the UI that would otherwise be disabled?

rushtong:
> The change looks OK but client side filtering based on role always makes me nervous. I assume that it's OK that a user can see all DACs even if they can't operate on them? So the filtering isn't for security, but to hide elements in the UI that would otherwise be disabled?

Yes exactly, all DACs are visible to all users. The APIs, of course, enforce role based restrictions as well.

PR #2657, opened 2024-08-29
DCJ-649: Use OidcBroker for sign-out functionality

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-649
* Partially addresses https://broadworkbench.atlassian.net/browse/DCJ-89
* Adds unit tests to start the coverage of `auth.ts`: https://broadworkbench.atlassian.net/browse/DCJ-328
* Adds unit tests to start the coverage of `oidcBroker.ts`: https://broadworkbench.atlassian.net/browse/DCJ-330

### Summary
- This PR tries to carry forward the effort started in https://github.com/DataBiosphere/duos-ui/pull/2611
- This implements a single feature - Sign Out via OIDC.
- `auth.ts` is a minimized version of [`auth.ts`](https://github.com/DataBiosphere/duos-ui/pull/2611/files#diff-12dab7e8b131f803fbdcb63c047768ac2c7c55e0d49d236cf42e13740277e993) in the above PR
- `oidcBroker.ts` is a minimized version of [`oidcBroker.ts`](https://github.com/DataBiosphere/duos-ui/pull/2611/files#diff-c22f0b724cae791bb887d11ce9ada280a1581625e133634e5c2356a5637d6980) in the above PR
- Due to testing, I updated the component test action to store failure screenshots for review.

Follow-on work is defined in a variety of tickets under the [DUOS-2534](https://broadworkbench.atlassian.net/browse/DUOS-2534) epic.

### Testing Notes
Testing should include reviewing the developer panel in a browser when logging into a local instance. 
- Signing in is unchanged so there should be no regressions
- Sign-out mimics the original implementation but additionally uses OidcBroker to invoke OIDC library functionality. This will be expanded in future work that layers in additional sign-in functionality
- Inspecting local session storage should show user info when logged in, and cleared out when signed out.

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

[...]

PR #2611, opened 2024-06-25
[DCJ-89] B2C Auth base

sjkobori:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-89
### Summary
https://github.com/DataBiosphere/duos-ui/pull/2512 has become pretty large an hard to parse. This PR will be the base PR which all tickets from https://broadworkbench.atlassian.net/browse/DCJ-319 which are dependent on this base PR will first be merged into.

PR #2667, opened 2024-09-09
DCJ-654: Sign-in via OIDC

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-654

### Summary
This PR completes the efforts started in https://github.com/DataBiosphere/duos-ui/pull/2611 and https://github.com/DataBiosphere/duos-ui/pull/2657. This replaces Google Sign-in with Oidc sign-in.

* Use `oidc-client-ts` library for sign-in
* Expand unit testing around auth, oidcBroker, and sign-in button components
* Minor refactoring around new user registration
* See also: https://github.com/broadinstitute/terra-helmfile/pull/5871

#### Full Screen
Screenshot 2024-09-17 at 11 37 41 AM

#### Minimized Screen
Screenshot 2024-09-17 at 11 37 50 AM

#### Our (current) B2C Tenant Screen
This will be changed to a DUOS version in https://broadworkbench.atlassian.net/browse/DCJ-165
Screenshot 2024-09-17 at 12 07 56 PM

----

- Verify all tests go green


cypress/component/Auth/auth.spec.ts
```diff
@@ -1,29 +1,47 @@
...
+import {mockOidcUser} from './mockOidcUser';
 
-describe('Auth', function () {
+describe('Auth Failure', function () {
+  it('Sign In error throws expected message', async function () {
+    cy.stub(OidcBroker, 'signIn').returns(null);
+    cy.on('fail', (err) => {
+      return err.message !== Auth.signInError();
+    });
+    Auth.signIn();
```
fboulnois, line 16: This function still seems to be async, does it need to be `await`ed?

fboulnois: This function still seems to be async, does it need to be `await`ed?

cypress/component/Auth/auth.spec.ts
```diff
@@ -1,29 +1,47 @@
...
       url: '/config.json',
       hostname: 'localhost',
     }, {'env': 'ci'});
     cy.stub(OAuth2, 'getConfig').returns({
-      'authorityEndpoint': 'authorityEndpoint',
+      'authorityEndpoint': Cypress.config().baseUrl,
       'clientId': 'clientId'
     });
-    await Auth.initialize();
+    Auth.initialize();
```
fboulnois, line 35: This function still seems to be async, does it need to be `await`ed?

fboulnois (approved):
overall looks good, questions relating to tests:

config/alpha.json
```diff
@@ -5,7 +5,6 @@
   "hash": "alpha",
   "apiUrl": "https://consent.dsde-alpha.broadinstitute.org",
   "ontologyApiUrl": "https://consent-ontology.dsde-alpha.broadinstitute.org/",
-  "clientId": "1020846292598-hd801vsmmbhh97vaf6aar17lu0q2evfj.apps.googleusercontent.com",
```
rjohanek, line 8: I thought alpha didn't exist anymore? Why do we need this config?

rushtong: This config file name is not important and is, unfortunately, an artifact of our build image process. Originally, it did reference real environment variables. I'll change this in a separate PR/ticket.

cypress/component/Auth/auth.spec.ts
rushtong: I was initially confused by this. Cypress gives you a warning if you use `async` in `beforeEach`:
```
Cypress Warning: Cypress detected that you returned a promise in a test, but also invoked one or more cy commands inside of that promise.

The test title was:

  > Auth Success "before each" hook

While this works in practice, it's often indicative of an anti-pattern. You almost never need to return both a promise and also invoke cy commands.

Cy commands themselves are already promise like, and you can likely avoid the use of the separate Promise.

https://on.cypress.io/returning-promise-and-commands-in-test
```

cypress/component/Auth/auth.spec.ts
rushtong: Good call, but `await`-ing breaks the test, I think I need to use `then`

config/alpha.json
rushtong: I created https://broadworkbench.atlassian.net/browse/DCJ-678 to cover this bit of tech debt.

src/components/SignInButton.tsx
```diff
@@ -139,11 +102,15 @@ export const SignInButton = (props: SignInButtonProps) => {
   const registerAndRedirectNewUser = async (redirectTo: string, shouldRedirect: boolean) => {
     const registeredUser: DuosUser = await User.registerUser();
     setUserRoleStatuses(registeredUser, Storage);
-    await onSignIn();
+    await syncSignInOrRegistrationEvent(eventList.userRegister);
+    history.push(`/tos_acceptance${shouldRedirect ? `?redirectTo=${redirectTo}` : ''}`);
```
rjohanek, line 106: why move this line from syncSignInOrRegistrationEvent to here?

src/index.tsx
```diff
@@ -4,12 +4,16 @@ import './index.css';
...
 import {Auth} from './libs/auth/auth';
+import {OidcBroker} from './libs/auth/oidcBroker';
 import {unregister} from './registerServiceWorker';
 import {BrowserRouter} from 'react-router-dom';
 
 const load = async () => {
   unregister();
   await Auth.initialize();
+  if (window.location.pathname.startsWith('/redirect-from-oauth')) {
+    await OidcBroker.getUserManager().signinPopupCallback(window.location.href);
```
rjohanek, line 24: what does this do?

rjohanek (approved):
overall, looks good! Just a couple questions for my own understanding

src/index.tsx
rushtong: Great question. In the original version of this work, there was a [whole component](https://github.com/DataBiosphere/duos-ui/pull/2611/files#diff-326a2368c43259048d25d6a6260488fed18c66c82678ef208bd6aa99020ed902) dedicated to servicing this single path. This is my attempt to simplify it. In short, it does this:

1. The OIDC client calls a popup window with a redirect url here: https://github.com/DataBiosphere/duos-ui/pull/2667/files#diff-c22f0b724cae791bb887d11ce9ada280a1581625e133634e5c2356a5637d6980L42
2. The user goes through the auth dance and is redirected to that url, but the popup doesn't go away.
3. On reload, we detect the path
4. Call the `signinPopupCallback` which does two things
    - It closes the open popup
    - Redirects the user to the original href, which is either the home page, or a sub-page inside the app to which we redirect.

src/components/SignInButton.tsx
rushtong: My goal with `syncSignInOrRegistrationEvent` was to isolate the Metrics calls. I wanted that to remain short and focused. Since the history modification is not related to that functionality, I pulled it out so it would be more clear as to where it is being used. What I found in going through this code is that there was a lot of dual-function helper methods that were very confusing to debug, especially the new user registration piece. By isolating functionality, I was better able to log and reason about what was happening in the different cases.

src/index.tsx
rushtong: I'm going to add some comments around this for future devs.

PR #2669, opened 2024-09-11
DCJ-667: Use better-npm-audit for CI instead of plain npm-audit

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-667

### Summary
This PR does two things:
1. Replace usages of `npm audit` with `better-npm-audit audit -p`
2. Adds an exclusion for the `path-to-regexp` finding

The `path-to-regexp` finding will be handled with https://broadworkbench.atlassian.net/browse/DCJ-294. In the meantime, this finding should not prevent other work from proceeding. This PR allows for that to happen and also provides a better way of managing potential exclusions in the future.

----

- Verify all tests go green

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "ajv": "8.17.1",
     "ajv-formats": "3.0.1",
     "axios": "1.7.7",
+    "better-npm-audit": "^3.11.0",
     "dompurify": "3.1.6",
     "http-proxy-middleware": "2.0.6",
     "js-file-download": "0.4.12",
@@ -43,6 +44,7 @@
     "voca": "1.4.1"
   },
   "scripts": {
+    "audit": "better-npm-audit audit -p",
     "analyze": "npm run build; source-map-explorer build/static/js/main.*",
     "start": "react-scripts start",
     "genschemas": "./scripts/compile-jsonschema.sh",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "ajv": "8.17.1",
     "ajv-formats": "3.0.1",
     "axios": "1.7.7",
+    "better-npm-audit": "3.11.0",
```
fboulnois, line 17: Although I can see the advantages of this package, I have a few concerns:

- It is itself an npm dependency that depends on a number of packages that may have vulnerabilities
- According to the [Readme](https://www.npmjs.com/package/better-npm-audit?activeTab=readme) , they are looking for collaborators, which suggests that it may not be well maintained (although I do see a recently published version)
- `npm` seems to change the audit command behavior between major versions, which also requires major changes to this package

Instead could we:

1. not depend on `npm audit` in Pull Requests?
2. run `npm audit` in a non-blocking way?
3. run `npm audit` on a schedule?

fboulnois (commented):
See below:

rushtong:
This PR is no longer necessary. https://github.com/DataBiosphere/duos-ui/pull/2673 updated the underlying problem behind this effort.

PR #2675, opened 2024-09-16
DCJ-676: Remove setupProxy and http-proxy-middleware

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-676

### Summary
This PR removes unnecessary configurations for local development.

----

- Verify all tests go green


fboulnois (approved):
nice cleanup 👍

DEVNOTES.md
```diff
@@ -12,8 +12,8 @@ nvm install 22.6.0
 npm install
 ```
 
-3. Install configs for an environment. This example is for the `alpha` environment, but you can use values from any environment by looking at the deployed configs in https://duos-k8s.dsde-{%ENV%}.broadinstitute.org/config.json where {%ENV%} is any of `dev`, `staging`, `alpha`, or `prod` 
-Remember to set the `env` value appropriately. We use `local` for running via npm, but under docker, we use a real env like `dev`.
+3. Install configs for an environment. This example is for the `alpha` environment, but you can use values from any environment by looking at the deployed configs in https://duos-k8s.dsde-{%ENV%}.broadinstitute.org/config.json where {%ENV%} is any of `dev`, `staging`, `alpha`, or `prod`
+Remember to set the `env` value appropriately, for example, `dev`. Certain features are not available on different environments and this will simulate that for local development.
```
rjohanek: is the point here that we want to point to dev backend for local development? Does "this" in this sentence refer to dev? This is a little unclear to me

rjohanek (approved):

DEVNOTES.md
rushtong: Thanks for the question Rae! `this` refers to the `env` variable. I'll take another stab at this language.

rjohanek: great thanks looks good!

PR #2683, opened 2024-10-07
DCJ-700: Update to use correct field name

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-700

### Summary
Update the component to use the correct field name. In https://github.com/DataBiosphere/duos-ui/pull/2646, we changed the dataset lookup to pull from Elastic Search instead of the Datasets APIs in Consent. The two objects are slightly different. A larger refactor is necessary to align the two models that will be addressed separately from this bug.

----

- Verify all tests go green


fboulnois (approved):
makes sense 👍

rjohanek (approved):

PR #2687, opened 2024-10-10
[DCJ-691] Add filter for primary data use

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DCJ-691

### Summary

Revamps the filtering process to be scalable for many different filters, and adds a filter for primary data use. In general, the filtering works by:

1. Updating the state of the filters based on selections (in `filterHandler`)
2. Constructing a query from the filter terms (in `assembleFullQuery`)
3. Querying elasticsearch for a filtered search (in `search`)
4. Refreshing the view with the new results (done by React)

What was changed here is that previously the selections for the (one and only) filter were hardcoded and added to an array, and then the state and query were constructed from that array. Now instead:

1. The filter selections are dynamically populated from the datasets
2. `filterHandler` also accepts a _category_ of filter, and these are stored in an object that corresponds to the filter category and the selections.
3. The query is assembled slightly differently because `{dataset}.dataUse.primary.code` is a deeply nested JSON object which ElasticSearch has indexed a particular way

In the future, the mechanism to add new filters should be straightforward:

1. Extract the filter selections from the available datasets
2. Create a new UI element for the filter (we could probably create a component for this)
3. Adjust the query to search for those particular filter selections

### DIFF
--- a/src/components/data_search/DatasetSearchTable.jsx
+++ b/src/components/data_search/DatasetSearchTable.jsx
@@ -41,9 +41,14 @@ const datasetTableHeader = [
   'Export to Terra',
 ];
 
+const defaultFilters = {
+  accessManagement: [],
+  dataUse: [],
+}
+
 export const DatasetSearchTable = (props) => {
   const { datasets, history, icon, title } = props;
-  const [filters, setFilters] = useState([]);
+  const [filters, setFilters] = useState(defaultFilters);
   const [filtered, setFiltered] = useState([]);
   const [tableData, setTableData] = useState({});
   const [selected, setSelected] = useState([]);
@@ -52,7 +57,9 @@ export const DatasetSearchTable = (props) => {
   const [showTranslatedDULModal, setShowTranslatedDULModal] = useState(false);
   const [dataUse, setDataUse] = useState();
   const searchRef = useRef('');
-  const isFiltered = (filter) => filters.indexOf(filter) > -1;
+
+  const isFiltered = (filter, category) => (filters[category]).indexOf(filter) > -1;
+  const numSelectedFilters = (filters) => Object.values(filters).reduce((sum, array) => sum + array.length, 0);
 
   const assembleFullQuery = (searchTerm, filters) => {
     const queryChunks = [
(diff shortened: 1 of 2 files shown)


rushtong (approved):
Looks great, thank you 👍🏽. Separately, there had recently been a UX effort to move away from lodash. I'm neutral on the subject, so this works for me.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -52,7 +57,9 @@ export const DatasetSearchTable = (props) => {
   const [showTranslatedDULModal, setShowTranslatedDULModal] = useState(false);
   const [dataUse, setDataUse] = useState();
   const searchRef = useRef('');
-  const isFiltered = (filter) => filters.indexOf(filter) > -1;
+
+  const isFiltered = (filter, category) => (filters[category]).indexOf(filter) > -1;
+  const numSelectedFilters = (filters) => Object.values(filters).reduce((sum, array) => sum + array.length, 0);
```
snf2ye, line 62: I'm not super familiar with the lodash reduce function, would you mind explaining what is happening here? Why couldn't we just the length of the array of filters?

snf2ye (approved):
I'm still new to this repo, but looks reasonable to me! 
Re: @rushtong 's comment about lodash - I thought we were trying to move away from fp/lodash, but not regular lodash helpers. But, I'm not totally up to speed on that subject.

src/components/data_search/DatasetSearchTable.jsx
fboulnois: We can definitely do something like `filters.accessManagement.length + filters.dataUse.length`, but this mechanism avoids having to remember to add new filters to this function. `reduce` is a [Javascript built-in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce), it runs a callback over each element of the array with an accumulator, with 0 (in this case) set as the initial value.

`Object.values(filters)` transforms `{ f1: ['a', 'b', 'c'], f2: ['d', 'e', 'f'] }` into `[ ['a', 'b', 'c'], ['d', 'e', 'f'] ]`
then `reduce` loops over each item in the array (which are subarrays), adds the length of the subarrays to the sum (which starts at zero), and so ultimately returns the total number of selected filters.

fboulnois: RE: lodash, that was also my understanding, moving specifically away from the `fp` version of lodash, not the regular version.

snf2ye: Thank for explaining that!

PR #2691, opened 2024-10-17
[DCJ-741] refactor dataset table into two dataset and studies table

s-rubenstein:
### Addresses

- Splits the collapsing table into two tables, studies and datasets

### Summary

- I added typing to the table, and abstracted it out into its own components
- I added optional tooltips if the text overflowed, rather than expanding the rows - this reduces screen jitter.

https://github.com/user-attachments/assets/66a1afff-e4b6-4db1-b293-4e08a8faee72

----

- Verify all tests go green


src/components/SimpleTable.jsx
```diff
@@ -104,7 +104,7 @@ const DataRows = ({rowData, baseStyle, columnHeaders, rowWrapper = ({renderedRow
           let output;
           //columnHeaders determine width of the columns,
           //therefore extract width from columnHeader and apply to cell style
-          const columnWidthStyle = columnHeaders[cellIndex].cellStyle;
+          const columnWidthStyle = { width: columnHeaders[cellIndex].cellStyle.width };
```
s-rubenstein, line 107: I didn't know where to check where this was used - but I was noticing we said we were extracting just the width, but actually pulling everything. I figured this would be cleaner.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -165,222 +150,21 @@ export const DatasetSearchTable = (props) => {
...
+    const darDraft = await DAR.postDarDraft({ datasetId: selected });
     history.push(`/dar_application/${darDraft.referenceId}`);
   };
 
   const clearSearchRef = () => {
     searchRef.current.value = '';
     filterHandler(null, datasets, '', '');
   };
-
-  const openTranslatedDUL = (dataUse) => {
```
s-rubenstein, line 259: This was unused in the new version - I figure we can add it back if need be. Not sure because I didn't see dataUse in the mocks: https://www.figma.com/design/MkffpVbUdP4lWvkN9OWiro/Data-Library?node-id=23-43&node-type=frame

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -463,14 +248,6 @@ export const DatasetSearchTable = (props) => {
           }
         </Box>
       </Box>
-      {
-        showTranslatedDULModal &&
-        <TranslatedDulModal
```
rjohanek, line 483: why did this go away?

src/types/model.ts
```diff
@@ -113,6 +113,69 @@ export interface Dataset {
...
+  dacId: number;
+  dacApproval: boolean;
+  accessManagement: string;
+  approvedUserIds: number[];
+  study: StudyTerm;
+  submitter: UserTerm;
+  updateUser: UserTerm;
+  dac: DacTerm;
+}
+
```
rjohanek, line 178: I'd love to hear more about the design of your approach. I'm wondering about the distinctions between this file and the constants file. Additionally, I'm wondering about the abstraction to the Display file.

src/components/data_update/DatasetUpdate.jsx
```diff
@@ -153,7 +153,7 @@ export const DatasetUpdate = (props) => {
         validators={[FormValidators.REQUIRED]}
         defaultValue={formData.properties.description}
         onChange={({ value }) => {
-          formData.properties.description = value;
+          formData.properties.duosId = value;
```
rjohanek: why are these changing from study columns to dataset columns? Did this work for datasets before because they were nested within a study?

src/types/model.ts
rjohanek: overall, this looks great and the video looks good, just looking for a bit more info for my understanding!

src/components/data_update/DatasetUpdate.jsx
```diff
@@ -198,7 +198,7 @@ export const DatasetUpdate = (props) => {
         validators={[FormValidators.REQUIRED]}
         defaultValue={formData.properties.species}
         onChange={({ value }) => {
-          formData.properties.species = value;
+          formData.properties.donorSize = value;
```
fboulnois: This looks unusual to me, especially since we are not changing the default value.

src/components/data_search/DatasetSearchTable.jsx
s-rubenstein: It wasn't used - I don't know if the new mocks have it or not, but I figure we can always readd it if need be

src/components/data_update/DatasetUpdate.jsx
s-rubenstein: Oh, these changes are accidental - good catch

rushtong:
One bit of visual feedback - it's not clear which tab one is on after clicking the tabs. It might be nice to have an active state style to make that more clear:

Screenshot 2024-10-18 at 12 57 13 PM

rushtong:
I think we lost Florian's recent data use filter panel in this PR

s-rubenstein:
Got it. I'll merge it in!

s-rubenstein:
https://github.com/user-attachments/assets/ab0d1c65-6441-427c-9025-54c305980787

src/types/model.ts
s-rubenstein: @rjohanek great question! The goal was to break up and separate concerns. Previously, the `DatasetSearchTable` handled a lot - I wanted to reduce what it handled. The first abstraction was to pull out the table. Eventually I thought pulling out the constants would be nice as well, so `DatasetSearchTableDisplay` just handled the rendering of the table, and `DatasetSearchTableConstants` handled the constant values describing the rendering.

`model.ts` is specifically the type definitions of the API responses, and doesn't include any scope beyond that.

src/components/Tooltips.tsx
rushtong, line 1: Nice work pulling this out

src/components/data_search/DatasetSearchTable.jsx
rushtong: Agreed - I'm usually in favor of aggressive pruning, it's easy enough to re-add. Especially for anything related to data use translations, we have some very inefficient code related to that functionality. If we do need it, we need to re-evaluate how we do it.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -180,222 +165,21 @@ export const DatasetSearchTable = (props) => {
...
-              };
-            }),
-          }
-        };
-      }),
-    };
-
-    setTableData(table);
-  }, [filtered, exportableDatasets, tdrApiUrl]);
+  }, [filtered]);
```
rushtong: I think we may have lost the export to TDR functionality here. Note, for example, that we're not getting the `tdrApiUrl` anywhere else in this PR.

rushtong (commented):
There are some really nice UX improvements here. We need to circle back and figure out where the export to TDR functionality fits now - I don't think it's in the mocks, but it is in the original version of the data library. I'm open to re-implementation in a different way if we can make that a priority.

fboulnois (approved):
After all the improvements this looks great! Thank you 👍

rjohanek (approved):
looks good! thanks for the changes and explanations

PR #2694, opened 2024-10-23
[dt-741] Fix default value for exportable datasets

s-rubenstein:
### Addresses
Fixes the default value for `exportableDatasets` to be an empty object so we don't error out before getting snapshots - we just can't export till we do. (Ideally we would have a loading spinner, but since this isn't the long term home for export I figured this was a good low effort fix)
### Summary

----

- Verify all tests go green


fboulnois (approved):
👍

snf2ye (approved):

rjohanek (approved):

PR #2700, opened 2024-10-28
[DT-899] Disable checkmarks and add tooltips for open or external datasets

s-rubenstein:
### Addresses
Disables open and external datasets, and studies which include at least one open or external dataset. Adds tooltips for these cases.

### Summary

https://github.com/user-attachments/assets/efbb0606-8067-41d7-8814-71c1a0a1d6a5

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTableConstants.tsx
+++ b/src/components/data_search/DatasetSearchTableConstants.tsx
@@ -2,9 +2,10 @@ import {DatasetTerm} from 'src/types/model';
 import _, {groupBy} from 'lodash';
 import {Checkbox, Link} from '@mui/material';
 import * as React from 'react';
-import {OverflowTooltip} from '../Tooltips';
+import {OverflowTooltip, tooltipStyle} from '../Tooltips';
 import {SnapshotSummaryModel} from 'src/types/tdrModel';
 import DatasetExportButton from './DatasetExportButton';
+import ReactTooltip from 'react-tooltip';
 
 export interface DatasetSearchTableTab<T> {
   key: string;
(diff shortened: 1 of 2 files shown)


rjohanek (approved):
with the commented out code in DatasetSearchTableConstants removed. tool tip text looks good. Overall, I am wondering if this is really a better user experience though? after selection, when access is applied for, the information for each dataset could be given then, instead of here in tooltips. I just wonder if this is the right way to display this information. But, as an incremental improvement I think it makes sense.

src/components/data_search/DatasetSearchTableConstants.tsx
```diff
@@ -72,22 +73,36 @@ export const makeStudyTableHeaders = (datasets: DatasetTerm[], selected: number[
...
       sortable: false,
       cellStyle: makeHeaderStyle(studyCellWidths.selected),
       cellDataFn: datasets => {
         const studyDatasetIds = datasets.map(dataset => dataset.datasetId);
         const numberSelected = _.intersection(studyDatasetIds, selected).length;
         const fullySelected = numberSelected === studyDatasetIds.length;
         const indeterminate = numberSelected > 0 && numberSelected < studyDatasetIds.length;
+        const checkboxId = `${datasets[0].study.studyId}-is-selected-checkbox`;
+        const isSelectableStudy = datasets.filter(isSelectable).length === datasets.length;
+        const tooltipText = 'This study contains some open or external datasets. Please select to apply for access on the individual dataset level.';
```
rushtong: The language here threw me for a loop at first. I would suggest something more like ...

> This study contains some open or external datasets. Select the 'View By Datasets' tab to apply for access to 'Controlled Access' datasets in this study.

rushtong (approved):
Thank you for this! See minor language comment inline, otherwise 👍🏽

PR #2701, opened 2024-10-28
[DT-841]Display error from Consent on Sign In

rjohanek:
### Addresses
Ticket: https://broadworkbench.atlassian.net/browse/DT-841

### Summary
- Pass through error from consent and display to the user
- Update other error messages to include the description as well (msg field was unused)
- Convert Alert to tsx so that we can use an optional prop to have an onClose event and an x in the corner of the alert

Screenshot 2024-10-31 at 1 13 20 PM

----

- Verify all tests go green


src/components/SignInButton.tsx
```diff
@@ -113,18 +114,47 @@ export const SignInButton = (props: SignInButtonProps) => {
...
+    }
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      result += decoder.decode(value);
+    }
+    return result;
+  };
```
fboulnois: This approach seems complicated, here's what I would suggest.
```suggestion
  const errorStreamToString = async (error: HttpError) => {
    return await new Response(error.body).text();
  };
```

fboulnois (commented):
A suggestion to simplify getting the body:

src/components/SignInButton.tsx
```diff
@@ -113,18 +114,34 @@ export const SignInButton = (props: SignInButtonProps) => {
     await Metrics.captureEvent(event);
   };
 
+  const errorStreamToString = async (error: HttpError) => {
+    const body = await new Response(error.body).json();
+    return body.message || JSON.stringify(body);
```
rjohanek, line 119: if there is a message return that, else return the whole error

snf2ye: What if that is really long? Do we truncate the message?

snf2ye (approved):

src/components/SignInButton.tsx
```diff
@@ -113,18 +114,34 @@ export const SignInButton = (props: SignInButtonProps) => {
...
+  };
+
+  const handleServerError = async (error: HttpError) => {
+    const errorMessage = await errorStreamToString(error);
+    setErrorDisplay({show: true, title: 'Error', description: errorMessage});
+    setTimeout(() => {
+      setErrorDisplay({});
+    }, 10000);
```
s-rubenstein: Can you `x` out of the error message?

s-rubenstein (approved):
Looks reasonable - I think there are a few questions but mine at least isn't blocking.

src/components/SignInButton.tsx
rushtong: IIRC, I don't think we have a close handler on the error message. Would be nice.

rushtong: I hacked this a little bit ... this code looks ugly (style wise), but works. Modify the alert div around line 221:
```
        : <div className="dialog-alert">
            <button onClick={() => setErrorDisplay({})}>X</button>
          <Alert
            id="dialog"
            type="danger"
            title={(errorDisplay as ErrorInfo).title}
            description={(errorDisplay as ErrorInfo).description}
          />
        </div>
```

rushtong: Spent some more time with this - I think the `Alert` component is too restrictive. Modifying it to take a conditional close handler might be more work than we want to take on due to how much usage it has. Here is my hacky attempt at a solution that uses `import CloseIcon from '@mui/icons-material/Close';`

```
        : <div id={`dialog_alert`} className={`alert-wrapper danger`} style={{border: '1px solid red', borderRadius: '5px'}}>
          <span
            style={{float: 'right', fontWeight: 'bolder', fontSize: 24, cursor: 'pointer'}}
            onClick={() => setErrorDisplay({})}><CloseIcon/></span>
          <h4 id={`dialog_title`} className="alert-title">{(errorDisplay as ErrorInfo).title}</h4>
          <span id={`dialog_description`}
            className="alert-description">{(errorDisplay as ErrorInfo).description}</span>
        </div>
```

Screenshot 2024-10-31 at 7 34 05 AM

src/components/SignInButton.tsx
rjohanek: I don't believe its truncated, I think it will display the whole thing. Would it be preferable to truncate it?

src/components/SignInButton.tsx
rjohanek: With some advice from Miguel, I decided to convert the Alert to tsx so I could explicitly define the props and make onClose optional with an undefined default value.

src/components/SignInButton.tsx
snf2ye: I'm not sure. This is possibly something we can just observe and address if it becomes a problem.

snf2ye (approved):

fboulnois (approved):
looks good 👍

src/components/Alert.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export interface AlertProps {
+  id: string;
+  type: string;
+  title: string;
+  description: string;
+  onClose?: () => void;
+}
+
+export const Alert = ({id, type, title, description, onClose}: AlertProps) => {
```
msilva-broad: breaking this out to a second line would help readability and match the recommended pattern (including the return type:
```

export const Alert = (props: AlertProps): React.ReactNode => {
  const { id, type, title, description, onClose } = props;
  //...
```

rushtong (approved):
👍🏽

PR #2702, opened 2024-10-30
DT-944: Fix chairperson edit DAC feature

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-944

### Summary
This PR makes a number of changes in function and style. The functional changes are intentional. As a business rule, we allow for Chairs to make changes to their own DACs in terms of both membership and other related details. There are also a number of minor style fixes.

Requires corresponding Consent PR: https://github.com/DataBiosphere/consent/pull/2419

### Testing Notes

1. Update the `env` in your local `config.json` to be `prod` so that you're seeing what we would see in production.
2. Ensure you're testing with a user that does not have Admin permissions. Admins can edit anything so you wouldn't see the current bug in that case.
3. There are actually two versions of this page, one in prod, one in non-prod. To see the same page in non-prod, change your `env` config to `dev` or `staging`.

#### Prod: New screen
Screenshot 2024-10-30 at 11 41 45 AM

#### Prod: Old screen
Screenshot 2024-10-30 at 11 43 36 AM

#### Non-prod: New screen
Screenshot 2024-10-30 at 12 23 09 PM

#### Non-prod: Old screen
Screenshot 2024-10-30 at 12 23 44 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/manage_dac/EditDac.jsx
+++ b/src/pages/manage_dac/EditDac.jsx
@@ -16,11 +16,10 @@ import { Styles } from '../../libs/theme';
 import DUOSUniformDataAccessAgreement from '../../assets/DUOS_Uniform_Data_Access_Agreement.pdf';
 import PublishIcon from '@mui/icons-material/Publish';
 import { UploadDaaModal } from '../../components/modals/UploadDaaModal';
+import {Storage} from '../../libs/storage';
 
 export const CHAIR = 'chair';
 export const MEMBER = 'member';
-const CHAIRPERSON = 'Chairperson';
-const ADMIN = 'Admin';
 
 export default function EditDac(props) {
   const [state, setState] = useState({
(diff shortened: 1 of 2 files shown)


src/pages/manage_dac/EditDac.jsx
```diff
@@ -83,44 +82,42 @@ export default function EditDac(props) {
...
-          if (daaFileData === null && selectedDaa.daaId !== broadDaa.daaId) {
-            handleErrors('Please select either the default agreement or upload your own agreement before saving.');
-            return;
-          } else if (daaFileData !== null && selectedDaa === undefined) {
-            currentDac = await DAC.create(currentDac.name, currentDac.description, currentDac.email);
-            const createdDaa = await DAA.createDaa(daaFileData, currentDac.dacId);
-            setCreatedDaa(createdDaa.data);
-          } else {
+          if (user.isAdmin) {
             currentDac = await DAC.create(currentDac.name, currentDac.description, currentDac.email);
```
rjohanek, line 100: I'm having trouble following the different cases here. If there is a dacId, update the DAC, else if there is non broad DAA selected without a file throw an error, else if the user is an admin and there is DAA file data, create a DAC and a DAA for it, else if the user is an admin create a dac. This is pretty complex and not super clear, I think it could benefit from either comments or helper functions or something to make this flow easier to return to in the future.

src/pages/manage_dac/EditDac.jsx
```diff
@@ -374,7 +371,6 @@ export default function EditDac(props) {
                         name="name"
                         className="form-control vote-input"
                         required={true}
-                        disabled={props.location.state.userRole === CHAIRPERSON}
```
rjohanek, line 377: why were these disabled for chairpersons?

src/pages/manage_dac/ManageEditDac.jsx
```diff
@@ -250,30 +248,37 @@ export default function ManageEditDac(props) {
   return (
     isLoading ?
       <Spinner/> :
-      <div className='container container-wide'>
-        <div className='row no-margin'>
+      <div className="container container-wide">
```
rjohanek: usage of " and ' is pretty inconsistent. is " the standard?

rushtong: Good catch, single quotes are the standard - not sure why the linter didn't pick this up

src/pages/manage_dac/EditDac.jsx
rushtong: I think it was an historic decision that doesn't make sense to me. I opted to remove it. Chairs **should** be able to update the values of their DAC.

src/pages/manage_dac/EditDac.jsx
rushtong: I'm trying to maintain the existing logic minus the chairperson restriction that existed before. I can take another stab at trying to refactor this, but my primary goal is bug fixing.

rjohanek (approved):
thanks for the responses to my comments

src/pages/manage_dac/EditDac.jsx
snf2ye: I don't want to prevent progress/bug fixing here if this is too big of a lift, but component tests for this sort of complex logic might be helpful to confirm it is working as expected.

snf2ye (approved):
Not a blocking comment - LGTM

src/pages/manage_dac/EditDac.jsx
rushtong: Good idea ... let me see what I can do here.

rushtong: @raejohanek - Another way to look at this is to hide the white space in the PR. That should make the actual changes much easier to see.

Screenshot 2024-10-31 at 1 03 06 PM

rushtong:
I created https://broadworkbench.atlassian.net/browse/DT-959 to address the lack of unit tests so we can get this bug fix into the next release.

PR #2703, opened 2024-10-30
[DT-900] Get search working again

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-900

### Summary

Fixes the search bar, and ensure that it works properly with any combination of filters and when clearing the search.

To do this the `searchRef` was removed and the search input is treated as if it was any other filter. The only difference is that the input changes must also be reflected to the user, so we need to set the state when the user types or clears the filters.

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTable.jsx
+++ b/src/components/data_search/DatasetSearchTable.jsx
@@ -34,6 +34,7 @@ const styles = {
 const defaultFilters = {
   accessManagement: [],
   dataUse: [],
+  search: []
 };
 
 export const DatasetSearchTable = (props) => {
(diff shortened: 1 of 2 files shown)


src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -76,12 +77,12 @@ export const DatasetSearchTable = (props) => {
...
 
-    // do not apply search modifier if there is no searchTerm
-    if (searchTerm !== '') {
+    // do not apply search modifier if there is no search term
+    if (filters.search.length > 0) {
       const searchModifier = [
         {
           'multi_match': {
-            'query': searchTerm,
+            'query': filters.search[filters.search.length - 1],
```
cinyecai, line 85: Why this change if searchTerm is being set in filterHandler?

fboulnois: For reasons I don't completely understand `searchTerm` at this line reflects the previous `searchTerm` state (for example if your current input was `hello`, `searchTerm` would be `hell`)

cinyecai: I'm wondering if something similar could be done everywhere searchTerm is used to eliminate the need for that variable at all - not sure if that would work right though

rushtong (approved):
👍🏽

snf2ye (approved):

PR #2704, opened 2024-10-31
DT-617: Hide spinner after loading failure

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-617

### Summary
If the dataset search endpoint fails, this PR closes out the loading spinner.

To test, I started up a local Consent (without a local ES instance) not on the VPN so that it would error out when querying ES. Then point local UI to local Consent and visit the data library page. The loading spinner is closed.

Something I did **_not_** do was to percolate the error down to the table itself. The table just sees an empty list of datasets and thinks there are none for the current library filter. This could be expanded if we would like to see more info in that condition, but since this is a fairly infrequent use case, I didn't pursue this.

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetSearch.jsx
+++ b/src/pages/DatasetSearch.jsx
@@ -298,6 +298,7 @@ export const DatasetSearch = (props) => {
           });
         } catch (error) {
           Notifications.showError({ text: 'Failed to load Elasticsearch index' });
+          setLoading(false);
         }
       }
     };


snf2ye (approved):

src/pages/DatasetSearch.jsx
```diff
@@ -298,6 +298,7 @@ export const DatasetSearch = (props) => {
           });
         } catch (error) {
           Notifications.showError({ text: 'Failed to load Elasticsearch index' });
+          setLoading(false);
```
fboulnois, line 301: I don't think this is desirable, since it shows an empty table which may be misleading when on a custom data library. This means that we will not be able to tell the difference between a data table that has no data vs a data library that failed to load.

fboulnois (commented):
See below:

src/pages/DatasetSearch.jsx
rushtong: This is true - the table doesn't know why there are no datasets, but it's also technically true that loading is complete. I noted in the PR description that I'm purposefully choosing not to try and modify the logic in the display to include an error state. If we were to go down that path, this would a larger refactor that I would prefer to do in a ticket with clear requirements in that case.

fboulnois: As a user I cannot take an action if the index fails to load, that's why I don't think it's desirable to go from a loading state (where it is clear that I cannot take an action) to an empty table (where it is unclear if I can take any actions)

rushtong: > As a user I cannot take an action if the index fails to load, that's why I don't think it's desirable to go from a loading state (where it is clear that I cannot take an action) to an empty table (where it is unclear if I can take any actions)

IIUC, is having the loading spinner remain visible in the error state the intended functionality? If so, I'll be happy to close this and close out the ticket.

fboulnois: Until we have a more defined approach to dealing with `index` failures (perhaps replacing the spinner with an error icon?), the spinner-as-error was the intended failure mode.

PR #2710, opened 2024-11-05
[DT-934] Create a banner similar to the Terra Banner on DUOS to inform users that chrome is our preferred browser

cinyecai:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-934
### Summary
Displays a warning alert for users opening DUOS on a browser that's not Chrome or Firefox, user cannot click the Sign In button until Alert is closed

https://github.com/user-attachments/assets/08a8e1b3-d60a-40a4-ba87-d2b0057a7aed

**Note:** Updated the appearance of the alert to be less extreme - see comment below

----

- Verify all tests go green


src/outdated-browser-message.js
```diff
@@ -0,0 +1,30 @@
...
+    Firefox: 60,
+    Vivaldi: false,
+    IE: false,
+  },
+  isUnknownBrowserOK: true,
+  messages: {
+    en: {
+      outOfDate: 'DUOS may not function correctly in this browser.',
+      update: {
+        web: `If you experience issues, please try ${window.chrome ? 'updating' : 'using'} Google Chrome.`,
```
pshapiro4broad: Does `window.chrome` generate a warning when compiled? It was flagged in the editor. If so, I wonder if there's a good / easy way to fix the warning.

pshapiro4broad (approved):
Looks good

src/outdated-browser-message.js
cinyecai: I don't see any warnings about this particular file when compiled

package.json
```diff
@@ -21,6 +21,7 @@
     "moment": "2.30.1",
     "noty": "3.2.0-beta",
     "oidc-client-ts": "3.1.0",
+    "outdated-browser-rework": "^3.0.1",
```
fboulnois: This package was last published 4 years ago and depends on another package that had multiple major revisions (and potentially security fixes).

I think we should evaluate whether we can do this without depending on an external package, and use our existing notification machinery to warn users.

fboulnois (dismissed):
See below:

package.json
rushtong: Agreed on this. Additionally, we generally pin versions explicitly due to historic problems with npm point updates.

src/outdated-browser-message.js
```diff
@@ -0,0 +1,30 @@
+import 'outdated-browser-rework/dist/style.css';
```
rushtong: This style, while it does look like Terra, is extreme. I'd like to ensure we get PM feedback/approval before merging.

public/index.html
```diff
@@ -40,6 +40,7 @@
     You need to enable JavaScript to run this app.
   </noscript>
   <div id="root"></div>
+  <div id="outdated"></div>
```
rushtong: Note that `cypress/component-index.html` looks a lot like this file so that component tests can have the same basic index to work with. I would love to see a test for this if possible.

package.json
cinyecai: I could probably do it using the Alert already in Duos, figuring out how to detect the browser being used will be the harder part - I'll look into it

cinyecai: @fboulnois @rushtong After some searching, I found an alternative approach could be to detect the browser using the user agent but this seems generally to be recommended **against** for being unreliable. I'm not sure if this would be better or worse than using this package, and I'm struggling to find other straightforward ways to do browser detection. Let me know your thoughts!

rushtong: This looks promising as an alternative: https://github.com/pure-js/browser-detection

cinyecai:
Updated visual:


https://github.com/user-attachments/assets/0e5ffa5f-e506-4200-9ea5-99cc35191ce2

package.json
```diff
@@ -14,6 +14,7 @@
     "ajv": "8.17.1",
     "ajv-formats": "3.0.1",
     "axios": "1.7.7",
+    "browser-version-detection": "^2.1.8",
```
fboulnois, line 17:
```suggestion
    "browser-version-detection": "2.1.8",
```

We should lock this version down so it doesn't automatically change.

This version hasn't seen [any updates for 7 years](https://www.npmjs.com/package/browser-version-detection), and the implementation is [straightforward](https://github.com/pure-js/browser-detection/blob/main/src/detect-browser-version.ts), so I would continue to advocate not depending on a package to do this and instead using a simple regex for browser detection (which is ultimately what this library and the previous one do).

In general, [browser detection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent) is not recommended, although I know we also do this on Terra UI. It might be better to identify the specific features that Safari doesn't support, and long term work to either stop depending on these features or use polyfills for those features that we really need. This is a separate issue from showing the alert, so in the meantime we might still want to show the alert regardless of whether we are doing browser or feature detection.

fboulnois (commented):
See below:

package.json
rushtong: I'm a little confused here ... the latest release version I see is [3.1.0](https://github.com/pure-js/browser-detection/releases) that is from last year. Is there a reason we're looking at 2.1.8 (which I don't even see in the list of releases)? I do see that NPM sees [2.1.8](https://www.npmjs.com/package/browser-version-detection), which is extra confusing.

rushtong: Aside from the library version question, I do agree in theory with Florian's feedback. To date, the only issue we've seen with Safari is variations in CSS rendering. I don't think we're using any browser-specific features, so to really tackle that question, we'd need a different kind of investigation effort.

PR #2711, opened 2024-11-05
DT-903: Change export button to link directly to terra-ui import page

pshapiro4broad:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-903

Instead of performing a two-step operation in the duos UI, use the new TDR import link to have terra-ui perform the TDR export step. The same APIs are called in both cases, so other than the UX differences there is no difference in behavior.

### Summary

Video clip showing behavior, note that this is trying to export an azure dataset so the final step is blocked, but everything else is working correctly.

https://github.com/user-attachments/assets/859bb8dd-171e-4b4e-91ca-51961eded4b3

### DIFF
--- a/src/components/data_search/DatasetExportButton.jsx
+++ b/src/components/data_search/DatasetExportButton.jsx
@@ -1,53 +1,22 @@
 import * as React from 'react';
-import { CircularProgress, IconButton, Link } from '@mui/material';
-import { useState } from 'react';
-import IosShareIcon from '@mui/icons-material/IosShare';
-import { TerraDataRepo } from '../../libs/ajax/TerraDataRepo';
+import {useEffect, useState} from 'react';
+import {Link} from '@mui/material';
+import {Config} from '../../libs/config';
 
 export const DatasetExportButton = (props) => {
-  const { snapshot, title } = props;
-  // The exportStatus flow is: initial -> prepping -> ready
-  // TODO: error handling?
-  const [exportStatus, setExportStatus] = useState('initial');
-  const [exportResult, setExportResult] = useState(null);
+  const {snapshot, title} = props;
 
-  // Not a supported export location
-  if (!snapshot) {
-    return null;
-  }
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
love the simplification 👍

rushtong (approved):
I can confirm that this works for the known positive case and fails in the known auth-domain case.

Working example:

Screenshot 2024-11-05 at 12 55 41 PM

Failing example:
Screenshot 2024-11-05 at 12 55 50 PM

I am seeing the export link twice when there are multiple snapshots, so we still need to address that at some point:
Screenshot 2024-11-05 at 12 57 58 PM

fboulnois:
> I am seeing the export link twice when there are multiple snapshots, so we still need to address that at some point:

At a minimum we probably want a bit of spacing between the links.

src/components/data_search/DatasetExportButton.jsx
```diff
@@ -1,53 +1,22 @@
...
-  }
-
-  if (exportStatus === 'ready') {
-    return (
-      <Link href={exportResult.terraImportLink} target="_blank" rel="noopener noreferrer" title={title} aria-label={title}>Export</Link>
-    );
-  }
-
-  return null;
+  const link = `${terraUrl}/#import-data?snapshotId=${snapshot.id}&format=tdrexport&tdrSyncPermissions=false`;
```
pshapiro4broad, line 17: Here's the code in TDR which also generates this link for import https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java#L475

The link's behavior is implemented in terra-ui: https://github.com/DataBiosphere/terra-ui/blob/dev/src/import-data/useImportRequest.ts#L95

pshapiro4broad:
I've added spacing between the links. There's not much room for multiple Exports, but we have other tickets for that issue, so I'm not going to address that now.

image

PR #2712, opened 2024-11-05
DT-966: Partition TDR Snapshot Query

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-966

### Summary
This PR breaks up the TDR snapshot query and re-combines the results to keep the url length manageable.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/TerraDataRepo.js
+++ b/src/libs/ajax/TerraDataRepo.js
@@ -16,16 +16,20 @@ export const TerraDataRepo = {
       roleMap: {},
       errors: []
     };
-    for await (const sublist of partitionedIdentifiers) {
-      // Note that TDR is expecting dataset identifiers, not dataset ids
-      const url = `${await Config.getTdrApiUrl()}/api/repository/v1/snapshots?duosDatasetIds=${sublist.join('&duosDatasetIds=')}`;
-      const res = await axios.get(url, Config.authOpts());
-      enumerateSnapshotModel.total = res.data.total;
-      enumerateSnapshotModel.filteredTotal += res.data.filteredTotal;
-      Object.assign(enumerateSnapshotModel.roleMap, res.data.roleMap);
-      enumerateSnapshotModel.items.push(...res.data.items);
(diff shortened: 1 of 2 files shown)


s-rubenstein (approved):
Looks reasonable to me.

src/libs/ajax/TerraDataRepo.js
```diff
@@ -1,12 +1,31 @@
...
+    // ~70 should a safe default at 1890 characters
+    const partitionedIdentifiers = partition(identifiers, 70);
+    let enumerateSnapshotModel = {
+      total: 0,
+      filteredTotal: 0,
+      items: [],
+      roleMap: {},
+      errors: []
+    };
+    for await (const sublist of partitionedIdentifiers) {
```
pshapiro4broad: This makes each async call synchronously. Since each call is independent, they could all be made in parallel which should make this code faster. Would it be hard to change this to use a `Promise.all()` construct instead?

pshapiro4broad (approved):
Looks ok, but it'd be better to perform the API calls in parallel.

Also, I've mentioned this before, but this fix changes one API call to multiple calls, and could run into scale issues as the number of datasets grows. A solution that wouldn't have this issue is to add a POST version of this API to TDR.

src/libs/ajax/TerraDataRepo.js
rushtong: > This makes each async call synchronously. Since each call is independent, they could all be made in parallel which should make this code faster. Would it be hard to change this to use a `Promise.all()` construct instead?

Yes, I think that shouldn't be too much of a lift

rushtong: @pshapiro4broad - PTAL, the requests are now all executed in parallel.

rushtong:
> Also, I've mentioned this before, but this fix changes one API call to multiple calls, and could run into scale issues as the number of datasets grows. A solution that wouldn't have this issue is to add a POST version of this API to TDR.

I created https://broadworkbench.atlassian.net/browse/DT-974 to cover this suggestion.

PR #2715, opened 2024-11-06
DT-781: Use logical AND to combine filters in the DUOS UI

pshapiro4broad:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-781

Users would like filters in the DUOS UI to combine as a set intersection operation, so an item is only shown if it's valid for all selected filters.

### Summary

By changing the root node of filter query to use `must` instead of `should`, the filters are combined using logical `AND` instead of `OR`.

Video of feature in action

https://github.com/user-attachments/assets/e401d739-9c74-4067-81f5-9920a0c69da6

### DIFF
--- a/src/components/data_search/DatasetSearchTable.jsx
+++ b/src/components/data_search/DatasetSearchTable.jsx
@@ -104,12 +104,12 @@ export const DatasetSearchTable = (props) => {
       queryChunks.push(...searchModifier);
     }
 
-    var filterQuery = {};
+    let filterQuery = {};
     if (numSelectedFilters(filters) > 0) {
-      const shouldTerms = [];
+      const filterTerms = [];
 
       filters.accessManagement.forEach(term => {
-        shouldTerms.push({
+        filterTerms.push({
           'term': {
             'accessManagement': term
           }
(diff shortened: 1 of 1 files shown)


src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -104,60 +104,48 @@ export const DatasetSearchTable = (props) => {
       queryChunks.push(...searchModifier);
     }
 
-    var filterQuery = {};
+    let filterQuery = {};
```
pshapiro4broad, line 107: I think `var` is not recommended now that we have `let` and `const`. The difference is that `var` has function scope, whereas `let` and `const` have block scope, which is usually what you want.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -104,60 +104,48 @@ export const DatasetSearchTable = (props) => {
...
-      };
-    } else {
-      return {
-        'from': 0,
-        'size': 10000,
-        'query': {
-          'bool': {
-            'must': queryChunks
-          }
+    return {
```
pshapiro4broad, line 144: This change wasn't required, but I thought it would be nice  to
- combine the code fragments to avoid duplication
- test `filterQuery` instead of recomputing `numSelectedFilters(filters)`, since `filterQuery` is the value actually used in the code block

rushtong (approved):
👍🏽

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -104,60 +104,48 @@ export const DatasetSearchTable = (props) => {
...
         });
       });
 
-      if (shouldTerms.length > 0) {
+      if (filterTerms.length > 0) {
         filterQuery = [
           {
             'bool': {
-              'should': shouldTerms
+              'must': filterTerms
```
fboulnois, line 137: I think we want to OR for selections within a filter, and AND selections between filters. For example, if you select both `open` and `external`, this will never match a dataset, since it is not logically possible for a dataset to have both access management properties simultaneously.

One approach to solve this problem is to store an array of arrays. The selections within a filter are `should`ed together, and then this is rolled up such that all the filters are `must`ed together.

fboulnois (changes requested):
See below:

src/components/data_search/DatasetSearchTable.jsx
pshapiro4broad: That's a good point. I noticed this while testing, but wasn't sure if the filter categories were always exclusionary, e.g. what about studies that have both open and controlled access data, or have all controlled access data but contain datasets with different data use codes?

OTOH I agree that I'd expect a filter on an e-commerce site to work in this way so that alone seems like an argument to use OR within a filter group.

pshapiro4broad: I've updated the code to use OR within a filter group

rushtong (approved):
👍🏽

fboulnois (approved):
looks good, thank you! 👍

PR #2717, opened 2024-11-07
[DT-744] Add option to modify email preferences

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-744

### Summary

Add option to modify email preferences:

https://github.com/user-attachments/assets/cba8b194-0f6b-4b29-9728-d7d9e596265d

----

- Verify all tests go green

### DIFF
--- a/cypress/component/UserProfile/user_profile.spec.js
+++ b/cypress/component/UserProfile/user_profile.spec.js
@@ -3,6 +3,8 @@
 import {mount} from 'cypress/react';
 import React from 'react';
 import {Storage} from '../../../src/libs/storage';
+import {User} from '../../../src/libs/ajax/User';
+import {Institution} from '../../../src/libs/ajax/Institution';
 import UserProfile from '../../../src/pages/user_profile/UserProfile';
 
 const duosUser = {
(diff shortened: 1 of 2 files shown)


rushtong (approved):
This looks great, thank you! I do see that there are no unit tests for this page, it might be nice to add one with a minimal amount of coverage so we can start fleshing it out in future work. Alternatively, we can ticket fully testing it which is not a small amount of work.

lstrano-broad:
LGTM - thx!

src/pages/user_profile/UserProfile.jsx
```diff
@@ -27,6 +27,7 @@ export default function UserProfile(props) {
   const [profile, setProfile] = useState({
     profileName: '',
     email: undefined,
+    emailPreference: false,
```
pshapiro4broad: Why `false` instead of `undefined`, as is used for `id` and `email`?

I would expect this page's content to block until the profile is loaded, is that not the case here?

src/pages/user_profile/UserProfile.jsx
```diff
@@ -179,7 +194,22 @@ export default function UserProfile(props) {
...
+      style={{
+        color: '#000',
+        fontSize: '16px',
+        fontWeight: '400',
+      }}
+    >
+      Send me email notifications
+    </p>
+    <FormField
+        type={FormFieldTypes.YESNORADIOGROUP}
```
pshapiro4broad, line 208: If the choices are Yes/No, does this need to use a radio button? In a case like this, I think a checkbox is equally clear and takes up less screen space

pshapiro4broad (approved):
The code look OK, one question about the UX.

cypress/component/UserProfile/user_profile.spec.js
```diff
@@ -15,4 +17,15 @@ describe('User Profile', () => {
     mount(<UserProfile/>);
     cy.get('h2').should('contain', 'Your Profile');
   });
+
+  it('Updates the user email preferences', () => {
```
fboulnois, line 34: to run these tests: `npm run cypress run --component --spec "**/user_profile.spec.js"`

cypress/component/UserProfile/user_profile.spec.js
```diff
@@ -0,0 +1,37 @@
...
+const duosUser = {
+    isSigningOfficial: false,
+};
+
+describe('User Profile', () => {
+  it('Renders the user profile page', () => {
+    cy.stub(Storage, 'getCurrentUser').returns(duosUser);
+    mount(<UserProfile/>);
+    cy.get('h2').should('contain', 'Your Profile');
+  });
```
rushtong: This test is making live calls to consent. I think you need to stub the same calls as you have in the next test:
```
    cy.stub(Institution, 'list').returns([]);
    cy.stub(User, 'getMe').returns(duosUser);
    cy.stub(User, 'getApprovedDatasets').returns([]);
    cy.stub(User, 'getAcknowledgements').returns({});
```

rushtong: Alternatively, the duplicate stubs could be in a `beforeEach`.

PR #2718, opened 2024-11-07
DT-530: Add conditional rendering around apply for access

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-530

## Summary
**Note to developers**: I suggest hiding whitespace due to the large number of lint changes. I modified the datasets seen in the screenshots so they have URL values. Most dev datasets do not have them.

This PR makes the following modifications to the individual dataset page:

1. Display the Dataset Identifier instead of the `alias` (this is a bug fix)
2. If Controlled Access, no changes
3. If Open Access, render specific text as seen in the ticket/mocks.
4. If External Access, render specific text as seen in the ticket/mocks.
5. New tests for this component and the 5 conditions around access.

In the Open/External cases, we only render a link (and relevant link language) if one exists as a URL property on the dataset.

### Controlled Access (no changes)
Screenshot 2024-11-07 at 12 25 20 PM

### Open Access
Screenshot 2024-11-07 at 12 24 29 PM

Screenshot 2024-11-07 at 12 26 49 PM

### External Access
Screenshot 2024-11-07 at 12 22 45 PM

Screenshot 2024-11-07 at 12 24 16 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetStatistics.jsx
+++ b/src/pages/DatasetStatistics.jsx
@@ -97,7 +97,7 @@ export default function DatasetStatistics(props) {
             <div style={Styles.TITLE}>Dataset Statistics</div>
             <div style={Styles.MEDIUM_ROW}>
               <div style={{fontWeight: '500', marginRight: '5px'}}>Dataset ID:</div>
-              <div>{dataset?.alias}</div>
+              <div>{dataset?.datasetIdentifier}</div>
             </div>
             <div style={Styles.MEDIUM_ROW}>
               <div style={{fontWeight: '500', marginRight: '5px'}}>Dataset Name:</div>
--- a/src/pages/DatasetStatistics.jsx
+++ b/src/pages/DatasetStatistics.jsx
@@ -9,6 +9,7 @@ import {find} from 'lodash/fp';
 import {ReadMore} from '../components/ReadMore';
 import {formatDate} from '../libs/utils';
 import {Button} from '@mui/material';
+import {AccessManagement} from '../libs/Models.ts';
 
 const LINE = <div style={{borderTop: '1px solid #BABEC1', height: 0}}/>;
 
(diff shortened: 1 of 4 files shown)


src/pages/DatasetStatistics.jsx
```diff
@@ -57,57 +57,82 @@ export default function DatasetStatistics(props) {
...
     } catch (error) {
-      Notifications.showError({ text: 'Error: Unable to retrieve dataset statistics from server' });
+      Notifications.showError({text: 'Error: Unable to retrieve dataset statistics from server'});
       setIsLoading(false);
     }
   };
 
+  const accessButton = () => {
+    const accessManagement = extract('Access Management').toLowerCase();
+    if (accessManagement === 'controlled') {
```
pshapiro4broad: Would a `switch` work here? I think using `switch` would make it clearer that each `if` is comparing against the same thing, and that each `if` clause is terminal.

If you used a functional style switch (like lodash https://lodash.com/docs/4.17.15#cond), you could say `return _.cond(...);`.

src/pages/DatasetStatistics.jsx
```diff
@@ -57,57 +57,82 @@ export default function DatasetStatistics(props) {
...
+    }
+    if (accessManagement === 'external') {
+      return <span>
+        This dataset is externally managed. Requests cannot be made via DUOS
+        {locationUrl &&
+          <span>, but must be made directly through the <a href={locationUrl}>dataset&apos;s host repository</a>.</span>
+        }
+      </span>;
+    }
+    return <div/>;
```
pshapiro4broad: I see `accessManagement` is type `string`, do we expect other values here?

src/pages/DatasetStatistics.jsx
```diff
@@ -57,57 +57,82 @@ export default function DatasetStatistics(props) {
       setDars(metrics.dars);
       setIsLoading(false);
     } catch (error) {
-      Notifications.showError({ text: 'Error: Unable to retrieve dataset statistics from server' });
+      Notifications.showError({text: 'Error: Unable to retrieve dataset statistics from server'});
       setIsLoading(false);
     }
   };
 
+  const accessButton = () => {
```
pshapiro4broad: Is this name correct? It only returns a button in one case.

rjohanek (approved):
changes and new text looks good. would be nice to add tests for these new different cases.

rushtong:
> would be nice to add tests for these new different cases.

Yes, I should do that!

pshapiro4broad (approved):
Looks good, a couple minor things

src/pages/DatasetStatistics.jsx
rushtong: Just string values come through. On the consent side, [we do have an enum](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/ConsentGroup.java#L815) so we can actually enforce this here as well.

src/pages/DatasetStatistics.jsx
rushtong: Yes, this could be cleaner with a switch. I will look into `cond` - have not used that before.

src/pages/DatasetStatistics.jsx
rushtong: I'll rename it to make it more reflective of what it's doing now that it's not just a button any more.

src/pages/DatasetStatistics.jsx
rushtong: I found some unintended behavior with using `cond` (hrefs were being removed from anchor elements) so I stuck with a switch approach.

s-rubenstein (approved):

PR #2719, opened 2024-11-07
[DT-980] Fix the enumerate snapshots to get more snapshots than the first 10

s-rubenstein:
### Addresses
This makes it so we load snapshots only once to avoid debounce issues and ensure that we have a comprehensive list of what can export always.

It also bumps up the limits on snapshots from 10 (the default) to 1000 - this should be a safe number given that we partition DUOS IDs into 70 at a time.
### Summary

----

- Verify all tests go green


rushtong (approved):
Nice catch on the limit parameter 👍🏽

fboulnois (approved):
good catch 👍

PR #2720, opened 2024-11-07
[DT-778] Filter by Data Access Committee

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-778

### Summary

Adds a filter for the Data Access Committee

https://github.com/user-attachments/assets/9dba680c-658c-4072-a51b-56327f377099

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetFilterList.jsx
+++ b/src/components/data_search/DatasetFilterList.jsx
@@ -15,7 +15,7 @@ export const DatasetFilterList = (props) => {
 
   const accessManagementFilters = uniq(compact(datasets.map((dataset) => dataset.accessManagement)));
   const dataUseFilters = uniq(compact(flatten(datasets.map((dataset) => dataset.dataUse?.primary))).map((dataUse) => dataUse.code));
-  const dacFilters = orderBy(uniq(compact(datasets.map((dataset) => dataset?.dac?.dacName))), (dac) => dac.toLowerCase(), 'asc');
+  const dacFilters = orderBy(uniq(compact(datasets.map((dataset) => dataset.dac?.dacName))), (dac) => dac.toLowerCase(), 'asc');
 
   return (
     <Box sx={{ bgcolor: 'background.paper' }}>
--- a/src/components/data_search/DatasetFilterList.jsx
+++ b/src/components/data_search/DatasetFilterList.jsx
@@ -10,6 +10,38 @@ import { Typography } from '@mui/material';
 import { Checkbox } from '@mui/material';
 import { flatten, uniq, compact, capitalize, orderBy } from 'lodash';
 
+export const FilterItemHeader = (props) => {
+  const { title, headerStyle = { fontFamily: 'Montserrat', fontWeight: '600', marginTop: '1em' } } = props;
+  return (
+    <Typography variant="h6" gutterBottom component="div" sx={headerStyle}>
+      {title}
+    </Typography>
+  );
+}
+
+export const FilterItemList = (props) => {
+  const { category, datasets, filter, filterHandler, isFiltered, filterNameFn } = props;
+  return (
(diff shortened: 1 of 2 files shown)


src/components/data_search/DatasetFilterList.jsx
```diff
@@ -8,13 +8,14 @@ import ListItemText from '@mui/material/ListItemText';
...
 import { Checkbox } from '@mui/material';
-import { flatten, uniq, compact, capitalize } from 'lodash';
+import { flatten, uniq, compact, capitalize, orderBy } from 'lodash';
 
 export const DatasetFilterList = (props) => {
   const { datasets, filters, filterHandler, isFiltered } = props;
 
   const accessManagementFilters = uniq(compact(datasets.map((dataset) => dataset.accessManagement)));
   const dataUseFilters = uniq(compact(flatten(datasets.map((dataset) => dataset.dataUse?.primary))).map((dataUse) => dataUse.code));
+  const dacFilters = orderBy(uniq(compact(datasets.map((dataset) => dataset?.dac?.dacName))), (dac) => dac.toLowerCase(), 'asc');
```
fboulnois: this does a case-ignore sort

rushtong (approved):
This looks really nice 👍🏽

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -131,6 +132,17 @@ export const DatasetSearchTable = (props) => {
         }
       });
 
+      filterTerms.push({
```
pshapiro4broad, line 135: With only two cases I didn't think it was worth refactoring, but with three maybe it's time? Something like
```
   const addTerms = (createFilters) => {
      filterTerms.push({
        'bool': {
          'should': createFilters()
        }
      });
  }

  addTerms(() => {
     filters.accessManagement.map(term => ({
                'term': {
                  'accessManagement': term
                }
              }))
  );
  addTerms(...);
  addTerms(...);
```

src/components/data_search/DatasetFilterList.jsx
pshapiro4broad: Why does this use `dataset?` but other cases don't? It seems like it should be consistent here. (Also, it would be nice to move to tsx to have better type information here.)

src/components/data_search/DatasetFilterList.jsx
```diff
@@ -64,6 +65,27 @@ export const DatasetFilterList = (props) => {
           })
         }
       </List>
+      <Typography variant="h6" gutterBottom component="div" sx={{ fontFamily: 'Montserrat' }} marginTop="1em">
```
pshapiro4broad: Could these elements be refactored to avoid code duplication? They look pretty similar to me.

pshapiro4broad (approved):
Looks OK, just a few minor comments

src/components/data_search/DatasetFilterList.jsx
fboulnois: Not all datasets will have DACs, but all datasets will have a primary data use and an access management policy.

src/components/data_search/DatasetFilterList.jsx
fboulnois: I definitely want to do this long term, let me see how much work it would be.

src/components/data_search/DatasetSearchTable.jsx
fboulnois: Hrm, I'm not sure this is clearer, and it's possible that we might have a filter that requires a `must` or some other boolean term or a more complex logic statement.

src/components/data_search/DatasetFilterList.jsx
fboulnois: It wasn't too challenging, I created two new components.

PR #2722, opened 2024-11-12
DT-979: Add Appcues Support

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-979

## Summary
This PR adds Appcues support to DUOS. This borrows heavily from Terra's implementation, similarly to how we borrow our Metrics implementation. See how Terra does this:
* Metrics: https://github.com/DataBiosphere/terra-ui/blob/048788c71c22bd524dc6a58e1516746f4dec7ec5/src/libs/ajax/Metrics.ts#L14
* Appcues: https://github.com/DataBiosphere/terra-ui/blob/048788c71c22bd524dc6a58e1516746f4dec7ec5/src/auth/auth.ts#L296

See also: https://github.com/broadinstitute/terra-helmfile/pull/5951

### Notable Features
* Typescriptify `Metrics`
  * Related to that, we needed a custom type for `@databiosphere/bard-client` since there is only a js client.
* Typescriptify `events`
* We now register an `Appcues` user on any sign-in or registration
* We now register an `Appcues` tracking event on any tracked `Metrics` event

### Local Testing
If you have [an account in Appcues](https://studio.appcues.com/), you can log in and see events in the "Events Explorer" page. The examples I have here are from signing in locally with a dev account.

Note that when I was testing locally, I was sending the Sam subject id. Since that time, I've updated the code to send the same OIDC user profile `sub` value that Terra uses.

Screenshot 2024-11-12 at 12 48 57 PM

Screenshot 2024-11-12 at 12 52 05 PM

----

- Verify all tests go green


public/index.html
```diff
@@ -22,7 +22,6 @@
   <link rel="icon" type="image/png" href="%PUBLIC_URL%/images/favicon/favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="%PUBLIC_URL%/images/favicon/favicon-16x16.png" sizes="16x16" />
   <link rel="mask-icon" href="%PUBLIC_URL%/images/favicon/safari-pinned-tab.svg">
-  <script src="https://accounts.google.com/gsi/client" async defer id="gsiClient"></script>
```
rushtong, line 25: Drive by fix - we removed "google sign in" in https://github.com/DataBiosphere/duos-ui/pull/2667

src/libs/ajax/Metrics.ts
```diff
@@ -1,15 +1,24 @@
...
+import axios, {AxiosRequestConfig} from 'axios';
 import {getDefaultProperties} from '@databiosphere/bard-client';
 
 import {Storage} from '../storage';
 import {getBardApiUrl} from '../ajax';
 import {Token} from '../config';
+import {MetricsEventName} from 'src/libs/events';
+
+// Set default timeout for all metrics calls to 30 seconds
+const defaultSignal: AbortSignal = AbortSignal.timeout(30000);
```
rushtong, line 10: Note that this was necessary for typescriptification to make all the axios calls compile. I'm open to alternative suggestions here but this seems like a safe timeout for bard & appcues calls.

src/components/SignInButton.tsx
```diff
@@ -107,9 +107,9 @@ export const SignInButton = (props: SignInButtonProps) => {
     history.push(`/tos_acceptance${shouldRedirect ? `?redirectTo=${redirectTo}` : ''}`);
   };
 
-  const syncSignInOrRegistrationEvent = async (event: String) => {
+  const syncSignInOrRegistrationEvent = async (event: MetricsEventName) => {
     Storage.setAnonymousId();
-    await Metrics.identify(Storage.getAnonymousId());
+    await Metrics.identify(`${Storage.getAnonymousId()}`);
```
s-rubenstein: If its just the value, why do you need to add the quotes?

src/libs/ajax/Metrics.ts
```diff
@@ -18,12 +27,19 @@ export const Metrics = {
  * @param {string} event - The event name.
  * @param {Object} [details={}] - The event details.
  * @param {AbortSignal} [signal] - The abort signal.
+ * @param refreshAppcues - The refresh Appcues flag.
  * @returns {Promise} - A Promise that resolves when the event is captured.
  */
-const captureEventFn = async (event, details = {}, signal) => {
+const captureEventFn = async (event: MetricsEventName, details: object = {}, signal: AbortSignal, refreshAppcues: boolean): Promise<any> => {
```
s-rubenstein: I think you can shorthand `object` as `{}` (so `details: {} = {}`)

s-rubenstein (approved):
Nothing here looks unreasonable!

src/components/SignInButton.tsx
rushtong: Typescript complains that it has to be a string, but `Storage.getAnonymousId()` is not typed.

cypress/support/component-index.html
```diff
@@ -5,8 +5,11 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width,initial-scale=1.0">
-  <script src="https://accounts.google.com/gsi/client" async defer id="gsiClient"></script>
```
rushtong, line 8: Another drive by fix - we removed "google sign in" in https://github.com/DataBiosphere/duos-ui/pull/2667

cypress/support/component-index.html
```diff
@@ -5,8 +5,11 @@
...
   <title>Components App</title>
+  <!-- AppCues Requirements  -->
+  <script type="text/javascript">
+    window.AppcuesSettings = { enableURLDetection: true };
+  </script>
```
rushtong, line 12: I'm open to testing suggestions here. I was unable to figure out a way to unit test the global `Appcues` window object.

s-rubenstein: I think at some point it is worth trusting the library. We could add tests that verify we're setting the right values, but as far as appcues itself, I think that's probably out of our purview to test? But happy to have that contested

fboulnois (approved):
looks good 👍

snf2ye (approved):

PR #2724, opened 2024-11-13
DT-988: Remove awaits for Metrics calls

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-988

## Summary
Remove all of the awaits for Metrics calls. We don't need to block the page load on these while we send usage stats to Bard.

### Why
Bard dev has been experiencing degraded performance which uncovered this bug in DUOS. We were await-ing all Metrics calls before continuing with the loading of the user's requested page. In dev, this delay hit 1 full minute per bard call. On sign-in, we need to make two calls, one to sync the user and one to register the sign-in or registration event.

### Follow-on Work
While investigating the use of Cypress intercepts, it became clear that we need to provide a systemic way of intercepting the application `config.json` file. It is not easily stub-able, but can be intercepted. This PR doesn't try to solve that problem globally, but does provide an example custom command that can be re-used in cypress loading code.

----

- Verify all tests go green


cypress/component/SignIn/sign_in_button.spec.js
```diff
@@ -42,14 +58,15 @@ describe('Sign In: Component Loads', function () {
   it('Sign In: On Success', function () {
     cy.viewport(600, 300);
     cy.stub(Auth, 'signIn').returns(Promise.resolve(mockOidcUser));
-    cy.stub(User, 'getMe').returns(duosUser);
+    cy.intercept({method: 'GET', url: '**/api/user/me'}, {statusCode: 200, body: duosUser}).as('getMe');
```
rjohanek, line 56: can you explain why we need to move from stubbing these calls to intercepting them?

rushtong: Previously, we were calling `then` on the `click` event. This would take advantage of the stubbing. That pattern is discouraged by Cypress (i.e., it is unsafe to chain calls off of click: https://docs.cypress.io/api/commands/click). So, to get around that unsafeness, instead we use the intercept + wait pattern seen on line 64 here. To do that, we can't use a stub, we have to wait for the result of the `getMe` call, and then check for the expectations we want to see.

rjohanek (approved):
looks good thanks for explaining the testing changes

s-rubenstein (approved):

PR #2726, opened 2024-11-14
[DT-776] Filter by Data Type

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-776

### Summary

Adds a filter for the Data Type (note that in dev we have a bunch of weird selections)

https://github.com/user-attachments/assets/2f160c31-000a-4de3-a907-58b6cfb0676d

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DataSearch/dataset_search_filters.spec.js
+++ b/cypress/component/DataSearch/dataset_search_filters.spec.js
@@ -1,0 +1,27 @@
+/* eslint-disable no-undef */
+
+import { mount } from 'cypress/react';
+import React from 'react';
+import {Storage} from '../../../src/libs/storage';
+import DatasetFilterList from '../../../src/components/data_search/DatasetFilterList';
+
+const duosUser = {
+    isSigningOfficial: false,
+};
+
+describe('Data Library Filters', () => {
+    // Intercept configuration calls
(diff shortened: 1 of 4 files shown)


s-rubenstein (commented):
How many data types do we expect? I know the long list is on dev, are we expecting much fewer on prod? If it is going to be long, I'd love to think about limiting the amount shown, and figuring out a `see more` story.

fboulnois:
> How many data types do we expect? I know the long list is on dev, are we expecting much fewer on prod? If it is going to be long, I'd love to think about limiting the amount shown, and figuring out a `see more` story.

We have 8 by default, but a user can add their own if they really want (although not all 8 may be shown since it's dynamically populated from the datasets). I agree that long term if the list gets long we should add a separate story to "see more".

s-rubenstein:
Great. I'll flag that as a potential future ticket for Victoria

s-rubenstein (approved):
LGTM

rjohanek (approved):
this looks good to me, although is there any kind of testing we want to do here?

fboulnois:
> this looks good to me, although is there any kind of testing we want to do here?

Good idea, let me work on increasing our unit test coverage.

PR #2733, opened 2024-11-21
[DT-777] Filter by participant count

rjohanek:
### Addresses
Jira Ticket: https://broadworkbench.atlassian.net/browse/DT-777

### Summary
Adds a filter for participant size 

DatasetSearchTable:
- generalize the filter handler to work for non-Array filters
- generalize the numFiltersSelected method to work for non-Array filters
-  update assembleFullQuery to include range chunk
- rename isFiltered to isFilteredArray because it is only used for array filters
- remove unused arguments to filterHandler and unused prop to DatasetFilterList

DatasetFilterList
- add new filter range constant
- add new filter to the filter list

### Testing
Unit tests
Note: cy.clock() is needed to mock the clock for debounced calls

### Visual

https://github.com/user-attachments/assets/d614e12b-8685-4a4d-a4ef-aa43cf232804

----

- Verify all tests go green


src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -49,8 +51,21 @@ export const DatasetSearchTable = (props) => {
   const [selectedTable, setSelectedTable] = useState(datasetSearchTableTabs.study);
   const [searchTerm, setSearchTerm] = useState('');
 
-  const isFiltered = (filter, category) => (filters[category]).indexOf(filter) > -1;
-  const numSelectedFilters = (filters) => Object.values(filters).reduce((sum, array) => sum + array.length, 0);
+  const isFilteredArray = (filter, category) => (filters[category]).indexOf(filter) > -1;
+
+  const numSelectedFilters = (filters) => {
```
pshapiro4broad: This could still use `reduce`, is there a reason not to use it here? It would look something like
```
filters.reduce((sum, category) => sum + isArray(filters[category]) ? filters[category].length : 1, 0)
```

I find `reduce()` clearer to read than using a `for`, as I don't have to read the code to understand how `sum` is generated at the end. It also avoids mutable state which I find easier to think about.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -49,8 +51,21 @@ export const DatasetSearchTable = (props) => {
...
+    var sum = 0;
+    for (const category in filters) {
+      if (isArray(filters[category])) {
+        sum += filters[category].length;
+      } else {
+        if (filters[category]) {
```
pshapiro4broad: I would combine an `if` after an `else` here
```suggestion
      } else if (filters[category]) {
```

src/components/data_search/DatasetSearchTable.jsx
pshapiro4broad: Looking at this further, the only use of this value is `if (numSelectedFilters(filters) > 0) {` below. Since that's the same as `anySelected`, replacing this code to instead find the first selected filter might even be simpler

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -155,6 +170,15 @@ export const DatasetSearchTable = (props) => {
         }
       });
 
+      filterTerms.push({
+        'range': {
+          'participantCount': {
+            'gte': filters.participantCountMin,
+            'lte': filters.participantCountMax,
```
pshapiro4broad, line 168: Is the plan to skip validation for now? These are text entry fields, so they might not be numbers. A user could enter min > max. They could enter negative numbers, etc.

Not sure if it was brought up during ticket refinement, but using a slider control could avoid the need for validation here.

src/components/data_search/DatasetFilterList.jsx
```diff
@@ -45,13 +45,34 @@ export const FilterItemList = (props) => {
...
 export const DatasetFilterList = (props) => {
-  const { datasets, filters, filterHandler, isFiltered, onClear } = props;
+  const { datasets, filterHandler, isFiltered, onClear } = props;
 
   const accessManagementFilters = uniq(compact(datasets.map((dataset) => dataset.accessManagement)));
   const dataUseFilters = uniq(compact(flatten(datasets.map((dataset) => dataset.dataUse?.primary))).map((dataUse) => dataUse.code));
   const dataTypeFilters = uniq(flatten(datasets.map((dataset) => dataset.study.dataTypes)));
   const dacFilters = orderBy(uniq(compact(datasets.map((dataset) => dataset.dac?.dacName))), (dac) => dac.toLowerCase(), 'asc');
+  // some participantCounts are undefined, so filter them out before calculating min and max
+  const participantCountMax = Math.max(...datasets.filter((dataset) => dataset.participantCount).map((dataset) => dataset.participantCount));
```
pshapiro4broad: I'm wondering about the efficiency of a spread operation here vs using `map().reduce()`. I would expect `reduce` to be more efficient but am not sure if that's true in javascript.

Also if you use `reduce()`, you could find min and max at the same time and avoid looping over `datasets` twice.

src/components/data_search/DatasetFilterList.jsx
```diff
@@ -101,15 +122,24 @@ export const DatasetFilterList = (props) => {
...
         filter={dataTypeFilters}
         filterHandler={filterHandler}
         isFiltered={isFiltered}
         filterNameFn={(filter) => filter}
       />
+      <FilterItemHeader title='Participant Count' />
+      <FilterItemRange
+        min={participantCountMin}
+        max={participantCountMax}
+        minCategory='participantCountMin'
```
pshapiro4broad, line 140: I see what's going on here but I do find it a little confusing that the local variable `participantCountMin` shares its name with a property on an object. If this was typscriptified it would be easier for me to follow.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -179,12 +203,16 @@ export const DatasetSearchTable = (props) => {
...
+      if (!isFilteredArray(filter, category) && filter !== '') {
+        newFilters[category] = filters[category].concat(filter);
+      } else {
+        newFilters[category] = filters[category].filter((f) => f !== filter);
+      }
     } else {
-      newFilters[category] = filters[category].filter((f) => f !== filter);
+      newFilters[category] = filter;
     }
     setFilters(newFilters);
```
pshapiro4broad: This code creates `newFilters` but doesn't need it until the end. Another way to write this:
```
    let newFilter;
    if (isArray(filters[category])) {
      if (!isFilteredArray(filter, category) && filter !== '') {
        newFilter = filters[category].concat(filter);
      } else {
        newFilter = filters[category].filter((f) => f !== filter);
      }
    } else {
      newFilter = filter;
    }
    const newFilters = _.clone(filters);
    newFilters[category] = newFilter;
    setFilters(newFilters);
```
To me this makes it a bit clearer that this method is doing two things; first it determines what the new filter value is, and then it sets the new value and updates the state.

pshapiro4broad (approved):
Looks OK, just a few misc comments.

src/components/data_search/DatasetFilterList.jsx
rushtong: I find this very difficult to read - Phil's reduce idea might be a more legible approach.

fboulnois (approved):
I like most of @pshapiro4broad 's suggestions! The only one I'm not sure about is the `filterHandler` refactor.

cypress/component/DataSearch/dataset_search_filters.spec.js
```diff
@@ -2,26 +2,22 @@
 
 import { mount } from 'cypress/react';
 import React from 'react';
-import {Storage} from '../../../src/libs/storage';
 import DatasetFilterList from '../../../src/components/data_search/DatasetFilterList';
 
-const duosUser = {
```
rjohanek, line 7: unused

cypress/component/DataSearch/dataset_search_table.spec.js
```diff
@@ -38,6 +41,31 @@ describe('Dataset Search Table tests', () => {
...
+          filtered = true;
+        }
+        request.reply({statusCode: 200, body:[]});
+      }
+
+      cy.intercept(
+        {method: 'POST', url: '**/api/dataset/search/index'}, handler).as('searchIndex');
+      mount(<DatasetSearchTable {...props} />);
+      cy.get('#participantCountMax-range-input').clear().type('50');
+      cy.wait(1000).then(() => {
```
rjohanek: when I wait on 'searchIndex' it doesn't work I think because there is another call to this api without the filters applied first

update: the other call (the first one without filters) was called by a previous test, adding cy.clock() to the other tests resolved this issue

src/components/data_search/DatasetFilterList.jsx
rjohanek: I'll rename

src/components/data_search/DatasetSearchTable.jsx
rjohanek: A little nuance to your first comment, if the filter is not an array, it does not mean add one, it means check if it is not null, and if it is not null add one

src/components/data_search/DatasetSearchTable.jsx
rjohanek: good shout out I forgot about this

cypress/component/DataSearch/dataset_search_table.spec.js
```diff
@@ -38,6 +43,47 @@ describe('Dataset Search Table tests', () => {
...
+    it('When a participant count filter is applied the query is updated', () => {
+      searchText ='{"range":{"participantCount":{"gte":null,"lte":"50"}}}';
+
+      cy.intercept(
+        {method: 'POST', url: '**/api/dataset/search/index'}, handler);
+      mount(<DatasetSearchTable {...props} />);
+      cy.get('#participantCountMax-range-input').clear().type('50');
+      cy.wait(3000).then(() => {
+        expect(filtered).to.be.true;
+      });
```
rushtong: The format of the intercept + wait approach should look something like this:
```
      cy.intercept(
        {method: 'POST', url: '**/api/dataset/search/index'}, handler).as('searchIndex');
      ...
      cy.wait('@searchIndex').then(() => {
        expect(filtered).to.be.true;
      });
```
We should avoid using timeouts in wait calls.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -49,8 +51,17 @@ export const DatasetSearchTable = (props) => {
   const [selectedTable, setSelectedTable] = useState(datasetSearchTableTabs.study);
   const [searchTerm, setSearchTerm] = useState('');
 
-  const isFiltered = (filter, category) => (filters[category]).indexOf(filter) > -1;
-  const numSelectedFilters = (filters) => Object.values(filters).reduce((sum, array) => sum + array.length, 0);
+  const isFilteredArray = (filter, category) => (filters[category]).indexOf(filter) > -1;
+
+  const anyFiltersSelected = (filters) => {
+    return Object.values(filters).some((filter) => {
```
pshapiro4broad: For a single line lambda, you can omit the `{ return }`
```suggestion
  const anyFiltersSelected = (filters) =>
    Object.values(filters).some((filter) => {
```

I know in terra-ui this was the standard, not sure if duos-ui has an opinion about this.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -49,8 +51,17 @@ export const DatasetSearchTable = (props) => {
...
+      if (isArray(filter)) {
+        return filter.length > 0;
+      } else {
+        return filter !== null;
+      }
```
pshapiro4broad: Not sure if it's clearer but you can express this using a single `return`, at which point this can also be written without the `{ }`s.
```suggestion
      return isArray(filter) ? filter.length > 0 : filter !== null
```

Actually I think lodash `isEmpty()` is equivalent to `filter.length > 0 || filter != null` so this should work
```
const anyFiltersSelected = (filters) => 
  Object.values(filters).some(filter => !isEmpty(filter));
```

cypress/component/DataSearch/dataset_search_table.spec.js
```diff
@@ -38,6 +42,56 @@ describe('Dataset Search Table tests', () => {
...
+        request.reply([]);
+      }
+    }
+
+
+    it('When a participant count filter is applied the query is updated', () => {
+      cy.intercept(
+        {method: 'POST', url: '**/search/index'}, (req) => {
+          return handler(req, '{"range":{"participantCount":{"gte":null,"lte":50}}}');
+        }).as('searchIndex1');
```
pshapiro4broad: Is it required that `searchIndex1` must be different than `searchIndex2` in the other test? I would expect the alias names to be independent across tests so the same name could be used for both.

pshapiro4broad (approved):
Looks good, glad to see that the flaky tests have been fixed

src/components/data_search/DatasetSearchTable.jsx
rjohanek: nice!

rjohanek: isEmpty doesn't work for this case, but the suggested change is still much cleaner!

cypress/component/DataSearch/dataset_search_table.spec.js
rjohanek: good question, no I don't think so I just changed them to detangle as part of the debugging effort but I can change them back now

cypress/component/DataSearch/dataset_search_table.spec.js
```diff
@@ -38,6 +42,56 @@ describe('Dataset Search Table tests', () => {
...
+        }).as('searchIndex');
+      mount(<DatasetSearchTable {...props} />);
+      // first clear the default value (100), without clearing first, type('50') would result in input of 10050
+      cy.get('#participantCountMax-range-input').clear().type('50');
+      cy.tick(150);
+      // this api call should have had a request that contained the searchText
+      cy.wait('@searchIndex').then((response) => {
+        expect(response.response.body[0]).to.equal('filtered');
+      });
+      cy.get('@searchIndex.all').should('have.length', 1);
```
rushtong: I can see that this test does pass, but I'm confused as to what the `.all` part of this alias selector is doing and where it is defined.

rushtong: I think I found why it is undocumented:
* https://github.com/cypress-io/cypress-documentation/issues/1573
* https://github.com/cypress-io/cypress/pull/3929#issuecomment-482372968

Since `.all` is experimental, I would recommend looking into a different approach to this specific case.

rushtong: Here's another approach, less concise, but uses documented cypress features:
```
      let count = 0;
      cy.wait('@searchIndex').then((response) => {
        expect(response.response.body[0]).to.equal('filtered');
        count++;
      });
      cy.get('@searchIndex').then(() => {
        expect(count).to.equal(1);
      });
```

cypress/component/DataSearch/dataset_search_table.spec.js
```diff
@@ -38,6 +42,56 @@ describe('Dataset Search Table tests', () => {
...
+
+
+    it('When a participant count filter is applied the query is updated', () => {
+      cy.intercept(
+        {method: 'POST', url: '**/search/index'}, (req) => {
+          return handler(req, '{"range":{"participantCount":{"gte":null,"lte":50}}}');
+        }).as('searchIndex');
+      mount(<DatasetSearchTable {...props} />);
+      // first clear the default value (100), without clearing first, type('50') would result in input of 10050
+      cy.get('#participantCountMax-range-input').clear().type('50');
```
rushtong: Cypress recommends not [chaining calls after clear](https://docs.cypress.io/api/commands/clear). It's a little more verbose, but safer to:
```
      const range = cy.get('#participantCountMax-range-input');
      range.clear();
      range.type('50');
```

cypress/component/DataSearch/dataset_search_table.spec.js
```diff
@@ -38,6 +42,56 @@ describe('Dataset Search Table tests', () => {
...
+      cy.tick(150);
```
rushtong, line 75: Cypress recommends that there be a `clock` command that precedes the `tick` command: https://docs.cypress.io/api/commands/tick

rjohanek: I use the clock command in the before each method, but I could move it into each individual test

rushtong: Thank you, I did miss that. No need if it's there, that should cover it.

rushtong (approved):
See comments inline, looks good, thank you 👍🏽

PR #2734, opened 2024-11-21
[DT-1055] Pin version of NodeJS to latest LTS

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1055

### Summary

This will prevent developers from accidentally installing incompatible version of NodeJS for the UI.

----

- Verify all tests go green

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,5 +92,8 @@
     "not dead",
     "not ie <= 11",
     "not op_mini all"
-  ]
+  ],
+  "engines": {
+    "node": ">=22.11.0"
+  }
 }
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -6,6 +6,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.11.0'
       - name: install dependencies
         run: npm ci
       - name: npm audit
--- a/.github/workflows/component-tests.yml
+++ b/.github/workflows/component-tests.yml
@@ -8,6 +8,9 @@ jobs:
         run: echo "${{ github.actor }}"
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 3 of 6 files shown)


rushtong (approved):
👍🏽

rushtong (approved):
👍🏽

package.json
```diff
@@ -92,5 +92,8 @@
     "not dead",
     "not ie <= 11",
     "not op_mini all"
-  ]
+  ],
+  "engines": {
+    "node": ">=22.11.0"
```
pshapiro4broad, line 97: Would it be wrong to use an exact match here too?

pshapiro4broad (approved):
This looks OK, although it's a shame we need to have the version both in `package.json` and in all the GitHub workflows. 

Also, can you also update `DEVNOTES.md` to reflect this change?

rushtong:
> This looks OK, although it's a shame we need to have the version both in `package.json` and in all the GitHub workflows.

It's also in [Dockerfile](https://github.com/DataBiosphere/duos-ui/blob/develop/Dockerfile#L2) ... I'd love to see a better way to specify this globally.

package.json
fboulnois: I was thinking we could do that, but any future developer would get stuck with newer versions of NodeJS (for example patch or minor). Perhaps that's a reasonable trade-off to make. Another thing to think about is that we probably do want to test newer versions, so it might be too much friction to lock it down completely.

fboulnois:
> > This looks OK, although it's a shame we need to have the version both in `package.json` and in all the GitHub workflows.
> 
> It's also in [Dockerfile](https://github.com/DataBiosphere/duos-ui/blob/develop/Dockerfile#L2) ... I'd love to see a better way to specify this globally.

I am hoping Dependabot is smart enough to update it everywhere. If not, we can use a slightly different approach and [extract the version](https://github.com/DataBiosphere/duos-ui/blob/develop/docs/getting-started.md#5-install-homebrew) like we do when installing Homebrew.

PR #2735, opened 2024-11-21
[DT-997] Move the Donor Size Header in DUOS further left so it aligns with the rest of the table

cinyecai:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-997
### Summary
Increase the spacing between the Donor Size and the Data Location columns in the Dataset table in the Data Library
image

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTableConstants.tsx
+++ b/src/components/data_search/DatasetSearchTableConstants.tsx
@@ -243,7 +243,7 @@ export const makeDatasetTableHeader = (datasets: DatasetTerm[], selected: number
     duosId: '10%',
     accessType: '10%',
     dataType: '15%',
-    donorSize: '7%',
+    donorSize: '10%',
     dataLocation: '13%',
     dataUse: '10%',
     exportToTerra: 100,
--- a/src/components/data_search/DatasetSearchTableConstants.tsx
+++ b/src/components/data_search/DatasetSearchTableConstants.tsx
@@ -25,6 +25,7 @@ interface DatasetSearchTableTabs {
 
 const makeHeaderStyle = (width: string | number): React.CSSProperties => ({
   width,
+  marginRight: 5
 });
 
 const makeRowStyle = (width: string | number): React.CSSProperties => ({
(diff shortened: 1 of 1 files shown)


rushtong (approved):
👍🏽

src/components/data_search/DatasetSearchTableConstants.tsx
```diff
@@ -243,7 +243,7 @@ export const makeDatasetTableHeader = (datasets: DatasetTerm[], selected: number
     duosId: '10%',
     accessType: '10%',
     dataType: '15%',
-    donorSize: '7%',
+    donorSize: '10%',
```
pshapiro4broad, line 247: Before the total %s equalled 100 when added up (25 + 10 + 10+ 15 + 7 + 13  + 10). Now it's 103%. Is that correct? If this % goes up, maybe another needs to go down?

pshapiro4broad (approved):
This looks OK but the underlying problem that I see is that this doesn't add padding between elements, so a "correct" appearance is still dependent on how a user sizes their window. For example, by resizing the window, you can create a state where `DATASET NAME` becomes too close to `STUDY NAME` in the dataset view.

It would be better if there's an overall way to specify between-cell spacing for the table, so we don't have to approximate the correct behavior by making columns wider.

src/components/data_search/DatasetSearchTableConstants.tsx
cinyecai: I thought that too, but it doesn't seem to cause any weird behavior in the rendering of the table itself, but I'll look more into it

rushtong: Widths can be > 100%. It can sometimes lead to overflow issues.

cinyecai:
> This looks OK but the underlying problem that I see is that this doesn't add padding between elements, so a "correct" appearance is still dependent on how a user sizes their window. 
> 
This is a good point - I checked and it looks like there's padding build into the cell styling, but not the header styling. Adding a right margin property in the header styling helps with the window size problem a little by forcing text wrapping before the headers get too close to each other, but eventually the whole page looks wonky if your window is too small, which I think is a larger problem

PR #2739, opened 2024-12-02
[DT-1072] Add Getz Lab data library

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1072

### Summary

Adds a Getz Lab data library that lives at `/datalibrary/getzlab`.

----

- Verify all tests go green


rushtong (approved):
👍🏽

rjohanek (approved):
looks good, is there any testing we want to do for this?

fboulnois:
> looks good, is there any testing we want to do for this?

Yes, do you have any thoughts on how I might write a test like "if it's this particular Data Library, then expect to see this title"?

rjohanek:
> > looks good, is there any testing we want to do for this?
> 
> Yes, do you have any thoughts on how I might write a test like "if it's this particular Data Library, then expect to see this title"?

Maybe you could use cypress.visit(url) to access the specific data library instead of using mountComponent like we usually do? If the specific library page info is stored via props to component, you could still use mountComponent, but I don't think it is right? so you might have to use visit to access the url directly. And then check for the title ? and maybe the filtered datasets too. I haven't done this before but I'd be happy to talk about it if you'd like! https://docs.cypress.io/api/commands/visit

rushtong:
> > > looks good, is there any testing we want to do for this?
> > 
> > 
> > Yes, do you have any thoughts on how I might write a test like "if it's this particular Data Library, then expect to see this title"?
> 
> Maybe you could use cypress.visit(url) to access the specific data library instead of using mountComponent like we usually do? If the specific library page info is stored via props to component, you could still use mountComponent, but I don't think it is right? so you might have to use visit to access the url directly. And then check for the title ? and maybe the filtered datasets too. I haven't done this before but I'd be happy to talk about it if you'd like! https://docs.cypress.io/api/commands/visit

Using `cy.visit` would require a running instance (along with a backing consent, populated ES instance, etc) which makes that an integration test, not a unit test. We might be able to pass page parameters as arguments to the overall page, but I think that's a non-trivial effort and best handled as a distinct task in and of itself.

rjohanek:
> Using `cy.visit` would require a running instance (along with a backing consent, populated ES instance, etc) which makes that an integration test, not a unit test. We might be able to pass page parameters as arguments to the overall page, but I think that's a non-trivial effort and best handled as a distinct task in and of itself.

That makes sense, if we aren't looking to add integration tests then maybe you could mount the DatasetSearch component with the query (as a prop) that would be specified by this route and then check for the filtered datasets

fboulnois:
I'm going to create a separate ticket for unit tests of the data libraries to unblock this ticket.

PR #2740, opened 2024-12-03
[DT-1079] Switch from MomentJS to DayJS

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1079

### Summary

Elliot is working on a date picker component that relies on DayJS, which is similar to MomentJS. We only rely on MomentJS for a single page, and the functionality we rely on is very minimal. Also, we already transitively depend on DayJS due to Cypress. Finally, MomentJS is substantially smaller, faster and less outdated.

This PR switches from MomentJS to DayJS.

----

- Verify all tests go green


package.json
```diff
@@ -14,11 +14,11 @@
     "ajv": "8.17.1",
     "ajv-formats": "3.0.1",
     "axios": "1.7.8",
+    "dayjs": "1.11.3",
```
fboulnois, line 17: I didn't switch to the latest version of `dayjs` to minimize the diffs in the `package-lock.json`.

rushtong (approved):
👍🏽

package.json
rushtong: I'm sure we'll get a dependabot PR for that, which will be nice to keep it isolated.

s-rubenstein (approved):

PR #2741, opened 2024-12-03
[DT-783] Add study details page

s-rubenstein:
### Addresses

https://github.com/user-attachments/assets/3296b42a-b16f-445e-9234-a4b03c1b0822

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/Routes.jsx
+++ b/src/Routes.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import {Redirect, Route, Switch} from 'react-router-dom';
+import { StudyDetails } from './components/study_details/StudyDetails';
 import AuthenticatedRoute from './components/AuthenticatedRoute';
 import {USER_ROLES} from './libs/utils';
 import ManageDac from './pages/manage_dac/ManageDac';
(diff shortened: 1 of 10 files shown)


src/Routes.jsx
```diff
@@ -47,65 +48,66 @@ import EditDac from './pages/manage_dac/EditDac';
...
+    <AuthenticatedRoute path='/data_submission_form' component={DataSubmissionForm} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.dataSubmitter]} />
+    <AuthenticatedRoute path='/study_update/:studyId' component={StudyUpdateForm} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.chairperson, USER_ROLES.dataSubmitter]} />
+    <AuthenticatedRoute path='/admin_manage_lc/' component={AdminManageLC} props={props} rolesAllowed={[USER_ROLES.admin]} />
+    <AuthenticatedRoute path='/admin_manage_dar_collections/' component={AdminManageDarCollections} props={props} rolesAllowed={[USER_ROLES.admin]} />
+    <AuthenticatedRoute path='/datalibrary/:query' component={DatasetSearch} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.all]} />
+    <AuthenticatedRoute path='/datalibrary' component={DatasetSearch} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.all]} />
+    <AuthenticatedRoute path='/studies/:studyId' component={StudyDetails} props={props} rolesAllowed={[USER_ROLES.admin, USER_ROLES.all]} />
```
s-rubenstein, line 104: This is the only new line here, other than that its all linting

src/components/DuosHeader.jsx
```diff
@@ -160,26 +160,26 @@ const NavigationTabsComponent = (props) => {
   return (
     <div className={`navbar-logged ${orientation === 'vertical' ? 'navbar-vertical' : ''}`}>
       {makeNotifications()}
-      <ul className="navbar-main">
```
s-rubenstein, line 163: Entirely linting changes

src/components/data_search/DatasetSearchFooter.tsx
```diff
@@ -1,11 +1,11 @@
 import * as _ from 'lodash';
 import {Button} from '@mui/material';
 import * as React from 'react';
-import {Dataset} from 'src/types/model';
+import {DatasetTerm} from 'src/types/model';
 
 interface DatasetSearchFooterProps {
   selectedDatasets: number[];
-  datasets: Dataset[];
+  datasets: DatasetTerm[];
```
s-rubenstein, line 8: For the elasticsearch terms, this is the right typing (I believe)

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -31,6 +31,21 @@ const styles = {
   },
 };
 
+export const applyForAccess = async (selected, history) => {
```
s-rubenstein, line 34: This got pulled out to be reused elsewhere. I could see this going to a libs or utils file as well - what do you all think?

rushtong: In another defunct DAA-related PR, I had pulled this out to a separate component, so I think it makes sense.

rushtong (approved):
This looks great, thank you! No code suggestions. The linting changes look good, nice catch. There are a couple other study properties we could be showing but would likely require a decision from product on whether we should be showing them here. We could look into adding the following fields. Some are currently indexed, but some may not be and we should look at adding those values to the index:

* study type
* data types 
* data submitter and/or email
* phs id, and any other NIH related info we capture

rjohanek (commented):
Does this maintain selections/search upon navigation back to the data library? I thought we decided on a side bar to avoid extra navigation?

rjohanek (commented):
can we add cypress component tests for this new component in ts?

s-rubenstein:
@rjohanek it does not maintain search or filters - this was on request from @lstrano-broad when we met with him - he prefers the extra page.

s-rubenstein:
I can add tests

rushtong:
> @rjohanek it does not maintain search or filters - this was on request from @lstrano-broad when we met with him - he prefers the extra page.

I think maintaining filter state would be a great feature, but we need to design that and it would be a completely new feature.

lstrano-broad:
I never requested the filters to reset. I agree the filter state needs to maintained when user click back.

rjohanek (approved):
tests look good thanks!!

s-rubenstein:
@lstrano-broad we'll need to do that as a separate ticket - right now we don't have a mechanism to store filter state between pages, so when you go back to the original URL it has the same filters

s-rubenstein:
(Separate page was Lou's request, to be clear - not the loss of filters. The loss of filters is a side effect that we can fix separately)

PR #2745, opened 2024-12-04
[doc] Update devnotes

snf2ye:
Hopefully clarifies a few of these steps for new users getting started on duos ui. The changes reflect areas that I was confused by. 
* Give more specifics to the use of`./render-configs` and make more prominent in the instructions. 
* Remove instructions that are covered by `./render-configs`. We can use the script as documentation for the steps so that we do not need to maintain this in two places. 
* Update `./render-configs` to differentiate between the parameters that set values (`project`) vs. boolean values (`env` -> `write_env`)
* Bug: I found that I had to be in the scripts directory before calling `./render-configs.sh`, so I've made that clear in the instructions. But, we may want to figure out how to run `./scripts/render-configs.sh` directly.

### DIFF
--- a/DEVNOTES.md
+++ b/DEVNOTES.md
@@ -14,6 +14,7 @@ npm install
 
 3. Generate configuration files and populate certificates by running the [render-configs.sh](scripts/render-configs.sh) script. This requires 
 connection to the Broad VPN. By default, this points the DUOS UI to the dev environment. 
+
 ```
 cd scripts
 ./render-configs.sh --write_env true --write_config true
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍🏽

fboulnois (approved):
Would you consider adding the newlines between coding blocks? Otherwise looks good and I'll close my PR.

snf2ye:
> Would you consider adding the newlines between coding blocks? Otherwise looks good and I'll close my PR.

For sure, done! Hopefully I didn't miss any.

PR #2749, opened 2024-12-09
[DT-1087] Date Picker initial implementation with tests.

otchet-broad:
### Addresses

[DT-1087](https://broadworkbench.atlassian.net/browse/DT-1087?atlOrigin=eyJpIjoiNDBlOWMxNDdiOGFlNDgzNjg5MzUyNDUzOGNjMTU5MDkiLCJwIjoiaiJ9)

### Summary
Adds a form field type of a date picker to Duos UI components per mockups from UX.  Visible in tests only.

----

- Verify all tests go green

[DT-1087]: https://broadworkbench.atlassian.net/browse/DT-1087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


otchet-broad:
There's an observed issue with Github Actions and the MUI date picker (generic) component rendering in "mobile" mode when being exercised in cypress under a github action.  To account for this, I opted to guarantee the Desktop date component since a Data Submitter working on a mobile device is not fully supported at this point.
See  [https://stackoverflow.com/questions/69799310/cypress-testing-a-material-ui-datepicker-is-not-working-on-github-actions](https://stackoverflow.com/questions/69799310/cypress-testing-a-material-ui-datepicker-is-not-working-on-github-actions).

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,197 @@
...
+  LocalizationProvider,
+  PickersActionBarProps,
+  PickersDay,
+  PickersDayProps
+} from '@mui/x-date-pickers';
+import {AdapterDayjs} from '@mui/x-date-pickers/AdapterDayjs';
+import {Button} from '@mui/material';
+import DialogActions from '@mui/material/DialogActions';
+import {Dayjs} from 'dayjs';
+import type {} from '@mui/x-date-pickers/themeAugmentation';
```
rushtong, line 14: It is interesting that intellij thinks this is unused.

Screenshot 2024-12-10 at 2 33 14 PM

rushtong (approved):
Thank you for the linting and test cleanup, looks reasonable to me 👍🏽

src/components/DuosDatePicker.tsx
otchet-broad: It is, and it definitely breaks when the import is removed.  :-)

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,197 @@
...
+          },
+          switchViewIcon:{
+            fontSize:'2.5rem',
+          },
+          label:{
+            fontFamily: 'Montserrat',
+            color: '#000',
+            fontSize: '13px',
+            fontWeight: '400',
+            padding: '7px 20px 7px 20px',
```
s-rubenstein: Nit: I think this could be `7px 20px`

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,197 @@
...
+      MuiPickersYear: {
+        styleOverrides: {
+          yearButton: {
+            fontFamily: 'Montserrat',
+            color: '#000',
+            fontSize: '13px',
+            fontWeight: '400',
+            padding: '7px 20px 7px 20px',
+            '&.Mui-selected': {
+              backgroundColor: '#216FB4',
```
s-rubenstein: Could we make an intermediate variable for this?

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,197 @@
...
+          sizeMedium: {
+            color:'#216FB4',
+          },
+        },
+      },
+      MuiDayCalendar:{
+        styleOverrides:{
+          weekDayLabel:{
+            fontFamily: 'Montserrat',
+            color: '#000',
```
s-rubenstein: How would you feel about standardizing this on `black`/using the color names where we have them?

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,197 @@
...
+  };
+
+  return <ThemeProvider theme={theme}>
+    <LocalizationProvider dateAdapter={AdapterDayjs}>
+      <DesktopDatePicker
+        closeOnSelect={false}
+        label={'Select a date'}
+        format={inputFormat}
+        value={value}
+        onAccept={(value)=>{onChange(value);}}
```
s-rubenstein: Nit: I think you can just provide the method reference here?

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,197 @@
...
+        onError={(error, value)=> onError(error, value)}
```
s-rubenstein: Same comment about the method reference

src/components/DuosDatePicker.tsx
otchet-broad: @s-rubenstein - Do you mean a CSS variable that calculates the value or a variable that assigns the backgroundColor to the theme?

src/components/DuosDatePicker.tsx
otchet-broad: I don't have strong feelings about this and would be happy to make that change, however we should be on the same page about colors in the context of MUI.  Do you expect to see [MUI color palettes](https://mui.com/material-ui/customization/color/#color-palette) or CSS ? For this PR, do you expect that standardization across all of DUOS-UI or just in the context of the DuosDatePicker?

src/components/DuosDatePicker.tsx
otchet-broad: I'll tweak the types on both onChange and onError from `Function` to `any` and update these.

Sorry for the back and forth edits on this.  I've got the types tweaked for this now to be just a method reference.  In making and testing, I found a bug with date validation in a test that I needed to fix.  I was passing in the wrong validator in the test.

src/components/DuosDatePicker.tsx
otchet-broad: One minor thing I've observed in just using the CSS color 'black' instead of '#000' is that the IDE removes the option for the color picker in the gutter.  Does that matter to you?

src/components/forms/formValidation.js
```diff
@@ -30,13 +31,19 @@ export const dateValidator = {
   msg: 'Please enter a date (YYYY-MM-DD), e.g. 2018-11-13',
 };
 
+export const dayJSValidator = {
+  id: 'dayjs',
+  isValid: (val) => isValidDayjsDate(val),
+  msg: 'Please select a valid date.',
```
rjohanek, line 37: Wondering about if there is a way to make this error message more helpful, if it is a pop up calendar, what is the case where a user selects an invalid date?

src/components/DuosDatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+            padding: '7px 20px',
+          },
+        }
+      },
+    },
+  });
+
+  const CancelSelectActionBar = (props:PickersActionBarProps) => {
+    // Quirk of this control's usage pattern is the need to destructure the unused onSetToday and onClear from 'other'
+    // props.  This is in part because per mockup, this control does not support 'clear' or 'go to today' style buttons.
```
rjohanek, line 145: can you link the mocks? we don't need a clear button because there is cancel? and we don't need go to today because today is the default value? is that right?

rjohanek (commented):
the changes look good, but I'm lacking context from the ticket or the PR description. I know you said it's only visible in tests rn, but I'd love to see what it looks like and where it is used.

src/components/forms/formValidation.js
otchet-broad: Great question!  The control includes a text field that users can type into.  The control does a good job at guarding against invalid dates, but they are still theoretically possible.  For example, you can type in 2/31/2025.

src/components/DuosDatePicker.tsx
otchet-broad: Yes!
Mocks:
image

Rendered component:
image

rjohanek (approved):
the date picker looks good and love the linting thank you!

fboulnois (approved):
tests and component look good 👍

src/components/DuosDatePicker.tsx
s-rubenstein: For black I don't strongly care - if we need to change it we can always move it to get the color picker back.

I would personally say let's steer towards unity across the codebase, so if the rest of the codebase uses `#000` then lets do that. However, I think right now there's existing inconsistency?

s-rubenstein (approved):
LGTM

PR #2754, opened 2024-12-11
[DT-1081] Fix filter bugs

s-rubenstein:
### Summary

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DataSearch/dataset_search_filters.spec.js
+++ b/cypress/component/DataSearch/dataset_search_filters.spec.js
@@ -1,8 +1,35 @@
 /* eslint-disable no-undef */
 
-import { mount } from 'cypress/react';
+import { makeDatasetTerm, makeStudyTerm } from '../test-utils';
+import { mount } from 'cypress/react18';
 import React from 'react';
+import { defaultFilters } from '../../../src/components/data_search/DatasetFilterConstants';
 import DatasetFilterList from '../../../src/components/data_search/DatasetFilterList';
+
+const datasets = [
+  makeDatasetTerm({
+    datasetId: 123456,
+    datasetIdentifier: `DUOS-123456`,
+    datasetName: 'Some Dataset 1',
+    participantCount: 100,
+    study: makeStudyTerm({
+      studyName: 'Some Study 1',
+      studyId: 1,
+      dataCustodianEmail: ['Some Data Custodian Email 1'],
+    })
+  }),
+  makeDatasetTerm({
+    datasetId: 123457,
+    datasetIdentifier: `DUOS-123457`,
+    datasetName: 'Some Dataset 2',
+    participantCount: 2,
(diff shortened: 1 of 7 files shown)


cypress/component/DataSearch/dataset_search_filters.spec.js
```diff
@@ -20,4 +47,23 @@ describe('Data Library Filters', () => {
...
+      datasets[0].participantCount);
+    cy.get('#participantCountMin-range-input').should('have.value',
+      datasets[1].participantCount);
+  });
+
+  it('calls the filter handler on range changes', () => {
+    const filterHandlerStub = cy.stub();
+    const props = { datasets, filters: { ...defaultFilters(datasets), participantCountMin: 2, participantCountMax: 5 }, filterHandler: filterHandlerStub, isFiltered: () => {}};
+    mount(<DatasetFilterList {...props} />);
+    cy.get('#participantCountMax-range-input').type('3').trigger('change');
```
rushtong: Chaining calls after `type` is [unsafe](https://docs.cypress.io/api/commands/type). A lot of our tests suffer from this pattern, so it is something we should be trying to clean up.

src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -303,7 +293,7 @@ export const DatasetSearchTable = (props) => {
         </Box>
         <Box sx={{display: 'flex', flexDirection: 'row', paddingTop: '2em'}}>
           <Box sx={{width: '14%', padding: '0 1em'}}>
-            <DatasetFilterList datasets={datasets} filterHandler={filterHandler} isFiltered={isFilteredArray} onClear={() => setFilters(defaultFilters)}/>
+            <DatasetFilterList datasets={datasets} filterHandler={filterHandler} filters={filters} isFiltered={isFilteredArray} onClear={() => setFilters(defaultFilters(datasets))}/>
```
rushtong, line 296: Something about this is breaking the `dataset_search_table.spec.js` tests, but it's not exactly clear what. I think the tests might be a little brittle as well in that they're expecting a populated `response.response.body[0]` object.

rushtong: I think I see it. The search filter text is slightly different in this PR which breaks the test. Changing line 68 in `dataset_search_table.spec.js` to:
```
return handler(req, '{"range":{"participantCount":{"gte":100,"lte":"50"}}}');
```
and line 91 to
```
return handler(req, '{"range":{"participantCount":{"gte":100,"lte":100}}}');
```
allows the string matching in the handler function to work correctly.

rushtong (approved):
Looks reasonable, addresses bugs in ticket. Once tests are passing, 👍🏽

fboulnois (approved):
looks good, thank you 👍

PR #2756, opened 2024-12-16
[DT-1071] Remove description column from studies

s-rubenstein:
### Summary

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2757, opened 2024-12-18
[DT-1077] typescriptify datasetstatistics and fix

s-rubenstein:
### Addresses
I noticed the issue appeared to be based on properties missing, so I added a test with no additional properties that still renders.

I opted not to put in the additional effort to typescriptify the test, because test files are much less impact as js vs ts, and if we decide to move unit tests to jest eventually, I wanted to prioritize effort. 

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTableConstants.tsx
+++ b/src/components/data_search/DatasetSearchTableConstants.tsx
@@ -71,7 +71,7 @@ export const makeStudyTableHeaders = (datasets: DatasetTerm[], selected: number[
     participants: '10%',
     phenotype: '20%',
     species: '15%',
-    piName: '15%',
+    piName: '10%',
     dataCustodians: '20%'
   };
   const isSelectable = (dataset: DatasetTerm) => dataset.accessManagement != 'open' && dataset.accessManagement != 'external';
--- a/src/components/data_search/DatasetSearchTableConstants.tsx
+++ b/src/components/data_search/DatasetSearchTableConstants.tsx
@@ -71,7 +71,7 @@ export const makeStudyTableHeaders = (datasets: DatasetTerm[], selected: number[
     participants: '10%',
     phenotype: '20%',
     species: '15%',
-    piName: '10%',
+    piName: '15%',
     dataCustodians: '20%'
   };
   const isSelectable = (dataset: DatasetTerm) => dataset.accessManagement != 'open' && dataset.accessManagement != 'external';
--- a/src/libs/Models.ts
+++ b/src/libs/Models.ts
@@ -1,6 +1,0 @@
-
-export enum AccessManagement {
-  OPEN = 'open',
-  CONTROLLED = 'controlled',
-  EXTERNAL = 'external'
-}
--- a/src/pages/DatasetStatistics.tsx
+++ b/src/pages/DatasetStatistics.tsx
@@ -9,15 +9,32 @@ import {find} from 'lodash/fp';
 import {ReadMore} from '../components/ReadMore';
 import {formatDate} from '../libs/utils';
 import {Button} from '@mui/material';
-import {AccessManagement} from '../libs/Models.ts';
+import {History} from 'history';
+import {Dataset, DatasetProperty} from '../types/model';
 
 const LINE = <div style={{borderTop: '1px solid #BABEC1', height: 0}}/>;
 
-export default function DatasetStatistics(props) {
+enum AccessManagement {
+  OPEN = 'open',
+  CONTROLLED = 'controlled',
+  EXTERNAL = 'external'
+}
+
+
+interface DatasetStatisticsProps {
+  history: History,
+  match: {
+    params: {
+      datasetIdentifier: string;
+    }
+  }
+}
+
+export default function DatasetStatistics(props: DatasetStatisticsProps) {
   const {history, match: {params: {datasetIdentifier}}} = props;
-  const [datasetId, setDatasetId] = useState();
-  const [dataset, setDataset] = useState();
-  const [dars, setDars] = useState();
+  const [datasetId, setDatasetId] = useState<number>();
+  const [dataset, setDataset] = useState<Dataset>();
+  const [dars, setDars] = useState<any>();
   const [isLoading, setIsLoading] = useState(true);
 
   const applyForAccess = async () => {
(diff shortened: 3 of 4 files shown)


src/libs/Models.ts
```diff
@@ -1,6 +0,0 @@
-
```
s-rubenstein, line 1: This file is side by side with a `models.js` file, and the capitalization being the only difference was causing issues. Given this was only used in one place I just moved it.

src/components/data_search/DatasetSearchTableConstants.tsx
```diff
@@ -127,20 +125,6 @@ export const makeStudyTableHeaders = (datasets: DatasetTerm[], selected: number[
         label: datasets[0].study.studyName
       })
     },
-    {
-      label: 'Description',
-      sortable: true,
-      cellStyle: makeHeaderStyle(studyCellWidths.description),
-      cellDataFn: (datasets) => ({
```
rjohanek: wasn't this exact change made in this PR too? https://github.com/DataBiosphere/duos-ui/pull/2756 Why is this being shown as a change here?

rjohanek (approved):
typescriptification looks good, love the additional test, is there a way to include pictures to show that the full page renders as expected?

src/components/data_search/DatasetSearchTableConstants.tsx
s-rubenstein: Oh, I think it may be because I messed up the git, lemme try to fix that.

s-rubenstein:
Screenshot 2024-12-19 at 10 35 58 AM
Here's a screenshot, but I don't think we can get much clarity on whether it worked without the specific prod data.

src/pages/DatasetStatistics.tsx
```diff
@@ -111,11 +128,11 @@ export default function DatasetStatistics(props) {
           </div>
           <div style={Styles.SUB_HEADER}>Dataset Information</div>
           <div style={{display: 'flex'}}>
-            <div style={Styles.DESCRIPTION_BOX}>
+            <div style={Styles.DESCRIPTION_BOX as React.CSSProperties}>
               <div style={{...Styles.MINOR_HEADER, paddingLeft: '10px'}}>Dataset Description:</div>
               {LINE}
               <div style={{fontSize: Theme.font.size.small, padding: '1rem'}}>
-                {extract('Dataset Description') || dataset?.description || dataset?.study?.description || 'N/A'}
+                {extract('Dataset Description') || dataset?.study?.description || 'N/A'}
```
snf2ye, line 135: Why is `dataset?.description` not needed here?

src/pages/DatasetStatistics.tsx
```diff
@@ -9,15 +9,32 @@ import {find} from 'lodash/fp';
...
+}
+
+export default function DatasetStatistics(props: DatasetStatisticsProps) {
   const {history, match: {params: {datasetIdentifier}}} = props;
-  const [datasetId, setDatasetId] = useState();
-  const [dataset, setDataset] = useState();
-  const [dars, setDars] = useState();
+  const [datasetId, setDatasetId] = useState<number>();
+  const [dataset, setDataset] = useState<Dataset>();
+  const [dars, setDars] = useState<any>();
```
snf2ye, line 37: Is it possible to type dars as well?

snf2ye (approved):

src/pages/DatasetStatistics.tsx
pshapiro4broad: It looks like it wouldn't be too hard, although this type doesn't exist yet in duos-ui. The Java type is a List of [DarMetricsSummary](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/MetricsService.java#L55l)

src/pages/DatasetStatistics.tsx
```diff
@@ -43,12 +60,12 @@ export default function DatasetStatistics(props) {
     });
   }, [datasetIdentifier]);
 
-  const extract = useCallback((propertyName) => {
-    const property = find({propertyName})(dataset.properties);
+  const extract = useCallback((propertyName: string) => {
+    const property = find({propertyName})(dataset?.properties) as DatasetProperty;
```
pshapiro4broad, line 64: Would it be too ambitious to add types to the Ajax APIs too? If there's a way to incrementally add types to just the APIs used here that would be an improvement.

src/pages/DatasetStatistics.tsx
```diff
@@ -9,15 +9,32 @@ import {find} from 'lodash/fp';
...
 import {formatDate} from '../libs/utils';
 import {Button} from '@mui/material';
-import {AccessManagement} from '../libs/Models.ts';
+import {History} from 'history';
+import {Dataset, DatasetProperty} from '../types/model';
 
 const LINE = <div style={{borderTop: '1px solid #BABEC1', height: 0}}/>;
 
-export default function DatasetStatistics(props) {
+enum AccessManagement {
```
pshapiro4broad, line 17: Should this be moved to `Model.ts?` Or this only an UI concept and not part of an API model?

src/pages/DatasetStatistics.tsx
pshapiro4broad: I think that was an error, the type `Dataset` doesn't have a `description` field. Maybe the API changed at some point and this was kept around for backward compatibility?

pshapiro4broad (approved):
Looks good, I'm looking forward to having more typescript in this code base

src/pages/DatasetStatistics.tsx
s-rubenstein: Its not actually a field on the typed object! I don't know why it was there before.

src/pages/DatasetStatistics.tsx
s-rubenstein: Ajax often has types, but extract doesn't because properties is ambiguous.

src/pages/DatasetStatistics.tsx
s-rubenstein: https://broadinstitute.slack.com/archives/C07T2DLHD24/p1734552165921369

src/pages/DatasetStatistics.tsx
s-rubenstein: It is not a part of the API model I believe. Or at least, it was in a separate file and `model.js` is a generated file. `Model.ts` was conflicting because of the import conflicts in js having issues with casing

PR #2758, opened 2024-12-20
[DT-791] Add count to top of data library

s-rubenstein:
### Addresses
Screenshot 2024-12-20 at 4 03 35 PM
Screenshot 2024-12-20 at 4 03 42 PM

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTableDisplay.tsx
+++ b/src/components/data_search/DatasetSearchTableDisplay.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import {Box} from '@mui/material';
 import {isEmpty} from 'lodash';
 import {DatasetTerm} from 'src/types/model';
 import SimpleTable from '../SimpleTable';
@@ -55,24 +54,23 @@ export const DatasetSearchTableDisplay = (props: DatasetSearchTableDisplayProps)
   const headers = tab.makeHeaders(filteredData, selected, onSelect, exportableDatasets);
   const rowData = tab.makeRows(filteredData, headers);
 
-  return isEmpty(filteredData) ? (
-    <Box sx={{
-      display: 'flex',
-      flexDirection: 'column', alignItems: 'center', justifyContent: 'center', height: '100%' }}>
-      <h1>There are no datasets that fit these criteria.</h1>
-    </Box>
-  )
-    : (
-      <>
-        <div style={{ fontWeight: 600, borderBottom: '1px solid black' }}>{rowData.length} {_.capitalize(rowData.length > 1 ? tab.plural : tab.singular)}</div>
+  return <>
+    <div style={{
+      fontWeight: 600,
+      borderBottom: '1px solid black'
+    }}>
+      {rowData.length} {_.capitalize(rowData.length !== 1 ? tab.plural : tab.singular)}
+    </div>
+    {
+      isEmpty(filteredData) ?
+        <div style={{fontWeight: 600, marginTop: '0.5rem'}}>There are no datasets that fit these criteria.</div> :
         <SimpleTable
           rowData={rowData}
           columnHeaders={headers}
           selected={selected}
           styles={styles}
           tableSize={10}
-          summary='faceted dataset search table'
-        />
-      </>
-    );
+          summary='faceted dataset search table' />
+    }
+  </>;
 };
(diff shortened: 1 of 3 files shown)


src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -62,13 +63,16 @@ export const DatasetSearchTableDisplay = (props: DatasetSearchTableDisplayProps)
...
-      <SimpleTable
-        rowData={rowData}
-        columnHeaders={headers}
-        selected={selected}
-        styles={styles}
-        tableSize={10}
-        summary='faceted dataset search table'
-      />
+      <>
+        <div style={{ fontWeight: 600, borderBottom: '1px solid black' }}>{rowData.length} {_.capitalize(rowData.length > 1 ? tab.plural : tab.singular)}</div>
```
rushtong: An interesting interaction is that if you filter to `0` records, the `isEmpty(filteredData)` condition kicks in and this block, along with the table, disappears. I think it would be cleaner to reflect that condition on this line, but it certainly isn't required. In that case, if we did show `0` records, the plural would be the correct language to use here.

rushtong (approved):
Looks good, thank you, see optional comment inline 👍🏽

rjohanek (approved):
Overall looks good, are there mocks for this? Does this reflect the filtered datasets/studies, or all? Does this work for the specific branded libraries too?

src/components/data_search/DatasetSearchTableDisplay.tsx
pshapiro4broad: I agree; I was expecting that the total would be outside the table, so it would be displayed in all cases. I don't see a specific UX proposal on the ticket so I may have missed a discussion about the UX in refinement.

Instead of showing a total inside the table, I think it would be better to show the total at the top, maybe on the same lines as the search (filter) text entry box. Also I would include both the number shown and the total number, so a user can tell how much of the total they're omitting. Text like "Showing X out of Y"/ "Showing all of Y"/"Showing none of Y" is one way to format this information.

s-rubenstein:
@rjohanek I can't think of why branding would impact it - they should be using the same component?

It does update when I filter.

s-rubenstein:
Mocks are here: https://www.figma.com/design/MkffpVbUdP4lWvkN9OWiro/Data-Library?node-id=0-1&p=f

src/components/data_search/DatasetSearchTableDisplay.tsx
s-rubenstein: I think for now my inclination is to leave the present `isEmpty` behavior, because it reduces redundant info. If necessary we can update it later.

s-rubenstein:
Screenshot 2025-01-06 at 10 56 43 AM

rushtong (approved):
Thanks for the updates 👍🏽

src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -54,21 +54,23 @@ export const DatasetSearchTableDisplay = (props: DatasetSearchTableDisplayProps)
...
+  return <>
+    <div style={{
+      fontWeight: 600,
+      borderBottom: '1px solid black'
+    }}>
+      {rowData.length} {_.capitalize(rowData.length !== 1 ? tab.plural : tab.singular)}
+    </div>
+    {
+      isEmpty(filteredData) ?
+        <div style={{fontWeight: 600, marginTop: '0.5rem'}}>There are no datasets that fit these criteria.</div> :
```
pshapiro4broad: Should this use the custom tab name instead?
```suggestion
        <div style={{fontWeight: 600, marginTop: '0.5rem'}}>There are no {tab.plural} that fit these criteria.</div> :
```

src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -54,21 +54,23 @@ export const DatasetSearchTableDisplay = (props: DatasetSearchTableDisplayProps)
...
+        <SimpleTable
+          rowData={rowData}
+          columnHeaders={headers}
+          selected={selected}
+          styles={styles}
+          tableSize={10}
+          summary='faceted dataset search table' />
```
pshapiro4broad:
```suggestion
          summary='faceted {tab.singular} search table' />
```

pshapiro4broad (approved):

PR #2764, opened 2025-01-08
DT-968: Speed up switching between dataset and study views

pshapiro4broad:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-968

### Summary

After profiling the page load, we found that the bulk of the time was spent adding and removing the react tooltip component used for every cell in this table. This approach replaces the react tooltip with a browser native tooltip, which doesn't have this performance issue.

With this change, the time to switch from studies to datasets goes from ~5-6 seconds to ~1-2 seconds.

Even with this change, there's still one tooltip which is still rendered using react tooltip, for the dataset data use column. This tooltip uses custom formatting which the browser native tooltip doesn't support -- it only supports a `string`.

Using a react virtual view could speed the view up even more, but getting it to work with the react tooltip was difficult. If react tooltip was completely removed, it should make this work easier if wanted to revisit it.

Before

https://github.com/user-attachments/assets/8555732b-404d-4838-8b1b-bb1a1a9140a0

After

https://github.com/user-attachments/assets/985cc721-6fa3-4b62-9ae4-b424f7c23568

----

- Verify all tests go green

### DIFF
--- a/src/components/Tooltips.tsx
+++ b/src/components/Tooltips.tsx
@@ -1,38 +1,14 @@
-import ReactTooltip from 'react-tooltip';
 import * as React from 'react';
-import {useEffect, useRef, useState} from 'react';
-
-// Copied from ReactTooltip Types
-type Place = 'top' | 'right' | 'bottom' | 'left';
-
-export const tooltipStyle: React.CSSProperties = {maxWidth: '30vw', textWrap: 'wrap'};
 
 interface OverflowTooltipProps {
   tooltipText: string;
   children: React.ReactNode | React.ReactNode[];
   id: string;
-  place: Place
 }
 export const OverflowTooltip = (props: OverflowTooltipProps) => {
-  const ref = useRef<HTMLDivElement>(null);
-  const [isOverflown, setIsOverflown] = useState(false);
-  useEffect(() => {
-    const element = ref.current!;
-    setIsOverflown(element.offsetWidth < element.scrollWidth);
-  }, []);
-
-  return <>
-    <div
-      data-tip='Full details'
-      data-for={props.id}
-      ref={ref}
+  const {id, tooltipText, children} = props;
+  return <div
+      data-for={id}
       style={{overflow: 'hidden', textOverflow: 'ellipsis'}}
-    >{props.children}</div>
-    <ReactTooltip
-      place={'top'}
-      effect={'solid'}
-      disable={!isOverflown}
-      scrollHide={true}
-      id={props.id}><div style={tooltipStyle}>{props.tooltipText}</div></ReactTooltip>
-  </>;
+    ><span title={tooltipText}>{children}</span></div>;
 };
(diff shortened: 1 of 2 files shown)


src/components/Tooltips.tsx
```diff
@@ -1,38 +1,14 @@
...
       style={{overflow: 'hidden', textOverflow: 'ellipsis'}}
-    >{props.children}</div>
-    <ReactTooltip
-      place={'top'}
-      effect={'solid'}
-      disable={!isOverflown}
-      scrollHide={true}
-      id={props.id}><div style={tooltipStyle}>{props.tooltipText}</div></ReactTooltip>
-  </>;
+    ><span title={tooltipText}>{children}</span></div>;
```
pshapiro4broad, line 13: Using `span` with a title allows this to work in both chrome and safari. Without this, the tooltip would only show in safari. This approach was copied from https://stackoverflow.com/a/16306846

s-rubenstein (approved):
This looks reasonable to me

src/components/data_search/DatasetSearchTableConstants.tsx
```diff
@@ -93,18 +92,12 @@ export const makeStudyTableHeaders = (datasets: DatasetTerm[], selected: number[
...
-              <Checkbox checked={fullySelected} indeterminate={indeterminate} disabled={!isSelectableStudy}
-                onClick={() => onSelect(fullySelected ? _.without(selected, ...studyDatasetIds) : indeterminate ? _.xor(_.without(selected, ...studyDatasetIds), studyDatasetIds) : [...selected, ...studyDatasetIds])}/>
-            </div>
-          </>,
+          data: <div data-for={checkboxId}>
+              <span title={isSelectableStudy ? '' : tooltipText}>
+                  <Checkbox checked={fullySelected} indeterminate={indeterminate} disabled={!isSelectableStudy}
+                            onClick={() => onSelect(fullySelected ? _.without(selected, ...studyDatasetIds) : indeterminate ? _.xor(_.without(selected, ...studyDatasetIds), studyDatasetIds) : [...selected, ...studyDatasetIds])}/>
+              </span>
+          </div>,
```
snf2ye, line 100: Just want to make sure I understand: Is this removing the tooltip in this case?

snf2ye (approved):
One question for my understanding, but LGTM

src/components/data_search/DatasetSearchTableConstants.tsx
pshapiro4broad: No, this change doesn't remove the tooltip. This case is similar to others in that it replaces `ReactToolTip` with a `span` with a `title`.

To make this case work, I used an empty tooltip text instead of using `disable`:

before:
```
            <ReactTooltip
              place={'top'}
              disable={isSelectableStudy}
```

after:
```
              <span title={isSelectableStudy ? '' : tooltipText}>
```

The logic here is that the tooltip is only shown in the study is not selectable. I've tested this out and it works when running locally.

PR #2765, opened 2025-01-08
[DT-1122] Apply `zizmor` suggestions

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1122

### Summary

* The Github actor is already logged on the action run page
* We should not persist credentials if we don't need to
* Use shell variable expansion instead of [template expansion](https://woodruffw.github.io/zizmor/audits/#template-injection)

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/component-tests.yml
+++ b/.github/workflows/component-tests.yml
@@ -4,8 +4,6 @@ jobs:
   cypress-run:
     runs-on: ubuntu-latest
     steps:
-      - name: Log Actor
-        run: echo "${{ github.actor }}"
       - name: Checkout
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/component-tests.yml
+++ b/.github/workflows/component-tests.yml
@@ -8,6 +8,8 @@ jobs:
         run: echo "${{ github.actor }}"
       - name: Checkout
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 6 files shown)


pshapiro4broad (approved):
Looks good. I wonder why `true` is the default, if it's not needed by most and could lead to this problem.

rushtong (approved):
👍🏽

.github/workflows/component-tests.yml
```diff
@@ -4,10 +4,10 @@ jobs:
   cypress-run:
     runs-on: ubuntu-latest
     steps:
-      - name: Log Actor
-        run: echo "${{ github.actor }}"
```
rjohanek, line 8: what does this change do?

rjohanek (approved):
looks good to me!

.github/workflows/component-tests.yml
fboulnois: When you run this line it will say something like `fboulnois` as the person who ran the GitHub Action, but that information is already available on the GitHub Action page for the particular workflow.

rushtong: This was leftover code from when it was a lot more difficult to track who was running what.

pshapiro4broad (approved):

PR #2791, opened 2025-02-18
[DT-1250][risk=no] Check for lint in the files that have changed between this PR and the destination branch.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1250](https://broadworkbench.atlassian.net/browse/DT-1250)

### Summary

Lint checking on only files that have changed in this PR.

### Screenshots

When there's no lint, the action passes as expected.  The file in this case is excluded because of linter configuration.
image

When there's lint, we get a failure:
image

When the lint is fixed:
image

Lint history tracked in PR history:
image

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,4 +16,6 @@ jobs:
       - name: install dependencies
         run: npm ci
       - name: npm lint
-        run: npx eslint $(git diff origin/${{ github.event.pull_request.base.ref }} --name-only | xargs)
+        run: npx eslint $(git diff "origin/$GITHUB_REF" --name-only | xargs)
+        env:
+                GITHUB_REF: ${{ github.event.pull_request.base.ref }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,17 @@
+on: [pull_request]
+name: changed files lint
+jobs:
+  scan:
+    name: npm lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.11.0'
+      - name: install dependencies
+        run: npm ci
+      - name: npm lint
+        run: npx eslint $(git diff ${{ github.event.pull_request.base.ref }} --name-only | xargs)
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,4 +16,4 @@ jobs:
       - name: install dependencies
         run: npm ci
       - name: npm lint
-        run: npx eslint $(git diff ${{ github.event.pull_request.base.ref }} --name-only | xargs)
+        run: npx eslint $(git diff origin/${{ github.event.pull_request.base.ref }} --name-only | xargs)
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,14 +10,12 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
           fetch-tags: true
-        env:
-          NODE_VERSION: $(cat Dockerfile | awk 'NR==2 {gsub(":","@",$2); print $2}' | awk '{split($0, array, "@"); print array[2]}')
+      -  run: echo "NODE_VERSION=$(cat Dockerfile | awk 'NR==2 {gsub(":","@",$2); print $2}' | awk '{split($0, array, "@"); print array[2]}')" >> $env:GITHUB_ENV
+      -  run: echo ${{ env.NODE_VERSION }}
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ env.NODE_VERSION }}
       - name: install dependencies
         run: npm ci
       - name: npm lint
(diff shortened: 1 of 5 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+name: changed files lint
+jobs:
+  scan:
+    name: npm lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 0
```
otchet-broad, line 11: Required to get the develop branch.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+          fetch-tags: true
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.11.0'
+      - name: install dependencies
+        run: npm ci
+      - name: npm lint
+        run: npx eslint $(git diff origin/${{ github.event.pull_request.base.ref }} --name-only | xargs)
```
otchet-broad: Requires origin prefix because the base ref does not include the fully qualified name of the ref as required by git diff.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.11.0'
```
rushtong: I wonder if there is a way to tie this to Dockerfile like we do here: https://github.com/DataBiosphere/duos-ui/blob/d281e1effa6233d711bcff89a0fed3ab9e423def/docs/getting-started.md?plain=1#L73

rushtong: However, we reference node directly in other actions as well, so definitely not a blocker.

rushtong (approved):
👍🏽

.github/workflows/lint.yml
fboulnois:
```suggestion
        run: npx eslint $(git diff "origin/$GITHUB_REF" --name-only | xargs)
        env:
                GITHUB_REF: ${{ github.event.pull_request.base.ref }}
```

Consider pulling out `${{ github.event.pull_request.base.ref }}` into an environment variable to avoid a template injection.

fboulnois (commented):
Looks good, one suggestion below:

.github/workflows/lint.yml
otchet-broad: Should we reference the `develop` branch or the branch the PR is on?  I'll look at extracting the file into an environment variable and l

.github/workflows/lint.yml
fboulnois: I see why the failure is happening, `GITHUB_REF` is already an environment variable: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

I would switch it to `GITHUB_BASE_REF` or something similar.

fboulnois:
I see why the failure is happening, `GITHUB_REF` is already an environment variable: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

I would switch it to `GITHUB_CUSTOM_EVENT_REF` or something similar. You may also be able to use `GITHUB_BASE_REF` directly, without using the env block.

.github/workflows/lint.yml
rushtong: > Should we reference the `develop` branch or the branch the PR is on

I think it makes the most sense to pull the version from the current PR branch.

otchet-broad: Huzzah!  Figured this out!

.github/workflows/lint.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  scan:
+    name: npm lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      -  run: echo "NODE_VERSION=$(cat Dockerfile | awk 'NR==2 {gsub(":","@",$2); print $2}' | awk '{split($0, array, "@"); print array[2]}')" >> $GITHUB_ENV
```
rushtong, line 12: Nice!

rushtong (approved):
👍🏽

.github/workflows/lint.yml
rushtong: Just in time for the node update in https://github.com/DataBiosphere/duos-ui/pull/2789 😄

otchet-broad: Do you think it makes sense to migrate the other workflows that use node-version to this methodology?

fboulnois (approved):
🎉

.github/workflows/lint.yml
rushtong: > Do you think it makes sense to migrate the other workflows that use node-version to this methodology?

I do, but that seems like a new ticket to me.

otchet-broad: Per standup, I've added the changes to this PR.

PR #2797, opened 2025-02-27
DT-1289: Update min participant count calculation in the data library

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1289

### Summary
On prod, there are datasets with participant counts of `0`, but the default min value shown in the filter defaults to `1`:

Screenshot 2025-02-27 at 3 05 03 PM

Manually setting that to `0` reveals a number of filtered datasets:

Screenshot 2025-02-27 at 3 06 10 PM

This PR fixes the calculation logic so that it doesn't exclude datasets with a `0` count.

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetFilterConstants.tsx
+++ b/src/components/data_search/DatasetFilterConstants.tsx
@@ -21,9 +21,9 @@ export const defaultFilters = (datasets: DatasetTerm[]): FiltersTypes => {
   };
 };
 
-export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce(() => {
+export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => {
   return {
     max: datasets.reduce((max, dataset) => Math.max(max, dataset.participantCount ? dataset.participantCount : 0), 0),
     min: 0
   };
-}, {max: 0, min: Infinity});
+}
--- a/src/components/data_search/DatasetFilterConstants.tsx
+++ b/src/components/data_search/DatasetFilterConstants.tsx
@@ -24,5 +24,5 @@ export const defaultFilters = (datasets: DatasetTerm[]): FiltersTypes => {
 export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce((acc, dataset) => {
   return {
     max: Math.max(acc.max, dataset.participantCount ? dataset.participantCount : 0),
-    min: Math.min(acc.min, dataset.participantCount ? dataset.participantCount : Infinity) };
+    min: 0 };
 }, {max: 0, min: Infinity});
--- a/src/components/data_search/DatasetFilterConstants.tsx
+++ b/src/components/data_search/DatasetFilterConstants.tsx
@@ -21,8 +21,9 @@ export const defaultFilters = (datasets: DatasetTerm[]): FiltersTypes => {
   };
 };
 
-export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce((acc, dataset) => {
+export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce(() => {
   return {
-    max: Math.max(acc.max, dataset.participantCount ? dataset.participantCount : 0),
-    min: 0 };
+    max: datasets.reduce((max, dataset) => Math.max(max, dataset.participantCount ? dataset.participantCount : 0), 0),
+    min: 0
+  };
 }, {max: 0, min: Infinity});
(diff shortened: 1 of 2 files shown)


src/components/data_search/DatasetFilterConstants.tsx
```diff
@@ -24,5 +24,5 @@ export const defaultFilters = (datasets: DatasetTerm[]): FiltersTypes => {
 export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce((acc, dataset) => {
   return {
     max: Math.max(acc.max, dataset.participantCount ? dataset.participantCount : 0),
-    min: Math.min(acc.min, dataset.participantCount ? dataset.participantCount : Infinity) };
+    min: 0 };
```
pshapiro4broad: Is this because some counts are undefined? Otherwise why wouldn't you want to use the lowest possible value?

If you only reduce over `max`, you should change the code to be explicit about that
```
    return { max: datasets.reduce((max, dataset) => 
            Math.max(max, dataset.participantCount ? dataset.participantCount : 0), 0), 
     min: 0 };
```

rushtong: > Is this because some counts are undefined? Otherwise why wouldn't you want to use the lowest possible value?

I'm not sure why it's failing to reduce down to 0, there are datasets with a populated value of 0, but it's only reducing down to 1.

I like the idea of simplifying the return.

rushtong: > Is this because some counts are undefined? Otherwise why wouldn't you want to use the lowest possible value?

I figured this out. When the count is `0`, the inner ternary is falsy, so it always returns `Infinity` instead of `0`

```
console.log(Math.min(Infinity, 0 ? 0 : Infinity));
```
yields `Infinity`

src/components/data_search/DatasetFilterConstants.tsx
```diff
@@ -21,8 +21,9 @@ export const defaultFilters = (datasets: DatasetTerm[]): FiltersTypes => {
   };
 };
 
-export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce((acc, dataset) => {
+export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce(() => {
```
otchet-broad: Do you still need this outer reducer?  Does this simplify and meet the need?
```suggestion
export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => {
  return {
    max: datasets.reduce((max, dataset) => Math.max(max, dataset.participantCount ? dataset.participantCount : 0), 0),
    min: 0
  };
}
```

rushtong: Good point, will try that

otchet-broad (approved):

rjohanek (approved):

fboulnois (commented):
This will show `0` even if all the participant counts in a filtered view are larger than zero. Is this behavior expected?

rushtong:
> This will show `0` even if all the participant counts in a filtered view are larger than zero. Is this behavior expected?

Yes, that is the requested behavior.

fboulnois:
> > This will show `0` even if all the participant counts in a filtered view are larger than zero. Is this behavior expected?
> 
> Yes, that is the requested behavior.

One issue is datasets with a count of 0 are being filtered out. I agree they shouldn't be, and this seems to be the behavior the ticket is trying to fix. Another issue is if all the datasets in a view (for example a data library) have a participant count that is greater than 0. In that case I would expect the default minimum to be whatever the smallest non-zero number of participants is, not zero.

rushtong:
> In that case I would expect the default minimum to be whatever the smallest non-zero number of participants is, not zero.

We can implement that approach - it just means adding a little bit more logic to the ternary, i.e.
```
...   max: Math.max(acc.max, dataset.participantCount ? dataset.participantCount : 0),
...   min: Math.min(acc.min, dataset.participantCount ? dataset.participantCount : Infinity)
```
needs to be re-written as something like this:
```
...   max: Math.max(acc.max, (dataset.participantCount >= 0) ? dataset.participantCount : 0),
...   min: Math.min(acc.min, (dataset.participantCount >= 0) ? dataset.participantCount : Infinity)
```

src/components/data_search/DatasetFilterConstants.tsx
pshapiro4broad: OK, so if you wanted to preserve the current behavior, using `??` instead of `?:` would be a fix
```
export const generateDefaultParticipantCountValues = (datasets: DatasetTerm[]) => datasets.reduce((acc, dataset) => {
  return {
    max: Math.max(acc.max, dataset.participantCount ?? 0),
    min: Math.min(acc.min, dataset.participantCount ?? Infinity) };
}, {max: 0, min: Infinity});
```

Although typescript doesn't think `participantCount` can be `undefined`, so maybe this logic isn't necessary?

PR #2799, opened 2025-02-27
[DT-1291] Remove `js-file-download`

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1291

### Summary

Replace `js-file-download`: this package has not received updates in 5 years, is very simple, bloats our package dependencies, has workarounds for unsupported browsers, doesn't have tests, and isn't fully in Typescript.

----

- Verify all tests go green

### DIFF
--- a/src/utils/FileDownload.ts
+++ b/src/utils/FileDownload.ts
@@ -8,10 +8,9 @@ export const fileDownload = (data: string | ArrayBuffer | ArrayBufferView | Blob
     a.href = blobUrl;
 
     document.body.appendChild(a);
-    a.click();
 
-    setTimeout(() => {
+    Promise.resolve(a.click()).then(() => {
         document.body.removeChild(a);
         URL.revokeObjectURL(blobUrl);
-    }, 100);
+    });
 }
--- a/src/utils/FileDownload.ts
+++ b/src/utils/FileDownload.ts
@@ -1,0 +1,17 @@
+export const fileDownload = (data: string | ArrayBuffer | ArrayBufferView | Blob, filename: string, mime?: string) => {
+    const blob = new Blob([data], { type: mime || 'application/octet-stream' });
+    const blobUrl = URL.createObjectURL(blob);
+
+    const a = document.createElement('a');
+    a.style.display = 'none';
+    a.download = filename;
+    a.href = blobUrl;
+
+    document.body.appendChild(a);
+    a.click();
+
+    setTimeout(() => {
+        document.body.removeChild(a);
+        URL.revokeObjectURL(blobUrl);
+    }, 100);
+}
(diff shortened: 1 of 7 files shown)


src/utils/FileDownload.ts
```diff
@@ -0,0 +1,16 @@
+export const fileDownload = (data: string | ArrayBuffer | ArrayBufferView | Blob, filename: string, mime?: string) => {
+    const blob = new Blob([data], { type: mime || 'application/octet-stream' });
```
otchet-broad, line 2: I appreciate the test coverage you've added for this!  Thank you!

Do we care about the mime type?  If the point of this code is to trigger "download" capability, 'application/octet-stream' will work in all cases because the browser will not be able to render the Blob and will always prompt to "[Save As](https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types#applicationoctet-stream)"

Looks like DAA.js and DAR.js don't pass mime type, so in all cases of application usage, It is effectively "application/octet-stream'

src/utils/FileDownload.ts
```diff
@@ -0,0 +1,16 @@
...
+    const a = document.createElement('a');
+    a.style.display = 'none';
+    a.download = filename;
+    a.href = blobUrl;
+
+    document.body.appendChild(a);
+
+    Promise.resolve(a.click()).then(() => {
+        document.body.removeChild(a);
+        URL.revokeObjectURL(blobUrl);
```
otchet-broad, line 14: A11Y guides I've followed in the past suggest leaving the link so that a screen reader doesn't loose context after the link is clicked.  Did you have a specific need to remove the link?

otchet-broad: I have a better appreciation of what's happening here now.

I'm almost wondering if it doesn't make sense to just refactor this code into DownloadLink to work with the element there so it can be refocused when the download is complete.

A different discussion (outside of this PR) might be switching to using signed URLs since the documents are already stored in GCS.

src/utils/FileDownload.ts
fboulnois: I rewrote it like this to keep maximum compatibility with the `js-file-download` package, without including the deprecated bits (like `bom`)

src/utils/FileDownload.ts
fboulnois: We should revisit how we do downloads to make it consistent across the UI, but I think we would need to do a larger evaluation of all the different ways we do downloads more generally. In the short term my goal is to deprecate the use of this package.

otchet-broad (approved):

snf2ye (approved):

rushtong (approved):
👍🏽

PR #2804, opened 2025-03-06
DT-1318: Show the custodian emails on the statistics page

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1318

### Summary
* Update the source of the Data Custodian value when we have it populated on the study.
* Update typing and linting warnings
* New test to ensure we're showing the custodian as defined in the study
* Clean up the initialization - we were originally making 2 calls for the dataset, one by identifier and one by id. They both return the same payload, so we didn't need both.

### Screens
**Before:**
Screenshot 2025-03-06 at 7 36 34 AM

**After:**
Screenshot 2025-03-06 at 7 36 52 AM

----

- Verify all tests go green

### DIFF
--- a/src/pages/DatasetStatistics.tsx
+++ b/src/pages/DatasetStatistics.tsx
@@ -32,7 +32,6 @@ interface DatasetStatisticsProps {
 
 export default function DatasetStatistics(props: DatasetStatisticsProps) {
   const {history, match: {params: {datasetIdentifier}}} = props;
-  const [datasetId, setDatasetId] = useState<number>();
   const [dataset, setDataset] = useState<Dataset>();
   const [dars, setDars] = useState<Array<DataAccessRequest>>();
   const [isLoading, setIsLoading] = useState(true);
@@ -51,7 +50,7 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
 
   const applyForAccess = async () => {
     try {
-      const draftResponse = await DAR.postDarDraft({datasetId: [datasetId]});
+      const draftResponse = await DAR.postDarDraft({datasetId: [dataset?.datasetId]});
       if (draftResponse.referenceId) {
         history.push(`/dar_application/${draftResponse.referenceId}`);
       } else if (draftResponse.message) {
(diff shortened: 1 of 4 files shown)


snf2ye (approved):

src/types/model.ts
```diff
@@ -181,7 +181,7 @@ export interface DatasetTerm {
 
 export interface AccessManagementSummary {
   name: string;
-  icon: any;
+  icon: unknown;
```
rjohanek, line 184: what does changing from any to unknown do?

src/pages/DatasetStatistics.tsx
```diff
@@ -148,19 +160,19 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
                   {extract('Principal Investigator(PI)') || dataset?.study?.piName}
                 </div>
               </div>}
-              {(extract('Data Depositor') || dataset?.createUser?.displayName) && <div style={{display: 'flex'}}>
+              {(extractStudyProp('dataCustodianEmail') || dataset?.createUser?.displayName) && <div style={{display: 'flex'}}>
                 <div style={Styles.SMALL_BOLD}>Data Custodian:</div>
                 <div style={Styles.SMALL_BOLD}>
-                  {extract('Data Depositor') || dataset?.createUser?.displayName}
+                  {extractStudyProp('dataCustodianEmail') || dataset?.createUser?.displayName}
```
rjohanek, line 166: this is the key change right?

rjohanek (approved):
this looks good!

src/types/model.ts
rushtong: This was a eslint suggestion. `any` triggers an action failure. See, for example [this failure](https://github.com/DataBiosphere/duos-ui/actions/runs/13698959849/job/38307800333):
```
Error:   180:28  error    Unexpected any. Specify a different type                                                                    @typescript-eslint/no-explicit-any
```
Normally, I'd try to make an interface for this, but I could not figure it out - it points to an svg icon file. I chose to follow the eslint suggestion.

src/pages/DatasetStatistics.tsx
rushtong: Essentially, yes, this is the bug fix. Everything else is just getting the page up to current typing/linting standards.

PR #2806, opened 2025-03-12
[DT-1288][risk=no] Create commonly used select options from typed objects

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1288](https://broadworkbench.atlassian.net/browse/DT-1288)

### Summary

Adding types and helpers for commonly filled form select elements to reduce duplicate code in upcoming form.  Currently only used in tests.

Adds test to demonstrate default value bug with select input.

### Screenshots
image

image

image

image

----

- Verify all tests go green


cypress/component/Forms/forms.spec.jsx
```diff
@@ -675,6 +675,33 @@ describe('FormField - Tests', () => {
...
+          value: [{displayName: 'Observational', displayText: 'Observational'},
+            {displayName: 'Observational', displayText: 'Observational'}],
+          isValid: true
+        });
+      });
+      cy.get('#studyType').type('ObS{enter}');
+      cy.get('#studyType').then(() => {
+        expect(props.onChange).to.be.calledWith({
+          key: 'studyType',
+          value: [{displayName: 'Observational', displayText: 'Observational'},
```
otchet-broad: The consequence of the bug: we get two of the same selected objects in the change event!

package-lock.json
otchet-broad, line 1: Changes needed to fix the moderate npm identified concern.

otchet-broad, line 1: Fix for npm identified moderate risk issue.

src/components/forms/DataTypes.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    static dataTypesList = [DataTypes.CITE,
+    DataTypes.HYB,
+    DataTypes.RNA,
+    DataTypes.SCRNA,
+    DataTypes.SPT,
+    DataTypes.SNRNA,
+    DataTypes.WGS,
+    DataTypes.WES]
```
fboulnois: Should this be pulled into a separate type? (Ditto for the other lists)

fboulnois (approved):
looks reasonable, one suggestion:

src/components/forms/DataTypes.ts
otchet-broad: Good question.  I may iterate on this and eventually pull it into a type, but didn't want to go there until I needed to do that.  The IDE correctly infers dataTypesList as an array of SelectOptionWithKeyNameAndAbbreviation. 
One of the things I'm thinking about is if these even stay in the client since they're shared between the front end for selection and backend for conformance checking.  In some ways, I really like how jade-data-repo leverages OpenAPI for sharing this information between various consumers.  I might be more inclined to go that route in the future with this - but it warrants some discussion.
I was starting to see where the larger form I was working on was getting polluted by long static lists and took a stab at minimizing those to get started.

cypress/component/Forms/forms.spec.jsx
otchet-broad: Memorialized in [https://broadworkbench.atlassian.net/browse/DT-1351](https://broadworkbench.atlassian.net/browse/DT-1351)

src/components/forms/DataTypes.ts
otchet-broad: I've added type declarations to each of the arrays so it's explicit for now.

cypress/component/Forms/SelectOptions.spec.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react'
+import {mount} from 'cypress/react'
+import {FormField, FormFieldTypes} from '../../../src/components/forms/forms';
+import {asIdAndDisplayText, getFormattedName} from '../../../src/components/forms/SelectOptionInterface';
+import {StudyType} from '../../../src/components/forms/StudyType';
+import {SecondaryDataUseTerms} from '../../../src/components/forms/SecondaryDataUseTerms';
+import {NIHInstitutesAndCenters} from '../../../src/components/forms/NIHInstitutesAndCenters';
```
pshapiro4broad: Is this a test of `FormField` or of `StudyType`, `SecondaryDataUseTerms`, `NIHInstitutesAndCenters`?

Unless the data objects are being tested, I think it's better to use data defined in the test code only, to isolate the test from the test data.

If the data objects do require a test, it would be better to test those in isolation as well

cypress/component/Forms/SelectOptions.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+            });
+        });
+        it('should load a list of Secondary Data Use Terms with options populated', () => {
+            const props = {
+                onChange: () => {
+                },
+                id: 'secondaryDataUse',
+                title: 'Secondary Data Use',
+                type: FormFieldTypes.SELECT,
+                selectOptions: asIdAndDisplayText(SecondaryDataUseTerms.secondaryDataUseTermsList),
```
pshapiro4broad: This is the kind of dependency I think it would be better to avoid. Unless this is testing the behavior of `SecondaryDataUseTerms.secondaryDataUseTermsList`, it would be better to use test-only data to validate the behavior of `FormField`. That way this test doesn't need to change if the data happens to change.

cypress/component/Forms/forms.spec.jsx
```diff
@@ -675,6 +675,33 @@ describe('FormField - Tests', () => {
 
     });
 
+    it('default value can allow second entry of the same type', () => {
```
pshapiro4broad: What is the purpose of this test? To validate a correct behavior, or prove that the code is working incorrectly?

If it's to show a bug, I don't think it should be committed as a test. Instead I would add it to the bug ticket so we know how to reproduce it in the future.

src/components/forms/DataTypes.ts
```diff
@@ -0,0 +1,48 @@
+import {SelectOptionWithKeyNameAndAbbreviation} from './SelectOptionInterface';
+
+
+export class DataTypes {
+    static CITE: SelectOptionWithKeyNameAndAbbreviation = {
```
pshapiro4broad, line 5: Why are these fields defined in addition to having them in the array below? It seems like their definition is redundant with the `key` field; if you instead defined them in the array, you could have a static method which returns the element given a `key`.

src/components/forms/DataTypes.ts
```diff
@@ -0,0 +1,48 @@
...
+        key: 'CITE',
+        name: 'CITE-seq'
+    };
+    static HYB: SelectOptionWithKeyNameAndAbbreviation = {
```
pshapiro4broad, line 9: Doesn't typescript already enforce this type due to this field's inclusion in the array below? I thought that if the array is typed, typescript would enforce that its elements conform to the array type (at least for static objects like these). It would be nice if we could avoid specifying the type for each member.

src/components/forms/DataTypes.ts
```diff
@@ -0,0 +1,48 @@
...
+        abbreviation:'WGS',
+        name: 'Whole Genome'
+    }
+    static WES: SelectOptionWithKeyNameAndAbbreviation = {
+        key: 'WES',
+        abbreviation: 'WES',
+        name: 'Whole Exome'
+    }
+
+    static dataTypesList:SelectOptionWithKeyNameAndAbbreviation[] = [DataTypes.CITE,
```
pshapiro4broad: This field is called `nihInstitutesAndCentersList` in `NIHInstitutesAndCenters`. If the fields have a similar meaning, maybe a generate name could be used so the name is consistent in all types. I don't think including the class name as part of the field name is necessary, since all uses of the field outside its definition must be explicitly scoped. What about `values` (or `VALUES` since it's a `static` member)?

pshapiro4broad (approved):
Looks OK although I have some suggestions for changes

src/components/forms/DataTypes.ts
otchet-broad: If the only access were through the list, that would be okay however, I don't expect the only access to be through the list due to conditional logic we have in validation rules that use these.

src/components/forms/DataTypes.ts
otchet-broad: I like the thought.  VALUE it is!

cypress/component/Forms/forms.spec.jsx
otchet-broad: The purpose was to first figure out why I was seeing the behavior - if it was code I had in flight, or something core to the element.  Turns out that it's something core to the element.  I've left it in to demonstrate to other developers in the code base something that behaves unexpectedly.

I'll gladly move it to the ticket.

src/components/forms/SelectOptionInterface.ts
```diff
@@ -0,0 +1,21 @@
+import {SelectEntry} from "src/components/forms/SelectOptions";
+
+export interface SelectOptionWithKeyNameAndAbbreviation {
+    key: string,
+    name?: string,
```
pshapiro4broad: It looks like all the values so far have `name`, can this be non-optional?

src/components/forms/SelectOptionInterface.ts
```diff
@@ -0,0 +1,21 @@
...
+    abbreviation?: string
+}
+
+export function getFormattedName(entry: SelectOptionWithKeyNameAndAbbreviation) {
+    if (entry.abbreviation && entry.name) {
+        return `${entry.name} (${entry.abbreviation})`;
+    }
+    if (entry.name) {
+        return `${entry.name}`;
```
pshapiro4broad: How is this different than `entry.name`? `name` is already `string?` and its type will be `string` since it's inside `if (entry.name) {`.
```suggestion
        return entry.name;
```

src/components/forms/DataTypes.ts
otchet-broad: It may seem like it, but we have a case where the key we're matching includes a "-".  We can't include that in the variable name, hence why I explicitly added the key field.
See GSMINUS in SecondaryDataUseTerms.ts

src/components/forms/SelectOptionInterface.ts
otchet-broad: It's a code style habit from another project.  Adjusted.

PR #2821, opened 2025-04-10
DT-1496: Remove institution selection/update

rushtong:
### Addresses
Front end work for https://broadworkbench.atlassian.net/browse/DT-1496

### Summary
* Remove the ability for users to update their institution
* Cleans up the unused `suggestedSigningOfficial` field and other unnecessary intermediate state objects
* Minor styling updates to distinguish better between the subheads and the content

Note that there will be a [follow-on PR](https://github.com/DataBiosphere/consent/pull/2477) in Consent to ensure users cannot update their institution without administrative support.

### Screens

When a user has an institution:
Screenshot 2025-04-10 at 1 24 06 PM

When a user does NOT have an institution:
Screenshot 2025-04-10 at 1 23 19 PM

----

- Verify all tests go green


s-rubenstein (approved):

src/pages/user_profile/AffiliationAndRoles.jsx
```diff
@@ -1,190 +1,39 @@
...
 
   useEffect(() => {
     const init = async () => {
       try {
-        const rolesList = [];
-        if (!isNil(user) && !isNil(user.roles)) {
-          for (let i = 0; i < user.roles.length; i++) {
-            const newRole = user.roles[i].name;
-            rolesList.push(newRole);
+        const allRoles = user?.roles?.map((role) => role.name).join(', ');
```
pshapiro4broad: This is good but it would be nice to typescriptfiy this too when we have time, otherwise it's hard to know if the `?`s are required or not

src/pages/user_profile/AffiliationAndRoles.jsx
```diff
@@ -197,62 +46,16 @@ export default function AffiliationAndRole(props) {
...
-      <p>Institution</p>
-      {!(profile.institutionId) && <div className={'alert-danger alert'} style={{fontWeight: 'normal'}}>
-        Please select an existing Institution. If you don&apos;t see your institution listed, please select the Contact
-        Us link to request that your institution be added.
-      </div>}
+    <div>
+      <p style={subHeadStyle}>Institution</p>
+      <div style={{ marginTop: '15px' }} />
+      {institution
+        ? <div data-cy='institutional-affiliation'>{institution?.name}</div>
```
pshapiro4broad: I believe the `?` isn't necessary as it's inside a block where `institution` is known to be defined.
```suggestion
        ? <div data-cy='institutional-affiliation'>{institution.name}</div>
```

pshapiro4broad (approved):
Looks good

fboulnois (approved):
looks good, I agree with @pshapiro4broad 's comment if we can Typescriptify a little and if it's not a big lift 👍

src/pages/user_profile/AffiliationAndRoles.jsx
rushtong: Good point, I think this was a leftover from an earlier version I had in place.

src/pages/user_profile/AffiliationAndRoles.jsx
rushtong: I'll give that a shot ... moving components jsx to tsx have typically been non-trivial in this codebase

fboulnois (approved):
👍

src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -0,0 +1,66 @@
...
+export default function AffiliationAndRole(props: AffiliationAndRoleProps) {
+
+  const {user} = props;
+  const [institution, setInstitution] = useState<Institution>();
+  const [roles, setRoles] = useState<string>('');
+
+  useEffect(() => {
+    const init = async () => {
+      try {
+        const allRoles = user?.roles?.map((role) => role.name).join(', ');
```
pshapiro4broad: Looking at the types, `user` and `user.role` are both non-optional. Are the types incorrect?
```suggestion
        const allRoles = user.roles.map((role) => role.name).join(', ');
```

src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -0,0 +1,66 @@
...
+        setRoles(allRoles);
+        if (user?.institutionId) {
```
pshapiro4broad:
```suggestion
        if (user.institutionId) {
```

pshapiro4broad (approved):
Looks good, always nice to see more typescript

src/pages/user_profile/AffiliationAndRoles.tsx
rushtong: Correct, I'll clean that up

src/pages/user_profile/AffiliationAndRoles.tsx
rushtong: `institutionId` is technically optional now. In the new world of limited institutional domains, that will continue to be true.

rushtong: Ah, but the user is not, thank you!

s-rubenstein (approved):
Pending Phil's recommendations.

PR #2827, opened 2025-04-15
DT-1528: Handle undefined user case

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1528

### Summary
This fixes a bug introduced in https://github.com/DataBiosphere/duos-ui/pull/2821
The user passed in from the parent component can be an empty object instead of a fully populated user. This causes an error message to flash while state is being updated and the component is being re-rendered.

----

- Verify all tests go green

### DIFF
--- a/src/pages/user_profile/AffiliationAndRoles.tsx
+++ b/src/pages/user_profile/AffiliationAndRoles.tsx
@@ -16,7 +16,7 @@ export default function AffiliationAndRole(props: AffiliationAndRoleProps) {
   useEffect(() => {
     const init = async () => {
       try {
-        const allRoles = user.roles.map((role) => role.name).join(', ');
+        const allRoles = user?.roles?.map((role) => role.name).join(', ');
         setRoles(allRoles);
         if (user.institutionId) {
           const institution: Institution = await InstitutionAPI.getById(user.institutionId);
--- a/src/pages/user_profile/AffiliationAndRoles.tsx
+++ b/src/pages/user_profile/AffiliationAndRoles.tsx
@@ -20,7 +20,7 @@ export default function AffiliationAndRole(props: AffiliationAndRoleProps) {
         // not honor that and pass an undefined entity during load.
         const allRoles = user?.roles?.map((role) => role.name).join(', ');
         setRoles(allRoles);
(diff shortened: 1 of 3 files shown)


src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -16,7 +16,9 @@ export default function AffiliationAndRole(props: AffiliationAndRoleProps) {
   useEffect(() => {
     const init = async () => {
       try {
-        const allRoles = user.roles.map((role) => role.name).join(', ');
+        // Note that although user is a required prop and roles is a required property, the parent JSX component may
+        // not honor that and pass an undefined entity during load.
```
pshapiro4broad, line 20: If the proper fix is to change this code, then shouldn't the typescript type for `user` and `roles` be updated to match this situation?

Or is there a way to fix the parent component so it initializes this data so there's no undefined data?

fboulnois (approved):
Is it possible to add a test for this case?

pshapiro4broad (approved):
Looks OK for a fix but it'd be nice to update the code so the types and usage agree.

src/pages/user_profile/AffiliationAndRoles.tsx
rushtong: I'm not sure what to do about conditionally optional properties in TS. User is a required prop, and roles is a required field on user, so the current model is correctly defined at present. I think the root problem is that we're mixing JSX and TS. Refactoring the parent component is, in my mind, a different ticket and out of scope.

rushtong:
> Is it possible to add a test for this case?

I have added a test for the error message display, but not specifically for this test case. The test class and component are both TS, so it won't compile if we try to pass an incomplete user object to the component. I think we're bumping up against the limits of JSX/TS interop.

rushtong:
> > Is it possible to add a test for this case?
> 
> I have added a test for the error message display, but not specifically for this test case. The test class and component are both TS, so it won't compile if we try to pass an incomplete user object to the component. I think we're bumping up against the limits of JSX/TS interop.

Actually, I do think I have a way of doing this - test case to come.

PR #2838, opened 2025-04-24
DT-1579: Show error message when ECM fails

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1579

## Summary
When ECM returns a 403, the error is swallowed and nothing is shown in the UI. This PR bubbles that error up so the user can understand what is happening. 

Includes new test to ensure error is displayed and adds some cleanup to existing syntax warnings.

### New error message:
Screenshot 2025-04-24 at 4 50 04 PM

----

- Verify all tests go green


fboulnois (approved):
looks good, thank you! 👍

src/components/ERACommons.jsx
```diff
@@ -135,7 +140,7 @@ export default function ERACommons(props) {
           <span style={{verticalAlign: '50%'}}>Authenticate your account</span>
         </a>
       )}
-      {nihError && <span className='era-cancel-color era-required-field-error-span'>{nihErrorMessage}</span>}
+      {nihError && <span data-cy='era-commons-error-span' className='era-cancel-color era-required-field-error-span'>{nihErrorMessage}</span>}
```
pshapiro4broad: The only use of `nihError` is here, to gate the showing of an error message. It looks like this change removes the message used by the other error cases (error on missing decoded token, error on delete account linkage error). I would fix both of these issues by removing `nihError` and use `nihErrorMessage` instead of `nihError` here, and add new error messages to the two other cases that are currently setting `nihError`.

cypress/component/ERACommons/era_commons.spec.jsx
```diff
@@ -101,10 +100,32 @@ describe('ERA Commons Component', function () {
...
+    cy.get('[data-cy=era-commons-header]').should('exist');
+    cy.get('[data-cy=era-commons-required]').should('exist');
+    cy.get('[data-cy=era-commons-authenticate-link]').should('not.exist');
+    cy.get('[data-cy=era-commons-id-value]').should('exist');
     cy.get('.required-field-error-span').should('not.exist');
   });
-});
\ No newline at end of file
+
+  it('shows an error when ECM fails required', function () {
```
pshapiro4broad: This change could break the display of an error message on delete failing or an incorrect token, since those also use the `nihError` code path. If it's not too much work, I would add tests for those cases too.

pshapiro4broad (approved):
The fix looks OK but I'm not sure if it breaks the other error case handling code.

rushtong:
@pshapiro4broad - I took another stab at consolidating the error handling and display as well as adding new tests for the untested error conditions.

src/components/ERACommons.jsx
```diff
@@ -94,55 +97,64 @@ export default function ERACommons(props) {
...
-      document.getElementById('era-commons-id').scrollIntoView({block: 'start', inline: 'nearest', behavior: 'smooth'});
+    } catch (_error) {
+      const errorMessage = 'Something went wrong removing your account, please submit an error report using the "Contact Us" form.';
+      displayError(errorMessage);
     }
   };
 
+  const displayError = (message) => {
+    setNihErrorMessage(message);
+    setNihError(true);
```
pshapiro4broad: This is fine, but if `nihError` is only `true` when `nihErrorMessage` is defined, we don't need both values. And in fact I think it's better to write `{nihErrorMessage && ... {nihErrorMessage} ...}` than `{nihError && ... {nihErrorMessage} ...}`.

pshapiro4broad (approved):
Great, nice to see the tests and it's good that the other cases are handled. I still think `nihError` is now redundant but this is a style issue, not one of correctness.

src/components/ERACommons.jsx
rushtong: Good point, I can clean that up.

PR #2843, opened 2025-04-29
[DT-1587] Add expires at column

s-rubenstein:
### Addresses

Screenshot 2025-04-25 at 2 00 06 PM

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/components/dar_collection_table/DarCollectionTable.jsx
+++ b/src/components/dar_collection_table/DarCollectionTable.jsx
@@ -5,7 +5,7 @@ import { Storage } from '../../libs/storage';
 import PaginationBar from '../PaginationBar';
 import { recalculateVisibleTable, goToPage as updatePage } from '../../libs/utils';
 import SimpleTable from '../SimpleTable';
-import cellData, {consoleTypes} from './DarCollectionTableCellData';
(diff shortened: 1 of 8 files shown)


rushtong (approved):
This looks good on all consoles. I do have one request which is that the search filter doesn't filter on this column. The current filter function is [here](https://github.com/DataBiosphere/duos-ui/blob/d2fff0f133d3eebd19e4734fbd0fac6c802d3a34/src/libs/utils.js#L327C14-L327C38). I believe this page is using the `darCollections` method to do the search bar filter. If we added the expiration date string to the search term array (see what we're doing to get a formatted submission date on line 367), I think this would work.

Edit: just tested, and yes, this will add it to the filter:
```
        filter(collection => {
          const {darCode, datasetCount, institutionName, name, researcherName, status, submissionDate, expiresAt} = collection;
          const formattedSubmissionDate = formatDate(submissionDate);
          const formattedExpiresAt = formatDate(expiresAt);
          const matched = find((phrase) => {
            const termArr = term.split(' ');
            return find(term => includes(toLower(term), toLower(phrase)))(termArr);
          })([darCode, datasetCount, institutionName, name, researcherName, status, formattedSubmissionDate, formattedExpiresAt]);
```

otchet-broad (approved):

PR #2844, opened 2025-04-29
[DT-1573 ]Allow admins to assign the service account role.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1573](https://broadworkbench.atlassian.net/browse/DT-1573)
### Summary
To enable email notifications, a new service endpoint is being established.  This endpoint can only be invoked by users with the role SERVICE_ACCOUNT.

This UI change adds the ability for an admin to assign the role SERVICE_ACCOUNT.

image

----

- Verify all tests go green

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -158,6 +158,7 @@ export const USER_ROLES = {
   alumni: 'Alumni',
   signingOfficial: 'SigningOfficial',
   dataSubmitter: 'DataSubmitter',
+  serviceAccount: 'ServiceAccount',
   all: 'All'
 };
 
@@ -192,6 +193,7 @@ export const setUserRoleStatuses = (user, Storage) => {
   user.isAlumni = currentUserRoles.indexOf(USER_ROLES.alumni) > -1;
   user.isSigningOfficial = currentUserRoles.indexOf(USER_ROLES.signingOfficial) > -1;
   user.isDataSubmitter = currentUserRoles.indexOf(USER_ROLES.dataSubmitter) > -1;
+  user.isServiceAccount = currentUserRoles.indexOf(USER_ROLES.serviceAccount) > -1;
   Storage.setCurrentUser(user);
   return user;
 };
--- a/cypress/component/SignIn/sign_in_button.spec.jsx
+++ b/cypress/component/SignIn/sign_in_button.spec.jsx
@@ -21,6 +21,7 @@ const duosUser = {
   isMember: false,
   isResearcher: false,
   isSigningOfficial: false,
+  isServiceAccount: false,
   roles: [{
     name: 'Admin'
   }]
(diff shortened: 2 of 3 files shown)


otchet-broad:
Needs [https://github.com/DataBiosphere/consent/pull/2505](https://github.com/DataBiosphere/consent/pull/2505) to land in consent before merging this PR.

rjohanek (approved):

rushtong (approved):
Once the back end role update is merged, 👍🏽

src/libs/utils.js
```diff
@@ -192,6 +193,7 @@ export const setUserRoleStatuses = (user, Storage) => {
   user.isAlumni = currentUserRoles.indexOf(USER_ROLES.alumni) > -1;
   user.isSigningOfficial = currentUserRoles.indexOf(USER_ROLES.signingOfficial) > -1;
   user.isDataSubmitter = currentUserRoles.indexOf(USER_ROLES.dataSubmitter) > -1;
+  user.isServiceAccount = currentUserRoles.indexOf(USER_ROLES.serviceAccount) > -1;
```
pshapiro4broad, line 196: I guess this is the way things work, but why do we need to precompute `indexOf()` for each role here? It seems like each of these booleans could be functions that accept `DuosUser`, or one function that accepts `User` and `Role`.

Also, why do we need to add a boolean for this role, if a user with this role will never log into the UI? Can we wait to add this until it's needed? `isAlumni` also appears to be unused in the UI.

pshapiro4broad (approved):
Changes look OK but I have questions about the `isXX` properties on `DuosUser`.

fboulnois (approved):
👍

fboulnois (commented):
We typically assign a service account directly in the system, not through the UI. If an admin accidentally assigns a user service account role, is the risk tolerable? Why wouldn't we do this directly through the Swagger UI?

src/libs/utils.js
otchet-broad: We have a test that ensures we can compute the presence of a role on a user.  That role presence is computed in this way, and without this, a test failed.

You raise a great point of if we should be considering doing something else, overall.  I'd like to revisit that in the future.

rushtong: `isAlumni` is a very legacy role - we haven't used that in at least 5 years.

otchet-broad:
@fboulnois - Good question!  IMHO, there's no more risk than an admin accidentally making other changes - like making another user an admin, for example.  The key point here is that we're adding a role for a fixed purpose.  I'm not opposed to  abandoning this change if we only want to do this via swagger.  The downside is that if you bring the user up in the frontend, you won't see the role assigned.

otchet-broad:
Per discussion in standup today, we'll land this work as-is and change if we need to.

PR #2845, opened 2025-04-30
[DT-1597] - Change Unreviewed in UI to Submitter for Researcher DAR table via update to utils.js

solideoglori:
DT-1597

Change Unreviewed in UI to Submitter for Researcher DAR table

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -256,7 +256,7 @@ export const outputCommaSeperatedElectionStatuses = (elections) => {
     elections.map((e) => processElectionStatus(e, e.votes, false))
   ).filter((status) => !isEmpty(status));
   if (isEmpty(statuses)) {
-    return 'Unreviewed';
+    return 'Submitted';
   }
   return statuses.join(', ');
 };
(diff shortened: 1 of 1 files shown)


rushtong (commented):
This requires some test changes. Look at:
https://github.com/DataBiosphere/duos-ui/blob/8f40c92f1f38b9cf1c783198cc4745f89993b7c5/cypress/component/utils/utils.spec.js#L367

and 

https://github.com/DataBiosphere/duos-ui/blob/8f40c92f1f38b9cf1c783198cc4745f89993b7c5/cypress/component/utils/utils.spec.js#L463

Those lines should be changed to expect the new values, i.e. `submitted`

src/libs/utils.js
```diff
@@ -256,7 +256,7 @@ export const outputCommaSeperatedElectionStatuses = (elections) => {
     elections.map((e) => processElectionStatus(e, e.votes, false))
   ).filter((status) => !isEmpty(status));
   if (isEmpty(statuses)) {
-    return 'Unreviewed';
+    return 'Submitted';
```
pshapiro4broad, line 259: I'm sure you tested this out locally, but I don't see how this text is being passed back to the display code. The only usage of this method that I see is in `CancelCollectionButton`, but that component appears to be unused.

To show that this works to fix the UI, in addition to a unit test, it would be good to have a UI test that verifies at the text of a rendered component.

rushtong: Right .. I'm pretty sure this PR isn't doing what the ticket describes. Running this locally, I still see `Unreviewed` in the researcher DAR table

rushtong:
Closing this PR in favor of https://github.com/DataBiosphere/consent/pull/2514 which addresses it on the back end.

PR #2852, opened 2025-05-06
[DT-105]Add remove datasets component for progress reports

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-105
We need a way to optionally remove datasets from a parent DAR for a progress report.

### Summary
Adds the remove datasets functionality and header to the form. reuses the SelectableDatasets component which already has the functionality to remove datasets. 
Adds state for the datasets to the progress report application
Adds parentDar as a prop, we will need this to populate information of the new progress report being created and we need its referenceId to submit the progress report
Adds small fixes to the use of the submitProgressReport component 
Makes a function to modify the form data in state and passes it into the SelectableDatasets component. 

The component looks like this:

https://github.com/user-attachments/assets/6c696f06-70ea-4277-b999-d0a28f00ed33

### Testing
The SelectableDatasets component is already tested thoroughly so this does not add new tests.
TODO: I could add tests for the logic I included on the progress report application page.

----

- Verify all tests go green


lstrano-broad:
LGTM, in the future we may want to revisit this. I would imagine some users may have a very long list of datasets so the UI may get very long.

rushtong:
> The mocks for this work show the FormField component used on the DAR application page which allows adding and removing datasets. Instead of making a new FormField for this purpose, I chose to reuse the existing Selectable Datasets component. So this work does not match the mocks. I'm wondering what people think about this decision.

This is the right decision, IMO - datasets should only be removable from PRs, never added.

src/components/RemoveDatasets.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface RemoveDatasetsProps {
+  datasets: Dataset[];
+  readOnlyMode: boolean;
+}
+
+interface FormStateInterface {
+    datasetIds: number[];
+}
+
+export default function RemoveDatasets(_props: RemoveDatasetsProps): React.JSX.Element {
```
rushtong: I'm confused on the naming here - this component is showing you `SelectableDatasets`. I'm also confused on the need to wrap this in a parent component.

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -571,7 +576,7 @@ const DataAccessRequestApplication = (props) => {
                   <ConditionalAccordion
                       condition={false}
                       title={`DAR Report ${reverseOrderedDARs.length}`}>
-                    <ProgressReportApplication readOnlyMode={false}/>
+                    <ProgressReportApplication readOnlyMode={false} datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)} parentDar={reverseOrderedDARs[0]}/>
```
rjohanek, line 579: for the application page, includes the datasets of the parentDar, where the parent is the most recent in the collection. this will need to change to include only the approved datasets included in the parentDar. and it will need to change when we can no longer assume the latest DAR is the parent.

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -587,7 +592,7 @@ const DataAccessRequestApplication = (props) => {
                                 condition={true}
                                 title={`DAR Report ${reverseOrderedDARs.length - index - 1}`}
                                 defaultExpanded={index === 0}>
-                              <ProgressReportApplication readOnlyMode={true} dar={dar?.data}  />
+                              <ProgressReportApplication readOnlyMode={true} datasets={filterForProgressReport(datasets, dar.datasetIds)} dar={dar?.data}/>
```
rjohanek, line 595: for the review page, includes just the datasets included in the submitted dar

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -19,10 +29,23 @@ export const ProgressReportApplication = ({dar, readOnlyMode=true}: ProgressRepo
...
+                        <SelectableDatasets
+                            disabled={readOnlyMode}
+                            datasets={datasets}
+                            setSelectedDatasets={(selectedDatasets) => {
+                                setFormState({...formState, datasetIds: selectedDatasets.map((ds) => ds.datasetId)})}
+                            }
+                        />
+                    </div>
+                </div>
+                {!readOnlyMode && parentDar && <div>
```
rjohanek, line 52: changed from `dar?.parentId &&` to `!readOnlyMode && parentDar &&` because the dar is only populated for the review page. when it is the review page we do not want to show this button. Instead when it is not read only and the parentDar is populated (which it is when the page is not read only) then we want to show this button

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -19,10 +29,23 @@ export const ProgressReportApplication = ({dar, readOnlyMode=true}: ProgressRepo
...
                   <SubmitProgressReport
                       progressReport={dar}
-                      parentId={dar.parentId}
+                      parentId={parentDar.referenceId}
```
rjohanek: the endpoint actually takes the parent reference id not the parent id (which is a database id). this is confusing because the DAR saves the database id not the reference id!

src/pages/dar_application/SelectableDatasets.jsx
```diff
@@ -11,7 +11,7 @@ export default function SelectableDatasets(props) {
     // Populate parent state with the current state of datasets to be saved to the DAR
     const newSelectedDatasets = datasets.filter(ds => !removedIds.includes(ds.datasetId));
     setSelectedDatasets(newSelectedDatasets);
-  }, [removedIds, datasets, setSelectedDatasets]);
```
rjohanek, line 14: this was causing an infinite loop

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: If I understood previous conversations correctly, there was going to be an effort to convert `dar.parentId` to the actual parent DAR's reference id. The original code was working on that assumption, but I don't know where that effort landed.

src/pages/dar_application/SelectableDatasets.jsx
```diff
@@ -11,7 +11,8 @@ export default function SelectableDatasets(props) {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rushtong: We should be able to work through this lint warning - will need to dig through older PRs to see what approach we've taken.

src/pages/dar_application/ProgressReportApplication.tsx
rjohanek: yes, we may eventually change the parentId to be a reference Id, but that is not the case currently and won't be the case when this page goes out. And even if we did make the change to referenceId before the page goes out, the dar.parentId won't be the value that we want because dar is only populated when the page is read only, so when we want the submit button to actually show up, that value won't be populated. Additionally, we will need to pass the parentDar in anyway to pre-populate other values, so I wasn't adding it as a prop just for this reason, but its convenient that we can just grab its referenceId.

rushtong: > Additionally, we will need to pass the parentDar in anyway to pre-populate other values, so I wasn't adding it as a prop just for this reason, but its convenient that we can just grab its referenceId.

Yes, this makes more sense than my original approach, thanks!

s-rubenstein (approved):

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -1,12 +1,22 @@
...
     readOnlyMode?: boolean
 };
-export const ProgressReportApplication = ({dar, readOnlyMode=true}: ProgressReportApplicationProps) => {
+
+interface FormStateInterface {
+    datasetIds: number[];
+}
+
+export const ProgressReportApplication = ({dar, parentDar, datasets, readOnlyMode=true}: ProgressReportApplicationProps) => {
+    const [formState, setFormState] = useState<FormStateInterface>({datasetIds: []});
```
fboulnois: This is where I would expect all of the Progress Report form state to be managed in the future, which works well with my mental model and design in my other PRs.

I would expect the initial `datasets` / `datasetIds` to also be loaded into the formState, however, for the `readOnly` view.

fboulnois (approved):
Looks good, one suggestion:

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -19,10 +29,23 @@ export const ProgressReportApplication = ({dar, readOnlyMode=true}: ProgressRepo
...
-                {dar?.parentId && <div>
+                <div data-cy='remove-datasets'>
+                    <div className='progress-report-step-card'>
+                        <h2>Step 3: Dataset(s) in this DAR</h2>
+                        <p style={{ marginBottom: '1rem' }}>Currently selected datasets:</p>
+                        <SelectableDatasets
+                            disabled={readOnlyMode}
+                            datasets={datasets}
+                            setSelectedDatasets={(selectedDatasets) => {
+                                setFormState({...formState, datasetIds: selectedDatasets.map((ds) => ds.datasetId)})}
```
fboulnois, line 40: I think this should load the initial form state as the datasets, and setFormState should modify the list of datasets.

src/pages/dar_application/ProgressReportApplication.tsx
rjohanek: I want to talk about this more. I agree I should load the initial datasetIds into the formState. But, I am hesitant to add datasets to the formState because it is not part of what will be submitted.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -19,10 +33,23 @@ export const ProgressReportApplication = ({dar, readOnlyMode=true}: ProgressRepo
...
```
snf2ye:
```suggestion
                                setFormState((prevState) => ({...prevState, datasetIds: selectedDatasets.map((ds) => ds.datasetId) }));
```

It looks like we need to use an "updater" function to use previous state to update the new state - https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

I _think_ this was the actual source of our infinite looping in the useEffect.

otchet-broad (approved):
Excellent.  Thanks for incorporating my suggestions.

rushtong (approved):
Thanks for the updates 👍🏽

src/pages/dar_application/ProgressReportApplication.tsx
rjohanek: ohhh that makes sense! thank you! Do you think I should make this change in place of what we did yesterday?

PR #2856, opened 2025-05-08
[DT-796] Sort Studies by Column Values

PintoGideon:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-796

### Summary

This PR implements the initial table–sorting functionality in the DUOS Data Library. There are a few edge cases I’d appreciate feedback on. Currently, all empty or undefined values are moved to the bottom of the table in both ascending and descending order.

https://github.com/user-attachments/assets/a1fd3806-f4f1-44f0-87fe-6357e104a4c6

----

- Verify all tests go green

### DIFF
--- a/src/components/data_search/DatasetSearchTableDisplay.tsx
+++ b/src/components/data_search/DatasetSearchTableDisplay.tsx
@@ -1,17 +1,15 @@
 import * as React from 'react';
-import { isEmpty, capitalize, isUndefined } from 'lodash';
-import { DatasetTerm } from 'src/types/model';
+import {isEmpty, capitalize, isUndefined} from 'lodash';
+import {DatasetTerm} from 'src/types/model';
 import SimpleTable from '../SimpleTable';
-import { Styles } from '../../libs/theme';
+import {Styles} from '../../libs/theme';
 import {
-  type CellData,
+ type CellData,
   DatasetSearchTableTab,
 } from './DatasetSearchTableConstants';
-import { SnapshotSummaryModel } from '../../types/tdrModel';
+import {SnapshotSummaryModel} from '../../types/tdrModel';
 import { Storage } from '../../libs/storage';
 
-// Storage key for persisting sort preferences
-const storageDatasetSearchSort = 'storageDatasetSearchSort';
 
 const styles = {
   baseStyle: {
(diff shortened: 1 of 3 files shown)


src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -48,10 +49,72 @@ interface DatasetSearchTableDisplayProps{
...
+
+        // Handle NaN values similar to undefined
+        if (isNaN(numA) && isNaN(numB)) return 0;
+        if (isNaN(numA)) return 1; // Treat NaN like undefined, to the bottom
+        if (isNaN(numB)) return -1; // Treat NaN like undefined, to the bottom
+
+        return (numA - numB) * sortState.dir;
+      }
+      const strA = String(aValue).toLowerCase();
+      const strB = String(bValue).toLowerCase();
```
rushtong, line 123: It looks like `aValue` and `bValue` are actually `string` types, so this works, but they could possibly be objects in the future. For example, date values might be added to the table in a future tasks.

src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -69,7 +132,10 @@ export const DatasetSearchTableDisplay = (props: DatasetSearchTableDisplayProps)
           selected={selected}
           styles={styles}
           tableSize={10}
-          summary={`faceted ${tab.singular} search table`} />
+          summary={`faceted ${tab.singular} search table`}
+          onSort={handleSort}
```
rushtong, line 160: Here's an example of where we try to remember the sort selection as part of the user's settings:
https://github.com/DataBiosphere/duos-ui/blob/045a7a36308a6093c78a7dbbf125327ef0224dec/src/components/dar_collection_table/DarCollectionTable.jsx#L349
It would be nice if we could support that, but not a strict requirement for this PR.

rushtong: And here, we try to retrieve it from user settings:
https://github.com/DataBiosphere/duos-ui/blob/045a7a36308a6093c78a7dbbf125327ef0224dec/src/components/dar_collection_table/DarCollectionTable.jsx#L190

rushtong (approved):
This looks good to me 👍🏽 Address comments as you see fit.

src/components/data_search/DatasetSearchTableDisplay.tsx
PintoGideon: Yeah I added String to be a bit defensive. 
String() function is very forgiving and won't throw errors for weird values. It can handle just about anything:

```js
String(undefined) → "undefined"
String(null) → "null"
String({}) → "[object Object]"
String([1,2,3]) → "1,2,3"
String(NaN) → "NaN"
String(function(){}) → function's source code as string
```

Even for React elements, String() will try to convert them by calling their toString() method, which might not be meaningful but won't throw an error.

src/components/data_search/DatasetSearchTableDisplay.tsx
PintoGideon: Oh I should definitely add this for good ux.

src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -1,13 +1,17 @@
...
+import { DatasetTerm } from 'src/types/model';
 import SimpleTable from '../SimpleTable';
-import {Styles} from '../../libs/theme';
+import { Styles } from '../../libs/theme';
 import {
   type CellData,
   DatasetSearchTableTab,
 } from './DatasetSearchTableConstants';
-import {SnapshotSummaryModel} from '../../types/tdrModel';
+import { SnapshotSummaryModel } from '../../types/tdrModel';
```
rushtong: nit: I think linting in your IDE is doing this and might need a configuration update. Locally, for me, when I reformat the files, all of these white space changes are reverted.

src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -39,6 +42,36 @@ const styles = {
   containerOverride: {}
 };
 
+interface Sort {
+    colIndex: number;
+    dir: number;
+  }
+
+  // Storage key for persisting sort preferences
+const storageDatasetSearchSort = 'storageDatasetSearchSort';
```
rushtong, line 51: nit: still seeing some formatting issues, this should be indented.

rushtong (approved):
Thanks for the changes, this will be nice when merged 👍🏽

snf2ye (approved):

PR #2859, opened 2025-05-12
DT-1650: Fix import shortening warning

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1650

### Summary
Add the correct path to the vite config so that imports can be shortened. Also added some pages for an example of how imports can be referenced from `src`

Note that I wanted to include a test, chose the ToS tests, but they needed updating as well. So, this PR includes some minor test changes to address lint warnings, as well as a ToS key fix, and import fixes for that file as well.

I tested this in two ways
1. Running locally via npm start
2. Running under docker with:
```
docker build . -t duos
docker compose up -d
```

----

- Verify all tests go green


pshapiro4broad (approved):
Looks good, I checked the lint warnings in IntelliJ and verified building and running locally. It will be nice to fix these as we update the code.

snf2ye (approved):

vite.config.ts
```diff
@@ -22,4 +22,9 @@ export default defineConfig({
     },
     open: true
   },
+  resolve: {
+    alias: {
+      src: '/src'
```
fboulnois, line 27:
```suggestion
      src: './src'
```

This alias looks strange to me, can you describe this in more detail?

fboulnois (changes requested):
See below:

vite.config.ts
rushtong: I spent some time looking over the docs to try and understand the normal vite pattern, and `./src` is the common usage. However, it doesn't work for us - vite can't seem to resolve that. It does work with `/src`. If there's a way to make `./src` work, I'm open to whatever that is.

fboulnois (approved):
Based on conversation today 👍

PR #2861, opened 2025-05-12
DT-1622: Allow for user to edit IRB expiration

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1622

### Summary
Allow for users to set the expiration for IRB document on upload using the new date picker widget.

#### Screens
New:
Screenshot 2025-05-12 at 10 59 32 AM

---

Old:
Screenshot 2025-05-12 at 9 18 55 AM

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/DataAccessRequest.jsx
+++ b/src/pages/dar_application/DataAccessRequest.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import { DataSet } from '../../libs/ajax/DataSet';
-import { DAR } from '../../libs/ajax/DAR';
+import {DataSet} from '../../libs/ajax/DataSet';
+import {DAR} from '../../libs/ajax/DAR';
 import {FormField, FormFieldTitle, FormFieldTypes, FormValidators} from '../../components/forms/forms';
 import {
   needsDsAcknowledgement,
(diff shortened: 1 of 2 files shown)


snf2ye:
Can the small text "select a date" be edited to something like "select an expiration date"? I want to make sure it is clear that this is the expiration date and not like today's date (which is common when signing forms).

rushtong:
> Can the small text "select a date" be edited to something like "select an expiration date"? I want to make sure it is clear that this is the expiration date and not like today's date (which is common when signing forms).

Really good question - I'll look into it, not sure how configurable it is, but I can add some surrounding text.

rushtong:
> > Can the small text "select a date" be edited to something like "select an expiration date"? I want to make sure it is clear that this is the expiration date and not like today's date (which is common when signing forms).
> 
> Really good question - I'll look into it, not sure how configurable it is, but I can add some surrounding text.

@snf2ye - Just added some explanatory text - the original made it clear in the default text that isn't possible in the new date widget.

snf2ye (approved):
Looks good 👍

rjohanek (approved):
this looks good, except, do we really want to provide a default value at all? this is a required field, we don't usually default populate the other required fields. I'm afraid if this has a default value, users will not change it. If we don't provide a default value, they must change it and will likely be more accurate.

rushtong:
> this looks good, except, do we really want to provide a default value at all?

@rjohanek - A default value is the existing behavior. I'm hesitant to change current functionality when trying to implement a bug fix.

PR #2864, opened 2025-05-13
[DT-1641] Implement the DUOS Data Library Card

PintoGideon:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1641

### Summary

- Replace the current "DUOS Overview" section on the homepage with a new "Data Libraries in DUOS" section featuring a three-column image layout that links to curated Data Library pages.
- Horizontal layout for screens larger than 904 px. Switches to vertical layout on smaller screens.
- MUI tooltips that display dataset names when the user hovers or focuses on the logos.

**Screenshots:**

Mobile (iPhone SE)

local dsde-dev broadinstitute org_3000_home(iPhone SE)

Laptop (XPS 13)

Screenshot From 2025-05-12 23-31-27

----

- Verify all tests go green

### DIFF
--- a/src/pages/Home.jsx
+++ b/src/pages/Home.jsx
@@ -85,18 +85,31 @@ const Home = (props) => {
     position: 'relative',     
   };
 
-  const disabledCard = {
-    ...baseCard,
-    opacity: 0.8,
-    cursor: 'not-allowed',
-  };
-
-
   const logoImg = {
     width: '100%',
     height: '100%',
     objectFit: 'contain',
     display: 'block',
+  };
+
+  const handleSignIn = (redirectPath) => {
+    // Set the redirectTo parameter without forcing a page reload
+    const currentUrl = new URL(window.location.href);
+    currentUrl.searchParams.set('redirectTo', redirectPath);
+    window.history.replaceState({}, '', currentUrl);
+    
+    // Find the existing sign-in button in the header and programmatically click it
+    // This will trigger the existing authentication flow with all proper session handling
+    const signInButtons = document.querySelectorAll('button');
+    const signInButton = Array.from(signInButtons).find(button => 
+      button.textContent && button.textContent.trim() === 'Sign In'
+    );
+    if (signInButton) {
+      signInButton.click();
+    } else {
+      // Fallback - scroll to top where the sign-in button is located
+      window.scrollTo({ top: 0, behavior: 'smooth' });
+    }
   };
 
   return (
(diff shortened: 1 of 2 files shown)


src/pages/Home.jsx
```diff
@@ -125,37 +143,42 @@ const Home = (props) => {
...
-                    {' '}
-                    <a href="https://www.ga4gh.org/genomic-data-toolkit/regulatory-ethics-toolkit/#:~:text=Machine%20Readable%20Consent%20Guidance&text=Machine%20readable%20consent%20language%20is,to%20for%20their%20research%20purposes" target="_blank" rel="noreferrer">Machine Readable Consent Guidance.</a>
-                    {' '}
-                    For help determining your data&apos;s permitted uses, try our
-                    {' '}
-                    <Link to="/consent_text_generator">Data Sharing Language Tool</Link>, which follows GA4GH guidelines.
```
rushtong, line 143: As a consequence of this, `ConsentTextGenerator.jsx` is no longer in use. Let's follow up with @jlawsonbroad about removing that component to keep the maintenance burden down.

fboulnois (commented):
Looks good, could we add some tests? Perhaps checking that the data library links now exist on the page, and maybe that the tooltips work?

src/pages/Home.jsx
```diff
@@ -125,37 +143,42 @@ const Home = (props) => {
...
-                    <p style={paragraph}>To evaluate the feasibility of using machine readable data use terms to interpret data use restrictions and access requests, we are piloting a trial of DUOS overseen by Partners’ Healthcare IRB. During the pilot, DACs comprised of governmental and non-governmental data custodians will pilot the use of DUOS, its ability to structure use limitations and access requests, and the accuracy of the DUOS algorithm. This aids us in improving the DUOS algorithm and providing feedback on the GA4GH Data Use Ontology based on experts’ feedback.</p>
+
+        <section style={{ margin: '5rem auto', padding: '0 2rem' }}>
+          <div style={{ maxWidth: '1200px', margin: '0 auto', textAlign: 'center' }}>
+            <p style={header}>Data Libraries in DUOS</p>
+            <p style={description}>Click the images below to view curated Data Libraries, and search and request access to data.</p>
+
+            <div style={logoGrid} className="logo-grid">
+              <Tooltip title="AnVIL" placement="bottom" arrow>
+                <a href="https://duos.org/datalibrary/anvil" target="_blank" rel="noreferrer" style={{ textDecoration: 'none' }}>
```
rushtong: Couple thoughts that apply to all of the `<a/>` elements on this page
1. I think this could be a `Link` component. 
2. Why is this opening in a new tab? Since we don't maintain login state across open tabs, the user would need to log in again.

fboulnois (approved):
See my previous comment, otherwise looks good.

src/pages/Home.jsx
```diff
@@ -125,37 +143,42 @@ const Home = (props) => {
...
+
+              <Tooltip title="Broad Institute" placement="bottom" arrow>
+                <a href="https://duos.org/datalibrary/broad" target="_blank" rel="noreferrer" style={{ textDecoration: 'none' }}>
+                  <div className="logo-card" style={{ ...baseCard, background: '#1F3B50', padding: '15px' }}>
+                    <img src={broadLogo} alt="Broad Institute" style={logoImg} />
+                  </div>
+                </a>
+              </Tooltip>
+       
+              <Tooltip title="Human Cell Atlas" placement="bottom" arrow>
```
rushtong: The tooltip text here is very difficult to read. Is there anything we can do to increase the font styling?
Screenshot 2025-05-13 at 8 44 55 AM

rushtong (approved):
Agreed that this component could use some tests. See comments inline, but nothing looks like a blocker, address as you see fit 👍🏽

src/pages/Home.jsx
```diff
@@ -125,37 +144,42 @@ const Home = (props) => {
...
-                    <p style={paragraph}>To evaluate the feasibility of using machine readable data use terms to interpret data use restrictions and access requests, we are piloting a trial of DUOS overseen by Partners’ Healthcare IRB. During the pilot, DACs comprised of governmental and non-governmental data custodians will pilot the use of DUOS, its ability to structure use limitations and access requests, and the accuracy of the DUOS algorithm. This aids us in improving the DUOS algorithm and providing feedback on the GA4GH Data Use Ontology based on experts’ feedback.</p>
+
+        <section style={{ margin: '5rem auto', padding: '0 2rem' }}>
+          <div style={{ maxWidth: '1200px', margin: '0 auto', textAlign: 'center' }}>
+            <p style={header}>Data Libraries in DUOS</p>
+            <p style={description}>Click the images below to view curated Data Libraries, and search and request access to data.</p>
+
+            <div style={logoGrid} className="logo-grid">
+              <OverflowTooltip id="anvil" tooltipText="AnVIL">
+                <Link to="/datalibrary/anvil" style={{ textDecoration: 'none' }}>
```
rushtong: The link behavior is confusing if the user is not signed in. For example, if you are not signed in, you are redirected back to this page, but not prompted in any way so it looks like nothing happened. After signing in, the user is directed to the correct link. I think we need some kind of prompt on these images if the user is not signed in.

PintoGideon: Yeah I agree. I'll fix this.

PintoGideon: @rushtong @solideoglori I ’ve added a visual indicator on the cards to show that these data libraries require login to access. However, after logging in, the UI always redirects users to the admin_dar_collections_page, preventing them from accessing the dataset they may want to select. We may need to rethink how this feature should work, or I might be missing something.

I added some visual indicators for unauthenticated users and disabled navigation. 

Screenshot From 2025-05-20 11-52-23

rushtong (approved):
Love the auto-sign-in solution 👍🏽

PR #2870, opened 2025-05-15
[DT-1629] Progress Report summary section

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1629

### Summary

- Adapts the collaborator component for publications to move quickly (I checked with Victoria and Jonathan beforehand and I added https://broadworkbench.atlassian.net/browse/DT-1671 as a follow-up)
- Add the Progress Report summary section to the form
- Renumbers the section headers

Screenshot 2025-05-15 at 15-42-17 Broad Data Use Oversight System

----

- Verify all tests go green

### DIFF
--- a/cypress/component/ProgressReports/collaborator_changes.spec.tsx
+++ b/cypress/component/ProgressReports/collaborator_changes.spec.tsx
@@ -44,10 +44,10 @@ describe('Collaborator Changes - Component Tests', () => {
 
   it('renders the component correctly', () => {
     cy.get('[data-cy=dar-closeout]').should('exist');
-    cy.contains('Step 2: Add or Remove Collaborators').should('be.visible');
-    cy.contains('2.1 Internal Lab Staff').should('be.visible');
-    cy.contains('2.2 Internal Collaborators').should('be.visible');
-    cy.contains('2.3 External Collaborators').should('be.visible');
+    cy.contains('Step 3: Add or Remove Collaborators').should('be.visible');
+    cy.contains('3.1 Internal Lab Staff').should('be.visible');
+    cy.contains('3.2 Internal Collaborators').should('be.visible');
+    cy.contains('3.3 External Collaborators').should('be.visible');
   });
 
   it('renders all three collaborator sections', () => {
--- a/src/components/publications_list/Publication.tsx
+++ b/src/components/publications_list/Publication.tsx
@@ -1,0 +1,10 @@
+export interface Publication {
+    title: string;
+    date: string;
+    bibliographic_citation: string;
+    did_cite: boolean;
+    pubmed_id: string;
+    authors: string;
+    dataset_citation: string;
+    link: string;
+}
--- a/src/components/publications_list/PublicationRow.tsx
+++ b/src/components/publications_list/PublicationRow.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import { Publication } from './Publication';
+import PublicationAddEdit from './PublicationAddEdit';
+import PublicationSummary from './PublicationSummary';
+
+interface PublicationRowProps {
+    readonly id: number;
+    readonly editMode: boolean;
+    publication: Publication;
+    readonly publicationText: string;
+    readonly publications: Publication[];
+    readonly columnsToShow: string[];
+    readonly editAction: () => void;
+    readonly deleteAction: () => void;
+    readonly closeAction: () => void;
+    readonly onPublicationChange: (publications: Publication[]) => void;
+    readonly disabled: boolean;
+}
+
+export default function PublicationRow(props: PublicationRowProps): React.JSX.Element {
+    const { id, editMode, publication, publicationText, publications, columnsToShow, editAction, deleteAction, closeAction, onPublicationChange, disabled } = props;
+
+    return (
+        <div>
+            {editMode && (
+                <PublicationAddEdit
+                    id={id}
+                    publication={publication}
+                    publicationText={publicationText}
+                    publications={publications}
+                    closeAction={closeAction}
+                    onPublicationChange={onPublicationChange}
+                />)}
+            {!editMode && (
+                <PublicationSummary
+                    publication={publication}
+                    columnsToShow={columnsToShow}
+                    editAction={editAction}
+                    deleteAction={deleteAction}
+                    disabled={disabled}
+                />)}
+        </div>
+    );
+}
--- a/src/components/publications_list/PublicationSummary.tsx
+++ b/src/components/publications_list/PublicationSummary.tsx
@@ -1,0 +1,75 @@
+import React, { useState } from "react";
+import PublicationDelete from "./PublicationDelete";
+import { Publication } from "./Publication";
+
+interface PublicationSummaryProps {
+    publication: Publication;
+    readonly columnsToShow: string[];
+    readonly editAction: () => void;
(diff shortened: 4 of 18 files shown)


src/components/publications_list/Publication.tsx
```diff
@@ -0,0 +1,10 @@
+export interface Publication {
```
fboulnois, line 1: Note: the entire `publications_list` component is essentially a duplicate of the collaborators component with different fields. In the future it might make sense to combine these components. As a result I haven't added tests for this component (practically, they would be nearly identical to those in the `collaborator_list` component).

src/components/publications_list/PublicationRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+                    closeAction={closeAction}
+                    onPublicationChange={onPublicationChange}
+                />)}
+            {!editMode && (
+                <PublicationSummary
+                    publication={publication}
+                    columnsToShow={columnsToShow}
+                    editAction={editAction}
+                    deleteAction={deleteAction}
+                    disabled={disabled}
```
rushtong, line 40: Could have have a case where `editMode` is false and `disabled` is true? IOW, could we default to `editMode`?

rushtong (approved):
Thank you, looks good 👍🏽

cypress/component/ProgressReports/collaborator_changes.spec.tsx
```diff
@@ -44,10 +44,10 @@ describe('Collaborator Changes - Component Tests', () => {
 
   it('renders the component correctly', () => {
     cy.get('[data-cy=dar-closeout]').should('exist');
-    cy.contains('Step 2: Add or Remove Collaborators').should('be.visible');
-    cy.contains('2.1 Internal Lab Staff').should('be.visible');
-    cy.contains('2.2 Internal Collaborators').should('be.visible');
-    cy.contains('2.3 External Collaborators').should('be.visible');
+    cy.contains('Step 3: Add or Remove Collaborators').should('be.visible');
```
pshapiro4broad, line 47: To avoid making continual updates to these tests, they could use regular expressions. Assuming that the exact outline level is not something we need to test.
```suggestion
    cy.contains(/^Step \d+: Add or Remove Collaborators$/).should('be.visible');
```

src/pages/progress_reports/SummarySection.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+                <div className='progress-report-row'>
+                    <FormField
+                        id='progressReportSummary'
+                        type={FormFieldTypes.TEXTAREA}
+                        title='1.1 Summary of Progress'
+                        description='Please summarize your research on this project since your initial request or most recent renewal in the space below. Please describe whether and how the dataset(s) was used, including referencing the dataset(s) by name in your summary.'
+                        placeholder='Please provide an update here.'
+                        rows={6}
+                        maxLength={2200}
```
pshapiro4broad: It would be nice to avoid a magic number here, in case this value changes in the future. Also I see `2200` in many other places in the UI code, so this appears to be an existing magic number.

src/pages/progress_reports/SummarySection.tsx
```diff
@@ -0,0 +1,120 @@
...
+                        type={FormFieldTypes.YESNORADIOGROUP}
+                        title='1.2 Intellectual Property'
+                        description={<span>Have you generated any <strong>intellectual property</strong> since your last renewal as a result of using the data?</span>}
+                        orientation='horizontal'
+                        onChange={({ key, value }: FormFieldChange) => {
+                            onFormChange({ [key]: value });
+                        }}
+                        disabled={readOnly}
+                    />
+                    {formState.intellectualPropertyYesNo === true && <FormField
```
pshapiro4broad, line 65: Why does this code use `=== true`? Is is possible that a non-boolean value will be stored in `intellectualPropertyYesNo`?

src/components/publications_list/PublicationSummary.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    data-tip='Delete dataset'
+                    data-for='tip_delete'
+                ></span>
+                <span style={{ marginLeft: '1rem' }}></span>
+            </a>
+            {/* delete modal */}
+            <PublicationDelete
+                publicationName={publication.title}
+                showDelete={showDeleteModal}
+                confirmAction={() => { deleteAction(); setShowDeleteModal(false); }}
```
pshapiro4broad, line 70: I'm surprised that eslint doesn't flag this, I would expect it to be formatted across lines
```suggestion
                confirmAction={() => {
                    deleteAction();
                    setShowDeleteModal(false);
                }}
```

src/components/publications_list/PublicationSummary.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    <div key={'publication_summary_column_' + index} style={{ flex: '1 1 100%', marginRight: '1.5rem' }}>
+                        <span>
+                            {columnContent}
+                        </span>
+                    </div>
+                );
+            })}
+            {/* edit button */}
+            <div className='collaborator-summary-edit-delete-buttons'>
+                <a
```
pshapiro4broad, line 40: Sonar flags this use of an anchor tag with a number of issues. For example, it says that without a keyboard listener users must use a mouse to interact with this element

pshapiro4broad (approved):
This is a large change, but looks OK as far as I can tell.

I worry that our exact matching on outline numbers is making our tests too fragile, requiring many updates when the text is changed. It would be even better if we had a CSS or some other way to automatically number outlines, since DUOS has so many of them.

src/components/publications_list/PublicationSummary.tsx
pshapiro4broad: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2870&id=DataBiosphere_duos-ui&open=AZbZWjdtzsbsOc2-Lz_i

src/pages/progress_reports/SummarySection.tsx
fboulnois: It's mainly for clarity, since our "Yes / No" radio buttons actually map to booleans, not "Yes" / "No" strings.

src/pages/progress_reports/SummarySection.tsx
fboulnois: Is the suggestion to pull out this value into a common constant?

src/components/publications_list/PublicationSummary.tsx
fboulnois: This seems ok to me, the line is short.

src/components/publications_list/PublicationSummary.tsx
fboulnois: This is an existing issue that we have with the old and new collaborator component (and now publication list component), but we can pull this into the larger refactoring ticket.

cypress/component/ProgressReports/collaborator_changes.spec.tsx
fboulnois: I agree that the tests are somewhat fragile due to the numbering scheme, but ultimately this form will change (and the numbers will probably be removed) which will require the tests to be updated anyways, so this seems like a reasonable stopgap until then.

PR #2874, opened 2025-05-19
DT-1678: Remove institution from Library Card creation

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1678

### Summary
Institutions are no longer part of Library Cards. This PR removes them from the Admin UI. They were already disabled in the Signing Official UI, so this cleans that up so the code doesn't have stray references to institutions where they're not necessary. The SO page was also doing some unnecessary institution id checks to determine if a user had a Library Card or not. This PR fixes that as well.

#### Notes

* When reviewing, I suggest hiding white space to remove the pure format changes.
* We don't have test coverage for these components. Will file a ticket so we can add testing and potentially migrate these to TypeScript, see https://broadworkbench.atlassian.net/browse/DT-1682
* Converting to TS will also address the sonarqube issues related to prop validation.
* There is no longer any action to take with **Update**, see https://broadworkbench.atlassian.net/browse/DT-1683

----

- Verify all tests go green

### DIFF
--- a/src/components/modals/LibraryCardFormModal.jsx
+++ b/src/components/modals/LibraryCardFormModal.jsx
@@ -3,18 +3,16 @@ import {cloneDeep, includes, isEmpty, isNil, isObject} from 'lodash/fp';
 import {Styles, Theme} from '../../libs/theme';
 import CloseIconComponent from '../CloseIconComponent';
 import Modal from 'react-modal';
-import {SearchSelect} from '../SearchSelect';
 import Creatable from 'react-select/creatable';
 import SimpleButton from '../SimpleButton';
 import {LibraryCardAgreementTermsDownload} from '../LibraryCardAgreementTermsDownload';
 
 const FormFieldRow = (props) => {
-  const { card, dropdownOptions, updateInstitution, updateUser, modalType, setCard } = props;
+  const { card, dropdownOptions, updateUser, modalType, setCard } = props;
 
   const cardlessOptions = dropdownOptions.filter((option) => {
     const libraryCards = option.libraryCards || [];
-    const savedCard = libraryCards.find(({institutionId}) => institutionId === card.institutionId);
-    return isNil(savedCard);
+    return isEmpty(libraryCards);
   });
   const [filteredDropdown, setFilteredDropdown] = useState(cardlessOptions);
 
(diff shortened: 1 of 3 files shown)


src/components/modals/LibraryCardFormModal.jsx
```diff
@@ -3,18 +3,16 @@ import {cloneDeep, includes, isEmpty, isNil, isObject} from 'lodash/fp';
...
 
 const FormFieldRow = (props) => {
-  const { card, dropdownOptions, updateInstitution, updateUser, modalType, setCard } = props;
+  const { card, dropdownOptions, updateUser, modalType, setCard } = props;
 
   const cardlessOptions = dropdownOptions.filter((option) => {
     const libraryCards = option.libraryCards || [];
-    const savedCard = libraryCards.find(({institutionId}) => institutionId === card.institutionId);
-    return isNil(savedCard);
+    return isEmpty(libraryCards);
```
pshapiro4broad: I think this can be simplified. Since `_.isEmpty(undefined)` is `true`, writing `|| []` is redundant
```suggestion
  const cardlessOptions = dropdownOptions.filter((option) => isEmpty(option.libraryCards));
```

src/pages/signing_official_console/SigningOfficialTable.jsx
```diff
@@ -120,11 +119,10 @@ const researcherFilterFunction = getSearchFilterFunctions().signingOfficialResea
 const LibraryCardCell = ({
   researcher,
   showConfirmationModal,
-  institutionId
 }) => {
   const id = researcher.userId || researcher.email;
   const card = !isEmpty(researcher.libraryCards)
-    ? find((card) => card.institutionId === institutionId)(researcher.libraryCards)
+    ? researcher.libraryCards[0]
     : null;
```
pshapiro4broad: Is this the same as lodash `head()`?
```
   const card = _.head(researcher.libraryCards);
```

pshapiro4broad (approved):
Looks OK, might be nice to have unit tests.

rushtong:
> Looks OK, might be nice to have unit tests.

I do want to tackle unit tests, but they will be non-trivial to write especially if we want to convert these to TypeScript. I have a ticket for that effort linked in the PR description. My hope is to get this fix in place first so Admins and SOs are able to work effectively with LCs in the short term.

s-rubenstein (approved):
Pending Phil's comments, this looks good.

PR #2877, opened 2025-05-19
[DT-1625] Add Data Use Acknowledgement component to Progress Report Form

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1625

### Summary
Refactors data use acknowledgements into their own component and reuses it on the DataAccessRequest page and the ProgressReportApplication page
Adds validation on submit to ensure the data use acknowledgments are checked
Makes selectedDatasets part of state because it is used in multiple places (instead of filtering the datasets each time)
Maintains current functionality on DAR application page

It said the screen recording is too big so here is a link: https://drive.google.com/file/d/1aeDiniYv_incMiL2iRunnRotgg17eT1H/view?usp=sharing 

----

- Verify all tests go green

### DIFF
--- a/cypress/component/DataAccessRequest/data_use_acknowledgements.spec.tsx
+++ b/cypress/component/DataAccessRequest/data_use_acknowledgements.spec.tsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import { mount } from 'cypress/react';
+import { DataUseAcknowledgements } from 'src/pages/dar_application/DataUseAcknowlegements';
+
+describe('DataUseAcknowledgements Component', () => {
+    let defaultProps: any;
+
+    beforeEach(() => {
+        defaultProps = {
+            title: 'Data Use Acknowledgements',
+            datasets: [],
+            dataUseTranslations: [],
+            formData: {},
+            readOnlyMode: false,
+            includeInstructions: true,
+            onChange: cy.stub(),
+            onValidationChange: cy.stub(),
+            validation: {}
+        };
+    });
+
+    it('renders the component with default props', () => {
+        mount(<DataUseAcknowledgements {...defaultProps} />);
+        cy.get('.data-use-acknowledgements').should('exist');
+    });
+
+    it('renders the GSO acknowledgement field when needed', () => {
+        const props = {
+            ...defaultProps,
+            datasets: [{ dataUse: {geneticStudiesOnly: true }}]
+        };
+        mount(<DataUseAcknowledgements {...props} />);
+        cy.get('#gsoAcknowledgement').should('exist');
+    });
+
+    it('renders the PUB acknowledgement field when needed', () => {
+        const props = {
+            ...defaultProps,
+            datasets: [{dataUse: { publicationResults: true }}]
+        };
+        mount(<DataUseAcknowledgements {...props} />);
+        cy.get('#pubAcknowledgement').should('exist');
+    });
+
+    it('renders the DS acknowledgement field when needed', () => {
+        const props = {
+            ...defaultProps,
+            dataUseTranslations: ['DS', 'DS2']
+        };
+        mount(<DataUseAcknowledgements {...props} />);
+        cy.get('#dsAcknowledgement').should('exist');
+    });
+
+    it('does not render fields when conditions are not met', () => {
+        mount(<DataUseAcknowledgements {...defaultProps} />);
+        cy.get('#gsoAcknowledgement').should('not.exist');
+        cy.get('#pubAcknowledgement').should('not.exist');
+        cy.get('#dsAcknowledgement').should('not.exist');
+    });
+
+    it('calls onChange when a checkbox is toggled', () => {
+        const props = {
+            ...defaultProps,
+            datasets: [{ dataUse: {geneticStudiesOnly: true }}]
+        };
+        mount(<DataUseAcknowledgements {...props} />);
+        cy.get('input[type="checkbox"]').check().then(() => {
+            expect(props.onChange).to.have.been.called;
+        });
+    });
+
+    it('disables fields in read-only mode', () => {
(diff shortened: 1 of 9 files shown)


cypress/component/DataAccessRequest/data_use_acknowledgements.spec.tsx
rushtong, line 1: Some day we need to standardize on test names

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -21,10 +21,7 @@ import { Storage } from '../../libs/storage';
 import { assign, cloneDeep, get, head, isEmpty, isNil, isString, keys, map } from 'lodash/fp';
 import './DataAccessRequestApplication.css';
 
-import {
-  validateDARFormData
-} from '../../utils/darFormUtils';
-import { isArray, set } from 'lodash';
+import {validateDARFormData} from '../../utils/darFormUtils';
```
rushtong: Imports can be shortened here and in other components as well
```suggestion
import {validateDARFormData} from 'src/libs/utils/darFormUtils';
```

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -3,28 +3,36 @@ import {AxiosError} from 'axios';
...
+  const {progressReport, parentReferenceId, onSuccess, onCancel, validateForm} = props;
 
-  const submit = async () => {
-    try {
-      const submittedPR = await ProgressReport.submitProgressReport(progressReport, parentReferenceId);
-      onSuccess(submittedPR);
-    } catch (error: unknown) {
-      handleError('Error: Unable to submit progress report: ', error);
+  const submit = async (e) => {
+    e.preventDefault();
```
rushtong: This shouldn't be necessary since https://github.com/DataBiosphere/duos-ui/pull/2876 hit

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -3,28 +3,36 @@ import {AxiosError} from 'axios';
...
+    if (validationFailed(validateForm())) {
+      Notifications.showError({text: "Form validation failed. Please check the form for errors."});
+    } else {
```
rushtong, line 24: I would pull this out of the submit function since it forks behavior. Instead, validation should be checked before the submit button is clickable, preferably in the parent.

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -3,28 +3,36 @@ import {AxiosError} from 'axios';
...
 import {ConsentError} from '../../types/responseTypes';
+import {validationFailed} from "src/utils/darFormUtils";
 
 
 interface SubmitProgressReportProps {
   readonly progressReport: object;
   readonly parentReferenceId: string;
   readonly onSuccess: (result: unknown) => void;
   readonly onCancel: (result: unknown) => void;
+  readonly validateForm: () => any;
```
rushtong: See my other comment about this - I think we're putting too much into this component, it should not be enabled unless validation has already happened and complete.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -71,6 +123,7 @@ export const ProgressReportApplication = ({ dar, parentDar, datasets, readOnlyMo
                     }}
                     onCancel={() => {
                     }}
+                    validateForm={validateForm}
```
rushtong, line 143: I would expect the result of this function to be called on line 118 so `SubmitProgressReport` doesn't need to call back to the parent. Maybe another function that looks something like:
```
const isValidProgressReport = () => {
  return !readOnlyMode 
      && parentDar 
      && validatePRFormData(formData, datasets, dataUseTranslations);
}
```

rushtong (commented):
Still reviewing, but some comments in the meantime

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -7,17 +7,25 @@ import CollaboratorChanges from 'src/pages/progress_reports/CollaboratorChanges'
 import DataManagementIncident from 'src/pages/progress_reports/DataManagementIncident';
 import DarCloseout from 'src/pages/progress_reports/DarCloseout';
 import SubmitProgressReport from 'src/pages/progress_reports/SubmitProgressReport';
+import {DataUseAcknowledgements} from "src/pages/dar_application/DataUseAcknowlegements";
+import {translateDataUseRestrictionsFromDataUseArray} from "src/libs/dataUseTranslation";
+import {validatePRFormData} from "src/utils/darFormUtils";
+import {FormValidationState} from "src/pages/dar_application/FormValidationState";
```
rushtong: minor nit: prefer single quotes. Not sure why eslint isn't reporting this, it should.

cypress/component/ProgressReports/SubmitProgressReport.spec.tsx
```diff
@@ -60,20 +66,54 @@ describe('SubmitProgressReport tests', () => {
...
+              onCancel={() => {
+              }}
+              validateForm={() => {
+                  return {darErrors: {datasetIds: {valid: false, failed: ['required']}}}
+              }}
+            />
+      );
+      cy.get('[data-cy=pr-submit-button]').should('exist');
+      cy.get('[data-cy=pr-submit-button]').click();
+      cy.get('@successHandler').should('not.have.been.called');
```
rjohanek, line 114: testing the successHandler is not called, but ideally I'd like to check that the API is not called either. Does anyone know how to do that?

rushtong (approved):

snf2ye (approved):

PR #2878, opened 2025-05-20
[DT-1657] Initial form state

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1657

### Summary

----

- Verify all tests go green

### DIFF
--- a/cypress/component/ProgressReports/collaborator_changes.spec.tsx
+++ b/cypress/component/ProgressReports/collaborator_changes.spec.tsx
@@ -57,7 +57,7 @@ describe('Collaborator Changes - Component Tests', () => {
   });
 
   it('displays preloaded internal lab staff', () => {
-    mountComponent({ internalLabStaff: initialCollaborators });
+    mountComponent({ labCollaborators: initialCollaborators });
     
     cy.contains('Test User 1').should('be.visible');
     cy.contains('Test User 2').should('be.visible');
@@ -82,7 +82,7 @@ describe('Collaborator Changes - Component Tests', () => {
   it('renders in read-only mode correctly', () => {
     const props = {
       readOnly: true,
-      formState: { internalLabStaff: initialCollaborators },
+      formState: { labCollaborators: initialCollaborators },
       onFormChange: onFormChangeSpy
     };
     
@@ -94,7 +94,7 @@ describe('Collaborator Changes - Component Tests', () => {
 
   it('handles empty collaborator lists', () => {
     mountComponent({
-      internalLabStaff: [],
+      labCollaborators: [],
       internalCollaborators: [],
       externalCollaborators: []
     });
--- a/src/pages/dar_application/ProgressReportApplication.tsx
+++ b/src/pages/dar_application/ProgressReportApplication.tsx
@@ -9,23 +9,32 @@ import DarCloseout from 'src/pages/progress_reports/DarCloseout';
 import SubmitProgressReport from 'src/pages/progress_reports/SubmitProgressReport';
 
 type ProgressReportApplicationProps = {
-    dar?: DataAccessRequest, // Dar will be empty if this is an application
-    parentDar?: DataAccessRequest, // Dar will be empty if this is view only
+    dar: DataAccessRequest, // corresponds either to the parent DAR for a new application or an existing readonly progress report
     datasets: Dataset[],
     readOnlyMode?: boolean
 };
 
-export const ProgressReportApplication = ({ dar, parentDar, datasets, readOnlyMode = true }: ProgressReportApplicationProps) => {
+export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true }: ProgressReportApplicationProps) => {
     const initialState = {
-        ...parentDar,
+        ...dar,
         // additional state for summary section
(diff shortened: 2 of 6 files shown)


cypress/component/ProgressReports/collaborator_changes.spec.tsx
```diff
@@ -57,7 +57,7 @@ describe('Collaborator Changes - Component Tests', () => {
   });
 
   it('displays preloaded internal lab staff', () => {
-    mountComponent({ internalLabStaff: initialCollaborators });
+    mountComponent({ labCollaborators: initialCollaborators });
```
fboulnois, line 60: Renamed to match the DAR form state.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -16,7 +16,19 @@ type ProgressReportApplicationProps = {
 };
 
 export const ProgressReportApplication = ({ dar, parentDar, datasets, readOnlyMode = true }: ProgressReportApplicationProps) => {
-    const [formState, setFormState] = useState<FormState>({});
+    const initialState = {
+        ...parentDar,
+        // additional state for summary section
+        intellectualPropertyYesNo: (parentDar?.intellectualPropertySummary?.length ?? 0) > 0,
```
fboulnois: Used to preselect true / false selections on the form.

src/pages/progress_reports/DataManagementIncident.tsx
```diff
@@ -28,6 +28,7 @@ export default function DataManagementIncident(props: DataManagementIncidentProp
                             titleStyle={titleStyle}
                             description='Have there been any incidents related to mismanagement or misuse of data?'
                             orientation='horizontal'
+                            defaultValue={formState.dmiYesNo}
```
fboulnois, line 31: Some missing default values.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -16,7 +16,19 @@ type ProgressReportApplicationProps = {
 };
 
 export const ProgressReportApplication = ({ dar, parentDar, datasets, readOnlyMode = true }: ProgressReportApplicationProps) => {
-    const [formState, setFormState] = useState<FormState>({});
+    const initialState = {
+        ...parentDar,
```
fboulnois: inherit almost all of the initial state from the parent dar

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: Per Jonathan these need to be undefined in the application view.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -9,14 +9,35 @@ import DarCloseout from 'src/pages/progress_reports/DarCloseout';
...
     readOnlyMode?: boolean
 };
 
-export const ProgressReportApplication = ({ dar, parentDar, datasets, readOnlyMode = true }: ProgressReportApplicationProps) => {
-    const [formState, setFormState] = useState<FormState>({});
+export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true }: ProgressReportApplicationProps) => {
+    const initialState = {
+        ...dar,
+        // additional state for summary section
+        ...(dar?.intellectualPropertySummary && {
```
fboulnois, line 21: I'm not sure of a more compact way to approach this. This says: if the value is not undefined (which is the case in the readonly view), then set the correct key with the true / false value so these can be selected in the form.

rjohanek (commented):
can we see what this looks like for a PR from an original DAR? and maybe for a second PR too?

src/pages/dar_application/ProgressReportApplication.tsx
snf2ye: I think this is slightly more readable.. but, I'm not sure if it works the same to explicitly set it to undefined. 
```
const initialState = {
    ...dar,
    intellectualPropertyYesNo: dar?.intellectualPropertySummary ? !!dar.intellectualPropertySummary : undefined,
    publicationsYesNo: dar?.publications ? (dar.publications.length > 0) : undefined,
    presentationsYesNo: dar?.presentations ? (dar.presentations.length > 0) : undefined,
    dmiYesNo: dar?.dataManagementIncident?.incidents ? (dar.dataManagementIncident.incidents.length > 0) : undefined,
    closeoutYesNo: dar?.closeOutSupplement ? !!dar.closeOutSupplement : undefined,
};
```

snf2ye (approved):

rushtong (approved):
Thank you! 👍🏽

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: I'm not sure either. I try to avoid setting stuff to undefined, since we've seen state issues with that before.

fboulnois: I think I'll keep it as is and if it's too confusing we can discuss another solution.

fboulnois:
> can we see what this looks like for a PR from an original DAR? and maybe for a second PR too?

This is hard to do while the model is changing, so I assumed that the existing model was correct (it is different right now on the form).

PR #2879, opened 2025-05-20
DT-108: Submit Progress Report Application

snf2ye:
### Addresses
Enables submitting a progress report

### Summary
- Mapped form state filled out by user to the format that consent expects and then made sure format matched again when displaying the read only values
- Updated submit component to send a multi-part form
- Updated types
- DAR closeout changed the most significantly: I've added yes/no radio options so that this section is only displayed if "yes" is selected and switched from mutli-select to radio options. 
image

image


src/components/collaborator_list/Collaborator.tsx
```diff
@@ -1,8 +0,0 @@
-export interface Collaborator {
-    uuid: string;
-    name: string;
-    eraCommonsId: string;
-    email: string;
-    title: string;
-    institution: string;
-}
```
snf2ye, line 8: Removed this type in favor of the "Collaborator" type in the main type file. This meant removing the "institution" field, which is not currently tracked in consent and we are moving away from tracking anyway. A corresponding change is made in the CollaboratorAddEdit component to no longer capture "institution".

src/pages/progress_reports/DarCloseout.tsx
```diff
@@ -18,77 +18,44 @@ export default function DarCloseout(props: DarCloseoutProps): React.JSX.Element
...
+                                    { text: 'The project is being transferred to a new Requestor at the same institution', name: CloseOutSupplement.PROJECT_TRANSFERRED },
+                                    { text: 'The project is being superseded by a new project', name: CloseOutSupplement.PROJECT_SUPERSEDED }
+                                ]}
+                                disabled={readOnly}
+                                defaultValue={formState.closeoutSupplement}
+                                onChange={({ key, value }: Partial<FormState>) => {
+                                    onFormChange({ [key]: value } as Partial<FormState>);
+                                }}
+                            />
+                        </div>
```
snf2ye, line 61: The back end currently expects a single value for the closeout field. So, this changes from a multi-select to a single select. We will revisit this when taking a closer look at the closeout process.

src/components/publications_list/PublicationOrPresentation.tsx
```diff
@@ -1,4 +1,4 @@
-export interface Publication {
+export interface PublicationOrPresentation {
```
snf2ye, line 1: Renamed this to make it clearer when we are using the main "publication" type vs. the type that is now shared with presentation.

src/types/model.ts
```diff
@@ -402,16 +409,16 @@ export interface Publication {
...
 
 export interface Collaborator {
   approverStatus: boolean;
   email: string;
   eraCommonsId: string;
   name: string;
-  title: number;
-  uuid: number;
+  title: string
+  uuid: string;
```
snf2ye, line 418: Changed these types to match types in consent

src/pages/progress_reports/DarCloseout.tsx
```diff
@@ -18,77 +18,44 @@ export default function DarCloseout(props: DarCloseoutProps): React.JSX.Element
...
+                        id={FormStateKey.CLOSEOUT_YES_NO}
+                        type={FormFieldTypes.YESNORADIOGROUP}
+                        title='5.1 Closeouts'
+                        description={<span>Are you ready to finish work on this project?</span>}
+                        orientation='horizontal'
+                        onChange={({ key, value }: Partial<FormState>) => {
+                            onFormChange({ [key]: value }  as Partial<FormState>);
+                        }}
+                        disabled={readOnly}
+                    />
```
snf2ye: I added a yes/no radio button to the closeout section. Otherwise, we don't really have a way to clear out the radio button selection.

snf2ye (commented):
Comments for reviewers

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -39,7 +44,7 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true }
         ...(dar?.closeOutSupplement && {
             closeoutYesNo: !!dar.closeOutSupplement
         }),
-    }
+    };
 
     const [formState, setFormState] = useState<FormState>(initialState);
```
snf2ye, line 57: The linter in my IDE is complaining about this because the initial state is actually the union of FormState and DataAccessRequest types.

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -568,7 +568,7 @@ const DataAccessRequestApplication = (props) => {
                   <ConditionalAccordion
                       condition={false}
                       title={`DAR Report ${reverseOrderedDARs.length}`}>
-                    <ProgressReportApplication readOnlyMode={false} datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)} dar={reverseOrderedDARs[0]}/>
+                    <ProgressReportApplication readOnlyMode={false} datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)} dar={reverseOrderedDARs[0]} history={props.history}/>
```
fboulnois: Consider pulling history out of props as a const, it makes it easier to see where props end up.

src/pages/progress_reports/DarCloseout.tsx
fboulnois: If we do this I think we also probably need to clear out all previously selected values.

src/pages/progress_reports/DarCloseout.tsx
fboulnois: Good idea!

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -5,35 +5,128 @@ import {Notifications} from 'src/libs/utils';
...
+    return publications.map((pub: PublicationOrPresentation) => {
+      const expectedPublication: Publication = {} as Publication;
+      expectedPublication.title = pub.title;
+      expectedPublication.pubmedId = pub.pubmed_id;
+      expectedPublication.date = pub.date;
+      expectedPublication.authors = pub.authors;
+      expectedPublication.bibliographicCitation = pub.bibliographic_citation
+      expectedPublication.datasetCitation = pub.dataset_citation;
+      expectedPublication.citation = pub.did_cite;
+      return expectedPublication;
```
fboulnois, line 40: Another solution here might be to refactor these values so they match what is expected, that way no transformation is required. (I can do this if you'd like)

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -5,35 +5,128 @@ import {Notifications} from 'src/libs/utils';
...
+    });
+  };
+
+  const getPresentationList = (formState: FormState): Presentation[] => {
+    const presentations: PublicationOrPresentation[] = formState.presentations ?? [];
+    return presentations.map((pub: PublicationOrPresentation) => {
+      const expectedPresentation: Presentation= {} as Presentation;
```
fboulnois, line 47: ditto here

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -5,35 +5,128 @@ import {Notifications} from 'src/libs/utils';
...
+      expectedForm.presentations = getPresentationList(formState);
+    }
+    expectedForm.labCollaborators = formState.labCollaborators ?? [];
+    expectedForm.internalCollaborators = formState.internalCollaborators ?? [];
+    expectedForm.externalCollaborators = formState.externalCollaborators ?? [];
+    if (formState.dmiYesNo) {
+      expectedForm.dataManagementIncident = getDataManagementIncidents(formState);
+    }
+    if (formState.closeoutYesNo) {
+      expectedForm.closeOutSupplement = formState.closeoutSupplement;
```
fboulnois: I can fix this up too, if that would be convenient.

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -5,35 +5,128 @@ import {Notifications} from 'src/libs/utils';
...
+    expectedForm.datasetIds = formState.datasetIds ?? [];
+    if (formState.publicationsYesNo) {
+      expectedForm.publications = getPublicationList(formState);
+    }
+    if (formState.presentationsYesNo) {
+      expectedForm.presentations = getPresentationList(formState);
+    }
+    expectedForm.labCollaborators = formState.labCollaborators ?? [];
+    expectedForm.internalCollaborators = formState.internalCollaborators ?? [];
+    expectedForm.externalCollaborators = formState.externalCollaborators ?? [];
```
fboulnois: These should default to `[]`, is this check necessary? If not I can fix this in the initial state.

src/pages/progress_reports/SubmitProgressReport.tsx
snf2ye: That would definitely be better, and I think breaking it out into a follow-on ticket/PR would be good.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -35,14 +39,23 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
+            dmiYesNo: (dar.dmi.incidents.length > 0),
+            dmiCombination: dar.dmi.incidents.includes(FormStateKey.DMI_COMBINATION),
+            dmiIdentification: dar.dmi.incidents.includes(FormStateKey.DMI_IDENTIFICATION),
+            dmiSharing: dar.dmi.incidents.includes(FormStateKey.DMI_SHARING),
+            dmiSecurity: dar.dmi.incidents.includes(FormStateKey.DMI_SECURITY),
+            dmiAcknowledgement: dar.dmi.incidents.includes(FormStateKey.DMI_ACKNOWLEDGEMENT),
+            dmiPublication: dar.dmi.incidents.includes(FormStateKey.DMI_PUBLICATION),
+            dmiFalsification: dar.dmi.incidents.includes(FormStateKey.DMI_FALSIFICATION),
+            dmiOther: dar.dmi.incidents.includes(FormStateKey.DMI_OTHER),
+            dmiDescription: dar.dmi.description
```
snf2ye: These fields are needed because we're mapping between the way the UI expects the data vs. how it is stored in the backend. We'll probably want to to rectify this at some point.

src/types/model.ts
```diff
@@ -372,12 +372,19 @@ export interface DataAccessRequest {
   intellectualPropertySummary: string;
   publications: Array<Publication>;
   presentations: Array<Presentation>;
-  dataManagementIncident: DataManagementIncident;
+  dmi: DataManagementIncident;
   researchPlans: string;
-  closeOutSupplement: 'PROJECT_COMPLETED' | 'REQUESTOR_MOVED_INSTITUTION' | 'PROJECT_TRANSFERRED' | 'PROJECT_SUPERSEDED';
+  closeoutSupplement: CloseOutSupplement;
```
snf2ye, line 377: Capitalization and exact name matters when mapping to the db model. The changes to dmi and closeoutSupplement match the state in consent.

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: This is great! Thank you!

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -5,35 +5,125 @@ import {Notifications} from 'src/libs/utils';
...
+    return expectedForm;
+  }
 
   const submit = async () => {
     if (validationFailed(validateForm())) {
       Notifications.showError({text: "Form validation failed. Please check the form for errors."});
     } else {
       try {
-        const submittedPR = await ProgressReport.submitProgressReport(progressReport, parentReferenceId);
+        const submittedPR = await ProgressReport.submitProgressReport(createMultiPartFormData(convertFormStateToDAR(formState)), parentReferenceId);
```
fboulnois:
```suggestion
        const multiPartFormData = createMultiPartFormData(convertFormStateToDAR(formState));
        const submittedPR = await ProgressReport.submitProgressReport(multiPartFormData, parentReferenceId);
```

Small change for clarity.

fboulnois (approved):
looks good 👍 ! One small suggestion:

src/pages/progress_reports/SubmitProgressReport.tsx
snf2ye: 👍 I'm not sure why I originally added these default states, but seems to work fine without.

src/pages/progress_reports/DarCloseout.tsx
snf2ye: Good point - I am making it so that the state doesn't get populated if closeoutYesNo = no, but clearing out the state is much clearer to the user

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -35,14 +39,23 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
             presentationsYesNo: (dar.presentations.length > 0)
         }),
         // additional state for dmi section
-        ...(dar?.dataManagementIncident?.incidents && {
-            dmiYesNo: (dar.dataManagementIncident.incidents.length > 0)
+        ...(dar?.dmi?.incidents && {
+            dmiYesNo: (dar.dmi.incidents.length > 0),
+            dmiCombination: dar.dmi.incidents.includes(FormStateKey.DMI_COMBINATION),
```
s-rubenstein: Optional: My brain is bouncing around some way to avoid the duplicated `dar.dmi.incidents.includes` bit, I think you might be able to do a list of keys, and then something like: `...(_.map(includedKeys, key => ({ [key]: dar.dmi.incidents.includes(key)})).reduce((acc, obj) => ({ ...acc, ...obj }), {})) `

s-rubenstein (approved):

src/pages/dar_application/ProgressReportApplication.tsx
s-rubenstein: But that might be an unreadable mess to anyone in 6 months, so your mileage may vary.

snf2ye: Nice, thank you! I fiddled with trying to get this populate by iterating through a list of the keys but never quite got it working. Your solution works and then copilot simplified a little further, so I think it is will be understandable in the future. (and I kinda hope we find some time to refactor this code in general so we don't have to do so many transformations between models as @fboulnois mentioned above.)

PR #2882, opened 2025-05-21
DT-1684: Add Researcher Identification section to PR Summary

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1684

## Summary
Adds a Researcher Identification panel to the Progress Report Application.

* In application mode, it allows for users to remove/re-authenticate with RAS/eRA Commons
* In read-only mode, it displays relevant researcher information in the same way that we show for submitted DARs or on the DAR Review page.
* Update the dar form utility to dynamically calculate RAS/eRA Commons validity
* Includes a fix for the `ResearcherInfo` component to hide the required `*` when in read-only mode.

### Notes
I suggest hiding white space due to the auto-format lint changes.

#### User Not RAS Authenticated
Screenshot 2025-05-21 at 9 29 30 AM

#### After RAS Authentication
Screenshot 2025-05-21 at 9 29 54 AM

#### Readonly View
Screenshot 2025-05-21 at 10 37 22 AM

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/DataAccessRequestApplication.jsx
+++ b/src/pages/dar_application/DataAccessRequestApplication.jsx
@@ -116,6 +116,8 @@ const DataAccessRequestApplication = (props) => {
     collaborationLetterName: '',
   });
 
+  const {location} = props;
+
   const [formValidation, setFormValidation] = useState({ researcherInfoErrors: {}, darErrors: {}, rusErrors: {} });
 
   const [nihValid, setNihValid] = useState(true);
@@ -572,7 +574,7 @@ const DataAccessRequestApplication = (props) => {
                       readOnlyMode={false}
                       datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)}
                       dar={reverseOrderedDARs[0]}
-                      location={props.location}
+                      location={location}
                       researcher={researcher}
                     />
                   </ConditionalAccordion>
@@ -622,7 +624,7 @@ const DataAccessRequestApplication = (props) => {
                   formValidationChange={(val) => formValidationChange('researcherInfoErrors', val)}
                   eRACommonsDestination={eRACommonsDestination}
                   formFieldChange={formFieldChange}
-                  location={props.location}
+                  location={location}
                   nihValid={nihValid}
                   onNihStatusUpdate={setNihValid}
                   showNihValidationError={showNihValidationError}
--- a/src/pages/dar_application/DataAccessRequestApplication.jsx
+++ b/src/pages/dar_application/DataAccessRequestApplication.jsx
@@ -568,7 +568,13 @@ const DataAccessRequestApplication = (props) => {
                   <ConditionalAccordion
                       condition={false}
                       title={`DAR Report ${reverseOrderedDARs.length}`}>
-                    <ProgressReportApplication readOnlyMode={false} datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)} dar={reverseOrderedDARs[0]}/>
+                    <ProgressReportApplication
+                      readOnlyMode={false}
+                      datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)}
+                      dar={reverseOrderedDARs[0]}
+                      location={props.location}
+                      researcher={researcher}
+                    />
                   </ConditionalAccordion>
                 </div>
             )}
@@ -584,7 +590,13 @@ const DataAccessRequestApplication = (props) => {
                                 condition={true}
                                 title={`DAR Report ${reverseOrderedDARs.length - index - 1}`}
                                 defaultExpanded={index === 0}>
-                              <ProgressReportApplication readOnlyMode={true} datasets={filterForProgressReport(datasets, dar.datasetIds)} dar={dar?.data}/>
+                              <ProgressReportApplication
+                                readOnlyMode={true}
+                                datasets={filterForProgressReport(datasets, dar.datasetIds)}
+                                dar={dar?.data}
+                                location={undefined}
+                                researcher={researcher}
+                              />
                             </ConditionalAccordion>
                           </div>);
                     }
--- a/src/pages/dar_application/ProgressReportApplication.tsx
+++ b/src/pages/dar_application/ProgressReportApplication.tsx
@@ -1,6 +1,7 @@
 import React, {useCallback, useEffect, useState} from 'react';
-import { DataAccessRequest, Dataset } from 'src/types/model';
-import { FormState } from 'src/pages/progress_reports/ProgressReportFormState';
+import {DataAccessRequest, Dataset, DuosUser} from 'src/types/model';
+import {Location} from 'history';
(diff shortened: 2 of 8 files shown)


src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -568,7 +568,13 @@ const DataAccessRequestApplication = (props) => {
                   <ConditionalAccordion
                       condition={false}
                       title={`DAR Report ${reverseOrderedDARs.length}`}>
-                    <ProgressReportApplication readOnlyMode={false} datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)} dar={reverseOrderedDARs[0]}/>
+                    <ProgressReportApplication
+                      readOnlyMode={false}
+                      datasets={filterForProgressReport(datasets, reverseOrderedDARs[0].datasetIds)}
+                      dar={reverseOrderedDARs[0]}
+                      location={props.location}
```
fboulnois: Can we pull `location` out as a const instead? It makes it more difficult to identify where certain variables are used otherwise.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -45,8 +48,9 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true }
     const [dataUseTranslations, setDataUseTranslations] = useState<string[]>([]);
     const [selectedDatasets, setSelectedDatasets] = useState<Dataset[]>(datasets);
     const [formValidation, setFormValidation] = useState<FormValidationState>(
-        {darErrors:
-                {gsoAcknowledgement: {}, pubAcknowledgement: {}, dsAcknowledgement: {}}});
+      {darErrors:
+            {gsoAcknowledgement: {}, pubAcknowledgement: {}, dsAcknowledgement: {}}});
+    const eRACommonsDestination = 'progress_report_application/' + dar?.collectionId;
```
fboulnois:
```suggestion
    const eRACommonsDestination = 'progress_report_application/' + dar.collectionId;
```

`dar.collectionId` should never be null.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -10,15 +11,17 @@ import SubmitProgressReport from 'src/pages/progress_reports/SubmitProgressRepor
...
 
 type ProgressReportApplicationProps = {
-    dar: DataAccessRequest, // corresponds either to the parent DAR for a new application or an existing readonly progress report
-    datasets: Dataset[],
-    readOnlyMode: boolean
+  readonly dar: DataAccessRequest, // corresponds either to the parent DAR for a new application or an existing readonly progress report
+  readonly datasets: Dataset[],
+  readonly readOnlyMode?: boolean
+  readonly location?: Location
+  readonly researcher: DuosUser
```
fboulnois: Do these need commas for consistency?

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -10,15 +11,17 @@ import SubmitProgressReport from 'src/pages/progress_reports/SubmitProgressRepor
...
-import {FormValidationState} from 'src/pages/dar_application/FormValidationState';
+import {FormValidationState, ValidationError} from 'src/pages/dar_application/FormValidationState';
 
 type ProgressReportApplicationProps = {
-    dar: DataAccessRequest, // corresponds either to the parent DAR for a new application or an existing readonly progress report
-    datasets: Dataset[],
-    readOnlyMode: boolean
+  readonly dar: DataAccessRequest, // corresponds either to the parent DAR for a new application or an existing readonly progress report
+  readonly datasets: Dataset[],
+  readonly readOnlyMode?: boolean
```
fboulnois: This should never be optional

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -10,15 +11,17 @@ import SubmitProgressReport from 'src/pages/progress_reports/SubmitProgressRepor
...
+  readonly location?: Location
```
fboulnois: Why is location optional?

src/pages/progress_reports/SummarySection.tsx
```diff
@@ -1,18 +1,24 @@
...
+import ERACommons from 'src/components/ERACommons';
+import {DuosUser} from 'src/types/model';
+import {Location} from 'history';
 
 interface SummarySectionProps {
     readonly readOnly: boolean;
     formState: FormState;
     onFormChange: (newState: Partial<FormState>) => void;
+    readonly eRACommonsDestination?: string
+    readonly location?: Location
```
fboulnois: Why are these optional?

fboulnois (approved):
Some comments and suggestions:

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: It's not necessary in the read-only mode. Location is what we use to get back to the current page after redirecting to the NIH auth page.

src/pages/progress_reports/SummarySection.tsx
rushtong: `eRACommonsDestination` is only necessary when being redirected away from DUOS to NIH for authentication, so we don't need it in read-only mode. Same for `location`.

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: I see a couple different approaches which all seem to work which is confusing. I see no separators, comma separators, and semi-colon separators. Looking at [some docs](https://www.typescriptlang.org/docs/handbook/utility-types.html), I think semi-colon is the expectation.

s-rubenstein: Does our linter not do anything here?

s-rubenstein (approved):
LGTM

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: > Does our linter not do anything here?

Unfortunately, I'm not seeing any warnings in my setup - I think we need to look into adding something for this.

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: It sounds like we will have a number of objects that will only exist if `readOnlyMode` is false. Can we make this more clear? This can be represented in Typescript using discriminated union types, for example:

```
type ConditionalObject = 
  | { readOnly: true; propertyWhenReadOnly: string }
  | { readOnly: false; propertyWhenReadOnly?: never };
```

Another possibility is to have a separate object just for `nonReadOnly` types, that way it makes it easier to scan what is expected vs non-expected types.

```
missingWhenReadOnly: {
    location?: Location,
}
```

src/pages/progress_reports/SummarySection.tsx
fboulnois: ditto my previous comment

PR #2883, opened 2025-05-21
DT-1683: Update Library Card Table to TypeScript and remove Update LC functionality

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1683

### Summary
* Convert updated components to TSX
* Add tests
* Remove update functionality
* Remove ajax call to update Library Cards
* Update the UI to more accurately reflect the data that exists in Library Cards
* General functionality is not changed - Admins can still add/remove LCs according to the previous logic

----

- Verify all tests go green


package.json
```diff
@@ -52,6 +52,7 @@
     "@types/lodash": "4.17.16",
     "@types/react": "18.3.12",
     "@types/react-dom": "18.3.1",
+    "@types/react-modal": "3.16.3",
```
fboulnois: What additional functionality does this bring that we do not already have? If we need this, should we refactor our other modals to use it?

src/components/library_card_table/LibraryCardTable.tsx
```diff
@@ -0,0 +1,431 @@
...
+  useEffect(() => {
+    setLibraryCards(props.libraryCards ?? []);
+    setUsers(props.users ?? []);
+    if (
+        !isNil(props.libraryCards) &&
+        !isNil(props.users)
+    ) {
+      setIsLoading(false);
+    }
+  }, [props.libraryCards, props.users]);
```
fboulnois, line 254: Do we need this hook? We are already setting these values in `setState` above. If we don't need this hook, then we can probably simplify this further and may not even need `setLibraryCards` / `setUsers`.

fboulnois (approved):
A few comments:

package.json
rushtong: [Line 5 in LibraryCardFormModal](https://github.com/DataBiosphere/duos-ui/pull/2883/files#diff-7df667067f81993f377c532ec08ecf4f0d4f728cc3d7adef5c95268945d057d0R4) requires this for typescript compilation. I'm sure that there are other usages of this but I didn't see any in TSX files yet, just JSX files.

src/components/library_card_table/LibraryCardTable.tsx
rushtong: This is almost an exact [copy of the JSX version](https://github.com/DataBiosphere/duos-ui/pull/2883/files#diff-42fc947cca7b1cb3004ad08ad36c5ac776be3898172cdcdee1e3efeb8bff204dL243) so I didn't look into trying to remove it. I can dig in a little further and try to do that.

package.json
fboulnois: I have a `ModalWrapper.jsx` that allows us to use our existing modals with Typescript without requiring additional imports.

snf2ye (approved):

package.json
rushtong: Thank you, that works!

src/components/library_card_table/LibraryCardTable.tsx
rushtong: I tested this out, we do need this - none of the table data loads if we don't call this hook.

PR #2884, opened 2025-05-23
[DT-108] Improve types flagged by typescript

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-108

### Summary

Improve some of the types flagged by `pnpm tsc`.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/Auth/mockOidcUser.ts
+++ b/cypress/component/Auth/mockOidcUser.ts
@@ -5,7 +5,7 @@ export const mockOidcUser: OidcUser = {
   get expires_in(): number | undefined {
     return undefined;
   },
-  session_state: undefined,
+  session_state: null,
   state: undefined,
   token_type: '',
   get expired(): boolean | undefined {
@@ -20,7 +20,7 @@ export const mockOidcUser: OidcUser = {
   profile: {
     jti: undefined,
     nbf: undefined,
-    sub: undefined,
+    sub: '',
     iss: '',
     aud: '',
     exp: 0,
--- a/cypress/component/DataAccessRequest/data_use_acknowledgements.spec.tsx
+++ b/cypress/component/DataAccessRequest/data_use_acknowledgements.spec.tsx
@@ -3,80 +3,67 @@ import { mount } from 'cypress/react';
 import { DataUseAcknowledgements } from 'src/pages/dar_application/DataUseAcknowlegements';
 
 describe('DataUseAcknowledgements Component', () => {
-    let defaultProps: object;
+    let onChangeSpy: () => void;
+    let onValidationChangeSpy: () => void;
 
-    beforeEach(() => {
-        defaultProps = {
+    const mountComponent = (customProps = {}) => {
+        const defaultProps = {
             title: 'Data Use Acknowledgements',
             datasets: [],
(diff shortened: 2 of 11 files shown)


cypress/component/Auth/mockOidcUser.ts
rushtong, line 1: Not sure if you're planning on it, but I also see that the import can be shortened and double quotes replaced with single quotes. At one point in time, we did have eslint showing double quotes as a warning but I'm not sure what happened to that.

fboulnois: I think that is a separate issue from Typescript typing.

rushtong (approved):
I like the improvements. Feel free to address imports and quoting as you see fit.

cypress/component/Auth/mockOidcUser.ts
fboulnois: I created a ticket for this: https://broadworkbench.atlassian.net/browse/DT-1700

src/pages/progress_reports/SummarySection.tsx
```diff
@@ -66,7 +66,7 @@ export default function SummarySection(props: SummarySectionProps): React.JSX.El
                         rows={6}
                         maxLength={FORM_TEXT_AREA_MAX_LENGTH}
                         defaultValue={formState.progressReportSummary}
-                        onChange={({ key, value }: Partial<FormState>) => {
+                        onChange={({ key, value }: AnyFormState) => {
                             onFormChange({ [key]: value } as Partial<FormState>);
```
snf2ye: Can you talk more about this change?
It seems like the `Partial<FormState>` type could still be useful because it should check if the key is a valid field in FormState and that the value matches that key's type.  On the other hand, `AnyFormState` doesn't really do any type checking since it has type "any" as the value.

fboulnois: `Partial<FormState>` isn't valid for these types, since `FormState` explicitly only allows certain key value pairs. `Partial` makes those optional, but not arbitrary.

I could potentially narrow `AnyFormState` to list out the possible value types.

cypress/component/DataAccessRequest/data_use_acknowledgements.spec.tsx
```diff
@@ -3,80 +3,67 @@ import { mount } from 'cypress/react';
...
 
     it('calls onChange when a checkbox is toggled', () => {
-        const onChangeSpy = cy.spy().as('onChangeSpy');
-        const props = {
-            ...defaultProps,
-            datasets: [{ dataUse: { geneticStudiesOnly: true } }],
-            onChange: onChangeSpy
-        };
-        mount(<DataUseAcknowledgements {...props} />);
+        mountComponent({ datasets: [{ dataUse: { geneticStudiesOnly: true } }] });
```
rjohanek, line 57: does this mount the component twice? Once in the beforeEach and then again here?

fboulnois: Cypress cleans up previous mounts when mounting again. However, it's a common pattern in component testing when initial states need to be tested.

rjohanek (approved):

PR #2886, opened 2025-05-27
DT-1697: Enable voting on progress reports

snf2ye:
### Addresses

This PR aims to address an issue where Progress reports could not be voted on because the UI thought the election was closed. 
image
We previously required that all elections in the DAR collection were open. However, it is neccessary for the original DAR to have elections closed in order to open a progress report. 

In other words, within a DAR collection that has a DAR and a progress report, we would have at least two elections:
**Dar Collection**
1. Original DAR - Approved with a closed election
2. Progress report - Open election

### Summary
These changes do two things:
1. Only require there to be one open election on a dataset/Data use in order for a user to submit a vote in the UI
2. Only submit votes for the open elections

### DIFF
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
@@ -4,7 +4,6 @@ import CollectionVoteYesButton from './CollectionVoteYesButton';
 import CollectionVoteNoButton from './CollectionVoteNoButton';
 import {Notifications} from '../../libs/utils';
 import { Votes } from '../../libs/ajax/Votes';
-import {toLower} from "lodash";
 
 const styles = {
   baseStyle: {
@@ -107,8 +106,8 @@ export default function CollectionSubmitVoteBox(props) {
 
   const updateVote = async (newVote, isChair) => {
     try {
-      const openElectionVotes = votes.filter(v => toLower(v.electionStatus) === 'open');
-      const voteIds = map(v => v.voteId)(openElectionVotes);
+      const openElectionVotes = votes.filter(v => v.electionStatus.toLowerCase() === 'open');
+      const voteIds = openElectionVotes.map(v => v.voteId);
       await Votes.updateVotesByIds(voteIds, {vote: newVote, rationale});
       setSubmitted(true);
       //call updateFinalVote for chairs in order to update source collection's votes and trigger sub-component re-render
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
@@ -112,9 +112,13 @@ export default function CollectionSubmitVoteBox(props) {
       await Votes.updateVotesByIds(voteIds, {vote: newVote, rationale});
       setSubmitted(true);
       //call updateFinalVote for chairs in order to update source collection's votes and trigger sub-component re-render
-      isChair ? updateFinalVote(bucketKey, {vote: newVote, rationale}, voteIds) : setVote(newVote);
+      if (isChair) {
+        updateFinalVote(bucketKey, {vote: newVote, rationale}, voteIds);
+      } else {
+        setVote(newVote);
+      }
       Notifications.showSuccess({text: 'Successfully updated vote'});
-    } catch (error) {
+    } catch (_error) {
       Notifications.showError({text: 'Error: Failed to update vote'});
     }
   };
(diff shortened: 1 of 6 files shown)


src/components/collection_voting_slab/MultiDatasetVoteSlab.jsx
```diff
@@ -98,7 +98,7 @@ export default function MultiDatasetVoteSlab(props) {
     const allOpenElections = flow(
       get('elections'),
       filter((election) => includes(election.electionId)(electionIds)),
-      every((election) => toLower(election.status) === 'open')
+      filter((election) => toLower(election.status) === 'open')
```
snf2ye: Question: can we handle not all elections being open?

This matches a check on the backend (so we would have to make a change in consent as well if we went through with this change): 
https://github.com/DataBiosphere/consent/blob/fc0ec261c1e8f3741bf8c76940aa4617f4cb4b52/src/main/java/org/broadinstitute/consent/http/service/VoteService.java#L453

**EDIT**: this check in the backend does not reflect this same requirement.

rushtong: I don't think we can. If we allow closed elections here, we might be showing votes for those elections which have already been voted on.

snf2ye: Ok - I think the second bit of code in this PR addresses that: We filter so only votes associated with open elections are included/submitted.

rushtong (approved):
I had a hard time testing this with the current back end validation bugs, but this looks reasonable to me.

otchet-broad (approved):
Nice work cleaning this up for progress reports!

src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
```diff
@@ -106,13 +107,18 @@ export default function CollectionSubmitVoteBox(props) {
 
   const updateVote = async (newVote, isChair) => {
     try {
-      const voteIds = map(v => v.voteId)(votes);
+      const openElectionVotes = votes.filter(v => toLower(v.electionStatus) === 'open');
+      const voteIds = map(v => v.voteId)(openElectionVotes);
```
fboulnois:
```suggestion
      const voteIds = openElectionVotes.map(v => v.voteId);
```

Avoids use of `lodash/fp` , which is no longer being updated and causes typing issues with Typescript.

src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
```diff
@@ -106,13 +107,18 @@ export default function CollectionSubmitVoteBox(props) {
 
   const updateVote = async (newVote, isChair) => {
     try {
-      const voteIds = map(v => v.voteId)(votes);
+      const openElectionVotes = votes.filter(v => toLower(v.electionStatus) === 'open');
```
fboulnois: I think there's also a built-in `toLower` function on strings that we could use instead of using `lodash`.

src/components/collection_voting_slab/MultiDatasetVoteSlab.jsx
```diff
@@ -98,7 +97,7 @@ export default function MultiDatasetVoteSlab(props) {
...
```
fboulnois: Can we use the built-in js approaches here? I think it's something like `elections.filter((election) => (election.status.toLower()) === 'open')` , but I'm not sure.

In general I think we should avoid proliferating additional use of `lodash/fp` since it is deprecated and causes typing issues with Typescript.

src/utils/DarCollectionUtils.js
```diff
@@ -18,8 +18,16 @@ export const processVotesForBucket = (darElections = []) => {
...
+    const {electionType, votes, status = []} = election;
+    // add field to each vote object to indicate election status
+    const updatedVotes = Object.values(votes).map(vote => ({
+      ...vote,
+      electionStatus: status
+    }));
+    updatedVotes.forEach(vote => {
+      votes[vote.voteId] = vote;
+    });
+    const dateSortedVotes = sortBy((vote) => vote.updateDate)(updatedVotes);
```
fboulnois:
```suggestion
    const dateSortedVotes = updatedVotes.sort((vote) => vote.updateDate);
```

I think we can do this instead.

fboulnois (approved):
A few suggestions mainly around use of `lodash/fp`:

src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
snf2ye: Gotcha! Agreed, directly using the map is better.

PR #2887, opened 2025-05-28
[DT-1687]Progress Report Validation and other small fixes

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1687 (Add form Validation for all components)
https://broadworkbench.atlassian.net/browse/DT-1688 (Move validation from on submit to on form update)

### Summary
- Moves validation from the submit button to on form update in the progress report application
- Disable submit button if form is not valid
- Adds validation to all components used in the Progress Report Application including
         - SummarySection, DataManagementIncident, DarCloseout, CollaboratorAddEdit, PublicationsAddEdit
- Adds missing fields (NIH ID and approverStatus) to CollaboratorAddEdit
- Makes a shared component for ApproverStatus
- Makes titles the same size, updates numbering, makes it so the titles turn red when their section is invalid
- Updates validation function to check for all required fields

## Video 
https://drive.google.com/file/d/1BlBn1FQnKVrDoKactxPIVIPWyUoalhCZ/view?usp=sharing

----

- Verify all tests go green


src/utils/darFormUtils.js
```diff
@@ -203,7 +203,55 @@ const calcDarErrors = (formData, datasets, dataUseTranslations, irbDocument, col
...
+  }
+  if (formData.publicationsYesNo && isEmpty(formData.publications)) {
+      errors.publications = requiredError;
+  }
+  if (isNil(formData.presentationsYesNo)) {
+        errors.presentationsYesNo = requiredError;
+  }
+  if (formData.presentationsYesNo && isEmpty(formData.presentations)) {
+      errors.presentations = requiredError;
+  }
```
fboulnois: Could we pull this into `CalcSummaryErrors` and `CalcDMIErrors`?

src/pages/progress_reports/SubmitProgressReport.tsx
```diff
@@ -136,18 +130,24 @@ export default function SubmitProgressReport(props: SubmitProgressReportProps) {
...
-            type={'button'}
+        <span>
+            <button type={'button'}
+                className='button button-blue'
+                style={{marginRight: '2rem', cursor: 'pointer', ...(disabled ? disabledStyle : {})}}
+                data-cy='pr-submit-button'
+                disabled={disabled}
+                title='Complete required form fields to enable submission.'
+                onClick={submit}>Submit
+            </button>
```
fboulnois:
```suggestion
                onClick={submit}>
                Submit
            </button>
```

Improve readability.

src/utils/darFormUtils.js
```diff
@@ -69,26 +69,26 @@ const isStringEmpty = (str) => {
 export const computeCollaboratorErrors = ({collaborator, needsApproverStatus=true}) => {
   const errors = {};
 
-  if (isStringEmpty(collaborator.name)) {
+  if (isStringEmpty(collaborator?.name)) {
```
fboulnois, line 72: I think if you add a `if(collaborator == undefined) { return }` all these `?` wouldn't be necessary.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -86,17 +84,17 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
+    useEffect(() => {
+        if (!readOnlyMode) {
+            const validation = validatePRFormData(
+                formState,
+                selectedDatasets,
+                dataUseTranslations
+            );
+            setFormValidation(validation);
+        }
+    }, [formState]);
```
fboulnois: I would avoid using `useEffect`, as this gets us into infinite loops.

src/pages/dar_application/collaborator/ApproverStatus.tsx
```diff
@@ -0,0 +1,47 @@
...
+                options={[
+                    {name: 'yes', text: 'Yes'},
+                    {name: 'no', text: 'No'}
+                ]}
+                validators={[FormValidators.REQUIRED]}
+                orientation='horizontal'
+                defaultValue={
+                    (approverStatus === true || approverStatus === 'yes') ? 'yes'
+                        : (approverStatus === false || approverStatus === 'no') ? 'no'
+                            : undefined
```
fboulnois: Instead of using multiple embedded `? :`, I would separate the conditions out clearly.

src/pages/dar_application/FormValidationState.tsx
```diff
@@ -33,6 +33,28 @@ export interface DarErrors {
     gsoAcknowledgement?: ValidationError,
     pubAcknowledgement?: ValidationError,
     dsAcknowledgement?: ValidationError,
+    progressReportSummary?: ValidationError,
```
fboulnois, line 37: (perhaps in a follow-up PR): I don't think any of the validation pieces should be optional. I think instead we should consider a model where things default validate to true, unless there's validation method which ends up validating to false.

This would avoid having to make all of these fields optional, and allow us to better type the validation.

src/components/publications_list/PublicationAddEdit.tsx
```diff
@@ -16,10 +18,32 @@ interface PublicationAddEditProps {
...
+}
 export default function PublicationAddEdit(props: PublicationAddEditProps): React.JSX.Element {
     const { id, publication, publicationText, publications, closeAction, onPublicationChange } = props;
 
     const [newPublication, setNewPublication] = useState(publication);
+    const [validation, setValidation] = useState<Validation>({});
+
+    useEffect(() => {
+        setValidation(calcPublicationOrPresentationErrors(newPublication, publicationText));
+    }, [newPublication]);
```
fboulnois: I think we should avoid using a useEffect here as well. When `newPublication` is created, do the validation at that stage.

src/components/collaborator_list/CollaboratorAddEdit.tsx
```diff
@@ -14,12 +18,35 @@ interface CollaboratorAddEditProps {
...
-    const { id, collaborator, collaboratorText, collaborators, closeAction, onCollaboratorChange } = props;
+    const { id, collaborator, collaboratorText, collaborators, closeAction, onCollaboratorChange, showApproverStatus } = props;
 
     const [newCollaborator, setNewCollaborator] = useState(collaborator);
+    const [validation, setValidation] = useState<Validation>({});
+    const accountLabel = nihAccountLabel();
+
+    useEffect(() => {
+        setValidation(computeCollaboratorErrors({collaborator: newCollaborator, needsApproverStatus: showApproverStatus}));
+    }, [setNewCollaborator]);
```
fboulnois: ditto here for collaborators.

fboulnois (commented):
A few suggestions:

src/pages/dar_application/ProgressReportApplication.tsx
rjohanek: if I don't use useEffect, but want to update anytime the form is updated, what would I use?

src/pages/dar_application/FormValidationState.tsx
rjohanek: that makes sense, I was just following the example of the DAR Application. If we change it for PR, we should change it for DARs too

src/components/publications_list/PublicationAddEdit.tsx
rjohanek: what do you mean when newPublication is created? are you suggesting just doing the validation on attempt to submit/add publication instead of on form update? if so, I think we wanted to take a more proactive approach with validation and not even allow an attempt to submit if it is not valid

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: Have we considered [`useFormStatus`](https://react.dev/reference/react-dom/hooks/useFormStatus) as a potential solution?

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -86,17 +85,18 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
+        if (!readOnlyMode) {
+            const validation = validatePRFormData(
+                nihValid,
+                formState,
+                selectedDatasets,
+                dataUseTranslations
+            );
+            setFormValidation(validation);
+        }
+    }, [formState, nihValid]);
```
snf2ye: What happens if you add the missing dependencies as suggested by the linter?

src/utils/darFormUtils.js
```diff
@@ -221,6 +284,32 @@ const calcDUAErrors = (formData, datasets, dataUseTranslations, errors) => {
...
+  }
+  if (isEmpty(newPublication?.date)) {
+    validation.date = requiredError;
+  } else if (!FormValidators.DATE.isValid(newPublication?.date)) {
+    validation.date = validationError('date');
+  }
+  if (isEmpty(newPublication?.authors)) {
+    validation.authors = requiredError;
+  }
+  if (publicationText === 'Publication') {
```
snf2ye: Since it is comparing a hard-coded string, it would be nice to pull this check into a shared utility method. This will help us identify all of the locations where we're doing this sort of check when we separate out the publication/presentation components

snf2ye (commented):
Overall looks reasonable to me - I would like to dig into the useEffects a bit more.

src/components/publications_list/PublicationAddEdit.tsx
fboulnois: We could do it `onChange` instead of on form update / continuously using `useEffect`.

src/pages/dar_application/ProgressReportApplication.tsx
rjohanek: I added them and it doesn't seem anything has changed

src/pages/dar_application/FormValidationState.tsx
otchet-broad: Agree with both of you on this.  It's probably also time to upgrade `forms.jsx` to typescript as well.  It's underlying all of this.

snf2ye (approved):

src/components/collaborator_list/CollaboratorAddEdit.tsx
```diff
@@ -14,12 +18,30 @@ interface CollaboratorAddEditProps {
...
 }
 
+interface Validation {
+    name?: ValidationError;
+    eraCommonsId?: ValidationError;
+    title?: ValidationError;
+    email?: ValidationError;
+    approverStatus?: ValidationError;
+}
 export default function CollaboratorAddEdit(props: CollaboratorAddEditProps): React.JSX.Element {
```
fboulnois:
```suggestion
}

export default function CollaboratorAddEdit(props: CollaboratorAddEditProps): React.JSX.Element {
```

tiny nit

fboulnois (approved):
Looks great! I think the validation is really easy to follow now. Thank you so much.

PR #2894, opened 2025-06-02
[DT-1758] Show approved datasets only

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1758

### Summary

Only show approved datasets from parent DAR. Also includes some additional refactors to avoid using `useEffect`.

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/ProgressReportApplication.tsx
+++ b/src/pages/dar_application/ProgressReportApplication.tsx
@@ -55,9 +55,13 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true, 
     };
 
     const [formState, setFormState] = useState<FormState>(initialState);
+    const [formValidation, setFormValidation] = useState<FormValidationState>({darErrors:{}});
     const [nihValid, setNihValid] = useState<boolean>(true);
     const [dataUseTranslations, setDataUseTranslations] = useState<string[]>([]);
     const [selectedDatasets, setSelectedDatasets] = useState<Dataset[]>(datasets);
+
+    const eRACommonsDestination = 'progress_report_application/' + dar.collectionId;
+
     const getValidation = (newState) => {
         if (!readOnlyMode) {
             return validatePRFormData(
@@ -69,8 +73,6 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true, 
         }
         return {darErrors: {}}
     }
-    const [formValidation, setFormValidation] = useState<FormValidationState>({darErrors:{}});
-    const eRACommonsDestination = 'progress_report_application/' + dar.collectionId;
 
     const onFormChange = (newState: Partial<FormState>) => {
         const setState = {...formState, ...newState};
--- a/src/pages/dar_application/ProgressReportApplication.tsx
+++ b/src/pages/dar_application/ProgressReportApplication.tsx
@@ -38,6 +38,10 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true, 
         ...(dar?.presentations && {
             presentationsYesNo: (dar.presentations.length > 0)
         }),
+        // additional state for datasets section populated by useEffect
+        datasets: [],
+        datasetIds: [],
+        selectedDatasets: [],
         // additional state for dmi section
         ...(dar?.dmi?.incidents && {
             dmiYesNo: (dar.dmi.incidents.length > 0),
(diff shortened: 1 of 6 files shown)


src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -81,22 +86,20 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
     useEffect(() => {
-        const selectedIds = selectedDatasets.map((ds => ds.datasetId));
-        translateDataUseRestrictionsFromDataUseArray(selectedDatasets.map((ds) => ds.dataUse)).then((translations) => {
-            setDataUseTranslations(translations);
-        });
-        onFormChange({ datasetIds: selectedIds });
-    }, [selectedDatasets]);
+        const approvedDatasets = datasets.filter((ds) => ds.dacApproval);
+        onFormChange({ datasets: approvedDatasets });
+        onSelectedDatasetChange(approvedDatasets);
```
fboulnois: This is the core change.

rushtong: I'm pretty sure that `ds.dacApproval` means that a DAC has approved the dataset from the data submitter. Whether or not the election for this dar-dataset is approved is stored on the election final vote value. Is that the intention here?

fboulnois: Maybe I misunderstood the ticket.

rushtong: Let me add that this filter is still a good thing to check. We check this value before publishing to the dataset library so users are only applying for datasets approved by a DAC, but after some period of time, this value may have changed.

rjohanek: I think there are a few possible approaches here. my suggestion was a new endpoint to get the approved datasets for a DAR, I think Elliot pointed out a DAR Collection may have this information already I'm not sure, and I think Greg had a suggestion of modifying the DAR when a dataset is approved to store that information on the DAR as well

rushtong: We already have all of the information available in the DAR Collection. It contains all elections so we could spelunk through the json and find out if the DAR-dataset is approved or not.

rushtong (approved):
Looks great, thank you, love the tests 🎉

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -81,22 +87,21 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
-        const selectedIds = selectedDatasets.map((ds => ds.datasetId));
-        translateDataUseRestrictionsFromDataUseArray(selectedDatasets.map((ds) => ds.dataUse)).then((translations) => {
-            setDataUseTranslations(translations);
-        });
-        onFormChange({ datasetIds: selectedIds });
-    }, [selectedDatasets]);
+        const approvedDatasetIds = getApprovedElectionDatasetIds(Object.values(dar.elections));
+        const approvedDatasets = datasets.filter((ds) => ds.dacApproval && approvedDatasetIds.includes(ds.datasetId));
+        onFormChange({ datasets: approvedDatasets });
+        onSelectedDatasetChange(approvedDatasets);
```
rjohanek, line 103: do we need to updated the datasetIds and the Data Use restrictions here too? Since datasets are updating and they are dependent on the datasets

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -81,22 +87,21 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
+    }, [datasets]);
```
rjohanek, line 104: what happens if you add the missing dependencies? does it still work as expected

rjohanek (approved):
this looks good ! just a few questions

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: Yes, the method calls `onSelectedDatasetChange` which updates those items.

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: No, unfortunately it causes some sort of infinite loop.

PR #2897, opened 2025-06-04
DT-1795: Handle undefined elections case

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1795

### Summary
When a Progress Report is submitted, there are no elections yet created for it. When a researcher navigates to the "Review" page in this state, the page errors out because there are no elections on the `dar` object. 

----

- Verify all tests go green


otchet-broad (approved):

rjohanek (approved):
do we want to add the missing dependencies?

rushtong:
> do we want to add the missing dependencies?

@rjohanek - I did try doing that but that seems to lock up the page so no actions can be taken. I think this `useEffect` needs to be refactored.

PR #2903, opened 2025-06-06
[DT-1772] Add signing official to progress report closeout

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1772

### Summary

Add signing official to progress report closeout:

Screenshot 2025-06-06 at 1 37 22 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/progress_reports/DarCloseout.tsx
+++ b/src/pages/progress_reports/DarCloseout.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import {FormField, FormFieldTypes, FormValidators} from 'src/components/forms/forms';
 import {ValidFormState, FormState, FormStateKey} from 'src/pages/progress_reports/ProgressReportFormState';
 import {DarErrors, ValidationError} from "src/pages/dar_application/FormValidationState";
 import {FORM_TEXT_AREA_MAX_LENGTH} from "src/components/forms/formConstants";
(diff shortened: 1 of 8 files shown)


rushtong (approved):
👍🏽

src/pages/progress_reports/DarCloseout.tsx
```diff
@@ -15,6 +17,27 @@ interface DarCloseoutProps {
...
+        const init = async () => {
+            const signingOfficials = await User.getSOsForCurrentUser();
+            setAllSigningOfficials(signingOfficials);
+            const closeoutSigningOfficial = signingOfficials.find(so => so.userId === formState.closeoutSigningOfficial?.userId);
+            if (closeoutSigningOfficial !== undefined) {
+                setDefaultSigningOfficial(closeoutSigningOfficial);
+            }
+        }
+        init();
+    }, []);
```
snf2ye, line 39: It seems reasonable to add the suggested dependency here (formState.closeoutSigningOfficial?.userId) - From some local testing, it doesn't seem to create an infinite loop.

snf2ye (approved):
One suggestion, but otherwise LGTM

src/pages/progress_reports/DarCloseout.tsx
fboulnois: I think we only want to run this once on component initialization. If we add the suggested dependency, it will get rerun any time a signing official is selected, which will set the `userId`.

PR #2908, opened 2025-06-12
[DT-1767][DT-1768] Add closeout review workflow in UI

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1767

### Summary

Add closeout review frame, buttons, and logic:

Screenshot 2025-06-12 at 9 35 35 AM

----

- Verify all tests go green


src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -115,6 +116,19 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
         onFormChange({ selectedDatasets: newDatasets, datasetIds: newDatasetIds });
     }
 
+    // Check if the DAR is a closeout review
+    // TODO: modify this logic for DAC chair when backend supports it
+    const isCloseoutReview = () => {
+        const user = Storage.getCurrentUser();
+        const isSigningOfficial = user.roles.some((role: UserRole) => role.name === 'SigningOfficial');
+        const isDacChair = user.roles.some((role: UserRole) => role.name === 'Chairperson');
```
rushtong: Note that this should return an instance of `DuosUser` which has all the user roles pre-calculated. For example, this is what I see in local storage for `Storage.getCurrentUser()`

Screenshot 2025-06-13 at 7 50 10 AM

src/types/model.ts
```diff
@@ -399,6 +399,8 @@ export interface DataAccessRequest {
   researchPlans: string;
   closeoutSupplement: Closeout;
   parentId?: number;
+  closeoutSigningOfficialApprovedDate: number;
```
rushtong, line 404: Can this be a `Date` type? I see mixed types for dates in this file so this could use some cleanup.

rushtong (approved):
👍🏽  Looks good, see comments inline

src/types/model.ts
fboulnois: I'm not sure since we are reporting a Unix timestamp I believe.

fboulnois: I don't think it is safe to switch this to a Date.

MatthewBemis (approved):

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -138,6 +142,35 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
+        try {
+            if (user.isSigningOfficial && !isCloseoutApproved) {
+                await DAR.approveCloseout(dar.referenceId);
+            } else {
+                const acknowledgement = 'dar_closeout_chair_ref_' + dar.referenceId;
+                await User.acceptAcknowledgments(acknowledgement);
+            }
+            Notifications.showSuccess({ text: 'Closeout review approved successfully.' });
+        } catch (e) {
+            const err = e as AxiosError<Record<string, string>>;
```
fboulnois, line 168: According to https://dev.to/mdmostafizurrahaman/handle-axios-error-in-typescript-4mf9 , this is a mechanism for handling Axios errors in `try-catch` statements.

rushtong: Nice! I had tried to do something similar [here](https://github.com/DataBiosphere/duos-ui/blob/develop/src/utils/ErrorUtils.ts). It would be nice to standardize this so we can be more consistent in error handling.

rushtong (approved):
👍🏽

MatthewBemis (approved):
👍

PR #2909, opened 2025-06-12
Populate "no" responses in read-only progress report

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1783

### Summary
This ensures that Yes / No fields in the Progress Report are populated with "No" when the submitter had supplied those values.  Corresponding properties in the API response are undefined, which had left them unchecked in read-only progress reports.

If the Progress Report is in edit mode, then undefined API response properties are not checked with "No" -- the Yes / No fields are unchecked as expected.

Here's how it looks.

Progress Report in read-only mode, showing checked "No" fields:

https://github.com/user-attachments/assets/13f88e2f-c6e7-4780-b552-4fa70e0ebae0

Progress Report in edit mode (i.e. `readOnly: false`), showing unchecked Yes / No fields:

https://github.com/user-attachments/assets/b30f2d06-0c3b-4438-8fd1-0362635722cd


src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -44,32 +44,24 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
...
-        }),
-        ...(dar?.publications && {
-            publicationsYesNo: (dar.publications.length > 0)
-        }),
-        ...(dar?.presentations && {
-            presentationsYesNo: (dar.presentations.length > 0)
-        }),
+        intellectualPropertyYesNo: !!dar.intellectualPropertySummary,
+        publicationsYesNo: (dar.publications?.length > 0),
+        presentationsYesNo: (dar.presentations?.length > 0),
```
fboulnois: I think all these states should be set but **only** in the read only view. In the create progress report view they should be unselected.

fboulnois (commented):
One comment, otherwise I appreciate adding the tests for `ProgressReportApplication.tsx`!

src/pages/dar_application/ProgressReportApplication.tsx
fboulnois: I believe the way the code is evaluating this condition will set these to `false`, because if `dar.publications?.length` evaluates to `undefined`, then the entire condition will evaluate to `undefined > 0`, which is `false`.

rushtong (approved):
Verified that the yes/no values are populated correctly in the read-only state. One thing I do see is that if the values are `Yes`, the content of the publications/presentations blocks are not populated. That may be in a different ticket, but I wanted to call that out.

Screenshot 2025-06-13 at 8 05 44 AM

eweitz:
Thanks Greg, yes, I also see this [on dev](https://duos-k8s.dsde-dev.broadinstitute.org/admin_review_collection/622).  I picked up DT-1817 to address that.

src/pages/dar_application/ProgressReportApplication.tsx
eweitz: Good catch!  Fixed.  I've also updated the test suite and demo videos above.

rjohanek (approved):

fboulnois (approved):
👍

PR #2912, opened 2025-06-12
DT-1829: Fix Dataset types

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1829

### Summary
Aligns the dataset type with what is really coming from Consent
See also: https://github.com/DataBiosphere/consent/pull/2566

----

- Verify all tests go green

### DIFF
--- a/src/types/model.ts
+++ b/src/types/model.ts
@@ -1,4 +1,3 @@
-import { DuosUserResponse } from 'src/types/responseTypes';
 import externalAccessIcon from 'src/images/external_access.svg';
 import openAccessIcon from 'src/images/open_access.svg';
 import controlledAccessIcon from 'src/images/controlled_access.svg';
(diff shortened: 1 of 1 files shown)


rjohanek (approved):
does this affect any usages? is there anywhere datasetName is being used that we want to update?

rushtong:
> does this affect any usages? is there anywhere datasetName is being used that we want to update?

There are no downstream usages here - I checked all of the usages of this field on this type.

src/types/model.ts
```diff
@@ -117,23 +116,26 @@ export interface Institution {
 
 export interface Dataset {
   name: string;
-  datasetName: string;
+  // datasetName is deprecated, use name instead
```
fboulnois:
```suggestion
  // @deprecated datasetName is deprecated, use name instead
```

I think you can use the `@deprecated` attribute: https://stackoverflow.com/a/62991642

fboulnois (approved):
thank you 👍

src/types/model.ts
rushtong: Nice!

PR #2913, opened 2025-06-13
[DT-1817] Show publications and presentations in read-only progress report

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1817

### Summary
Previously, summaries of any publications or presentations added in a progress report were not shown when the progress report was in read-only mode.  Now, any such publications or presentations are shown.

Here's how it looks.

Before: presentation and publication added, but not shown.  (The oddly unshown datasets seem to be a manifestation of [DT-1818](https://broadworkbench.atlassian.net/browse/DT-1818), a bug ticket I just picked up.)

Publications_and_presentations_added_but_not_shown__DUOS-UI_2025-06-13

After: presentation and publication added and shown

Publications_and_presentations_added_and_shown__DUOS-UI_2025-06-13

[DT-1818]: https://broadworkbench.atlassian.net/browse/DT-1818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rushtong (approved):
Nice! 👍🏽

MatthewBemis (approved):
👍

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -41,8 +41,8 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
         closeoutProjectTransferred: false,
         closeoutProjectSuperseded: false,
         closeoutOther: false,
-        publications: [],
-        presentations: [],
+        publications: dar.publications,
+        presentations: dar.presentations,
```
fboulnois: I think this may not even be needed, because the initial state is first populated with `...dar`, so these fields should become available (as long as the model is the same).

eweitz: Good point, I removed these unneeded assignments and confirmed the fields still display as expected.

PR #2920, opened 2025-06-18
[DT-1812] Make PI/Submitter the same value 

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1812

### Summary
When the page is in edit mode, change the pi name and email field to be read only and populate it with the researcher's information. Remove validation from these fields as they cannot be edited by users. 
In read only mode, the form will show whatever was submitted as the pi name and email even if it was different from the researcher. 
Adds unit tests.

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/DataAccessRequestApplication.jsx
+++ b/src/pages/dar_application/DataAccessRequestApplication.jsx
@@ -273,12 +273,14 @@ const DataAccessRequestApplication = (props) => {
     }
 
     formData.researcher = isNil(researcher) ? '' : researcher.displayName;
+    formData.piName = props.existingDarsReadOnlyMode ? formData.piName : formData.researcher;
+    formData.piEmail = props.existingDarsReadOnlyMode ? formData.piEmail : (isNil(researcher) ? '' : researcher.email);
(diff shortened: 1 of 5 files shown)


eweitz (approved):

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -277,12 +277,14 @@ const DataAccessRequestApplication = (props) => {
     }
 
     formData.researcher = isNil(researcher) ? '' : researcher.displayName;
+    formData.piName = props.existingDarsReadOnlyMode ? formData.piName : formData.researcher;
+    formData.piEmail = props.existingDarsReadOnlyMode ? formData.piEmail : (isNil(researcher) ? '' : researcher.email);
```
fboulnois: Can we pull `existingDarsReadOnlyMode` out of `props` directly instead? It makes it easier to track use inside components.

fboulnois (commented):
See below:

fboulnois (approved):
👍

PR #2921, opened 2025-06-18
[DT-1609] [DT-1851] Add support for bulk-issuing library cards

MatthewBemis:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1609
https://broadworkbench.atlassian.net/browse/DT-1851

### Summary

Adds frontend support for issuing library cards in bulk, accessible through the SO Console and the Admin Console. Since the number of library cards being issued in one go is expected to be small, this was only implemented on the frontend. But a future optimization could be to implement a bulk version of the existing createLibraryCard endpoint.

The two videos are pretty much the same because they use the same modal, but here's both of them in action:

SO Console:

https://github.com/user-attachments/assets/7ef9c794-9e40-44ff-83fe-badb27c57cc5

Admin Console:

https://github.com/user-attachments/assets/da35d218-717d-46d2-bf22-e71db1d03a59

----

- Verify all tests go green

### DIFF
--- a/src/components/library_card_table/LibraryCardTable.tsx
+++ b/src/components/library_card_table/LibraryCardTable.tsx
@@ -22,7 +22,6 @@ import TableIconButton from 'src/components/TableIconButton';
 import {AxiosError} from 'axios';
 import {ConsentError} from 'src/types/responseTypes';
 import {LibraryCard} from 'src/types/model';
-import {extractError} from 'src/utils/ErrorUtils';
 import { processLibraryCards } from 'src/utils/LibraryCardUtils';
 
 interface UserData {
--- a/src/components/library_card_table/LibraryCardTable.tsx
+++ b/src/components/library_card_table/LibraryCardTable.tsx
@@ -22,7 +22,7 @@ import TableIconButton from 'src/components/TableIconButton';
 import {AxiosError} from 'axios';
 import {ConsentError} from 'src/types/responseTypes';
 import {LibraryCard} from 'src/types/model';
-import {extractError} from "src/utils/ErrorUtils";
+import {extractError} from 'src/utils/ErrorUtils';
 
 interface UserData {
   userId: number;
@@ -306,7 +306,7 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
     if (hasDuplicateCard) {
       Notifications.showError({text: 'One or more of the library cards already exist.'});
     } else {
-      // Execute library card update with payload, get the updated card, and
+      // Execute library card updates with payload, get the updated card, and
       // add (with sort afterwards) library card to libraryCards (reference list)
       const successfulCards = [];
       const failedCards = [];
@@ -324,16 +324,16 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
       const updatedList = cloneDeep(libraryCards);
       libraryCards.push(...successfulCards);
 
-      updatedList.sort((a: LibraryCard, b: LibraryCard) => {
-        const dateA = new Date(a.createDate ?? '');
-        const dateB = new Date(b.createDate ?? '');
-        return dateB.getTime() - dateA.getTime();
+      updatedList.sort((a, b) => {
+        return dayjs(b.createDate).valueOf() - dayjs(a.createDate).valueOf();
       });
+
       setLibraryCards(updatedList);
       setShowModal(false);
+
       if(failedCards.length > 0) {
-        const errorMessages = failedCards.map(failure => `${failure.card.userEmail}: ${failure.error}`).join('\n');
-        Notifications.showError({text: `Failed to create the following library cards:\n${errorMessages}`});
+        const errorMessages = failedCards.map(failure => `${failure.error}`).join(', ');
+        Notifications.showError({text: `${errorMessages}`});
       }
     }
   };
@@ -386,7 +386,7 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
                       flexDirection: 'row',
                       alignItems: 'flex-end',
                       justifyContent: 'center',
-                      width: '300px'
+                      width: '380px'
                     }}
                 >
                   <button
@@ -406,7 +406,7 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
                           )
                       }
                   >
-                    <span>Add Library Card</span>
+                    <span>Bulk Issue / Add Users</span>
                   </button>
                 </div>
               }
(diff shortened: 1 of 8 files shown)


src/pages/signing_official_console/SigningOfficialTable.jsx
```diff
@@ -426,10 +444,10 @@ export default function SigningOfficialTable(props) {
       />
       <LibraryCardFormModal
         showModal={showModal}
-        createOnClick={(card) => issueLibraryCard(card, researchers)}
+        createOnClick={(cards) => issueLibraryCards(cards, researchers)}
         closeModal={() => setShowModal(false)}
         card={selectedCard}
-        users={onlyResearchersWithoutCardFilter(researchers)}
+        users={researchers.filter(onlyResearchersWithoutCardFilter)}
```
MatthewBemis, line 437: This line specifically addresses https://broadworkbench.atlassian.net/browse/DT-1851

src/components/library_card_table/LibraryCardTable.tsx
```diff
@@ -294,35 +295,51 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
...
+      const failedCards = [];
+      for (const card of newLibraryCards) {
+        try {
+          const newCard = await LibraryCardAPI.createLibraryCard(card);
+          successfulCards.push(newCard);
+        } catch (error: unknown) {
+          const errorMessage = extractError(error);
+          failedCards.push({ card, error: errorMessage });
+        }
       }
```
MatthewBemis: I could rewrite this to just fire these off all at once and then sort through the results once they're all complete, as long as there aren't any performance concerns with the underlying API?

Doing them serially doesn't seem slow at all from my testing, but happy to refactor this if anyone reviewing has a preference.

src/components/library_card_table/LibraryCardTable.tsx
```diff
@@ -294,35 +295,51 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
...
-        });
-        setLibraryCards(updatedList);
-        setShowModal(false);
+  const addLibraryCards = async (newLibraryCards: LibraryCard[]): Promise<void> => {
+    // Check if any cards already exist, show error if any do
+    const duplicateLibraryCards = newLibraryCards.filter((card) => {
+        return findIndex(
+            (element: LibraryCard) => isEqual(element.userEmail)(card.userEmail),
+            libraryCards
+        ) > -1;
```
fboulnois: Consider replacing this `lodash/fp` functionality with the Javascript built-ins: `lodash/fp` doesn't work well with Typescript.

src/components/library_card_table/LibraryCardTable.tsx
fboulnois: This initial approach seems fine to me. I think the long term solution is probably a bulk-upload endpoint instead.

fboulnois (approved):
looks good, a few suggestions:

src/components/library_card_table/LibraryCardTable.tsx
fboulnois: We've also been moving away from `lodash/fp` since it is abandoned.

src/components/library_card_table/LibraryCardTable.tsx
s-rubenstein: I think it is totally fine to do this for now, but I would display a loading spinner or disable the button to provide some immediate feedback. There is a small delay between clicking the button and the completion, based on the video

src/components/library_card_table/LibraryCardTable.tsx
```diff
@@ -410,7 +427,7 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
         />
         <LibraryCardFormModal
             showModal={showModal}
-            createOnClick={addLibraryCard}
+            createOnClick={(cards) => addLibraryCards(cards)}
```
s-rubenstein: I think you can still just use a method reference here, rather than a lambda

src/pages/signing_official_console/SigningOfficialTable.jsx
```diff
@@ -313,37 +314,54 @@ export default function SigningOfficialTable(props) {
...
 
-  const issueLibraryCard = async (selectedCard, researchers) => {
-    let messageName;
-    try {
+  const issueLibraryCards = async (cards, researchers) => {
+    const successfulCards = [];
+    const failedCards = [];
+
+    // Process each card individually
+    for (const card of cards) {
```
s-rubenstein: Given we do this multiple times, it might be worth making a util function?

src/components/library_card_table/LibraryCardTable.tsx
rushtong: Agree with Sky here. We do have something a bit old that I needed to ensure serial promise execution in [util.js](https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/utils.js#L200) but that might be more than what you want.

rjohanek (commented):
I'm wondering about the text on the button. bulk issue / add users seems unnecessarily wordy. Could we just say issue library cards or something instead of two things with a slash? I know this text was specified in the ticket but maybe @lstrano-broad could weigh in? Does the button display need to change at all? Maybe we could just stick with Add?

src/components/modals/LibraryCardFormModal.tsx
```diff
@@ -18,15 +19,14 @@ interface UserOption {
...
   dropdownOptions: UserOption[];
-  updateUser: (value: SingleValue<UserOption>) => void;
-  setCard: React.Dispatch<React.SetStateAction<LibraryCard>>;
+  updateUsers: (values: MultiValue<UserOption>) => void;
 }
 
 export interface LibraryCardFormModalProps {
   showModal: boolean;
-  createOnClick: (card: LibraryCard) => void;
+  createOnClick: (cards: LibraryCard[]) => void;
```
otchet-broad:
```suggestion
  createOnClick: (cards: LibraryCard[]) => Promise<never>;
```

otchet-broad: This might address some of the Sonar noise you're seeing.

src/components/modals/LibraryCardFormModal.tsx
```diff
@@ -18,15 +19,14 @@ interface UserOption {
...
   closeModal: () => void;
   users: UserOption[];
   card: LibraryCard;
```
otchet-broad: I realize it's not part of your change set, but the card property looks like it's safe to remove.  The test doesn't really use it either - seems to build an empty object.

otchet-broad:
Nice iteration through the requirements and appreciate the videos in the pull request description.
Added some minor feedback about Sonar and some ways to tame the issues it finds.  Address as you see fit.

rushtong (approved):
Looks reasonable to me. Elliot's suggestions for addressing the sonar warnings look like the right approach.

otchet-broad (approved):

PR #2926, opened 2025-06-23
DT-1866: Update DAR models used in Progress Reports and Dataset Statistics pages

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1866

### Summary
This PR updates the underlying DAR models to be more explicit about what it is comprised of. Previously, the Progress Report Application component was populated with either a [DataAccessRequest](https://github.com/DataBiosphere/consent/blob/412d337fcd019cbcebf94cc4f15907747ba53b09/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java#L29) or a [DataAccessRequestData](https://github.com/DataBiosphere/consent/blob/412d337fcd019cbcebf94cc4f15907747ba53b09/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java#L16) object depending on what context it was used in. Inside the component, both items were spread-assigned into the same `dar` object. 

In this approach, we now merge the two objects into a combined DAR model before passing it down to `ProgressReportApplication`.  This allows us to move the `DataAccessRequestData` fields out of the DAR model and into a new, combined model, which should be easier to grok for future developers.

As a side effect of this, the `DatasetStatistics` component became out of date so this PR updates that component as well.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
+++ b/cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
@@ -1,7 +1,7 @@
 import React, { ReactNode } from 'react';
 import { mount } from 'cypress/react';
 import { ProgressReportApplication } from 'src/pages/dar_application/ProgressReportApplication';
-import { DataAccessRequest, Dataset, DuosUser, FileStorageObject } from 'src/types/model';
+import {CombinedDataAccessRequest, Dataset, DuosUser, FileStorageObject} from 'src/types/model';
 import { History, Location, Action } from 'history';
 import { Storage } from 'src/libs/storage';
 
@@ -153,7 +153,7 @@ describe('ProgressReportApplication - Component Tests', () => {
     }
   ];
 
-  const baseDar: Partial<DataAccessRequest> = {
+  const baseDar: Partial<CombinedDataAccessRequest> = {
     userId: 1,
     projectTitle: 'Test Project',
     draft: false,
@@ -161,15 +161,14 @@ describe('ProgressReportApplication - Component Tests', () => {
     referenceId: 'DAR-123',
     collectionId: 1,
     elections: {},
-    darCode: 'DAR-123',
-    createDate: '2023-10-01T00:00:00Z',
-    sortDate: '2023-10-01T00:00:00Z',
-    submissionDate: '2023-10-01T00:00:00Z',
-    updateDate: '2023-10-01T00:00:00Z'
+    createDate: 1748736000,
+    sortDate: 1748736000,
+    submissionDate: 1748736000,
+    updateDate: 1748736000
   };
 
-  const mountComponent = (dar: Partial<DataAccessRequest> = {}, readOnly = true) => {
-    const fullDar = { ...baseDar, ...dar } as DataAccessRequest;
+  const mountComponent = (dar: Partial<CombinedDataAccessRequest> = {}, readOnly = true) => {
+    const fullDar = { ...baseDar, ...dar } as CombinedDataAccessRequest;
 
     const props = {
       dar: fullDar,
@@ -177,7 +176,8 @@ describe('ProgressReportApplication - Component Tests', () => {
       readOnlyMode: readOnly,
       history: mockHistory,
       location,
-      researcher
+      researcher,
+      countriesOfOperation: []
     };
 
     return mount(<ProgressReportApplication {...props} /> as ReactNode);
--- a/cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
+++ b/cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
@@ -572,7 +572,7 @@ describe('ProgressReportApplication - Component Tests', () => {
     };
 
     // Mount component with datasets and elections
-    const fullDar = { ...baseDar, ...darWithElections } as unknown as DataAccessRequest;
+    const fullDar = { ...baseDar, ...darWithElections } as unknown as CombinedDataAccessRequest;
     const props = {
       dar: fullDar,
       datasets: testDatasets,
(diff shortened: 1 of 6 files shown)


cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
```diff
@@ -153,31 +153,31 @@ describe('ProgressReportApplication - Component Tests', () => {
...
-  const baseDar: Partial<DataAccessRequest> = {
+  const baseDar: Partial<CombinedDataAccessRequest> = {
     userId: 1,
     projectTitle: 'Test Project',
     draft: false,
     datasetIds: [1],
     referenceId: 'DAR-123',
     collectionId: 1,
     elections: {},
-    darCode: 'DAR-123',
```
rushtong, line 164: Note that `darCode` does not exist on a DAR, but instead on the DarCollection object.
Update: It also exists on the data field of `DataAccessRequest`

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -30,7 +30,7 @@ import {ConditionalAccordion} from 'src/components/forms/ConditionalAccordion.js
 import {ProgressReportApplication} from 'src/pages/dar_application/ProgressReportApplication';
 import {ScrollableTabs} from 'src/pages/dar_application/ScrollableTabs';
 import {validateDARFormData, validationFailed} from 'src/utils/darFormUtils.js';
-import {isArray, set} from 'lodash';
+import {assign, cloneDeep, head, isArray, isEmpty, isNil, isString, keys, merge, set} from 'lodash';
```
rushtong, line 33: Note: I only needed to add `merge` here, but it was a no-op to replace many of the fp-based functions with plain lodash versions. Replacing `get` and `map` will require larger refactoring that I didn't want to tackle in this PR.

src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -52,8 +52,8 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
           ? {
             // In read-only mode, check "No" when undefined
             intellectualPropertyYesNo: !!dar.intellectualPropertySummary,
-            publicationsYesNo: (dar.publications?.length > 0),
-            presentationsYesNo: (dar.presentations?.length > 0),
+            publicationsYesNo: ((dar.publications?.length ?? 0) > 0),
```
eweitz, line 55: This is functionally equivalent to `(dar.publications?.length > 0)`, right?  The former style seems a bit visually cleaner, but perhaps there's a case I'm overlooking.

src/pages/DatasetStatistics.tsx
```diff
@@ -22,8 +21,8 @@ enum AccessManagement {
 
 
 interface DatasetStatisticsProps {
-  history: History,
-  match: {
+  readonly history: History,
+  readonly match: {
```
eweitz, line 25: I wasn't aware of the `readonly` keyword!  Neat.  Like `const` but for properties in interfaces or classes.

eweitz (approved):

src/types/model.ts
```diff
@@ -331,23 +331,25 @@ export interface Acknowledgement {
...
-  elections: Record<number, Election>;
+  nonTechRus: string;
+  referenceId: string;
+}
+
+/**
+ * This interface combines the DataAccessRequest with all fields from the DataAccessRequestData object stored
+ * in DataAccessRequest.data field. This model simplifies usages in ProgressReport forms.
+ */
+export interface CombinedDataAccessRequest extends DataAccessRequest {
```
otchet-broad, line 350: Nice!  Thanks for the comment!

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: It is similar, but linting complains due to the change of `publications` being an optional type now.

src/pages/DatasetStatistics.tsx
rushtong: This was added due to linting advice

src/pages/dar_application/ProgressReportApplication.tsx
rushtong: Here is what the linter complains about:

Screenshot 2025-06-23 at 5 04 40 PM

src/pages/dar_application/DataAccessRequestApplication.jsx
```diff
@@ -584,7 +584,7 @@ const DataAccessRequestApplication = (props) => {
                     <ProgressReportApplication
                       readOnlyMode={false}
                       datasets={datasets}
-                      dar={reverseOrderedDARs[0]}
+                      dar={merge(reverseOrderedDARs[0]?.data, reverseOrderedDARs[0])}
```
otchet-broad, line 587: I like the preservation of the outer object's (second parameter) values as truth!  I think that's the right call.

rushtong: Thank you - I did mean to comment on that detail.

otchet-broad (approved):
Thanks for the change.  This should also fix an issue with DarCloseout.tsx that became visible to me this afternoon!

src/pages/dar_application/ProgressReportApplication.tsx
eweitz: Ah, thanks!  Given that, I agree this slightly complex expression is the best call for now.

The expression evaluates to false even if `publications` is undefined (`foo = {bar: 1}; foo.publications?.length > 0; // false`).  Ideally we'd be able to suppress with by modifying TypeScript configuration to _selectively_ ignore this scenario with optional chaining (foo?.length) in boolean expressions.  But I doubt we can, so the approach here seems pragmatic.

PR #2930, opened 2025-06-25
[DT-1808] Fix broken links in current NIH Library Card Agreement

rjohanek:
### Addresses
Ticket: https://broadworkbench.atlassian.net/browse/DT-1808

### Summary
Make new pdf with links and other minor formatting updates
Upload pdf and replace all references to it

### Proof the links work now

https://drive.google.com/file/d/1Rw9athdySfygbx4355bKthYPu8Psomc7/view?usp=sharing 

----

- Verify all tests go green


MatthewBemis (approved):
👍

src/components/SigningOfficialDaaAgreementWrapper.jsx
```diff
@@ -1,6 +1,6 @@
 import { Notifications } from '../libs/utils';
 import BroadLibraryCardAgreementLink from '../assets/Library_Card_Agreement_2023_ApplicationVersion.pdf';
-import NihLibraryCardAgreementLink from '../assets/NIHLibraryCardAgreement08012024.pdf';
+import NihLibraryCardAgreementLink from '../assets/NIHLibraryCardAgreement06252025.pdf';
```
rushtong: If we're touching this file, it would be nice to update the imports

src/components/LibraryCardAgreementTermsDownload.jsx
```diff
@@ -3,7 +3,7 @@ import {
   NIHDataUseCertificationAgreement
 } from '../components/external_docs/NIHDataUseCertificationAgreement';
 import NihLibraryCardAgreementLink
-  from '../assets/NIHLibraryCardAgreement08012024.pdf';
+  from '../assets/NIHLibraryCardAgreement06252025.pdf';
```
rushtong: Similar feedback, maybe we can update all of these too.

src/components/SigningOfficialDaaAgreementWrapper.jsx
rushtong, line 1: It doesn't look like we're using this component any more - I can't find any usages. To test, I went into my account as an SO and deleted all of my acknowledgements. Previously, this page would load and prompt the user to acknowledge these documents. It seems like we've refactored that away completely. We should confer with product to ensure that we aren't losing any expected functionality here.

rushtong (approved):
Looks good to me. I think, depending on product's requirements, we might be able to remove the wrapper component.

src/components/SigningOfficialDaaAgreementWrapper.jsx
rjohanek: oh okay I'll check it out but I am wondering will we need this when we start using DAAs?

src/components/SigningOfficialDaaAgreementWrapper.jsx
rjohanek: update them to reference src directly?

src/components/SigningOfficialDaaAgreementWrapper.jsx
rjohanek: It looks like its still referenced in routes

rjohanek: but I agree I'm not seeing it on the so console dar request page which I think it is supposed to be on, maybe this is a bug?

fboulnois (approved):
looks good, thanks for also cleaning up the imports 👍

PR #2933, opened 2025-06-26
DT-1859: Show Chairperson Closeout Approval

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1859

## Summary
This PR populates user acknowledgements and looks for one that the user has approved for the current DAR closeout. If populated, we hide the "Approve" button. Note that this PR does NOT modify the user's console where they get an option to "Review Closeout". That should be handled by a back-end change to the available actions on the `DarCollectionSummary` object for which a closeout exists.

### Before Approval
Screenshot 2025-06-26 at 12 44 09 PM

### After Approval
Screenshot 2025-07-01 at 12 39 23 PM

----

- Verify all tests go green


src/pages/dar_application/ProgressReportApplication.tsx
```diff
@@ -200,6 +201,21 @@ export const ProgressReportApplication = ({ dar, datasets, readOnlyMode = true,
       onSelectedDatasetChange(approvedDatasets);
     }, [datasets]);
 
+    // Required to get Chairperson acknowledgments of closeouts
+    useEffect(() => {
+      async function fetchAcknowledgements() {
+        if (readOnlyMode) {
+          const acknowledgementMap: AcknowledgementMap = await User.getAcknowledgements();
```
otchet-broad: You can `get` a single acknowledgement:
/api/user/acknowledgements/{key}

You know the key already:
dar_closeout_chair_ref_'${dar.referenceId}'

rushtong: Great point, will refactor this call.

otchet-broad (commented):
Looks good and I appreciate the tests.

I left you one thought about the acknowledgement - you can target the request so you don't have to iterate through the list.

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -1,113 +1,176 @@
...
-        <div className="progress-report-step-card" style={{
-            border: '1px solid #4D72AA',
+
+  const [acknowledgement, setAcknowledgement] = useState<Acknowledgement | undefined>(undefined);
+
+  // Required to get Chairperson acknowledgments of closeouts
+  useEffect(() => {
+    // Fetch the acknowledgement for the given referenceId
+    const fetchAcknowledgement = async () => {
+      if (referenceId) {
```
otchet-broad, line 26: Does it make sense to use a promise pattern here?

Have getAcknowledgement return a promise-like response and handle the catch with a standard error object?

pseudocode:
User.getAcknowledgement(myKey).then(response => {
    setAcknowledgement(response.data)
}).catch( error => {
if (error.response && error.response.status != 404) {
Notifications.showError({...fill in the error to display})
}
});

rushtong: Looking at our current API classes, we generally don't do that. I'm hesitant to start using a different pattern here without having the team comfortable with going in a different direction.

otchet-broad (approved):

otchet-broad:
@rushtong - I went looking through code to see if we were using that pattern after I wrote the comment, realizing we didn't, I removed it and then your comment showed up.
LGTM!

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -1,113 +1,176 @@
...
+import {User} from 'src/libs/ajax/User';
+import {Notifications} from 'src/libs/utils';
+import {extractConsentError, extractError} from 'src/utils/ErrorUtils';
 
 interface CloseoutReviewProps {
-    onApprove?: () => void;
-    onReturn?: () => void;
+  onApprove?: () => void;
+  onReturn?: () => void;
+  referenceId?: string;
```
fboulnois: How can `referenceId` be optional? You are always passing it in the `CloseoutReview`.

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -1,113 +1,176 @@
...
-                            lineHeight: '1.4'
-                        }}>
-                            If there are issues with the content in this closeout report, please contact the researcher.
-                        </p>
-                    </div>
+  const approvedButton =
+      (acknowledgement: Acknowledgement) => {
+        let approvalDate = '';
+        if (acknowledgement.lastAcknowledged) {
+          approvalDate += new Date(acknowledgement.lastAcknowledged).toISOString().substring(0, 10);
```
fboulnois: perhaps pull this item into a separate one line function instead of repeating the ISO conversion.

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -1,113 +1,176 @@
...
+            style={{
+              backgroundColor: '#cccccc',
+              color: 'black',
+              border: '1px solid gray',
+              borderRadius: '4px',
+              padding: '10px 20px',
+              cursor: 'default',
+              minWidth: '136px'
+            }}>
+          Approved on {approvalDate}
```
fboulnois: Should we even show the button when approved? No action can be taken.

fboulnois (commented):
A few suggestions:

src/pages/progress_reports/CloseoutReview.tsx
fboulnois: In the ticket acceptance criteria it says:

> If [the acknowledgment] exists, hide the button

src/pages/progress_reports/CloseoutReview.tsx
rushtong: Good point, this is left over from refactoring when an acknowledgement was optionally passed in - will update this.

src/pages/progress_reports/CloseoutReview.tsx
rushtong: I can remove it. The problem is that the user has a link to get to this page to review the closeout. If we hide the button, it will be confusing to see nothing at all to review. I wanted to at least show the user something rather than nothing.

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -1,113 +1,147 @@
...
-                            If there are issues with the content in this closeout report, please contact the researcher.
-                        </p>
-                    </div>
+  const [acknowledged, setAcknowledged] = useState<boolean | undefined>(undefined);
+
+  // Required to get Chairperson acknowledgments of closeouts
+  useEffect(() => {
+    // Fetch the acknowledgement for the given referenceId
+    const fetchAcknowledgement = async () => {
+      if (referenceId) {
```
fboulnois: I don't think this `if` check is required anymore since `referenceId` will always be defined.

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -1,113 +1,147 @@
...
+                color: '#333F52',
+                fontSize: '14px',
+                lineHeight: '1.4'
+              }}>
+                If there are issues with the content in this closeout report, please contact the researcher.
+              </p>
+            </div>
+
+            <div style={{display: 'flex', gap: '12px', marginLeft: '16px'}}>
+              {/* If there is a closeout acknowledgement, show the approved button with the approved date */}
```
fboulnois: Is this comment still accurate? We do _not_ show the button when there is an acknowledgement.

fboulnois (commented):
Additional comments:

eweitz (approved):

PR #2940, opened 2025-06-27
[DT-1907] Read only NHGRI datasets

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1907

### Summary

When prepared to do so, **only** revert the following commit: 2b43979324ecb54abe3ef4dd5dee30f874853691

Full information below:

## Story

**As an** NHGRI DAC chair, **I want** to temporarily prevent users from requesting datasets to my DAC, **so that** DUOS remains compliant with security requirements.

## Requirements

1. Temporarily disallowing DARs from being created on NHGRI (AnVIL) datasets in the UI
2. Datasets should not disappear from the UI, they should not be selectable instead
3. The approach should be easy to undo

## What

Due to administrative processes on the NHGRI side and to remain compliant with new requirements around controlled access datasets, we need to prevent researchers from making Data Access Requests (DARs) against NHGRI datasets.

## When

To continue being compliant, the solution needs to be implemented by Monday June 30th, 2025.

## How

To keep the implementation extremely simple, all datasets in the NHGRI DAC are switched in the UI from “controlled” access to “external” access. This will effectively prevent researchers from making requests against those datasets.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/DataSet.js
+++ b/src/libs/ajax/DataSet.js
@@ -1,8 +1,7 @@
-import { mergeAll } from 'lodash/fp';
-import { Config } from '../config';
 import axios from 'axios';
-import { getApiUrl, fetchOk } from '../ajax';
-
+import { mergeAll } from 'lodash/fp';
+import { Config } from 'src/libs/config';
+import { getApiUrl, fetchOk } from 'src/libs/ajax';
 
 export const DataSet = {
   getDatasetNames: async () => {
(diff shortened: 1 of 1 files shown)


eweitz (approved):

MatthewBemis (approved):

rushtong (approved):
Great approach 👍🏽

otchet-broad (approved):
Nice work!

PR #2943, opened 2025-06-30
[DT-1834] Dataset details page redesign

MatthewBemis:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1834

### Summary

* Redesigns the dataset details page to be more in line with the study details page
* Split out some functionality into util functions
* Added note when there are no active DARs

Before (no active DARs)
screencapture-duos-k8s-dsde-dev-broadinstitute-org-dataset-DUOS-001193-2025-06-30-09_51_21

After (no active DARs)
screencapture-local-dsde-dev-broadinstitute-org-3000-dataset-DUOS-001193-2025-06-30-10_03_45

Before (with active DARs)
screencapture-duos-k8s-dsde-dev-broadinstitute-org-dataset-DUOS-000688-2025-06-30-09_50_36

After (with active DARs)
screencapture-local-dsde-dev-broadinstitute-org-3000-dataset-DUOS-000688-2025-06-30-09_50_45

----

- Verify all tests go green


src/pages/DatasetStatistics.tsx
```diff
@@ -66,6 +83,31 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
     const init = async () => {
       try {
         const dataset: Dataset = await DataSet.getDatasetByDatasetIdentifier(datasetIdentifier);
+        const datasetTerms: DatasetTerm[] = await DataSet.searchDatasetIndex({ query: {
```
MatthewBemis: I might be missing something, but it seems like in order to populate all of the information on this page, we unfortunately need both the Dataset _and_ the DatasetTerms

src/components/dac_dataset_table/DACDatasetTableCellData.jsx
```diff
@@ -72,38 +72,9 @@ export function dataCustodianCellData({dataset, label = 'dataCustodianCellData'}
 }
 
 export function dataUseCellData({dataset, label = 'dataUseCellData', divClass = style['cell-data'], spanClass = style['data-use'], cellWidth = styles.cellWidths.dataUse, tooltipPlace = 'right'}) {
-  const codesAndDescriptions = dataset.dataUse?.primary ? dataset.dataUse.primary.map((dataUse) => {
```
MatthewBemis, line 75: split these out into DataUseUtils in order to re-use them in the DatasetStatistics component

src/pages/DatasetStatistics.tsx
```diff
@@ -113,61 +150,55 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
...
+              {getDataLocationLink(datasetTerm.dataLocation, datasetTerm.url)}
+            </LabeledSection>
+            <LabeledSection label={'Phenotype'}>
+                {datasetTerm.study?.phenotype ?? 'N/A'}
+            </LabeledSection>
+            <LabeledSection label={'Participants'}>
+                {datasetTerm.participantCount}
+            </LabeledSection>
+            <LabeledSection label={'Principal Investigator(s)'}>
+                {dataset?.study?.piName}
```
rushtong: I see this value on the term as well: `datasetTerm.study?.piName`

src/pages/DatasetStatistics.tsx
```diff
@@ -113,61 +150,55 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
...
+            </LabeledSection>
+            <LabeledSection label={'Data Custodian'}>
+              {extractStudyProp('dataCustodianEmail') ?? datasetTerm.createUserDisplayName}
+            </LabeledSection>
+            <div style={{paddingTop: '20px'}}>
+              {dataset?.study?.description ?? 'N/A'}
```
rushtong: Same here: `datasetTerm.study?.description`

src/pages/DatasetStatistics.tsx
rushtong: If you're still missing some required details, let's discuss. We can easily add new values to the `DatasetTerm` object that is indexed in ES.

MatthewBemis: Thanks Greg, looks like this was a false alarm from me. All of the fields I thought I was missing were available on the `study` property as you pointed out 🙂

src/pages/DatasetStatistics.tsx
```diff
@@ -65,10 +79,32 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
...
+          'bool': {
+            'must': [
+              {
+                'match': {
+                  '_type': 'dataset'
+                }
+              },
+              {
+                'match': {
+                  'datasetIdentifier': datasetIdentifier
```
MatthewBemis, line 92: Does this query look right? It returns the dataset that I expect, but also returns a few others in the list, hence the additional filter below. Would prefer to clean that up, if it's possible to do it server-side.

For example, I see this happen with the identifier `DUOS-001193`

rushtong: You should be able to match on the ID itself ... see the example in the [API docs](https://consent.dsde-prod.broadinstitute.org/#/Dataset/post_api_dataset_search_index). If that isn't available, try this exact match query:
```
{
  "query": {
    "match_phrase": {
      "datasetIdentifier": "DUOS-001193"
    }
  }
}
```

MatthewBemis: Thanks, that match query solves the issue!

For context, the numerical ID isn't available in this context because we're pulling the identifier out of the URL path, which uses the "DUOS-001193"-style identifiers instead.

rushtong: Another alternative would be to traverse the list of dataset terms returned from the original query and look for the one that in the list. Definitely sub-optimal, but it is a feasible solution.

rushtong (approved):
Thank you for the additional tests!

cypress/component/Dataset/datasetTerm.json
```diff
@@ -0,0 +1,48 @@
+{
+  "datasetId": 1975,
+  "name": "ExteranlAccessTestJL1",
+  "datasetName": "ExteranlAccessTestJL1",
```
eweitz: For higher fidelity test data where these values match the corresponding value in study.studyName in line 34:

```suggestion
  "name": "ExternalAccessTestJL1",
  "datasetName": "ExternalAccessTestJL1",
```

src/pages/DatasetStatistics.tsx
```diff
@@ -113,61 +136,55 @@ export default function DatasetStatistics(props: DatasetStatisticsProps) {
...
+              {getDataLocationLink(datasetTerm.dataLocation, datasetTerm.url)}
+            </LabeledField>
+            <LabeledField label={'Phenotype'}>
+                {datasetTerm.study?.phenotype ?? 'N/A'}
+            </LabeledField>
+            <LabeledField label={'Participants'}>
+                {datasetTerm.participantCount}
+            </LabeledField>
+            <LabeledField label={'Principal Investigator(s)'}>
+                {datasetTerm.study.piName}
```
eweitz: The "(s)" entails a study can have more than one PI.  Is that indeed possible?

If so, then the template value (right now an always-singular `datasetTerm.study.piName`) should presumably account for possibly having multiple PIs.  

If not -- which AIUI is the case -- then `Principal Investigator(s)` should be `Principal Investigator`.

```suggestion
            <LabeledField label={'Principal Investigator'}>
                {datasetTerm.study.piName}
```

eweitz (approved):
Looks good overall!  I suggest two non-blocking refinements.  

As a later improvement, we might want to consider using a two-column layout like before.  That would reduce need to scroll in more viewports, and make information more glanceable.  Even without that, this is a nice enhancement as is!

PR #2950, opened 2025-07-01
DT-1874: Ensure that missing form values are propagated to the post submission

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1874

### Summary
This PR ensures that the data use acknowledgement and IRB document form fields are propagated from the parent to the submitted Progress Report POST body. It also fixes some typing issues discovered along the way.

Here's an example of where the data use acknowledgements have been preserved across the original DAR, the first progress report, and now in the application for the second progress report.

Screenshot 2025-07-01 at 12 10 58 PM

----

- Verify all tests go green


s-rubenstein (commented):
Can we add a test to ensure that we don't make a regression later?

rushtong:
> Can we add a test to ensure that we don't make a regression later?

I was trying to think of a way to do that, but have yet to come up with something reasonable. It's not that we have a display issue, it's more of an underlying data population issue.

s-rubenstein:
> > Can we add a test to ensure that we don't make a regression later?
> 
> I was trying to think of a way to do that, but have yet to come up with something reasonable. It's not that we have a display issue, it's more of an underlying data population issue.

Could you add a test which makes an edit or calls an onChange, and then clicks the submit button, and ensures that the mocked API call is called with the right parameters?

rushtong:
> > > Can we add a test to ensure that we don't make a regression later?
> > 
> > 
> > I was trying to think of a way to do that, but have yet to come up with something reasonable. It's not that we have a display issue, it's more of an underlying data population issue.
> 
> Could you add a test which makes an edit or calls an onChange, and then clicks the submit button, and ensures that the mocked API call is called with the right parameters?

I took the route of extracting the population functions into a utility class so they can be tested independently from the calling component.

eweitz (approved):
Code looks good, and my manual tests passed.  Thanks for helping spot the edge case with providing a 0-byte file as an IRB document!

MatthewBemis (approved):
👍

PR #2957, opened 2025-07-10
[DT-1915] Better UX for conflicts in voting.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1915](https://broadworkbench.atlassian.net/browse/DT-1915)

### Summary

When two chair people are voting on the data access request or progress report, the experience, it can lead to a poor user experience because when one user votes, the page of the other user isn't updated.

To mitigate this poor experience, when the second user votes - a conflict error will now be thrown by the backend indicating that the election is closed.  This specific error type then causes the UI to refresh the loaded data on the page so that the user can see that the vote was already recorded.

BEFORE:

image

AFTER:

image

----

- Verify all tests go green

### DIFF
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
@@ -119,7 +119,7 @@ export default function CollectionSubmitVoteBox(props) {
       Notifications.showSuccess({text: 'Successfully updated vote'});
     } catch (error) {
       if (error && error.status === 409) {
-        Notifications.showError({text: error.response.data.message + ' Chair vote not submitted, updating chair votes'});
+        Notifications.showError({text: error.response.data.message + ' Chair vote not submitted, updating chair votes.'});
         reloadFn();
       } else {
         Notifications.showError({text: 'Error: Failed to update vote'});
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
@@ -119,7 +119,8 @@ export default function CollectionSubmitVoteBox(props) {
       Notifications.showSuccess({text: 'Successfully updated vote'});
     } catch (error) {
       if (error && error.status === 409) {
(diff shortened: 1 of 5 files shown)


src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
```diff
@@ -117,8 +117,13 @@ export default function CollectionSubmitVoteBox(props) {
         setVote(newVote);
       }
       Notifications.showSuccess({text: 'Successfully updated vote'});
-    } catch (_error) {
-      Notifications.showError({text: 'Error: Failed to update vote'});
+    } catch (error) {
+      if (error && error.status === 409) {
+        Notifications.showError({text: error.response.data.message + ' Chair vote not submitted, updating chair votes.'});
```
MatthewBemis: minor wording nit: `updating chair votes` could be a bit confusing/ambiguous to a user since it seemingly contradicts the text preceding it

I really like the wording you chose in the other error message below for rationales because it lets the user know that it's just the content on the screen that's being updated/refreshed. Maybe something like `Chair vote not submitted, updating vote display.`?

MatthewBemis (approved):
nice UX improvement! just have one minor suggestion about error message text

rushtong (approved):
Agree with Matt's comments, nice work 👍🏽

src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
otchet-broad: Thanks for the feedback.

The AC had the specific wording for what to display, and I'll always use that when provided.  The other case below wasn't provided so I played with the wording and am glad you noticed the change.  Since we've three of us on board with the change, I'll make it!

eweitz: Also, can this error occur in non-chair votes?  If so, a refinement might help:

```suggestion
        const voteKind = isChair ? 'Chair vote' : 'Vote'
        Notifications.showError({text: `${error.response.data.message} ${voteKind} not submitted, updating vote display.`});
```

eweitz (approved):
Code and screenshot look good!

PR #2959, opened 2025-07-11
[DT-1939] "Edit Institution" page

MatthewBemis:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1939

### Summary

Adds a new "Edit institution" page, accessible from the institutions list.

https://github.com/user-attachments/assets/dc2ae08a-dd4d-4a6e-ac88-5b41982d89b1

Notes:

* This needs a final UX review from Victoria and other stakeholders ✅ 
* As Elliot noted, [DT-1940](https://broadworkbench.atlassian.net/browse/DT-1940) is dessert for this ticket, so I'll have a follow-on PR to this one which will add support for creating new institutions (and re-using the same components to do so).

----

- Verify all tests go green

[DT-1940]: https://broadworkbench.atlassian.net/browse/DT-1940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/institution_table/components/InstitutionDomainEditor.tsx
+++ b/src/components/institution_table/components/InstitutionDomainEditor.tsx
@@ -36,7 +36,7 @@ export const InstitutionDomainEditor = ({ domains = [], editMode, onDomainsChang
         <div style={{paddingTop: 20}}>
             <div style={{fontWeight: 600, marginBottom: '0.5rem', fontSize: 18}}>Domains</div>
             <div>
-                {domains.map((domain, idx) => (
+                {domains.map(domain => (
                     <DomainChip
                         key={domain}
                         domain={domain}
--- a/src/components/institution_table/components/InstitutionDomainEditor.tsx
+++ b/src/components/institution_table/components/InstitutionDomainEditor.tsx
@@ -38,7 +38,7 @@ export const InstitutionDomainEditor = ({ domains = [], editMode, onDomainsChang
             <div>
                 {domains.map((domain, idx) => (
                     <DomainChip
-                        key={idx}
+                        key={domain}
                         domain={domain}
                         editMode={editMode}
                         onDelete={() => handleDomainDelete(domain)}
--- a/src/components/institution_table/components/InstitutionDomainEditor.tsx
+++ b/src/components/institution_table/components/InstitutionDomainEditor.tsx
@@ -9,6 +9,7 @@ interface DomainEditorProps {
 
 export const InstitutionDomainEditor = ({ domains = [], editMode, onDomainsChange }: DomainEditorProps) => {
     const [tempDomain, setTempDomain] = useState<string>('');
+    const [errorMessage, setErrorMessage] = useState<string>('');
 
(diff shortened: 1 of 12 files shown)


src/components/institution_table/components/InstitutionDomainEditor.tsx
```diff
@@ -0,0 +1,119 @@
...
+            const updatedDomains = domains.filter(domain => domain !== domainToDelete);
+            onDomainsChange(updatedDomains);
+        }
+    };
+
+    const handleDomainAdd = () => {
+        const trimmedDomain = tempDomain.trim();
+
+        if(allInstitutionDomains.includes(trimmedDomain)) {
+            setErrorMessage('This domain is associated with another institution');
```
otchet-broad: Nice work here.  I'm wondering if it makes sense to add the name of the institution to the message so it's clearer to the user which domain already has the value set.  Our API seems to let the user know about all of the institutions.  If it's passed back to the client, we might as well display the name.

src/types/model.ts
```diff
@@ -122,6 +122,7 @@ export interface Institution {
   updateUser: number;
   updateDate: Date;
   signingOfficials: SimplifiedDuosUser[];
+  domains?: string[];
```
otchet-broad, line 125: Is this really optional?

otchet-broad (approved):
Great overall.  Just a couple of comments that you can address as you see fit.

src/types/model.ts
MatthewBemis: Yes- there are a number of institutions returned that don't have a `domains` property at all, i.e.:

```
    {
        "id": 370,
        "name": "Bloomsburg University Of Pennsylvania",
        "createDate": "Jun 14, 2022",
        "createUserId": 3351,
        "createUser": {
            "userId": 3351,
            "email": "grushton@broadinstitute.org",
            "displayName": "Greg Rushton Admin",
            "createDate": 1452101967000,
            "emailPreference": true,
            "eraCommonsId": "grushton@broadinstitute.org"
        },
        "updateUser": {}
    },
```

I think that's how the backend handles all institutions that don't have any domains set, because I didn't notice any institutions that had a `domains` value set to an empty array.

src/components/institution_table/components/InstitutionDomainEditor.tsx
MatthewBemis: That's a great idea- I'll update this!

src/types/model.ts
rushtong: This is true. In general, null values are not returned by the API. This can be changed, but it's the default. There are many institutions that do not have domains yet.

rushtong (approved):
Nice, thank you! This should simplify a blocker I have on the main page.

src/types/model.ts
otchet-broad: I didn't realize that we weren't populating domains for all institutions.  Thanks for taking the time to explain!

PR #2966, opened 2025-07-15
[DT-1114] Moves storage from session to local.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1114](https://broadworkbench.atlassian.net/browse/DT-1114)

### Summary

DUOS' OIDC broker is already writing to local storage.  This change moves DUOS's use of storage from session storage to local storage.   This allows for DUOS to open in new tabs without requiring users to log in a second time.
When a user logs out in one tab, the user will effectively be logged out of all tabs however other tabs will not automatically return to the login screen until an action is taken in them.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
+++ b/cypress/component/ProgressReports/ProgressReportApplication.spec.tsx
@@ -277,7 +277,7 @@ describe('ProgressReportApplication - Component Tests', () => {
           link: ''
         }
       ]
-    };
+    } as unknown as CombinedDataAccessRequest;
 
     mountComponent(darWithPublications, true);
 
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍🏽

rjohanek (approved):
tested locally and it looks good! Is the "as unknown as CombinedDataAccessRequest" a drive by change? what is that about?

otchet-broad:
@rjohanek - Great question.  It became necessary because the test wasn't changed when the type was and this file had to be modified for a storage change.  As a result, I needed it fixed to pass the linter.  :-)

I'll be happier when @s-rubenstein puts us in the position to lint the entire project in each pull request.

MatthewBemis (approved):
nice! 🎉

eweitz (approved):
This is a significant UX improvement!  I manually tested it in my local environment.

PR #2969, opened 2025-07-16
[DT-1954] Use Object instead of React State variable.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1954](https://broadworkbench.atlassian.net/browse/DT-1954)

### Summary

The first implementation of this fix relied on a react state-hook managed value.  This value is being mutated in this same function and can create a loop with unset and then set values causing issues.   This alternate approach relies on the set of DARs already in the value and sorts them ascending so that the first entry is the oldest DAR in the collection and then uses that to load the correct DAR.

Before:
Intermittent endless spinner when loading the "Full DAR" view in reviews.
image

After:

Properly loaded DAR, consistently.
image

----

- Verify all tests go green


rushtong (approved):
👍🏽

MatthewBemis (approved):

PR #2970, opened 2025-07-17
DT-1938, DT-1940, DT-2007: Update Admin Manage Institutions Page

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1938
https://broadworkbench.atlassian.net/browse/DT-1940
Also fixes some fallout from lint updates: https://broadworkbench.atlassian.net/browse/DT-2007

### Summary
There are a number of large changes in this PR:
* Convert Admin Manage Institutions to TypeScript
  * Otherwise, not much is changed here
  * Added tests
* Convert Institution Table to TypeScript
  * Moved many functions out of the component into a helper component
  * Added tests
* Slight change to data loading in `InstitutionDetails`
  * Previously, we could pass the full list of institutions in location.state. That's not possible in the `SimpleTable` case since each row only knows about the individual data entity
  * Instead, query for the full list for validation purposes and load the requested institution from that list
* New institution search function that looks at the values of all possible fields. We have similar methods for other table search features so this follows the standard pattern.
* Model updates to reflect what fields are optional on `Institution `

----

- Verify all tests go green


src/pages/AdminManageInstitutions.tsx
```diff
@@ -0,0 +1,97 @@
...
+    setSearchTerm(value);
+    setFilteredList(filter(institutionList, searchTerm));
+  };
+
+  const filter = (list: Institution[], value: string): Institution[] => {
+    if (value) {
+      return getSearchFilterFunctions().institutions(value, list);
+    }
+    return list;
+  };
```
rushtong, line 49: Previously, we were using the institution json to use for matching. That would allow for users to filter on the property names such as `createDate`, `signingOfficials`, etc, which will match on every institution with those fields populated.

src/components/institution_table/InstitutionTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const goToPage = (page: number) => {
+    if (page > 0 && page < pageCount + 1) {
+      setCurrentPage(page);
+    }
+  };
+
+  return (
+      <div data-cy="institution-table">
+        <SimpleTable
```
rushtong, line 63: Using this component is the meat of the change. Originally, we were writing out table elements directly. This change allows us to use the same styling (e.g. material-ui) as the rest of our console pages.

otchet-broad:
image

Not sure if this is a symptom of your PR or just how the table works, but text can become hard to read.  Do we need to consider wrapping and cell width?

cypress/component/Institutions/AdminManageInstitutions.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+          cy.contains(domain).should('exist');
+        });
+      }
+    });
+  });
+
+  /*
+   * Note: SearchBar functionality via cypress looks finicky. It works normally in the browser,
+   * but to see the onChange effect, we need to add an extra space after the search term. That component
+   * is outside the scope of this test, so we will leave it as is.
```
otchet-broad: Do we have a ticket to look into this?

rushtong:
> image
> Not sure if this is a symptom of your PR or just how the table works, but text can become hard to read. Do we need to consider wrapping and cell width?

Thanks for that Elliot, I didn't see that locally. I'll look at wrapping this if possible.

cypress/component/Institutions/AdminManageInstitutions.spec.tsx
rushtong: Not at this time. `SearchBar` does not have its own set of tests. I'll put a ticket into the backlog.

src/components/institution_table/InstitutionTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+                  currentPage={currentPage}
+                  tableSize={tableSize}
+                  goToPage={goToPage}
+                  changeTableSize={changeTableSize}
+              />
+            }
+            sort={sort}
+            onSort={(sort: SortType) => {
+              Storage.setCurrentUserSettings(storageInstitutionSort, {
+                field: columns[sort.colIndex],
```
otchet-broad, line 81: With the the changes to use local storage, this means sort order will be preserved between sessions.  You good with that?

src/components/institution_table/InstitutionTableUtils.tsx
```diff
@@ -0,0 +1,291 @@
...
+    label: 'Institution',
+    cellStyle: {width: columnWidths.name},
+    cellDataFn: (row: Institution) => {
+      if (row) {
+        return <Link
+            to={{pathname: `/admin_manage_institutions/institutions/${row.id}`}}
+            style={{
+              overflow: 'hidden',
+              whiteSpace: 'nowrap',
+              textOverflow: 'ellipsis',
```
otchet-broad: I'm guessing this isn't working as expected.  See my screenshot.

src/components/institution_table/InstitutionTableUtils.tsx
```diff
@@ -0,0 +1,291 @@
...
+    },
+    sortable: true,
+    sortValueFn: (row: Institution) => row.name,
+  },
+  domains: {
+    label: 'Domains',
+    cellStyle: {width: columnWidths.domains},
+    cellDataFn: (row: Institution) => {
+      if (row?.domains) {
+        return row.domains.join(', ');
```
otchet-broad: If we sort them in the cell can we then turn on sorting in the header?

src/libs/utils.js
```diff
@@ -421,6 +421,31 @@ export const getSearchFilterFunctions = () => {
...
         includes(loweredTerm, toLower(status));
     }, targetList),
+    institutions: (term, targetList) => filter(institution => {
+      const loweredTerm = toLower(term);
+      const soStrings = institution.signingOfficials?.map((so) => {
+        return so.displayName + ' ' + so.email;
+      }).join(' ');
+      const domains = institution.domains?.join(' ') || '';
+      return includes(loweredTerm, toLower(institution.name)) ||
+        includes(loweredTerm, toLower(institution.id)) ||
```
otchet-broad, line 431: Am I understanding that we'll call toLower on each cell in the table, each time the filter changes?

otchet-broad (commented):
Thanks for the test coverage and solid work, @rushtong.  I left some observations, questions and comments.

src/components/institution_table/InstitutionTableUtils.tsx
rushtong: Header sorting is on the column. We can certainly do that, I didn't see the utility of that.

src/libs/utils.js
rushtong: Correct. You can see how we're using the same pattern in all of the other filter methods in this utility.

src/components/institution_table/InstitutionTableUtils.tsx
rushtong: I'll take another look, I think this is copied over from the original style.

src/components/institution_table/InstitutionTable.tsx
rushtong: Yes, we do this in quite a few other places as well so that change affects all of our user-specific sorting preferences.

eweitz (commented):
These are some nice UX refinements!  Being able to sort columns is useful, and I think this is a good start to a better solution for long multi-valued table cells like "Signing officials".

Two notable issues:

1.  Long institution names can overlap domains, and make the latter unreadable.  I saw this while browsing at full width on a 14" MacBook Pro laptop.  _Update: I just noticed this was pointed out in another review -- sorry for the repetition!_

2.  Institution search cannot be cleared (without a page refresh).  Previously, deleting the text in the search box would clear search results.

I also noticed a few minor issues:

3.  Signing officials cannot be copied if any text extends beyond 1 row, because the full text is only accessible via a tooltip.  Adding a "Copy" icon to copy table cell text upon click, and which renders only upon hovering over "Signing officials" cells, might solve this.

4.  The "Updated on" column header flows across 2 rows, due to narrow column width for that text.  Relatedly, the "ID" column looks quite wide.  Could we shrink the "ID" column and thus make space for a wider "Updated on" column such that the header doesn't split across 2 lines on a 14" MBP?

I'd be happy to open tickets for those minor issues if they seem better to tackle in future work.

For convenient reference, here are some videos and screenshots adding context to the above functional notes.

Width and overflow:
Image

Cannot clear institution search:

https://github.com/user-attachments/assets/a7dae8be-c79c-474c-b959-83038f6a7f47

Signing officials not copyable:

https://github.com/user-attachments/assets/3e5e8fbe-c87b-4953-919a-4ad1340f7df5

cypress/component/Institutions/AdminManageInstitutions.spec.tsx
rushtong: https://broadworkbench.atlassian.net/browse/DT-2008

rushtong: Turns out that the flakiness was in the search filter and how it as using state. This is fixed now.

src/components/institution_table/InstitutionTableUtils.tsx
rushtong: Fixed.

src/libs/utils.js
otchet-broad: A future improvement could be to cache a copy of the data once lower cased if we need to deal with lower power devices.

src/components/institution_table/InstitutionTableUtils.tsx
rushtong: I added sorting both within the column and on the column values

rushtong:
Thank you @eweitz !

> Two notable issues:
> 
> 1. Long institution names can overlap domains, and make the latter unreadable.  I saw this while browsing at full width on a 14" MacBook Pro laptop.  _Update: I just noticed this was pointed out in another review -- sorry for the repetition!_

I believe this is fixed now, PTAL.

> 2. Institution search cannot be cleared (without a page refresh).  Previously, deleting the text in the search box would clear search results.
 
I wasn't able to reproduce this, but looking more closely, I did see a different bug. The search function was setting the search term via `useState` and then trying to immediately use it to filter which led to buggy behavior. This now is working for me correctly.

> I also noticed a few minor issues:
> 
> 3. Signing officials cannot be copied if any text extends beyond 1 row, because the full text is only accessible via a tooltip.  Adding a "Copy" icon to copy table cell text upon click, and which renders only upon hovering over "Signing officials" cells, might solve this.

This is a great idea, I'll file a follow-on ticket for that.

> 4. The "Updated on" column header flows across 2 rows, due to narrow column width for that text.  Relatedly, the "ID" column looks quite wide.  Could we shrink the "ID" column and thus make space for a wider "Updated on" column such that the header doesn't split across 2 lines on a 14" MBP?

I did adjust the widths a bit to better accommodate the last two columns, PTAL.

eweitz:
Looks great!  I manually tested your refinements locally, and confirmed they work -- including institution search clearing.

eweitz (approved):

otchet-broad (approved):
Nice changes.  Thank you!

src/libs/utils.js
rushtong: See https://broadworkbench.atlassian.net/browse/DT-2013

rushtong:
> Thank you @eweitz !
> 
> > 3. Signing officials cannot be copied if any text extends beyond 1 row, because the full text is only accessible via a tooltip.  Adding a "Copy" icon to copy table cell text upon click, and which renders only upon hovering over "Signing officials" cells, might solve this.
> 
> This is a great idea, I'll file a follow-on ticket for that.

Following up on this comment, see: https://broadworkbench.atlassian.net/browse/DT-2014

PR #2974, opened 2025-07-18
[DT-1999] Compute logged in state from OIDC response

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1999](https://broadworkbench.atlassian.net/browse/DT-1999)

### Summary

Examine the OidcUser `expires_at` to figure out when a user is logged out.

----

- Verify all tests go green

### DIFF
--- a/src/libs/auth/auth.ts
+++ b/src/libs/auth/auth.ts
@@ -46,19 +46,19 @@ declare global {
   interface Window {
     Appcues?: {
       /** Identifies the current user with an ID and an optional set of properties. */
-      identify: (userId: string, properties?: never) => void;
+      identify: (userId: string, properties?: unknown) => void;
       /** Notifies the SDK that the state of the application has changed. */
       page: () => void;
       /** Forces specific Appcues content to appear for the current user by passing in the ID. */
       show: (contentId: string) => void;
       /** Fire the callback function when the given event is triggered by the SDK */
-      on: ((eventName: Exclude<string, 'all'>, callbackFn: (event: never) => void | Promise<void>) => void) &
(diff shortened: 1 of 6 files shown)


src/libs/auth/auth.ts
```diff
@@ -49,19 +46,19 @@ declare global {
   interface Window {
     Appcues?: {
       /** Identifies the current user with an ID and an optional set of properties. */
-      identify: (userId: string, properties?: any) => void;
+      identify: (userId: string, properties?: never) => void;
```
fboulnois: Can you explain this change? This allows `never | undefined`, which seems unusual. Should it be `object` or `function` (for the one below) instead?

fboulnois (approved):

src/libs/auth/auth.ts
otchet-broad: I was thinking about @s-rubenstein 's PR and waiting to see what was being done there.

I didn't want to spend too much time fretting and **never** lets us tell typescript we don't know.
https://www.typescriptlang.org/docs/handbook/type-compatibility.html#any-unknown-object-void-undefined-null-and-never-assignability

In the draft @s-rubenstein pushed, it looks like they're setting ES lint rules to ignore the nags about **any**.  I'm fine with that and update these to match that if you'd prefer that approach.

otchet-broad: Per our conversation, we've gone with the 'unknown' type.

fboulnois (approved):
👍

rushtong (approved):
Tested locally by manipulating the expiration 👍🏽 As a side note, even if a user were to extend the expiration in local storage, the APIs would still recognize the token expiration.

PR #2977, opened 2025-07-22
[DT-1997] Update bucket logic to work on most recently submitted DAR/Progress report

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1997](https://broadworkbench.atlassian.net/browse/DT-1997)

### Summary

Prior to this update, the DAC Console would display votes from prior elections when reviewing progress reports.  This became evident in closeout progress reports.

BEFORE:

image

AFTER:

image

Member and Chairperson tabs are not visible because no election is open.
----

- Verify all tests go green

### DIFF
--- a/src/utils/BucketUtils.js
+++ b/src/utils/BucketUtils.js
@@ -16,10 +16,12 @@ import {
   pick,
   toLower,
   uniq,
-  values
+  values,
 } from 'lodash/fp';
-import { Match } from '../libs/ajax/Match';
-import {translateDataUseRestrictionsFromDataUseArray} from '../libs/dataUseTranslation';
+import {Match} from '../libs/ajax/Match';
+import {
+  translateDataUseRestrictionsFromDataUseArray,
+} from '../libs/dataUseTranslation';
 import {processVotesForBucket} from './DarCollectionUtils';
 import {processMatchData} from './VoteUtils';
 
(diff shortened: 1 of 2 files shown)


rushtong (approved):
Nice 👍🏽

src/utils/BucketUtils.js
```diff
@@ -16,10 +16,12 @@ import {
...
   uniq,
-  values
+  values,
 } from 'lodash/fp';
-import { Match } from '../libs/ajax/Match';
-import {translateDataUseRestrictionsFromDataUseArray} from '../libs/dataUseTranslation';
+import {Match} from '../libs/ajax/Match';
+import {
+  translateDataUseRestrictionsFromDataUseArray,
+} from '../libs/dataUseTranslation';
```
eweitz, line 24: Can we standardize on a format, to denoise PRs across author?  Let me know if this seems worthwhile, and I can open a tech health ticket.

src/utils/BucketUtils.js
```diff
@@ -45,14 +47,11 @@ import {processMatchData} from './VoteUtils';
 export const binCollectionToBuckets = async (collection, dacIds = []) => {
 
   const buckets = [];
+  //  Find the most recent DAR
+  const recentDar = collection.dars !== undefined ? Object.values(collection.dars).sort((a, b) => b.id - a.id).at(0) : [];
```
eweitz, line 51: I wasn't aware of [`at()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at).  Using `myArray.at(-1)` is certainly more intuitive than `myArray.slice(-1)[0]`.

eweitz (approved):
Code and screenshots look good!

src/utils/BucketUtils.js
otchet-broad: Looked into this a bit and we definitely will want to do this in an independent ticket so that we can both update the eslint.config.js with the formatting rule and then update all of the files to the standard.

PR #2979, opened 2025-07-22
[dt-2006] Fix TSC

s-rubenstein:
### Summary

----

- Verify all tests go green


eweitz (approved):
Thanks for enabling holistic TypeScript checks!

Code looks good.  I manually tested this locally, and functionality seems OK.  In manual tests I noticed that [DT-2003](https://broadworkbench.atlassian.net/browse/DT-2003) persists -- it does not seems fixed by this as we speculated it might.

src/utils/darFormUtils.js
```diff
@@ -315,15 +315,31 @@ export const calcPublicationOrPresentationErrors = (newPublication, publicationT
...
+  if (isEmpty(newPublication?.bibliographicCitation)) {
+    validation.bibliographicCitation = requiredError;
+  }
+
+  return validation;
+}
+
+export const calcPresentationErrors = (newPresentation) => {
+  const validation = {};
+  if (isEmpty(newPresentation?.title)) {
```
rjohanek, line 335: maybe some of this duplicated code can be extracted: title, date, and authors

rjohanek (commented):
is there a ticket to describe and track this work?

rjohanek (commented):
I'm seeing an issue with publication validation when I test this out locally. I can't successfully add a publication, but presentations seem to work.
Screenshot 2025-07-22 at 3 47 13 PM

src/utils/darFormUtils.js
s-rubenstein: Potentially, but I think we don't want to rely on the types having the same fields. The one I think makes sense to abstract is date, because it has more complexity. I can do that.

s-rubenstein:
@raejohanek the ticket number is in the title, but also it is this ticket: https://broadworkbench.atlassian.net/browse/DT-2006

s-rubenstein:
@rjohanek I need someone to show me how to get to the progress report submission. I think I continue to be missing permissions to test locally. I've been trying to chase that down for a while. Maybe after standup we can try to fix this for me?

s-rubenstein:
Got help to fix it. We were keying off of ID, which isn't ideal. We should change that.

rjohanek (approved):
looks good, I tested out the bug and it works well now!

PR #2984, opened 2025-07-23
[DT-2002] Hide `Approve Closeout` button after click

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2002](https://broadworkbench.atlassian.net/browse/DT-2002)

### Summary

The `Approve Closeout` button wasn't disappearing after the user clicked it.  Now, after each user clicks it, the button will disappear.  The button also becomes disabled while the request is processing so that it can't be clicked again while a request is in flight.

BEFORE:

https://github.com/user-attachments/assets/ef5e2397-c771-4c2c-8b24-8ba483626922

AFTER:

https://github.com/user-attachments/assets/7d80cd48-4296-4e00-b056-5eb4fc3fc9c6

----

- Verify all tests go green

### DIFF
--- a/cypress/component/ProgressReports/closeout_review.spec.tsx
+++ b/cypress/component/ProgressReports/closeout_review.spec.tsx
@@ -1,18 +1,19 @@
 import React from 'react';
 import {mount} from 'cypress/react';
 import {CloseoutReview} from 'src/pages/progress_reports/CloseoutReview';
-import {Acknowledgement} from 'src/types/model';
+import {Acknowledgement, DataAccessRequest} from 'src/types/model';
 import {User} from 'src/libs/ajax/User';
+import {Storage} from 'src/libs/storage';
+import {DAR} from 'src/libs/ajax/DAR';
 
 describe('CloseoutReview - Component Tests', () => {
-  let onApproveSpy: () => void;
   let onReturnSpy: () => void;
+  const user = {isSigningOfficial: true};
 
   const mountComponent = (props = {}) => {
     const defaultProps = {
-      onApprove: onApproveSpy,
       onReturn: onReturnSpy,
-      referenceId: 'DAR-UUID',
+      dar: {referenceId:'DAR-UUID'} as DataAccessRequest,
       ...props
     };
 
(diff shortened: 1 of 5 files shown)


s-rubenstein (approved):

src/pages/progress_reports/CloseoutReview.tsx
```diff
@@ -3,27 +3,50 @@ import InfoIcon from '@mui/icons-material/Info';
...
-    onApprove,
-    onReturn,
-    referenceId
+    dar, onReturn
 }) => {
 
   const [acknowledged, setAcknowledged] = useState<boolean | undefined>(undefined);
+  const [pendingApproval, setPendingApproval] = useState<boolean> (false);
 
+    const onApprove = async () => {
```
fboulnois, line 23: 👍 , should we also move `onReturn` inside this component?

cypress/component/ProgressReports/closeout_review.spec.tsx
```diff
@@ -1,30 +1,32 @@
...
   };
 
   beforeEach(() => {
-    onApproveSpy = cy.stub().as('approveStub');
     onReturnSpy = cy.stub().as('returnStub');
   });
 
   it('renders the component correctly', () => {
+    cy.stub(Storage, 'getCurrentUser').returns(user);
+    cy.stub(User, 'getAcknowledgement').returns(null);
```
fboulnois: If this is in all the tests, would it make sense to move it into the `beforeEach`?

fboulnois (approved):
A few questions:

cypress/component/ProgressReports/closeout_review.spec.tsx
otchet-broad: Unfortunately, cypress doesn't allow functions to be re-wrapped so I can do this for the getCurrentUser since all of the tests rely on the same response.  I can't do that for the getAcknowledgement because that value needs to change.

src/pages/progress_reports/CloseoutReview.tsx
otchet-broad: The onReturn function is dependent on a prop so we'd still have to pass another parameter for this to work.  I'm good leaving it in the parent to figure out where to take the user.

PR #2985, opened 2025-07-23
[DT-1960] Improve validation for institution name and domains

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1960

### Summary
Previously, frontend validation only checked that domain names were unique.  Now the client validates more robustly.

New validation for institution names:
* Is unique
* Trims spaces
* Replace single curly quotes with single straight quotes
* No double quotes

New validation for domain name:
* Has a top-level domain (TLD)
* No spaces 
* No underscores
* No wildcards
* Total [length](https://en.wikipedia.org/wiki/Domain_Name_System) < 254 characters, with labels (parts separated by dots) < 64 characters
* Domain labels can't start or end with a hyphen or a period

Like before, Unicode input is handled.

Here's how it looks!

https://github.com/user-attachments/assets/d3cc1119-f13c-4a83-87a9-dd224c856eb6

----

- Verify all tests go green


src/pages/AdminManageInstitutions.tsx
```diff
@@ -34,7 +34,7 @@ export default function AdminManageInstitutions() {
       }
     };
     init();
-  }, []);
+  }, [searchTerm]);
```
rushtong: Adding this dependency causes a reload of the institutions API call when the search term changes.

src/pages/dar_application/ScrollableTabs.tsx
```diff
@@ -53,12 +53,10 @@ export const ScrollableTabs = ({ applicationTabs, formSelectedTabId }: Scrollabl
     };
 
     useEffect(() => {
-        // eslint-disable-next-line -- codacy says event listeners are dangerous
```
rushtong, line 56: Nice cleanup

package.json
```diff
@@ -30,7 +30,8 @@
     "react-tooltip": "4.5.1",
     "stackdriver-errors-js": "0.12.0",
     "tss-react": "4.9.18",
-    "uuid": "11.1.0"
+    "uuid": "11.1.0",
+    "validator": "^13.15.15"
```
rushtong, line 34: Minor suggestion - we usually pin dependencies exactly
```suggestion
    "validator": "13.15.15"
```

package.json
```diff
@@ -49,11 +50,12 @@
   "devDependencies": {
     "@emotion/styled": "11.14.1",
     "@eslint/js": "9.31.0",
-    "@types/node": "24.0.13",
     "@types/lodash": "4.17.20",
+    "@types/node": "24.0.13",
     "@types/react": "18.3.12",
     "@types/react-dom": "18.3.1",
     "@types/react-router-dom": "5.3.3",
+    "@types/validator": "^13.15.2",
```
rushtong, line 58:
```suggestion
    "@types/validator": "13.15.2",
```

package.json
rushtong: I wonder if we could reuse this library in many of our form component validations. See, for example, https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/utils.js#L103

rushtong (approved):
Fantastic update, thank you! See minor suggestions inline 👍🏽

src/pages/AdminManageInstitutions.tsx
rushtong: Looking into this a little more. I think if we pulled out the search term update into a separate useEffect we could address the dependency issue and prevent repeated API calls:

```
  useEffect(() => {
    const init = async () => {
      try {
        setIsLoading(true);
        const listOfInstitutions = await InstitutionAPI.list();
        setInstitutionList(listOfInstitutions);
        setFilteredList(listOfInstitutions);
        setIsLoading(false);
      } catch (error) {
        const message = extractError(error);
        Notifications.showError({text: 'Error: Unable to retrieve institutions from server: ' + message});      } finally {
        setIsLoading(false);
      }
    };
    init();
  }, []);

  useEffect(() => {
    setFilteredList(filter(institutionList, searchTerm));
  }, [searchTerm, institutionList]);
```

src/components/institution_table/components/InstitutionDomainEditor.tsx
```diff
@@ -20,6 +21,49 @@ export const InstitutionDomainEditor = ({ domains, isEditing, onDomainsChange, i
         });
     });
 
+    const validateDomain = (domain: string): string | null => {
+        try {
+            // Check if it's a valid FQDN (Fully Qualified Domain Name)
+            if (!validator.isFQDN(domain, {
```
fboulnois, line 27: Since we're only using `validator` here, instead of pulling in another dependency would it be better to copy the method over? https://github.com/validatorjs/validator.js/blob/master/src/lib/isFQDN.js#L13

fboulnois (commented):
Suggestion:

cypress/component/Institutions/InstitutionDomainEditor.spec.tsx
```diff
@@ -166,4 +167,196 @@ describe('Institution Domain Editor Tests', () => {
...
+                'test.测试',
+                'università.it',
+                'тест.рф'
+            ];
+
+            const expectedDomains: string[] = [];
+            internationalDomains.forEach((domain) => {
+                cy.get('input').clear().type(domain);
+                cy.contains('button', 'Add').click();
+                expectedDomains.push(domain);
```
otchet-broad: Are you expecting all of the domains to be added to this list?  I don't see a test that uses this value.

cypress/component/Institutions/InstitutionDomainEditor.spec.tsx
otchet-broad, line 1: Nice tests!  Thanks for adding the coverage.

One of the things these make me wonder about are the existing domain values we have stored in consent.  Do you think it's worthwhile to scan them to see if we have any stored that might not be in compliance with these validation rules?

rjohanek (approved):
looks good! I love the tests! It would be nice to have a test of the normalization of domains as well. Like show that capital letters are saved lowercase for example

src/components/institution_table/components/InstitutionDomainEditor.tsx
eweitz: Greg [suggests](https://github.com/DataBiosphere/duos-ui/pull/2985#discussion_r2228416423) using `validator` for more email validation as well.  Think it's worthwhile soon?  If so, it might be most efficient to keep this dependency.

src/components/institution_table/InstitutionDetails.tsx
```diff
@@ -64,6 +65,41 @@ export const InstitutionDetails = (props: InstitutionDetailsProps) => {
         loadInstitution();
     }, [formMode, institutionId]);
 
+    const normalizeInstitutionName = (name: string): string => {
+        // Trim whitespace from start and end
+        let normalized = name.trim();
+
+        // Replace curly single quotes with straight straight Iquotes
+        normalized = normalized.replace(/[‘’]/g, "'");
```
otchet-broad, line 73: I'm wondering if we need to communicate to the user that we've changed their input.

src/components/institution_table/components/InstitutionDomainEditor.tsx
otchet-broad: I could see where we could replace some of the other input validation rules we have with this library.  Are you worried about package size bloat or have a different concern?

src/components/institution_table/InstitutionDetails.tsx
eweitz: Good suggestion, done in 54b9bc9.  Demo:

https://github.com/user-attachments/assets/5b91093d-4795-41bb-a59b-19c2ba340454

src/pages/AdminManageInstitutions.tsx
eweitz: Nice find and optimization!  Done in 90af7ac.

eweitz:
> It would be nice to have a test of the normalization of domains as well. Like show that capital letters are saved lowercase for example

Agreed, @rjohanek, done in 3586443.

cypress/component/Institutions/InstitutionDomainEditor.spec.tsx
eweitz: Nice catch, I had been this particular array in a check for Punycode conversion.  I removed most of that code when we opted to store Unicode user input, but overlooked removing this.  I've cleaned that up in 6830a6d.

cypress/component/Institutions/InstitutionDomainEditor.spec.tsx
eweitz: I recall some discussion around that in an earlier standup -- let's check in again as a parking lot topic.

package.json
eweitz: Sorry, I overlooked committing this suggestion, will do in my next PR.

cypress/component/Institutions/InstitutionDomainEditor.spec.tsx
eweitz: Following up: yes, we've planned this in https://broadworkbench.atlassian.net/browse/DT-1990.  Thanks @rjohanek for finding the ticket!

PR #2989, opened 2025-07-24
[DT-2001] Display missing DAC info on DAR addendum

MatthewBemis:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2001

### Summary

Previously, the DAC name was not displayed in the DAR addendum. Now, that information is retrieved and displayed to the user.

Before:
Screenshot 2025-07-24 at 1 54 31 PM

After:
Screenshot 2025-07-24 at 1 53 39 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_application/DucAddendum.jsx
+++ b/src/pages/dar_application/DucAddendum.jsx
@@ -127,6 +127,11 @@ export default function DucAddendum(props) {
   }, [datasets])
 
   const buildDucAddendumTable = useCallback(async () => {
+    const getDacName = (dacId) => {
+      const dac = dacs.find(dac => dac.dacId === dacId)
+      return dac?.name || 'Unknown DAC'
+    }
+
     const tableChunks = buckets.map((bucket) => {
       const dataUseCodes = bucket.label
       const dataUseSummary = bucket.dataUses.map(dataUse => dataUse.description).join('. ')
@@ -157,7 +162,7 @@ export default function DucAddendum(props) {
             style: columnStyles,
           },
           {
-            data: dacs.find(dac => dac.dacId === dataset.dacId)?.name || 'Unknown DAC',
+            data: getDacName(dataset.dacId),
             id: dataset.datasetId,
             style: columnStyles,
           },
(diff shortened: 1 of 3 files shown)


src/pages/dar_application/DucAddendum.jsx
```diff
@@ -106,7 +109,29 @@ export default function DucAddendum(props) {
     getBuckets()
   }, [getBuckets])
 
+  useEffect(() => {
+    const loadDacs = async () => {
+      const dacIds = [...new Set(datasets.map(dataset => dataset.dacId))]
+      const dacPromises = dacIds.map(dacId => DAC.get(dacId))
```
rushtong: This endpoint is currently restricted to 
```
@RolesAllowed({ADMIN, MEMBER, CHAIRPERSON})
```
so Researchers will get an error. I think we might need a different approach here.

rushtong: If we populated the DAC name on the dataset, we could read it directly from the datasets in the collection that is queried.

MatthewBemis: Shoot, sorry for the oversight. Good catch, I'll look at re-working this.

Will throw this PR back into draft mode.

MatthewBemis: I've updated the `DucAddendum` component to just fetch the DatasetTerms (which include more information about the relevant DAC) instead of trying to load the DACs using the restricted endpoint.

It's another API call, but it's a quick one (<100ms), and it's still fewer calls than the original implementation here since we can at least load the DatasetTerms in bulk.

I took a look at the parent component (`DataAccessRequestApplication`) to see if I could refactor that to just use DatasetTerms and pass those down here to avoid the API request entirely, but that proved to be wayyy too thorny. So I went with this intermediate approach instead.

src/pages/dar_application/DucAddendum.jsx
```diff
@@ -106,7 +109,50 @@ export default function DucAddendum(props) {
...
+          bool: {
+            must: [
+              {
+                match: {
+                  _type: 'dataset',
+                },
+              },
+              {
+                terms: {
+                  _id: datasets.map(dataset => dataset.datasetId),
```
rushtong, line 125: Interesting, didn't realize we could pass in a list of ids here.

rushtong (approved):
Very nice work, thank you! 👍🏽

fboulnois (approved):
👍

PR #2995, opened 2025-07-28
DT-1985: Show RADAR Votes

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1985

## Summary
* Convert `BucketUtils` and test to Typescript
* Add RADAR vote group to buckets
* Refactor data use parsing to query from ES rather than use UI dev library that does something similar
* Refactor bucketing logic to use the new values from ES
* Refactor the `DucAddendum` DAC call to `BucketUtils` so we aren't making duplicate API calls on the same page. Also allows for simpler data population and testing.
* Fixes a bug for member voting. If a different chair had approved the dataset bucket, the member tab would still show the bucket as votable. This adds more logic to ensure member votes are not possible when the election is closed.

### Before View
Screenshot 2025-08-01 at 1 40 53 PM

### After View of a RADAR Approval
Screenshot 2025-08-01 at 1 42 28 PM

### After View of a Chairperson Approval
Screenshot 2025-08-01 at 2 49 07 PM

### Incomplete Read-Only View
Screenshot 2025-08-01 at 1 55 15 PM

----

- Verify all tests go green


src/utils/BucketUtils.ts
```diff
@@ -0,0 +1,384 @@
...
+  label: string
+  datasets: Dataset[]
+  datasetIds: number[]
+  dataUse?: DataUseSummary
+  dataUses?: DataUseTerm[]
+  elections: Election[]
+  votes: Record<string, VoteGroup>[]
+  matchResults: MatchResult[]
+  algorithmResult?: AlgorithmResult
+  isRP?: boolean
```
eweitz: I hadn't seen the "RP" initialism in DUOS before, and it wasn't immediately obvious what it meant.  Can a brief description be added here?

```suggestion
  isRP?: boolean // Whether this is an RP (Research Purpose) vote bucket
```

src/utils/BucketUtils.ts
```diff
@@ -0,0 +1,384 @@
...
+
+    // Step 5: Generate bucket key and label
+    b.key = 'bucket-' + b.datasetIds.join('-')
+    b.label = b.dataUses?.map((du: DataUseTerm) => du.code).join(', ') || 'Undefined Data Use'
+
+    // Step 6: Coalesce match results into a single result per bucket
+    b.algorithmResult = calculateAlgorithmResultForBucket(b)
+  })
+
+  // Step 7: Populate RUS Vote bucket with RP votes and move to the top of the bucket array
```
eweitz, line 147: I like how this function outlines all the steps in the docstring, then again piecewise near the relevant implementation.  Nicely maintainable handling of domain-specific complexity.

src/utils/BucketUtils.ts
```diff
@@ -0,0 +1,384 @@
...
+        createDate,
+        rationales,
+        id,
+      }
+    }
+  }
+  else {
+    // Different match values? Provide a custom message
+    return {
+      result: 'Unable to determine a system match',
```
eweitz, line 243: Do we define "system match" in prose anywhere for users?  This phrase is probably worth using in forthcoming RADAR docs, at least.

src/utils/BucketUtils.ts
```diff
@@ -0,0 +1,384 @@
...
+ *
+ * Four potential cases:
+ *  1. No matches
+ *  2. Exactly one match or N matches that are all the same - easy case
+ *  3. Abstain is true and match is false - decision is abstained
+ *  4. N matches - not all the same - very confusing case
+ *
+ * @private
+ * @param bucket
+ * @returns {AlgorithmResult}
```
eweitz: Such docstring elements are often nice in JavaScript, but seem unnecessarily duplicative in TypeScript.

src/utils/BucketUtils.ts
```diff
@@ -0,0 +1,384 @@
...
+    const rpVoteGroup: VoteGroup = {
+      chairpersonVotes: [],
+      memberVotes: [],
+      finalVotes: [],
+    }
+    forEach((v: Vote) => {
+      const lowerCaseType = v.type.toLowerCase()
+      switch (lowerCaseType) {
+        case 'chairperson':
+          // 'Chairperson' votes count as final votes for 'RP' elections. This is not true for 'DataAccess' votes
```
eweitz: Helpful clarification!  I mistakenly thought Chair votes were automatically final in _all_ types of elections.  

Is this more correct?

```suggestion
          // 'Chairperson' votes count as final votes for 'RP' elections. This is not true for 'DataAccess' elections
```

src/utils/DarCollectionUtils.js
```diff
@@ -42,10 +43,14 @@ export const processVotesForBucket = (darElections = []) => {
       targetMember = dataAccess.memberVotes
       targetChair = dataAccess.chairpersonVotes
       targetFinal = dataAccess.finalVotes
+      targetRadar = dataAccess.radarVotes
     }
-    forEach((vote) => {
+    dateSortedVotes.forEach((vote) => {
```
eweitz, line 48: Nice readability improvement.

eweitz (approved):
Code and screenshots look good!  I learned a lot about DUOS votes and elections here.

I suggest a few minor, non-blocking maintainability refinements.

cypress/component/utils/bucket_utils.spec.ts
```diff
@@ -371,7 +373,7 @@ const dar_collection = {
       dacId: 5,
     },
   ],
-}
+} as unknown as DarCollection
```
fboulnois: `as unknown as x` should be avoided since it disables Typescript checking on the object. Consider directly casting the object `as DarCollection`, if this is not possible it means that sub-objects are not properly typed which may indicate a real issue.

cypress/component/utils/bucket_utils.spec.ts
```diff
@@ -396,10 +398,69 @@ const match_results = [
...
+    },
+    dacId: 4,
+  },
+  {
+    datasetId: 5,
+    datasetName: 'ds 5',
+    datasetIdentifier: 'DUOS-000005',
+    dacId: 5,
+  },
+] as unknown as DatasetTerm[]
```
fboulnois: ditto here, cast each individual object in the array to `as DatasetTerm`

cypress/component/utils/bucket_utils.spec.ts
```diff
@@ -981,23 +1004,91 @@ describe('BucketUtils', () => {
...
+        datasetIdentifier: 'DUOS-000005',
+        dataUse: {
+          primary: [{
+            code: 'HMB',
+            description: 'Health, Medical and Biomedical Research',
+          }],
+        },
+        dacId: 5,
+      },
+    ] as unknown as DatasetTerm[]
```
fboulnois: ditto here

src/types/model.ts
```diff
@@ -183,14 +183,16 @@ export interface StudyTerm {
...
   description: string
+  manualReview?: boolean
+  type?: string
 }
 
-interface DataUseSummary {
-  primary: DataUseTerm[]
-  secondary: DataUseTerm[]
+export interface DataUseSummary {
+  primary?: DataUseTerm[]
```
fboulnois: I thought the primary should never be optional?

src/types/model.ts
```diff
@@ -251,30 +253,30 @@ export const getAccessManagementSummary = (accessManagement: string): AccessMana
...
-  gender: string
-  pediatric: boolean
-  population: boolean
-  illegalBehavior: boolean
-  sexualDiseases: boolean
-  stigmatizeDiseases: boolean
-  vulnerablePopulations: boolean
-  psychologicalTraits: boolean
-  notHealth: boolean
+  generalUse?: boolean
```
fboulnois, line 270: Should all of these be optional?

cypress/component/utils/bucket_utils.spec.ts
```diff
@@ -981,23 +1004,91 @@ describe('BucketUtils', () => {
           dacId: 5,
         },
       ],
-    }
-    cy.wrap(binCollectionToBuckets(collection)).then((buckets) => {
+    } as unknown as DarCollection
```
fboulnois: ditto here

src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
```diff
@@ -218,3 +251,24 @@ export default function CollectionSubmitVoteBox(props) {
     </div>
   )
 }
+
+CollectionSubmitVoteBox.propTypes = {
```
fboulnois: shouldn't we put the type directly on the component?

fboulnois (commented):
A few suggestions:

src/utils/BucketUtils.ts
rushtong: Great point. `system match`, in this context, is the Ontology algorithm match described [here](https://github.com/DataBiosphere/consent-ontology/blob/develop/docs/Algorithms.md). We definitely could use some better instructional language in the UI around this point.

src/utils/BucketUtils.ts
rushtong: I had to go back to the back end code to verify this, but yes, Chair votes do count as final for RP elections. We don't capture a FINAL vote type for RP elections, we only do that for DataAccess elections. Good language clarification, thank you 👍🏽

cypress/component/utils/bucket_utils.spec.ts
rushtong: Thank you, will go back to this - I believe the issue I was struggling with was the sub-object typing.

src/types/model.ts
rushtong: I believe that is true, will take a second look at why I made this change.

src/types/model.ts
rushtong: Individually, yes, each field can be optional. Holistically, at least one of `generalUse`, `hmbResearch`, or `diseaseRestrictions` must be present to ensure a primary data use. I'll look into seeing if TS can handle that.

src/components/collection_vote_box/CollectionSubmitVoteBox.jsx
rushtong: This was added to address sonarqube warnings. I think we'd need to convert this to TS to move it to the component. Will look at that.

fboulnois (approved):
👍

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
```diff
@@ -170,19 +239,20 @@ export default function CollectionSubmitVoteBox(props) {
...
                   isVotingDisabled={isVotingDisabled}
+                  isRadar={isRadar}
                 />
                 <div style={styles.voteButtons}>
                   {!isVotingDisabled && (
                     <CollectionVoteYesButton
-                      onClick={() => updateVote(true, !isNil(updateFinalVote))}
-                      disabled={isVotingDisabled || isApprovalDisabled || isLoading}
+                      onClick={() => updateVote(true, !!updateFinalVote)}
+                      disabled={isVotingDisabled || isApprovalDisabled || isLoading || isElectionClosed}
```
rushtong, line 248: Here and line 255 is a bug fix. Previously, members might see an active vote Yes/No button even if the election were closed.

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
```diff
@@ -39,44 +63,89 @@ const styles = {
...
+    ? (
+        <img
+          data-cy="radar-icon"
+          className="radar-icon"
+          src={radarIcon}
+          alt="Rule Automated Data Access Decision"
+          title="Rule Automated Data Access Decision"
+        />
+      )
+    : null
```
rushtong, line 109: This, along with the collection of radar votes in `BucketUtils`, is the meat of this PR

cypress/component/MultiDatasetVoteTab/collection_submit_vote_box.spec.tsx
```diff
@@ -300,49 +391,106 @@ describe('CollectionSubmitVoteBox - Tests', function () {
...
     )
-    cy.get('[data-cy=vote-subsection-heading]').should('exist').contains('The final vote is: YES')
+    cy.get('[data-cy=vote-subsection-heading]').should('exist')
+    cy.get('[data-cy=vote-subsection-heading]').contains('The final vote is: YES')
     cy.get('textarea').should('be.disabled')
     cy.get('[datacy=yes-collection-vote-button]').should('not.exist')
     cy.get('[datacy=no-collection-vote-button]').should('not.exist')
   })
+
+  it('shows the radar vote icon when the vote type is radar', () => {
```
rushtong, line 455: Net new test for radar identifier

cypress/component/MultiDatasetVoteTab/collection_submit_vote_box.spec.tsx
```diff
@@ -300,49 +391,106 @@ describe('CollectionSubmitVoteBox - Tests', function () {
...
+        adminPage={true}
+        bucketKey="collection-submit-vote-box"
+        updateFinalVote={() => {}}
+        reloadFn={() => {}}
+      />,
+    )
+    cy.get('[data-cy=radar-icon]').should('exist')
+  })
+
+  it('shows disabled Yes/No buttons as member when election is closed', () => {
```
rushtong, line 473: Net new test for votable closed election bug

cypress/component/DarCollectionTable/dar_dataset_table.spec.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react'
 import { mount } from 'cypress/react'
-import { DarDatasetTable } from '../../../src/components/dar_dataset_table/DarDatasetTable.jsx'
+import { DarDatasetTable } from 'src/components/dar_dataset_table/DarDatasetTable.jsx'
```
otchet-broad:
```suggestion
import { DarDatasetTable } from 'src/components/dar_dataset_table/DarDatasetTable'
```

cypress/component/DarCollectionTable/dar_dataset_table.spec.jsx
```diff
@@ -1,8 +1,9 @@
...
 import darCollection from './darCollection.json'
-import { Match } from '../../../src/libs/ajax/Match.js'
+import { Match } from 'src/libs/ajax/Match.js'
```
otchet-broad:
```suggestion
import { Match } from 'src/libs/ajax/Match'
```

cypress/component/DarCollectionTable/dar_dataset_table.spec.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react'
 import { mount } from 'cypress/react'
-import { DarDatasetTable } from '../../../src/components/dar_dataset_table/DarDatasetTable.jsx'
+import { DarDatasetTable } from 'src/components/dar_dataset_table/DarDatasetTable.jsx'
 import darCollection from './darCollection.json'
```
otchet-broad:
```suggestion
import darCollection from './darCollection'
```

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
otchet-broad: Do you think we need a RADAR green and RADAR red version of the icon so we can change the icon and the alt text based on the approval (or future denial)?

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
```diff
@@ -86,44 +155,42 @@ export default function CollectionSubmitVoteBox(props) {
...
       }
       else {
         setVote(newVote)
       }
       Notifications.showSuccess({ text: 'Successfully updated vote' })
     }
-    catch (error) {
-      if (error && error.status === 409) {
+    catch (error: unknown) {
+      const consentError = extractConsentError(error)
```
otchet-broad, line 190: That extractConsentError method is helpful!

src/types/model.ts
otchet-broad, line 270: Looking at the type definition, I'm thinking that DataUse has a "one or more of the following" type interface.  Is that right?  It might be good to document that in a comment within the code so users of DataUse understand that expectation.

src/utils/BucketUtils.ts
```diff
@@ -0,0 +1,354 @@
...
+  const secondaryOther = dataUse?.secondary?.some((dut: DataUseTerm) => dut.code === 'OTHER') || false
+  return primaryOther || secondaryOther
+}
+
+/**
+ * Calculate abstention for a data use. There are a number of cases where there should
+ * not be an algorithm decision if a field is true, including any "Other" state.
+ */
+export const shouldAbstain = (dataUse?: DataUseSummary): boolean => {
+  const codeList: string[] = ['OTHER', 'POP-M', 'POP-F', 'COL', 'IRB', 'GSO', 'PUB', 'MOR', 'POP-PD']
```
otchet-broad: Should we move this to a constants file so we can use it in other places?

otchet-broad (approved):
Nice work!  Left some comments for your consideration.

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
rushtong: > Do you think we need a RADAR green and RADAR red version of the icon so we can change the icon and the alt text based on the approval (or future denial)?

I like the idea. We can definitely revisit that when we have denials.

src/types/model.ts
rushtong: Yes, one of the three primaries have to be present. I didn't see a clean way to do this with Union types, so will add a comment here.

PR #3015, opened 2025-08-13
[DT-2022] Create Voting History Tab & Update Review Header

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2022 

### Summary
Updates the ReviewHeader to show the approved dataset names
- shows all to admins, shows only datasets for that DAC to DAC chairs/members
- QUESTION: Do I need to gate on the DAC chair being a chair is on a relevant DAC or is that already gated? I don't see any logic for that? Currently, can any chair view the dar collection review page for any dar collection if they just modify the url directly?

Updates the DarCollectionReview to have a VotingHistory Tab 
- page shows if for admins or if the user is a DAC Chair/member

Creates a new Voting History Tab
- displays Chair vote details in one table
- displays members' election overview in one table with expandable rows to show individual member votes
- only shows elections for the datasets in the list of relevant dacIds

Adds unit tests for header, review page, chair vote table, and election with member votes table
Also adds unit tests for sorting functionality in utils

Additional changes:
Change default sort order of rows in vote summary table to show populated votes first (yes, then no)
Change vote summary table to not show create date of votes, only update date if it is present

----

- Verify all tests go green


src/utils/DarCollectionUtils.js
```diff
@@ -175,7 +175,7 @@ const collapseVotes = ({ votes }) => {
   const collapsedVotes = {}
   votes.forEach((vote) => {
     const matchingVote = collapsedVotes[`${vote.vote}`]
-    const lastUpdate = vote.updateDate || vote.createDate
```
rjohanek, line 178: do not show vote create date in any vote summary / vote history tables, only show update date if it exists

src/components/vote_summary_table/VoteSummaryTable.jsx
```diff
@@ -142,7 +142,7 @@ function rationaleCellData({ rationale = '- -', voteId, label = 'rationale' }) {
 }
 
 export default function VoteSummaryTable(props) {
-  const [sort, setSort] = useState({ colIndex: 0, dir: 1 })
+  const [sort, setSort] = useState({ colIndex: 0, dir: -1 })
```
rjohanek, line 145: put populated votes first, yes, then no

src/libs/utils.js
```diff
@@ -520,7 +520,7 @@ export const setDivAttributes = (disabled, onClick, style, dataTip, onMouseEnter
 // each item in the list is an array of metadata representing a single table row
 // the metadata for each cell needs a data (exactly what is displayed in the table)
 // or value (string or number alternative) property which determines sorting
-export const sortVisibleTable = ({ list = [], sort }) => {
+export const sortVisibleTable = ({ list, sort }) => {
```
rjohanek, line 523: causing typescript errors so default values are used where it is called

cypress/component/VoteHistoryTable/voting_history.spec.tsx
```diff
@@ -0,0 +1,509 @@
...
+      alias: 999,
+      datasetIdentifier: 'DUOS-00999',
+      dataUse: {},
+      dacId: 1,
+      translatedDataUse: 'Samples are restricted for use under the following conditions:\nData use is limited for studying: sleep apnea [DS]\nFuture use for population origins or ancestry research is prohibited. [POA]\nCommercial use is not prohibited.\nData use for methods development research irrespective of the specified data use limitations is not prohibited.\nFuture use as a control set for diseases other than those specified is prohibited. [NCTRL]',
+      deletable: false,
+      properties: [],
+      dacApproval: true,
+      study: {
+        piName: 'Lisa Simpson, Betty White',
```
otchet-broad, line 386: I recognize those two.

src/components/vote_history_table/ElectionWithMemberVotesTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+}
+
+const processVoteSummary = (memberVotes: Vote[]) => {
+  if (!memberVotes || memberVotes.length === 0
+    || memberVotes.every(v => v.vote === null || v.vote === undefined)) {
+    return 'No votes cast'
+  }
+  const positives = memberVotes.filter(v => v.vote === true).length
+  const negatives = memberVotes.filter(v => v.vote === false).length
+  return positives + ' Approved, ' + negatives + ' Denied'
```
otchet-broad: Should we use the words 'YES' and 'NO' to match the words on the buttons the user clicks to vote?

image

src/pages/dar_collection_review/DarCollectionReview.jsx
```diff
@@ -75,14 +77,58 @@ const tabsForUser = (user, buckets, adminPage = false) => {
...
+    return dacIds.includes(dataset?.dacId)
+  }).map((datasetId) => {
+    const dataset = darCollection.datasets?.find(ds => ds.datasetId === datasetId)
+    return dataset?.name
+  })
+  return approvedDatasetNames || []
+}
+
+const userIsDacUser = (user) => {
+  return user.roles?.some(role => role.roleId === 2 || role.roleId === 1)
```
otchet-broad, line 123: I'm wondering if you should just let the backend do the filtering and display what's returned.  There are a number of edge cases because the user can be in multiple DACs and have different roles in each DAC.  Given you have a ticket to cover the issue you found, I'm not sure it makes sense to try to filter like this in the UI.

src/types/model.ts
```diff
@@ -525,14 +608,14 @@ export interface Collaborator {
 export interface Election {
   electionId: number
   electionType: string
-  finalVote: boolean
+  finalVote?: boolean
   status: string
-  createDate: string
-  lastUpdate: string
-  finalVoteDate: string
+  createDate: string | number
```
otchet-broad: Seems odd it can sometimes be a string or a number.  The response from the backend looks like it's always a number.  Do we create these in other ways?

src/components/vote_history_table/ElectionWithMemberVotesTable.tsx
otchet-broad: Using Yes and No also matches the wording used in the tables on the member and chair tabs.

image

src/components/vote_history_table/ElectionWithMemberVotesTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+interface ElectionWithMemberVotesTableProps {
+  electionsWithMemberVotes: ElectionWithMemberVotes[]
+}
+
+interface RowData {
+  data: string
+  cellStyle: React.CSSProperties
+  label: string
+  id: number | string
```
fboulnois: Why can the id be both a string and number?

src/pages/dar_collection_review/DarCollectionReview.jsx
```diff
@@ -92,28 +138,39 @@ export default function DarCollectionReview(props) {
...
+      setApprovedDatasets(approvedDatasetNames)
       setTabs(tabsForUser(user, processedBuckets, adminPage))
+      setDacIds(dacIds)
       setIsLoading(false)
       setSubcomponentLoading(false)
       setReferenceIdForDocuments(referenceIdForDocuments)
     }
-    catch (_error) {
+    catch (error) {
+      console.error('Error initializing DarCollectionReview:', error)
```
fboulnois:
```suggestion
```

src/types/model.ts
```diff
@@ -525,14 +608,14 @@ export interface Collaborator {
...
+  lastUpdate?: string | number
+  finalVoteDate?: string
   referenceId: string
-  finalRationale: string
-  finalAccessVote: boolean
+  finalRationale?: string
+  finalAccessVote?: boolean
```
fboulnois, line 618: Why can all of these values be optional?

src/types/model.ts
```diff
@@ -482,6 +482,89 @@ export interface DataAccessRequest {
   closeoutSigningOfficialApprovedUserId: number
 }
 
+export interface DataAccessRequestData {
```
fboulnois, line 485: Do we have this defined somewhere else?

fboulnois (approved):
A few questions:

src/pages/dar_collection_review/DarCollectionReview.jsx
rjohanek: Do you mean the filtering for the approved dataset names list here? Or, the filtering for the elections in VotingHistory? Or, both? Either way, we still need the method userIsDacUser to determine to show the VotingHistory Tab.

src/types/model.ts
rjohanek: We have CombinedDataAccessRequest which is a flattened version of DataAccessRequest and DataAccessRequestData. CombinedDataAccessRequest extends DataAccessRequest which now includes the DataAccessRequestData definition. Idk if this messes anything up for CombinedDataAccessRequest. I'm not seeing any errors but these fields are now duplicated in those two objects. Any thoughts on this?

src/types/model.ts
rjohanek: I believe if the election doesn't have a final vote these fields will not exist.

src/types/model.ts
rjohanek: Yeah I think it should just be a number

src/components/vote_history_table/ElectionWithMemberVotesTable.tsx
rjohanek: yeah I think it should just be a number

src/pages/dar_collection_review/DarCollectionReview.jsx
rjohanek: I'm not sure I feel comfortable removing the filtering until we have the backend fix in place.

src/types/model.ts
fboulnois: That seems reasonable then. I think long term we probably want to split these out since we have issues with all the 
"optional" version of objects.

src/utils/DarCollectionUtils.js
rushtong: This makes sense. The vote create date is usually the same date as the election create date. They can be different if a user was added to a DAC after election creation though.

rushtong:
Interim feedback on the overall UI. It's nice to be able to show the votes, but it's challenging to understand what the votes are for and, as an admin, which DAC the votes belong to. We should iterate on this and think about bringing in the following:

* Which dataset the votes are for
* Data Access or RUS election type
* Which DAC the vote belongs to
* Action that triggered the vote, e.g.
  * Open Election action
  * RADAR action
  * Re-open

The phrase "Initial DAR" is not clear. In the case of a DAR being opened/reopened/canceled/opened again, all of the votes are given the same type.

cypress/component/utils/bucket_utils.spec.ts
```diff
@@ -16,12 +46,90 @@ const dar_collection = {
...
+      rationale: '',
+      rus: {},
+      status: 'submitted',
+      userId: 1,
+      eraCommonsId: 'test-era-commons-id',
+      expiresAt: new Date().getTime(),
+      createDate: new Date().getTime(),
+      sortDate: new Date().getTime(),
+      submissionDate: new Date().getTime(),
+      updateDate: new Date().getTime(),
```
rushtong: `new Date().getTime()` could be a constant which would simplify a lot of usages in simple object construction. 

Separately, it is very concerning to me that so much additional code is required solely due to updates to `model.ts`. Locally, I reverted all of this code and the tests run with no changes. So much boilerplate feels like an anti-pattern to me. I wonder if it's time to think about genericizing obect construction to a cypress helper class.

src/utils/DarCollectionUtils.js
```diff
@@ -71,7 +71,7 @@ export const processVotesForBucket = (darElections = []) => {
 // Gets data access votes from this bucket by members of this user's DAC
 // Note that filtering by DAC does not occur if user is viewing on admin review page
 export const extractDacDataAccessVotesFromBucket = (bucket, user, adminPage) => {
-  const votes = !isNil(bucket) ? bucket.votes : []
+  const votes = !isNil(bucket?.votes) ? bucket.votes : []
```
rushtong: This can probably be simplified further:
```
  const votes = bucket?.votes ?? []
```

rushtong (approved):
I'm comfortable with this moving forward. We have a plan to iterate on the view and perhaps add additional functionality, for example - it might be nice to have a specialized view of just a single election.

rjohanek:
> Interim feedback on the overall UI. It's nice to be able to show the votes, but it's challenging to understand what the votes are for and, as an admin, which DAC the votes belong to. We should iterate on this and think about bringing in the following:
> 
> * Which dataset the votes are for

Great idea!

> * Data Access or RUS election type

All elections in the voting history table are Data Access

> * Which DAC the vote belongs to

Yeah that would be helpful for admins or users that are members of multiple DACs

> * Action that triggered the vote, e.g.
>   
>   * Open Election action
>   * RADAR action
>   * Re-open
> 
> The phrase "Initial DAR" is not clear. In the case of a DAR being opened/reopened/canceled/opened again, all of the votes are given the same type.

While its true that Initial DAR phrase would be the same for a first and second election on a DAR, the election date would be different and that could be enough of a distinguishing factor? Thoughts? Is it a common use case that elections are canceled and reopened in prod?

cypress/component/utils/bucket_utils.spec.ts
rjohanek: totally agree, the tests do run without these changes but they cause a lint error. I love the idea of having a class to help with test object construction. I see you already made a tech debt ticket for this issue. I think that is great thanks!

rushtong:
> > The phrase "Initial DAR" is not clear. In the case of a DAR being opened/reopened/canceled/opened again, all of the votes are given the same type.
> 
> While its true that Initial DAR phrase would be the same for a first and second election on a DAR, the election date would be different and that could be enough of a distinguishing factor? Thoughts? Is it a common use case that elections are canceled and reopened in prod?

This is actually an uncommon case in prod, but a common case in dev :-( The dates do make a difference. The one I was looking at in particular is `DAR-602`, `Lou Test`. It has 10 datasets so if there were multiple re-openings or PRs on that one, it will get very cluttered/confusing very fast.

PR #3025, opened 2025-08-21
DT-2113: Prevent the button from disappearing until complete

rushtong:
### Addresses
Fix up for https://broadworkbench.atlassian.net/browse/DT-2113

### Summary
Minor update to the Save/Attest buttons so they don't disappear until the action is complete

----

- Verify all tests go green


fboulnois (approved):
👍 , I wonder if we should switch the default on the async spinner button to be `false`.

rushtong:
> 👍 , I wonder if we should switch the default on the async spinner button to be `false`.

@fboulnois - I'm open to that suggestion. Would you like me to make that change in this PR?

rjohanek (approved):

PR #3036, opened 2025-09-05
DT-2142: Use the same selectable datasets component for DARs and PRs

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2142

## Summary
This PR replaces the autocomplete dataset selection widget on the primary Data Access Request Application page with a newer dataset selection widget that is used on the Progress Report Application page. This changes existing functionality in the following way:
* Previously, users could add new datasets to a draft DAR even though those datasets may not be appropriate for adding. This is due to the underlying autocomplete API that does not apply the more recent logical filters that we have in place in the Data Library.
* After this is merged, users can only remove datasets from a draft DAR. This means that users will need to pre-select all potential datasets from the Data Library page before proceeding.

The current functionality where datasets cannot be added or removed from submitted DARs is not changed.

### Old view
Screenshot 2025-09-03 at 1 14 21 PM

### New view
Screenshot 2025-09-03 at 1 14 15 PM

----

- Verify all tests go green


kevinmarete (approved):
Very nice!

fboulnois (approved):
looks good 👍

PR #3037, opened 2025-09-05
[DT-2162] Error not displayed when cannot set role.

otchet-broad:
### Addresses

[DT-4010](https://broadworkbench.atlassian.net/browse/DT-4010).

Security risk: no.

### Summary

`MatchAlgorithm.version` was package-private and non-final with a public `setVersion`. Enum constants are JVM-wide singletons, so one call anywhere would change the version stamped on every subsequent match row for the life of the process. The field is now private and final and the setter is gone; nothing called it.

The three-argument `matchFailure` and four-argument `matchSuccess` defaulted to `V4` with no caller in main or test — `MatchService` passes `V5` explicitly at both sites. A future caller would have stamped `v4` on a row produced by V5 logic. Both are deleted and the explicit-algorithm overloads cover every use.

`V1` through `V3` stay declared: production still holds `v1` and `v2` rows and duos-ui matches on those literals. No API contract change, so `api-docs.yaml` is untouched.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Verify all tests go green

[DT-4010]: https://broadworkbench.atlassian.net/browse/DT-4010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/libs/ajax/User.ts
+++ b/src/libs/ajax/User.ts
@@ -1,7 +1,7 @@
 import { cloneDeep, flow, unset, mergeAll } from 'lodash/fp'
 import { Config } from '../config'
 import axios from 'axios'
-import { getApiUrl, fetchOk, fetchAny } from '../ajax'
+import { getApiUrl, fetchOk } from '../ajax'
 import { CreateDuosUserResponse, DuosUserResponse, UpdateDuosUserResponse } from 'src/types/responseTypes'
 import { CreateDuosUserRequest, UpdateDuosUserRequestV1, UpdateDuosUserRequestV2 } from 'src/types/requestTypes'
 import {
--- a/src/libs/ajax/User.ts
+++ b/src/libs/ajax/User.ts
@@ -130,20 +130,14 @@ export const User = {
 
   addRoleToUser: async (userId: number, roleId: number): Promise<DuosUserResponse> => {
     const url = `${await getApiUrl()}/api/user/${userId}/${roleId}`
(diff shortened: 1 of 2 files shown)


kevinmarete (approved):

src/libs/ajax/User.ts
```diff
@@ -130,20 +130,14 @@ export const User = {
 
   addRoleToUser: async (userId: number, roleId: number): Promise<DuosUserResponse> => {
     const url = `${await getApiUrl()}/api/user/${userId}/${roleId}`
-    const res = await fetchAny(
-      url,
-      mergeAll([Config.authOpts(), { method: 'PUT' }]),
-    )
-    return res.json()
+    const res = await axios.put(url, null, Config.authOpts())
```
rushtong, line 133: See https://broadworkbench.atlassian.net/browse/DT-2086
Not sure we want to add axios here if we are going to move away from it

rushtong (approved):
Looks good, thank you.

src/libs/ajax/User.ts
otchet-broad: There's a good bit to we'll need to add in order to extract the error from the body with fetch.  I didn't want to hold the fix up with that scope.

PR #3056, opened 2025-09-22
[DT-2038] accessibility discernible name

brack2891:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2038

### Summary
updating name within images for clarity
----

- Verify all tests go green


config/base_config.json
rushtong, line 1: Changes to this file should be reverted.

fboulnois (approved):
I agree with @rushtong 's comment, otherwise looks good.

src/pages/researcher_console/DatasetSubmissionsTable.jsx
```diff
@@ -134,6 +134,7 @@ export default function DatasetSubmissionsTable(props) {
                 className="glyphicon glyphicon-trash"
                 onClick={() => handleClick(term)}
                 to="#"
+                aria-label={`Delete set '${selectedTerm.datasetIdentifier}'`}
```
rushtong: This link is operating on `term` ... is there a reason we're showing the identifier for `selectedTerm` instead? I do see that we're using `selectedTerm` in the following `ConfirmationDialog` block, but that might be a bug, can you confirm that we're using the correct variable in this div?
```suggestion
                aria-label={`Delete dataset '${term.datasetIdentifier}'`}
```

otchet-broad: ‸‸‸‸. (agree with Greg's point about using `term` instead of `selectedTerm`)

otchet-broad (changes requested):

src/pages/researcher_console/DatasetSubmissionsTable.jsx
brack2891: @rushtong and @otchet-broad - updating to reflect suggested changes

otchet-broad (approved):
👍

PR #3058, opened 2025-09-23
[DT-2197] Add NIH Institutional Certification link and download capability.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2197](https://broadworkbench.atlassian.net/browse/DT-2197)

### Summary

Adds the download link to the UI
Wires the service to the download link, fetching the filename from the dataset using the API
Adds boolean column sorting to simple table (with tests)

AFTER:

image

----

- Verify all tests go green

### DIFF
--- a/src/components/dac_dataset_table/DACDatasetTableCellData.jsx
+++ b/src/components/dac_dataset_table/DACDatasetTableCellData.jsx
@@ -27,11 +27,7 @@ export function nihCertificationLinkData({
           <div>
             <button
               onClick={async () => {
-                const datasetInfo = await DataSet.getDataSetsByDatasetId(
-                  dataset.datasetId)
-                const fileName = datasetInfo.nihInstitutionalCertificationFile.fileName
-                DataSet.getNIHInstitutionalCertification(dataset.datasetId,
-                  fileName)
+                DataSet.getNIHInstitutionalCertification(dataset.datasetId)
               }}
               className="button button-white"
               style={{ padding: '10px 12px' }}
--- a/src/components/dac_dataset_table/DACDatasetTableCellData.jsx
+++ b/src/components/dac_dataset_table/DACDatasetTableCellData.jsx
@@ -3,6 +3,7 @@ import style from 'src/pages/DACDatasets.module.css'
 import { styles } from './DACDatasetConstants'
 import DACDatasetApprovalStatus from './DACDatasetApprovalStatus'
 import { createDataUseDisplay, processDataUseCodes } from 'src/utils/DataUseUtils.js'
+import { DataSet } from 'src/libs/ajax/DataSet.js'
 
 export const consoleTypes = { CHAIR: 'chair' }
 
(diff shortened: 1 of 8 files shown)


src/libs/utils.js
```diff
@@ -529,7 +529,7 @@ export const sortVisibleTable = ({ list, sort }) => {
     return list.sort((a, b) => {
       const aVal = a[sort.colIndex].value || a[sort.colIndex].data
       const bVal = b[sort.colIndex].value || b[sort.colIndex].data
-      if (typeof aVal === 'number') {
+      if (typeof aVal === 'number' || typeof aVal === 'boolean') {
```
otchet-broad, line 532: This adds the ability to sort on boolean values so you can sort on the presence of a file to download.

src/components/dac_dataset_table/DACDatasetsTable.jsx
```diff
@@ -1,12 +1,11 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useCallback } from 'react'
 import { Storage } from '../../libs/storage'
```
rushtong: It would be nice to update these imports, but definitely not required.

src/components/dac_dataset_table/DACDatasetTableCellData.jsx
```diff
@@ -16,6 +17,37 @@ export function duosIdCellData({ dataset, label = 'duosIdCellData' }) {
...
+    data: dataset.hasInstitutionCertification
+      ? (
+          <div>
+            <button
+              onClick={async () => {
+                const datasetInfo = await DataSet.getDataSetsByDatasetId(
+                  dataset.datasetId)
+                const fileName = datasetInfo.nihInstitutionalCertificationFile.fileName
+                DataSet.getNIHInstitutionalCertification(dataset.datasetId,
+                  fileName)
```
rushtong: Could these two calls be combined into a single method in `DataSet.js`? IOW, could `DataSet.getNIHInstitutionalCertification` just take an ID, make the call to `getDataSetsByDatasetId` for the file name, and then call the actual file download? It's a little bit of a different pattern for us, but it would simplify this button construction and hide the abstraction behind `getNIHInstitutionalCertification`.

rushtong (approved):
Looks pretty good, thank you 👍🏽

src/components/dac_dataset_table/DACDatasetTableCellData.jsx
otchet-broad: They can definitely be combined - and you're right in that it hasn't been our pattern to do that.  I'll put them into a single method and we can see if we like it better.

otchet-broad: @rushtong - let me know what you think of the change.

rushtong: @otchet-broad Nice 👍🏽

kevinmarete (approved):

PR #3060, opened 2025-09-23
DT-2246: Bug fix for Chair/Member Voting Tabs

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2246

### Summary
Currently, we are using the endpoint to get a collection with a full set of election history items to populate similar structures in different tabs. Previously, we populated the vote tabs with the singular dar collection (most recent one), but started using the full collection (with history) when we added the vote history tab. This made a degree of sense since the singular collection is a subset of the full collection with history. However, the bucketing utility doesn't correctly parse out the current set of elections from the full history. It was putting extra votes into the voting tabs that were for previously closed elections which created a number of UI bugs.

This PR:
* Uses the singular dar collection API to populate the voting tabs
* Continues to use the collection with history to populate the vote history tab
* Small update to the vote button functionality to set buttons to disabled while a call to the vote API is taking place. This prevents users from submitting conflicting votes in quick succession which can trigger an error in the chairperson vote tabs.
* Small bug fix to always set the vote state after it has been updated instead of in an else block.

----

- Verify all tests go green

### DIFF
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
@@ -192,6 +192,7 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props) 
   }
 
   const onVoteError = (error: unknown, isChair: boolean) => {
+    setVoteInProgress(false)
     const consentError = extractConsentError(error)
     if (consentError && consentError.code === 409) {
       const voteText = isChair ? 'Chair vote' : 'Vote'
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
@@ -132,6 +132,7 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props) 
   const [isVotingDisabled, setIsVotingDisabled] = useState<boolean>(false)
   const [isRadar, setIsRadar] = useState<boolean>(false)
   const [isElectionClosed, setIsElectionClosed] = useState<boolean>(false)
+  const [voteInProgress, setVoteInProgress] = useState<boolean>(false)
   const {
     question,
     votes,
@@ -173,6 +174,7 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props) 
   }, [votes])
 
   const updateVote = async (newVote: boolean, isChair: boolean) => {
+    setVoteInProgress(true)
     const openElectionVotes = votes.filter(v => v.electionStatus?.toLowerCase() === 'open')
     const voteIds = openElectionVotes.map(v => v.voteId)
     await Votes.updateVotesByIds(voteIds, { vote: newVote, rationale })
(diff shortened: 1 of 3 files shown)


src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
```diff
@@ -181,14 +183,14 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props)
     if (isChair && updateFinalVote && bucketKey) {
       updateFinalVote(bucketKey, { vote: newVote, rationale }, voteIds)
     }
-    else {
-      setVote(newVote)
-    }
+    setVote(newVote)
```
rushtong, line 186: This is a small bug fix. After updating a vote value on the back end, we should always set the local state so the button shows up correctly.

src/pages/dar_collection_review/DarCollectionReview.jsx
```diff
@@ -163,12 +165,9 @@ export default function DarCollectionReview(props) {
   const init = useCallback(async () => {
     const user = Storage.getCurrentUser()
     try {
-      let collection
+      const collection = await Collections.getCollectionById(collectionId)
```
rushtong, line 168: This is the bulk of the bug fix.

src/pages/dar_collection_review/DarCollectionReview.jsx
```diff
@@ -221,8 +220,9 @@ export default function DarCollectionReview(props) {
       setSubcomponentLoading(true)
       init()
     }
-    catch (_error) {
-      Notifications.showError({ text: 'Failed to initialize collection' })
+    catch (error) {
+      const message = extractError(error)
+      Notifications.showError({ text: 'Failed to initialize collection: ' + message })
     }
```
rushtong, line 226: Drive by fix for sonar warning

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
```diff
@@ -247,15 +249,15 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props)
                     <CollectionVoteYesButton
                       onClick={async () => await updateVote(true, !!updateFinalVote)}
                       onError={(error: unknown) => onVoteError(error, !!updateFinalVote)}
-                      disabled={isVotingDisabled || isApprovalDisabled || isLoading || isElectionClosed}
+                      disabled={voteInProgress || isVotingDisabled || isApprovalDisabled || isLoading || isElectionClosed}
```
rushtong, line 252: Here and in the `CollectionVoteNoButton`, this fixes the problem where a slow API call can leave both buttons enabled after clicking either one. If the user does click the other button, there can be a conflict on the back end if the vote is a final vote. Instead, we disable when an API call is actively operating and reset that state after the API completes.

otchet-broad (approved):
Good find!
Thanks for the other fixes!

src/pages/dar_collection_review/DarCollectionReview.jsx
eweitz: Nice debuggability refinement!

eweitz (approved):

PR #3063, opened 2025-09-25
[DT-2269] Don't redirect the /api/user/me call if it fails with a 401 

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2269](https://broadworkbench.atlassian.net/browse/DT-2269)

### Summary

The backend will return a 401 when /api/user/me is called.  The SigninButton.tsx code depends on the logic continuing to flow so we must exclude this call from the traditional 401 handling so that the user account can be created.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -18,7 +18,9 @@ axios.interceptors.response.use(function (response) {
 }, function (error) {
   // Default to a 502 when we can't get a real response object.
   const status = getOr(502)('response.status')(error)
-  if (status === 401) {
+  const url = error.config.url
+  const data = error.config.data
+  if (status === 401 && !(url.endsWith('/api/user/me') && data === undefined)) {
     redirectOnLogout()
   }
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -20,9 +20,8 @@ axios.interceptors.response.use(function (response) {
   const status = getOr(502)('response.status')(error)
   const reportUrl = getOr(null)('response.config.url')(error)
   const method = getOr(null)('response.config.method')(error)
-  console.error(reportUrl, method)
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -18,7 +18,9 @@ axios.interceptors.response.use(function (response) {
 }, function (error) {
   // Default to a 502 when we can't get a real response object.
   const status = getOr(502)('response.status')(error)
-  if (status === 401) {
+  const url = error.config.url
+  const data = error.config.data
+  if (status === 401 && !(url.endsWith('/api/user/me') && data === undefined)) {
```
kevinmarete: Nit suggestions..

```suggestion
    const { url, data } = error.config ?? {};
    const isMeCheck = url?.endsWith("/api/user/me") && data === undefined;
    
    if (status === 401 && !isMeCheck) {
      redirectOnLogout();
    }
```

kevinmarete (approved):

src/libs/ajax.js
rushtong: A bit of history here, the `status === 401` check has caused us grief in the past, especially when integrating with external services ... if they throw a 401, we log the user out incorrectly.

rushtong (approved):

src/libs/ajax.js
otchetbluehalo: Started doing this refactor after I reviewed the PR for myself.  Take a look at how it ended up.

otchet-broad: @kevinmarete - I started doing this refactor after I reviewed the PR for myself. Take a look at how it ended up.

kevinmarete (approved):
Looks good!

PR #3064, opened 2025-09-25
adding Dataset ID to the main level of the Voting History Tab

brack2891:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2247

### Summary
The Dataset ID needed to be added to the Voting History Tab under the Data Access Request Review. Also updated the Cypress tests.

*New Screenshots*
Screenshot 2025-09-26 at 3 38 24 PM

But you will notice that the `VOTE TYPE` field is overlapping `RATIONALE`, so I also adjusted the column width to correct this overlap. Now it looks like:
Screenshot 2025-09-26 at 3 46 49 PM

----

- Verify all tests go green

### DIFF
--- a/src/pages/dar_collection_review/VotingHistory.tsx
+++ b/src/pages/dar_collection_review/VotingHistory.tsx
@@ -78,7 +78,7 @@ const extractChairVotes = (darCollection: DarCollection, filteredDatasetIds: num
           if (isChairVote(vote)) {
             votesByRole.push({
               ...vote,
-              darTitle: dar.data.projectTitle,
+              datasetId: election.datasetId,
               progressReport: dar.progressReport,
               electionDate: election.createDate,
             })
@@ -103,7 +103,7 @@ const extractElectionsWithMemberVotes = (darCollection: DarCollection, filteredD
       .forEach((election: Election) => {
         const electionWithMemberVotes: ElectionWithMemberVotes = {
           ...election,
-          darTitle: dar.data.projectTitle,
+          datasetId: election.datasetId,
           progressReport: dar.progressReport,
           memberVotes: [],
         }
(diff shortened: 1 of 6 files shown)


eweitz (approved):
Code and [screenshot](https://broadinstitute.slack.com/archives/C07T2DLHD24/p1758807174791069) with noted refinements look good!

src/pages/dar_collection_review/VotingHistory.tsx
```diff
@@ -103,7 +103,7 @@ const extractElectionsWithMemberVotes = (darCollection: DarCollection, filteredD
       .forEach((election: Election) => {
         const electionWithMemberVotes: ElectionWithMemberVotes = {
           ...election,
-          darTitle: dar.data.projectTitle,
+          datasetId: election.datasetId,
```
fboulnois: I believe this line is redundant since `...election` will already expand the `datasetId`.

fboulnois (commented):
See below:

src/pages/dar_collection_review/VotingHistory.tsx
brack2891: Agreed, removing. Thank you

rushtong:
I'm confused by the value we're showing on screen. The database ID doesn't really tell me which dataset the vote is for. I think a better choice is to show the dataset identifier. Additionally, we should link it to the dataset page.

brack2891:
> I'm confused by the value we're showing on screen. The database ID doesn't really tell me which dataset the vote is for. I think a better choice is to show the dataset identifier. Additionally, we should link it to the dataset page.

@rushtong is this something that would be worth discussing in a standup as a parking lot?

rushtong:
> > I'm confused by the value we're showing on screen. The database ID doesn't really tell me which dataset the vote is for. I think a better choice is to show the dataset identifier. Additionally, we should link it to the dataset page.
> 
> @rushtong is this something that would be worth discussing in a standup as a parking lot?

Great idea, let's do that.

rushtong (approved):
Thank you for the updates 👍🏽

PR #3065, opened 2025-09-25
[DT-2255] Add AI/LLM use to DAR

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2255

### Summary

Add AI and LLM use to DAR. Tested with unit tests. This is the first part that adds the question, validates the response, and records the answer. Banners will be added in a separate PR.

----

- Verify all tests go green


src/libs/dataUseTranslation.js
```diff
@@ -59,6 +59,12 @@ export const srpTranslations = {
     manualReview: false,
     type: ControlledAccessType.modifiers,
   },
+  aiLlmUse: {
+    code: 'AI',
+    description: 'The research involves the use of Artificial Intelligence (AI) or Large Language Models (LLMs).',
+    manualReview: true,
```
fboulnois, line 65: I wasn't sure about this field value, since it's not a "real" DUO term.

src/pages/dar_application/ResearchPurposeStatement.jsx
```diff
@@ -68,6 +68,16 @@ export default function ResearchPurposeStatement(props) {
 
           <h4>I am proposing to:</h4>
 
+          <ResearchPurposeRow
+            title="Do research involving AI or Large Language Models (LLMs). If yes, please explain in your Research Use Statement."
```
fboulnois, line 72: This is the core change.

src/libs/dataUseTranslation.js
rushtong: `manualReview` is somewhat of an outdated field. This is the right thing to do in this case. We do need to re-evaluate the usages and see if we can remove it.

rushtong (approved):
Looks good, thank you 👍🏽

eweitz (approved):

PR #3070, opened 2025-09-29
DT-2282: Force the response object to be a DuosUser

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2282

### Summary
Small type fix to correctly pass a `DuosUser` to the extract auth state function.

This uncovers a larger issue with the User.ts class that is returning `DuosUserResponse` objects which is incorrect, they should be `DuosUser` objects, but that can be tackled separately from this bug.

----

- Verify all tests go green


otchet-broad (approved):
👍

kevinmarete (approved):

PR #3071, opened 2025-09-29
DT-2284: Standardize on DuosUser

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2284

### Summary
Use `DuosUser` in all places where we were using `DuosUserResponse`, `CreateDuosUserResponse` and `UpdateDuosUserResponse`.

----

- Verify all tests go green


src/types/model.ts
```diff
@@ -2,6 +2,11 @@ import externalAccessIcon from 'src/images/external_access.svg'
 import openAccessIcon from 'src/images/open_access.svg'
 import controlledAccessIcon from 'src/images/controlled_access.svg'
 
+export interface ConsentError {
+  readonly message?: string
+  readonly code?: number
+}
```
rushtong, line 8: Copied from `responseTypes.ts` where is was the only interface left after cleanup.

kevinmarete (approved):
Nice cleanup!

src/components/SignInButton.tsx
```diff
@@ -14,7 +14,6 @@ import { StackdriverReporter } from '../libs/stackdriverReporter'
 import { History } from 'history'
 import { OidcUser } from '../libs/auth/oidcBroker'
```
fboulnois: nit: update imports

src/libs/ajax/User.ts
```diff
@@ -52,7 +51,7 @@ export const User = {
     }
   },
 
-  updateSelf: async (payload: UpdateDuosUserRequestV1): Promise<UpdateDuosUserResponse> => {
+  updateSelf: async (payload: UpdateDuosUserRequestV1): Promise<DuosUser | false | undefined> => {
```
fboulnois, line 54: why can these return `false` or `undefined`?

fboulnois (approved):
See below:

src/libs/ajax/User.ts
rushtong: That's what they do now: https://github.com/DataBiosphere/duos-ui/blob/343e84c16100b9cd2a14976b3094caa65eb2141a/src/types/responseTypes.ts#L15
In an effort to minimize unexpected behaviors, I kept that same pattern here.

PR #3088, opened 2025-10-02
[DT-2307] Downgrade Node LTS to 24.7.0 for DevContainer

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2307

### Summary

Downgrades Node LTS to 24.7.0 to fix the DevContainer build and locks the OS to Debian 13 Trixie.

----

- Verify all tests go green

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "not op_mini all"
   ],
   "engines": {
-    "node": ">=24.9.0 <25"
+    "node": ">=24.7.0 <25"
   },
   "type": "module"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
     "name": "Node.js",
-    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:24",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "not op_mini all"
   ],
   "engines": {
-    "node": ">=24.9.0 <25"
+    "node": ">=24.7.0 <25"
```
kevinmarete, line 83: What happens when dependabot upgrades this to 24.9.0? Shouldn't we lock it to `24.7.0`

kevinmarete (commented):
Small clarification!

package.json
fboulnois: This block of config is manually set, and Dependabot does not change it. Previous updates (such as #3084 ) only modify the Dockerfile. This block only specifies a minimum required LTS version, if an LTS is newer than it should work without issue.

Perhaps a better solution would be to also ensure the DevContainer uses the latest available Node LTS. I can ticket that as a follow-up.

kevinmarete (approved):

otchet-broad (approved):
👍
Thank you!

rushtong (approved):

PR #3089, opened 2025-10-03
[DT-2310] Ensure DevContainer build version matches Dockerfile

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2310

### Summary

Sorry for the version change fatigue. This sets up the DevContainer to use the same Node version as the one in the Dockerfile.

----

- Verify all tests go green

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -70,7 +70,7 @@ gcloud auth application-default login
 gcloud auth configure-docker
 
 # install the appropriate version of nodejs and its dependencies
-NODE_VERSION=$(curl -L https://raw.githubusercontent.com/DataBiosphere/duos-ui/develop/Dockerfile | awk 'NR==2 {gsub(":","@",$2); print $2}')
+NODE_VERSION=$(curl -L https://raw.githubusercontent.com/DataBiosphere/duos-ui/develop/Dockerfile | awk 'NR==2 {gsub(/node:|-.*/,"", $2); print "node@" $2}')
 volta setup && volta install ${NODE_VERSION}
 ```
 
--- a/scripts/setup-devcontainer.sh
+++ b/scripts/setup-devcontainer.sh
@@ -17,6 +17,12 @@ install_gcloud_cli() {
   sudo apt install -y google-cloud-cli
 }
 
+install_node_with_nvm() {
+  NODE_VERSION=$(awk 'NR==2 {gsub(/node:|-.*/,"", $2); print $2}' Dockerfile)
+  . "${NVM_DIR}/nvm.sh"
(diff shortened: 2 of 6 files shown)


docs/getting-started.md
```diff
@@ -70,7 +70,7 @@ gcloud auth application-default login
 gcloud auth configure-docker
 
 # install the appropriate version of nodejs and its dependencies
-NODE_VERSION=$(curl -L https://raw.githubusercontent.com/DataBiosphere/duos-ui/develop/Dockerfile | awk 'NR==2 {gsub(":","@",$2); print $2}')
+NODE_VERSION=$(curl -L https://raw.githubusercontent.com/DataBiosphere/duos-ui/develop/Dockerfile | awk 'NR==2 {gsub(/node:|-.*/,"", $2); print "node@" $2}')
```
fboulnois, line 73: This is needed now that we specify the OS version.

eweitz (approved):

scripts/setup-devcontainer.sh
```diff
@@ -17,6 +17,12 @@ install_gcloud_cli() {
   sudo apt install -y google-cloud-cli
 }
 
+install_node_with_nvm() {
+  NODE_VERSION=$(awk 'NR==2 {gsub(/node:|-.*/,"", $2); print $2}' Dockerfile)
+  . "${NVM_DIR}/nvm.sh"
+  nvm install "$NODE_VERSION"
```
otchet-broad: Silly question, but why are we using both volta and nvm?

otchet-broad (approved):

scripts/setup-devcontainer.sh
fboulnois: The Microsoft DevContainer build has `nvm` built-in, and it would require several more steps to install `volta`. I would still recommend `volta` for our personal deployments, though, since it has some other advantages that I've laid out previously in Slack.

otchet-broad: No disagreement with the benefits of volta.  I was just surprised to see us using both.

fboulnois: Would it be preferable to use Volta in the container, then? I wasn't able to get this working since I think the built-in `nvm` command interferes with how `volta` installs Node, but I could spend additional time trying to get this to work.

otchet-broad: IMHO, I think the fewer things we have to maintain over the long haul, the better.  I'd be in favor of you spending some time to move us from nvm to volta.

otchet-broad (approved):
👍
Nice!

rushtong (approved):
👍🏽

PR #3095, opened 2025-10-09
[DT-2086] Replace Axios with Built-in Fetch

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2086

### Summary

Update code to use Node.js’s built-in fetch instead of Axios, following 2025 best practices. This removes an external dependency and modernizes our HTTP request handling.

----

- Verify all tests go green


src/libs/ajax.js
```diff
@@ -1,37 +1,15 @@
...
-axios.interceptors.response.use(function (response) {
-  return response
-}, function (error) {
-  // Default to a 502 when we can't get a real response object.
-  const status = getOr(502)('response.status')(error)
-  const reportUrl = getOr(null)('response.config.url')(error)
-  const method = getOr(null)('response.config.method')(error)
-  const isMeCheck = reportUrl?.endsWith('/api/user/me') && (method?.toLowerCase() === 'get')
-  if (status === 401 && !isMeCheck) {
-    redirectOnLogout()
```
eweitz, line 25: This handling looks relevant, but I don't see it ported in the new code.

src/libs/ajax.js
```diff
@@ -1,37 +1,15 @@
...
 const redirectOnLogout = () => {
   Auth.signOut()
   window.location.href = `/home?redirectTo=${window.location.pathname}`
 }
 
-axios.interceptors.response.use(function (response) {
-  return response
-}, function (error) {
-  // Default to a 502 when we can't get a real response object.
-  const status = getOr(502)('response.status')(error)
```
eweitz, line 20: This case also seems relevant but dropped in the updated code.  Specifically, the new code seems like it wouldn't trigger the `status >= 500` handling when we can't get a real response object (i.e., I assume where a `status` HTTP response header isn't set).

src/libs/ajax/fetchAdapter.js
```diff
@@ -0,0 +1,88 @@
+function buildUrlWithParams(url, params = {}) {
+  const query = new URLSearchParams(params).toString()
+  return query ? `${url}?${query}` : url
+}
+
+async function handleResponse(res, url, responseType) {
+  if (!res.ok) {
+    if (res.status >= 500) { // Alternative for interceptor's reportError
+      reportError(url, res.status)
+    }
```
eweitz: This seems like a bug.  I'm aware of two `reportError` functions in DUOS UI:

1.  The custom function `reportError` that's defined in `ajax.js`, and which reports to StackDriver
2. The global function [`reportError`](https://developer.mozilla.org/en-US/docs/Web/API/Window/reportError) that's a web API built-in, which does not

I think the intent is (1).

src/libs/ajax/DataSet.js
```diff
@@ -71,20 +71,25 @@ export const DataSet = {
 
   updateStudy: async (studyId, studyObject) => {
     const url = `${await getApiUrl()}/api/dataset/study/${studyId}`
-    return await axios.put(url, studyObject, Config.multiPartOpts())
+    const res = await fetchMultipart(url, studyObject, Config.multiPartOpts(), 'PUT')
+    return res.data
```
eweitz, line 75: I like this incidental outcome of assigning the completed promise, then returning its resolved `data` property.  Returning without the interstitial assignment is more common in some codebases, but I think the easier debuggability here is preferable to the briefer and denser prior code.

src/libs/ajax/DAR.js
```diff
@@ -1,65 +1,63 @@
...
-    return await res.json()
+    const res = await fetchGet(url, Config.authOpts())
+    return res.data
   },
 
   // v2, v3 Draft DAR Update
   updateDarDraft: async (dar, referenceId) => {
-    // noinspection ES6MissingAwait
-    Metrics.captureEvent(eventList.dar, { action: 'update' })
+    await Metrics.captureEvent(eventList.dar, { action: 'update' })
```
eweitz: Let's not `await` here.  It's preferable to have the Metrics request initiate and process in a separate thread in a non-blocking manner.

```suggestion
    // noinspection ES6MissingAwait
    Metrics.captureEvent(eventList.dar, { action: 'update' })
```

src/libs/ajax/DAR.js
```diff
@@ -1,65 +1,63 @@
...
-    const res = await axios.put(url, dar, Config.authOpts())
+    const res = await fetchPut(url, dar, Config.authOpts())
     return res.data
   },
 
   // v2, v3 Draft DAR Creation
   postDarDraft: async (dar) => {
-    // noinspection ES6MissingAwait
-    Metrics.captureEvent(eventList.dar, { action: 'draft' })
+    await Metrics.captureEvent(eventList.dar, { action: 'draft' })
```
eweitz:
```suggestion
    // noinspection ES6MissingAwait
    Metrics.captureEvent(eventList.dar, { action: 'draft' })
```

eweitz (commented):
I'm glad we're migrating to standard `fetch`!

I note some potential issues and refinements to consider.

src/libs/ajax.js
kevinmarete: Handled here: https://github.com/DataBiosphere/duos-ui/pull/3095/commits/1a22d72c2de9c9ac7d6e82558222eff9ea401508

src/libs/ajax.js
kevinmarete: Handled here: https://github.com/DataBiosphere/duos-ui/pull/3095/commits/1a22d72c2de9c9ac7d6e82558222eff9ea401508

src/libs/ajax/fetchAdapter.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+async function handleResponse(res, url, responseType, method = 'GET') {
+  const isMeCheck = url?.endsWith('/api/user/me') && (method?.toLowerCase() === 'get')
+  if (!res.ok) {
+    if (res.status === 401 && !isMeCheck) {
+      redirectOnLogout()
+    }
+    if (res.status >= 500) {
+      await reportError(url, res.status)
```
eweitz: If a function only does reporting, logging, analytics, or metrics, then blocking execution flow on the return of that function is generally not needed.  The cost is unnecessary slowness -- potentially quite a bit of slowness if the logging service is having issues -- and there's typically no benefit as the function doesn't return any value.  So omitting `await` in such cases helps speed and graceful degradation.


```suggestion
      // Don't `await`, as we don't need the result
      reportError(url, res.status)
```

src/libs/ajax/fetchAdapter.js
```diff
@@ -0,0 +1,68 @@
...
+    headers: isMultipart ? headers : { 'Content-Type': 'application/json', ...headers },
+    credentials,
+    body: getRequestBody(data, isMultipart),
+  }
+  try {
+    const res = await fetch(fullUrl, fetchOptions)
+    return handleResponse(res, fullUrl, responseType, method)
+  }
+  catch {
+    await reportError(fullUrl, 502) // Default to a 502 when we can't get a real response object.
```
eweitz:
```suggestion
    reportError(fullUrl, 502) // Default to a 502 when we can't get a real response object.
```

src/libs/ajax/fetchAdapter.js
```diff
@@ -0,0 +1,68 @@
...
+  fetchRequest({ url, data, ...config, method: 'POST' })
+
+export const fetchPut = (url, data, config = {}) =>
+  fetchRequest({ url, data, ...config, method: 'PUT' })
+
+export const fetchPatch = (url, data, config = {}) =>
+  fetchRequest({ url, data, ...config, method: 'PATCH' })
+
+export const fetchDelete = (url, config = {}) =>
+  fetchRequest({ url, data: config.data, ...config, method: 'DELETE' })
```
eweitz, line 62: Elegant simplification!

src/libs/ajax/AuthenticateNIH.js
```diff
@@ -1,56 +1,45 @@
...
       return res.data
     }
     return undefined
   },
 
   getSyncedUser: async () => {
     const url = `${await getApiUrl()}/api/nih/sync`
-    const res = await axios.get(url, merge(Config.authOpts(), { headers: { 'Content-Type': 'application/json' } }))
-    return await res.data
+    const config = merge(Config.authOpts(), { headers: { 'Content-Type': 'application/json' } })
```
eweitz: `'Content-Type': 'application/json'`  is a default in `fetchRequest` so it's not needed here, right?

src/libs/ajax/Support.js
```diff
@@ -17,20 +17,22 @@ export const Support = {
 
   createSupportRequest: async (ticket) => {
     const url = `${await getApiUrl()}/support/request`
-    return await axios.post(url, ticket, { headers: { 'Content-Type': 'application/json' } }).catch(
-      function (error) {
-        return Promise.reject(error.response)
-      },
-    )
+    try {
+      return await fetchPost(url, ticket, { headers: { 'Content-Type': 'application/json' } })
```
eweitz: `'Content-Type': 'application/json'`  is a default in `fetchRequest` so it's not needed here, right?

package.json
```diff
@@ -9,7 +9,6 @@
     "@mui/x-date-pickers": "8.12.0",
     "ajv": "8.17.1",
     "ajv-formats": "3.0.1",
-    "axios": "1.12.2",
```
eweitz, line 12: 🎉

eweitz (approved):
Nice refinements!  I suggest a few more for performance, robustness, and maintainability, but this overall looks good to me.

otchet-broad:
General though here with some testing:
We've got error handling that likely needs to be addressed in this PR.  The screenshot below demonstrates this.  We had a compliance requirement to return specific text from the backend in this case.  As you can see, with this change that no longer is happening.

image

In this specific case, we have code that relied on the axios behavior of exposing the error status code and response message so we could display the error from the backend.  The code affected in this case is in [accessUtils.tsx, line 16](https://github.com/DataBiosphere/duos-ui/blob/1a2e17b8cacd29117164df2b975c698940550a6c/src/utils/accessUtils.tsx#L16).  I'm almost certain we have other cases of this as well.

src/libs/ajax/Support.js
```diff
@@ -17,20 +17,22 @@ export const Support = {
...
-    return await axios.post(url, file, { headers: { 'Content-Type': 'application/binary' } }).catch(
-      function (error) {
-        return Promise.reject(error.response)
-      },
-    )
+    try {
+      return await fetchPost(url, file, { headers: { 'Content-Type': 'application/binary' } })
+    }
+    catch (error) {
+      return Promise.reject(error)
```
otchet-broad: Maybe a pattern of throwing a new consent Error object decorated with status code and message might be a way to deal with the issue we've been discussing AND the sonar feedback on this pattern.

kevinmarete: https://github.com/DataBiosphere/duos-ui/pull/3095/commits/28610a6c25a4346edb9e82cbd68ae4a898bf9f6b

kevinmarete:
Fixed here: https://github.com/DataBiosphere/duos-ui/pull/3095/commits/d164f409025127f7f949a430f43f3369e3449777

Screenshot 2025-10-10 at 3 07 16 PM


> General though here with some testing: We've got error handling that likely needs to be addressed in this PR. The screenshot below demonstrates this. We had a compliance requirement to return specific text from the backend in this case. As you can see, with this change that no longer is happening.
> 
> image
> In this specific case, we have code that relied on the axios behavior of exposing the error status code and response message so we could display the error from the backend. The code affected in this case is in [accessUtils.tsx, line 16](https://github.com/DataBiosphere/duos-ui/blob/1a2e17b8cacd29117164df2b975c698940550a6c/src/utils/accessUtils.tsx#L16). I'm almost certain we have other cases of this as well.

src/components/library_card_table/LibraryCardTable.tsx
```diff
@@ -144,9 +143,11 @@ const deleteOnClick = (
...
   }
   catch (error: unknown) {
-    const axiosError = error as AxiosError
-    const consentError = axiosError?.response?.data as ConsentError
-    const serverError = consentError.message ?? 'Error: Failed to delete library card'
+    // fetch-based errors are generic Error objects
+    let serverError = 'Error: Failed to delete library card'
+    if (error instanceof Error && error.message) {
+      serverError = error.message
+    }
```
rushtong: This is looks a lot like `extractError` in `ErrorUtils`. IIRC, the old code here was the precursor to `ErrorUtils` and was never replaced by that utility.

src/libs/ajax/AuthenticateNIH.js
```diff
@@ -1,56 +1,43 @@
...
-import { Config } from '../config'
-import { getECMUrl, getApiUrl, reportError } from '../ajax'
-import { get, isNil, merge } from 'lodash'
+import { Config } from 'src/libs/config'
+import { getECMUrl, getApiUrl } from 'src/libs/ajax'
+import { fetchGet, fetchPost, fetchDelete } from 'src/libs/ajax/fetchAdapter'
 
 /**
- * ECM has several different providers such as `era-commons`, `ras`, `github`, `fence`, and others. DUOS has
+ * ECM has several different providers such as `era-commons`, `ras`, `GitHub`, `fence`, and others. DUOS has
```
rushtong: This is incorrect, the correct value is `github` ... see https://externalcreds.dsde-prod.broadinstitute.org/#/oauth/listProviders for where these come from.

rushtong (approved):
Love the adapter pattern 👍🏽 Lots of great updates here, see minor comments inline, address as you see fit.

otchet-broad (approved):
Nice work!

Thanks for addressing the concerns!

PR #3113, opened 2025-10-15
DT-400: Replace validator.isFQDN with local copy

rushtong:
### Addresses

### Summary
Remove the validator library and copy current usage to a local utility function. See https://github.com/validatorjs/validator.js/blob/6f436be36945e460ee624bf72a935a06daded859/src/lib/isFQDN.js, https://github.com/validatorjs/validator.js/blob/6f436be36945e460ee624bf72a935a06daded859/src/lib/util/assertString.js, and https://github.com/validatorjs/validator.js/blob/6f436be36945e460ee624bf72a935a06daded859/src/lib/util/merge.js for the original source.

This PR can be reverted once the validator library vulnerabilities are updated.

----

- Verify all tests go green

### DIFF
--- a/src/utils/isFQDN.js
+++ b/src/utils/isFQDN.js
@@ -1,3 +1,5 @@
+// Code from https://github.com/validatorjs/validator.js
+// Original license: https://github.com/validatorjs/validator.js/blob/master/LICENSE
 const default_fqdn_options = {
   require_tld: true,
   allow_underscores: false,
--- a/src/utils/isFQDN.js
+++ b/src/utils/isFQDN.js
@@ -1,0 +1,87 @@
+const default_fqdn_options = {
+  require_tld: true,
+  allow_underscores: false,
+  allow_trailing_dot: false,
+  allow_numeric_tld: false,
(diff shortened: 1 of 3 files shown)


src/utils/isFQDN.js
```diff
@@ -0,0 +1,87 @@
+const default_fqdn_options = {
```
otchet-broad, line 3: Should we indicate in the source that includes code from validator.js and that it was directly included because of the issue with the dependency?

rushtong: Great point, I'll add some comments here

otchet-broad, line 3:
```suggestion
// Code from https://github.com/validatorjs/validator.js
// Original license: https://github.com/validatorjs/validator.js/blob/master/LICENSE
// Will return to npm module when https://github.com/validatorjs/validator.js/pull/2603 is resolved and dependency is safe.
const default_fqdn_options = {
```

otchet-broad (approved):
Thanks!

fboulnois (approved):
👍

PR #3125, opened 2025-10-23
[DT-2347] Define new architecture for data submission form

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2347](https://broadworkbench.atlassian.net/browse/DT-2347)

### Summary

----

- Verify all tests go green


kevinmarete (approved):
Great work, love to see some tests.

src/pages/data_submission/v2/DataSubmissionFormV2.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useCallback, useEffect, useState } from 'react'
+import { useParams } from 'react-router'
+import { DataSet } from 'src/libs/ajax/DataSet'
+import { cloneDeep } from 'lodash/fp'
+import { set } from 'lodash'
```
fboulnois: Avoid using `lodash` if possible, it is deprecated and doesn't interact well with Typescript generally.

src/pages/data_submission/v2/GeneralStudyInformation.tsx
```diff
@@ -0,0 +1,144 @@
...
+          'snRNA-Seq',
+          'Whole Genome (WGS)',
+          'Whole Exome (WES)',
+        ]}
+        defaultValue={formData?.dataTypes}
+        onChange={onChange}
+      />
+      {generateFormTextField(formData, onChange, new PhenotypeIndication())}
+      {generateFormTextField(formData, onChange, new Species())}
+      <FormField
```
fboulnois, line 95: Is it worth converting all the fields over to functions even if used once for consistency?

src/Routes.jsx
```diff
@@ -50,6 +50,9 @@ import EditDac from 'src/pages/manage_dac/EditDac'
 import ControlledAccessGrants from 'src/pages/user_profile/ControlledAccessGrants'
 import { InstitutionDetails } from 'src/components/institution_table/InstitutionDetails.js'
 import { FORM_MODES } from 'src/components/institution_table/InstitutionFormMode.js'
+import {
+  DataSubmissionFormV2,
+} from 'src/pages/data_submission/v2/DataSubmissionFormV2.js'
```
fboulnois: The `.js` extension shouldn't be necessary, I think.

src/pages/data_submission/v2/v2-models.ts
```diff
@@ -0,0 +1,258 @@
...
+      delete obj.studyId
+    }
+    if (!obj.studyPropertyId) {
+      delete obj.studyPropertyId
+    }
+    return obj
+  }
+}
+
+export class StringStudyProperty extends StudyProperty {
```
fboulnois, line 37: I think there are different approaches to implement this functionality -- I can imagine a more functional approach that returns a specific study property like `StringStudyProperty` instead of the class-based approach, but let's continue with this approach and see if it is an effective mechanism for type safety, extensibility, and refactoring older components.

fboulnois (approved):
A few minor comments. This is a huge improvement over the older approach.

src/Routes.jsx
otchet-broad: I've cleaned up the others as well.

src/pages/data_submission/v2/GeneralStudyInformation.tsx
otchet-broad: Maybe.  I'm doing some refactoring in DT-2381 as I work through more form elements.  I'm liking a small adjustment I've made to the base class and think it might make sense to do this in a wider approach.  I'm definitely feeling the refactor of `generateFormTextField` and some of the other work I'm doing on DT-2381 that expands on that effort, but the jury is still out while I go through the Study elements.  I'll look for more feedback and discussion in that PR.

src/pages/data_submission/v2/DataSubmissionFormV2.tsx
otchet-broad: Good catch.  Thank you.  Updated here.

src/pages/data_submission/v2/GeneralStudyInformation.tsx
otchet-broad: Work in progress on DT-2381 is pushed here: https://github.com/DataBiosphere/duos-ui/pull/3130

eweitz: Is there a net benefit of using standard functions over (functional) components?  The `generateFormTextField` approach strikes me as unconventional compared to `FormField`.

On that topic, what's your sense on the value of migrating these `FormField` components to the newer, more specific components from `formComponents.jsx`?  I recall we began that migration effort given typing / static analysis productivity roadbumps with `FormField`.

src/pages/data_submission/v2/GeneralStudyInformation.tsx
```diff
@@ -0,0 +1,144 @@
...
+          // The top properties were extracted from the prod database and deduplicated using the query:
+          // SELECT property_value, COUNT(*) FROM dataset_property WHERE property_key = 2 GROUP BY property_value ORDER BY COUNT(*) DESC;
+          'CITE-seq',
+          'Hybrid Capture',
+          'RNA-Seq',
+          'scRNA-Seq',
+          'Spatial Transcriptomics',
+          'snRNA-Seq',
+          'Whole Genome (WGS)',
+          'Whole Exome (WES)',
```
eweitz, line 88: Would it be worth a ticket to make these values not hard-coded?  There are 8 "data type" values here, and well over 50 in the "data types" facet searchable via [data library](https://duos-k8s.dsde-dev.broadinstitute.org/datalibrary).

src/pages/data_submission/v2/GeneralStudyInformation.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <FormField
+        id="studyType"
+        title="Study Type"
+        type={FormFieldTypes.SELECT}
+        selectOptions={[
+          'Observational', 'Interventional', 'Descriptive',
+          'Analytical', 'Prospective', 'Retrospective',
+          'Case report', 'Case series', 'Cross-sectional',
+          'Cohort study',
+        ]}
```
eweitz, line 50: Are these fields searchable via GUI now?  If not, and if we don't intend to enable that soon, it might be worth removing this field from the input UI for now.

Single Cell Portal has a few fields in our data submission forms where we aspired to make them searchable, but they remain unsearchable after several years.  So in SCP's case it's a more costly upload UX -- unnecessarily time-consuming for the user -- and a more costly DX due to having to maintain superficially "used" code that isn't actually used in any sense of value.

If SCP were doing a big data submission refactor then I'd advocate removing those fields unless it were paired with a corresponding search UI update soon after.  It seems opportune to consider the same here!

src/pages/data_submission/v2/v2-models.ts
```diff
@@ -0,0 +1,258 @@
...
+export class PiInstitution extends StudyProperty {
+  constructor(value: number, studyId?: number, studyPropertyId?: number) {
+    super('piInstitution', 'Number' as StudyPropertyType, value, studyId, studyPropertyId)
+  }
+}
+
+export class DataCustodianEmail extends StudyProperty {
+  constructor(value: string[], studyId?: number, studyPropertyId?: number) {
+    super('dataCustodianEmail', 'Json' as StudyPropertyType, value, studyId, studyPropertyId)
+  }
```
eweitz, line 96: How can front-end validation be hooked into this?  It doesn't jump out to me.

eweitz (approved):
Looks generally good!  I don't notice any blocking issues given this will undergo refinement in future PRs before being generally available (GA), AIUI.  

I suggest a few refinements.  The hard-coded data types strike me as somewhat of a lurking bug that we should resolve before GA, though not necessarily before merge.

src/pages/data_submission/v2/DataSubmissionFormV2.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useCallback, useEffect, useState } from 'react'
+import { useParams } from 'react-router'
+import { DataSet } from 'src/libs/ajax/DataSet'
+import { cloneDeep, set } from 'lodash/fp'
```
rushtong, line 4: We've been trying to get away from `lodash/fp`, can these be done in pure js or plain lodash?

rushtong (approved):
👍🏽

src/pages/data_submission/v2/GeneralStudyInformation.tsx
otchet-broad: @eweitz - This field purported to be required for schema validation in the backend.  Evidence of some data makes me suspect on the field being required, but I'm inclined to leave it in now so that we can get past the first draft of the end-to-end form submission and revisit this after we've been able to get that completed.

src/pages/data_submission/v2/GeneralStudyInformation.tsx
otchet-broad: Yes.  We could load a facet and allow for new entries.

src/pages/data_submission/v2/GeneralStudyInformation.tsx
otchet-broad: The advantage of the functions wrapping the formComponents is that we're reducing the readability and standardizing the input experience across similar fields which has been an issue in this particular form.

IMHO, typing the underlying FormField components is needed and these functions are helping demonstrate the kinds of typing that needs to happen for them.

src/pages/data_submission/v2/v2-models.ts
otchet-broad: I think we'd need to go deeper on the types of properties.  I'm taking a walk down this path in the DT-2381.

src/pages/data_submission/v2/DataSubmissionFormV2.tsx
otchet-broad: Both are tricky to do in pure JS but it's possible.

An updated _typescriptified_ version of these exist in [radashi](https://radashi.js.org/). Would a ticket to migrate us to that be acceptable?

rushtong: Even plain `lodash` would be better (which does have both of these methods) but yes, tickets to migrate are also good.

otchet-broad: Ticket here: [https://broadworkbench.atlassian.net/browse/DT-2393](https://broadworkbench.atlassian.net/browse/DT-2393)

PR #3126, opened 2025-10-24
DT-294: Upgrade from react router dom 5 -> 6

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-294

## Summary
Upgrade react router dom from 5 -> 6

### Major Changes include
* Use of `Navigate` and `useNavigate` in place of history and the use of `useParams` in place of `match` params
  * These updates required extensive changes in tests to wrap components in `MemoryRouter` or `BrowserRouter` as needed. `useParams` requires `MemoryRouter` and `useNavigate` requires at least `BrowserRouter`
  * This resulted in a significant reduction in component properties which cleaned up a ton of code.
* Partial path parameters (i.e. `/dataset/DUOS-:datasetIdentifier`) are not supported so we moved Dataset and Study permalinks into `NotFound`
* New route handlers for:
  * Authenticated routes
  * Environment specific routes
  * Role based access (RoleBAC) routes
  * SO Acknowledged routes

### Non-routing changes
* Refactored `UserProfile` and `RequestForm` to simplify props
* Non-trivial simplifications to the Signing Official wrapper class
* ~There was an infinite loop in a use effect in `DatasetSearchTableDisplay` so I updated that component to memoize the rows and headers.~ This was also fixed in [Eric's PR](https://github.com/DataBiosphere/duos-ui/pull/3128/files#diff-2eda37b0bad00ca3cd231ce72ecb6a750086ec9749caf29c3fa055a5cabf38fe)

----

- Verify all tests go green


src/routing/AppRoutes.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <Route path="/dar_vote_review/:collectionId" element={<DarCollectionReview readOnly={true} />} />
+        <Route path="/dar_collection/:collectionId" element={<DarCollectionReview />} />
+        <Route path="/dataset_submissions" element={<DatasetSubmissions />} />
+        <Route path="/dataset_update/:datasetId" element={<DatasetUpdateForm />} />
+        <Route path="/data_submission_form" element={<DataSubmissionForm />} />
+        <Route path="/data_submission_form2/" element={<DataSubmissionFormV2 />}>
+          <Route path=":studyId" element={<DataSubmissionFormV2 />} />
+        </Route>
+        <Route path="/study_update/:studyId" element={<StudyUpdateForm />} />
+        {/* NOTE: Previous support for Chairperson adding DACs should not have been allowed */}
```
otchet-broad, line 142: 👍

src/pages/signing_official_console/SigningOfficialTable.jsx
```diff
@@ -369,7 +369,7 @@ export default function SigningOfficialTable(props) {
 
   return (
     <>
-      <div style={{ display: 'flex', justifyContent: 'space-between', alignItems: 'flex-end', width: '112%', marginLeft: '-6%' }}>
+      <div style={{ display: 'flex', justifyContent: 'space-between', alignItems: 'flex-end' }}>
```
fboulnois, line 372: Can you describe why this change was required?

src/pages/AdminManageInstitutions.tsx
```diff
@@ -74,10 +74,8 @@ export default function AdminManageInstitutions() {
             <div>
               <Link
                 id="btn_addInstitution"
-                to={{
-                  pathname: '/admin_manage_institutions/create_new',
-                  state: { institutionList },
-                }}
+                to="/admin_manage_institutions/create_new"
```
fboulnois, line 77: Can you describe this change in more detail?

src/components/SigningOfficialDaaAgreementWrapper.jsx
```diff
@@ -1,145 +1,89 @@
...
-      onAccept()
-    }
-  }
-
-  useEffect(() => {
-    if (isLoading) {
-      spinnerService.showAll()
+    try {
+      await acceptAcknowledgments(Acknowledgments.broadLcaAcknowledgement, Acknowledgments.nihLcaAcknowledgement)
+      window.location = '/signing_official_console/library_cards'
```
fboulnois, line 20: Does this need to be switched to `navigate` instead?

fboulnois (approved):
A few questions, but otherwise this looks great, particularly the additional tests!

src/pages/signing_official_console/SigningOfficialTable.jsx
rushtong: The width and margin settings caused the content of the LC page to overflow past the left edge of the screen I'm not completely sure why that was the case, but this fix allowed it to fit the screen:

Image

src/pages/AdminManageInstitutions.tsx
rushtong: The signature of `Link` [changed slightly](https://reactrouter.com/6.30.1/components/link#state) so that `state` needs to be a separate property instead of part of a `to` object as it was previously.

src/components/SigningOfficialDaaAgreementWrapper.jsx
rushtong: I'm looking into this. I do remember testing this and found that using navigate here did not actually navigate. I am looking into whether I was not implementing it correctly, so taking a second look.

otchet-broad: FYSA: We've got other usages of setting window.location in our code.  It might be good to make a new ticket about migrating those cases of doing that to `navigate` where possible.

otchet-broad (approved):
👍
Sonar left some feedback for minor cleanups.  Feel free to disposition those as you see fit.  I've run through a number of UI components and things look good on my end.
Thank you!

src/components/SigningOfficialDaaAgreementWrapper.jsx
rushtong: I spent some time looking into this and wasn't able to come to a workable solution. Going to punt on this specific question and take Elliot's advice to create a ticket for reviewing `window.location`

rushtong:
> Sonar left some feedback for minor cleanup

Thanks @otchet-broad! I did look into the sonar warnings a bit. They are all existing issues that I don't want to conflate with the intent of this PR.

PR #3128, opened 2025-10-28
[DT-2298, DT-2337] Enrich data libraries on home page

eweitz:
# Addresses
* https://broadworkbench.atlassian.net/browse/DT-2298
* https://broadworkbench.atlassian.net/browse/DT-2299
* https://broadworkbench.atlassian.net/browse/DT-2337

# Summary
This upgrades how data libraries are displayed on the DUOS home page, https://duos.org.  Previously, only three data libraries were shown, and only with a graphic logo.  

Now an arbitrary number of libraries can be featured for display, and each includes both a graphic and a label.  The display order is configurable, and the display is responsive to different viewport sizes.

## Before

https://github.com/user-attachments/assets/7e76ee87-873b-438f-b224-9125b5b75200

## After

https://github.com/user-attachments/assets/8699d239-0742-4e6b-a5da-69f3f6541576

----

- Verify all tests go green


kevinmarete (approved):

src/libs/libraryVersions.ts
```diff
@@ -0,0 +1,524 @@
+import broadIcon from 'src/logo.svg'
```
fboulnois, line 1: Thank you for this refactoring! Should we switch this to a `.ts` file instead?

cypress/component/libs/libraryVersions.spec.ts
```diff
@@ -0,0 +1,156 @@
+import { getLibraryVersions } from 'src/libs/libraryVersions'
```
fboulnois, line 1: Ditto here?

fboulnois (approved):
This is great!

Another question: any concerns about the uneven sizes between the different logos and text? Should we resize them or emphasize the text?

src/libs/libraryVersions.ts
eweitz: Done: 058c7de

cypress/component/libs/libraryVersions.spec.ts
eweitz: Done: 058c7de

eweitz:
> any concerns about the uneven sizes between the different logos and text? Should we resize them or emphasize the text?

Good suggestion -- I slightly enlarged the font.  

Screenshot 2025-10-29 at 12 40 22 PM

PR #3129, opened 2025-10-28
[DT-2358] Updated Publication Components with New Fields

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2358

### Summary

This refactors the existing Publication forms, adding an updated model schema and TypeScript interface with field-level validation and updated data bindings.

**New**
New

**Edit**
Edit

**List**
List

----

- Verify all tests go green

### DIFF
--- a/src/components/publications_list/PublicationAddEdit.tsx
+++ b/src/components/publications_list/PublicationAddEdit.tsx
@@ -1,22 +1,12 @@
 import React, { useState } from 'react'
-import { FormField, FormFieldTypes, FormValidators } from 'src/components/forms/forms'
+import { FormField, FormValidators } from 'src/components/forms/forms'
 import { ValidationError } from 'src/pages/dar_application/FormValidationState'
-import { validationFailed, calcPublicationErrors } from 'src/utils/darFormUtils'
-import { Publication } from 'src/types/model'
-
-const defaultPublication: Publication = {
-  title: '',
-  date: '',
-  authors: '',
-  pubmedId: '',
-  bibliographicCitation: '',
-  datasetCitation: '',
-  citation: false,
-}
+import { validationFailed, calcPublicationErrors, ORCID_REGEX } from 'src/utils/darFormUtils'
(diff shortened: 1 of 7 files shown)


rushtong (approved):
👍🏽

src/components/publications_list/PublicationAddEdit.tsx
```diff
@@ -29,118 +19,335 @@ interface PublicationAddEditProps {
...
+    }
+    else {
+      updatePublication({ ...newPublication, [key]: value })
+    }
+  }
+
+  // Enable Add only when all current authors have name + valid ORCID
+  const disableAddAuthor = newPublication.authors.some(
+    a => a.name.trim() === '' || a.orcId.trim() === '' || !ORCID_REGEX.test(a.orcId),
+  )
```
eweitz: If we want ORCID to be optional as I suspect, I think we'd want something closer to the following (which would also be worth a quick local UI check):

```suggestion
  // Enable Add only when all current authors have name + valid ORCID
  const disableAddAuthor = newPublication.authors.some(
    a => a.name.trim() === '' || (a.orcId.trim() !== '' && !ORCID_REGEX.test(a.orcId)),
  )
```

eweitz (commented):
Looks broadly good!  I note a concern and a potential future enhancement.

ORCID seems required given the implementation here, which strikes me as a significant UX barrier that's likely unintended.  Can you confirm with product that we want ORCID to be required?  

Many -- most? -- authors don't have an ORCID, and I don't think e.g. PubMed requires it.  So AIUI the current code here would block most publications from being entered.  I note a possible way to make this field option in a code suggestion, but I think the product question is upstream of that.

Separately, for consideration outside the scope of this PR, I've opened a ticket to populate publication data upon ID entry: https://broadworkbench.atlassian.net/browse/DT-2387.  It's the same general idea as proposed for clinical trials.

eweitz (approved):

PR #3134, opened 2025-10-30
[DT-2097] Improve Progress Report Form Validation

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2097

### Summary

Updates the form to validate fields only after user interaction or on submit, instead of showing errors when the form first loads. This provides a smoother experience and highlights only the fields that need attention.

**Before**

https://github.com/user-attachments/assets/e46647b7-1461-4148-95a6-c1387af31d69

**After**

https://github.com/user-attachments/assets/6d3afa77-03bb-4a79-8e47-c3dc6839ffce

----

- Verify all tests go green


fboulnois (commented):
When a field is entered, this solution activates validation for the whole form instead of the individual component. I think this is a potential solution but we should discuss.

kevinmarete:
> When a field is entered, this solution activates validation for the whole form instead of the individual component. I think this is a potential solution but we should discuss.

Fixed here: https://github.com/DataBiosphere/duos-ui/pull/3134/commits/d1d2a5a6172fc6d7ad076ff775c98f6d9fb531b0


https://github.com/user-attachments/assets/9c8aef68-4923-4b4c-a63d-d4cd033e4f6d

eweitz (approved):
Nice refinement!  

This UX is better than before, but still seems somewhat unconventional.  Later, it might be preferable if validation issues were shown only for the _current field on blur_ (which would presumably be uncommon, as most checks confirm simple presence of a value), and for _all fields on submit_.  This further refinement doesn't strike me as necessary to do as part of this PR, as the changes here strictly improve the current state.

fboulnois (approved):
👍 based on our previous conversation and bugfix

PR #3176, opened 2025-11-14
DT-2474: Self-host fonts

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2474

### Summary
Self-host fonts that were previously pulled from `fonts.gstatic.com`

----

- Verify all tests go green


fboulnois (commented):
We are using a huge number of font files! Is there an opportunity for simplification?

rushtong:
> We are using a huge number of font files! Is there an opportunity for simplification?

That is true, this is a lot of content. It is likely that we can simplify, but I think that would require going through every Montserrat and Roboto font usage across the site which is non-trivial. I'm open to a separate effort on that front. The intent behind this PR is to stop calling out to Google for every app load.

fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3202, opened 2025-11-21
[DT-2503] Fix existing study registration submission.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2503](https://broadworkbench.atlassian.net/browse/DT-2503)

### Summary

Only attempts to add files when they exist; Let the browser set the proper content-type header when multipart.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/fetchAdapter.js
+++ b/src/libs/ajax/fetchAdapter.js
@@ -36,6 +36,10 @@ async function fetchRequest({ url, method = 'GET', data, params, headers = {}, c
     credentials,
     body: getRequestBody(data, isMultipart),
   }
+  if (isMultipart) {
+    // Let the browser set the correct multipart boundary.
+    delete fetchOptions.headers['Content-Type']
+  }
   try {
     const res = await fetch(fullUrl, fetchOptions)
     return handleResponse(res, fullUrl, responseType, method)
--- a/src/libs/ajax/fetchAdapter.js
+++ b/src/libs/ajax/fetchAdapter.js
@@ -36,10 +36,6 @@ async function fetchRequest({ url, method = 'GET', data, params, headers = {}, c
     credentials,
     body: getRequestBody(data, isMultipart),
   }
-  if (isMultipart) {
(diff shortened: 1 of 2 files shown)


src/pages/data_submission/DataSubmissionForm.jsx
```diff
@@ -99,9 +99,11 @@ export const DataSubmissionForm = () => {
         if (Array.isArray(files)) {
           files.forEach((file, idx) => {
             const [[duosFileType, duosFileObj]] = Object.entries(file)
-            // Consent has code to look for the string as formatted below
-            const fieldKey = `${field}[${idx}].${duosFileType}`
-            multiPartFormData.append(fieldKey, duosFileObj, duosFileObj.fileName)
+            if (duosFileObj !== null) {
```
otchet-broad, line 102: We should be filtering the object elsewhere so if the value is undefined we don't get into this logic, but we're not so this is the path of least resistance to fix this.

src/libs/ajax/fetchAdapter.js
```diff
@@ -46,6 +46,28 @@ async function fetchRequest({ url, method = 'GET', data, params, headers = {}, c
...
+async function fetchMultipartRequest({ url, method = 'POST', data, params, headers = {}, credentials }) {
+  const fullUrl = params ? buildUrlWithParams(url, params) : url
+  const fetchOptions = {
+    method,
+    headers,
+    credentials,
+    body: getRequestBody(data, true),
+    isMultipart: true,
+  }
+  if (headers['Content-Type']) {
```
otchet-broad, line 58: The header for the content type should be handled by the browser when using form multipart.
We would have been able to do this on all requests when multipart is true however consent requires a specific content-type when uploading attachments from the support form.  When we fix that, we can combine the logic in this method into the fetchRequest method.  For now, it's a clean break when dealing with multipart.

fboulnois (approved):
👍

kevinmarete (approved):

PR #3227, opened 2025-12-11
[DT-2402] Disable specific fields and actions when updating a saved study registration.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2402](https://broadworkbench.atlassian.net/browse/DT-2402)

### Summary

When editing a study registration that has been saved, this PR will disable:
* Deleting a consent group/dataset
* Changing the DAC assigned to a controlled access consent group/dataset
* Changing the Access Method from controlled access to any other option (because it would necessitate clearing the DAC assigned and that field isn't currently editable)

Disabled Delete:
image

Disabled Dac Assignment on a controlled access dataset:
image

Disabled Access Method adjustment:
image

Enabled Access Method adjustment when the Access Method isn't controlled:
image

Note: The mouse pointer change is the visual clue that the Access Method enabled/disabled styling has taken place.  I can't capture that in a screen shot but have manually verified the behavior of both cases.

----

- Verify all tests go green

### DIFF
--- a/src/pages/data_submission/v2/DataSubmissionFormV2.tsx
+++ b/src/pages/data_submission/v2/DataSubmissionFormV2.tsx
@@ -99,7 +99,7 @@ export const DataSubmissionFormV2 = () => {
         <NihAnvilUseRelated study={study} setStudy={setStudy} />
         <NihAdministrativeInformation study={study} setStudy={setStudy} />
         <NihDataManagement study={study} setStudy={setStudy} />
-        <StudyAssetManagement study={study} setStudy={setStudy} />
+        <StudyAssetManagement study={study} setStudy={setStudy} isEditingExistingStudy={isEditing} />
         {!isEditing && (
           <AsyncSpinnerButton
             onClick={onSubmitStudy}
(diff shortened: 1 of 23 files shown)


kevinmarete (approved):

src/pages/data_submission/v2/DataSubmissionFormV2.tsx
eweitz, line 94: Incidentally, this `description` is inaccurate when `isEditingExistingStudy` (i.e. `isEditing` here) is `true`, right?

eweitz (approved):
Can you help me understand why deleting a consent group/dataset is disabled when editing a study?  

I understand the other two limitations per our helpful chat at standup, but I don't recall covering this one, and the rationale isn't immediately obvious to me.

src/pages/data_submission/v2/DataSubmissionFormV2.tsx
otchet-broad: No, because you can use this form to submit new datasets to an existing study as well.  That being said, I think there's some work we could do to this that's also inclusive of the other assets we're now taking in.  For what it's worth, we've designed those other pieces so they can  created, updated, or deleted at every point in the creation or editing lifecycle.

otchet-broad:
@eweitz wrote "Can you help me understand why deleting a consent group/dataset is disabled when editing a study?"

Consent has logic that prevents removing datasets from studies using this API.  We do have other APIs that could be invoked that **may** be able to delete the dataset from the study, but only if certain conditions are met (there's a different logic tree for that).  Wiring all of that up was beyond the 'thin slice' we're doing here.

PR #3237, opened 2025-12-16
[DT-2746] Enable form for both DAC chairs and data submitter flows in local/dev.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2746](https://broadworkbench.atlassian.net/browse/DT-2746)

### Summary

- Updates logic in AppRoutes to add routes based on environment
- Adds new option to direct the S4R Form route after successful saving so that the Data Submitter routes back to the correct location in the application.

----

- Verify all tests go green

### DIFF
--- a/src/routing/AppRoutes.tsx
+++ b/src/routing/AppRoutes.tsx
@@ -3,7 +3,7 @@ import { Route, Routes } from 'react-router-dom'
 import Home from 'src/pages/Home'
 import UserProfile from 'src/pages/user_profile/UserProfile'
 import Authenticated from 'src/routing/Authenticated'
-import { envGroups } from 'src/utils/EnvironmentUtils'
+import { checkEnv, envGroups } from 'src/utils/EnvironmentUtils'
 import HealthCheck from 'src/pages/HealthCheck'
 import Status from 'src/pages/Status'
 import BackgroundSignIn from 'src/pages/BackgroundSignIn'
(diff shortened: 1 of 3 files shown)


src/routing/AppRoutes.tsx
```diff
@@ -100,14 +100,22 @@ const AppRoutes = (props: AppRoutesProps) => {
         </Route>
         <Route element={<RoleBAC rolesAllowed={[USER_ROLES.dataSubmitter, USER_ROLES.chairperson, USER_ROLES.admin]} />}>
           <Route path="/dataset_submissions" element={<DatasetSubmissions />} />
-          <Route path="/data_submission_form" element={<DataSubmissionForm />} />
-          <Route element={<EnvRoute env={envGroups.NON_STAGING} />}>
-            <Route path="/data_submission_form2" element={<DataSubmissionFormV2 />}>
-              <Route path=":studyId" element={<DataSubmissionFormV2 />} />
-            </Route>
-          </Route>
+          {checkEnv(envGroups.NON_STAGING) && (
```
otchet-broad, line 103: Couldn't use the EnvRoute component because we need duplicate paths and ReactRouter had issues when the same paths were possible.

rushtong: I think one way that might get past that is to have the single path, but have a conditional that returns different components inside `element`. I had to do something similar with `SOAcknowledged`.

otchet-broad: Agreed - I thought about that option but since this is temporary, I didn't think we'd keep them around long and they'd end up being deleted when we're ready to make this "the form."

eweitz, line 103: Nit: I naively assumed "production" would be part of `NON_STAGING`.  But I see it's `['local', 'dev']`, which is also equivalent with `DEV`.  

The latter seems more intuitive; perhaps it's worth using here.  I defer to you!


```suggestion
          {checkEnv(envGroups.DEV) && (
```

eweitz (approved):

src/routing/AppRoutes.tsx
otchet-broad: @eweitz - Great point!  We already had a usage of NON_STAGING in this file, so I kept it consistent.  It might be worth a larger refactor around the name 'NON_STAGING' throughout the codebase.  I'll make a ticket and we can discuss at refinement.

kevinmarete (approved):

PR #3253, opened 2026-01-09
[DT-2435] Speed up Data Library by removing redundant index request

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2435

### Summary
This improves time-to-final render -- and eliminates a confusing interstitial state -- by removing a redundant request to the /index endpoint.  It keeps the initial index request, which is more specific and seems to be the intended one.

While manually testing I incidentally found an issue with the visibility filter.  See #3256 for that fix.  That fix is substantially different from this related-but-distinct issue, so I've kept the PRs separate.

#### Before

https://github.com/user-attachments/assets/e8300bff-332e-4e3b-9d75-9a2c4d26d685

### After

https://github.com/user-attachments/assets/23460ebe-c88a-4fae-90b7-c94c3de808b7

----

- Verify all tests go green


src/components/data_search/DatasetSearchTable.jsx
```diff
@@ -223,6 +224,12 @@ export const DatasetSearchTable = (props) => {
   const handleSearchChange = useCallback(searchTerms => setSearchTerm(searchTerms))
 
   useEffect(() => {
+    if (!hasRunInitialSearch.current) {
```
otchet-broad, line 251: It seems to me that the root of this problem is caused by a second query coming in while one is running.  Why is the second query running?  Are there cases where the second query is the **correct** query to run instead of the first?

I think you might have mentioned that the second query is missing filters, shouldn't we be ensuring those filters are always present?  Do we need a ticket to investigate that?  Maybe there are multiple bugs here.

I would have expected something like an abort controller to abort the first request and process the second request so we don't leak requests on the client while the user clicks around with search filters.

rushtong: I agree that there is likely a hidden bug with the duplicate call. The initial search is called in a `useEffect` where we've seen to be problematic on re-renders.

rushtong (approved):
This is definitely an improvement. I do think there's more work do to here. 👍🏽

src/components/data_search/DatasetSearchTable.jsx
eweitz: > It seems to me that the root of this problem is caused by a second query coming in while one is running. Why is the second query running? Are there cases where the second query is the correct query to run instead of the first?

The query is run both when initializing `DatasetSearch` and when initializing its child component `DatasetSearchTable`.  I'm not aware of a case where the second query _at initialization_ is desired, and it also seems non-straightforward to refactor the `useEffect` in `DatasetSearchTable` to not fire at initialization, so the `hasRunInitialSearch` guard strikes me as reasonable.

> I think you might have mentioned that the second query is missing filters, shouldn't we be ensuring those filters are always present? Do we need a ticket to investigate that? Maybe there are multiple bugs here.

Yup!  While manually testing I incidentally found an issue with the visibility filter. See #3256 for that fix. That fix is substantially different from this related-but-distinct issue, so I've kept the PRs separate.

> I would have expected something like an abort controller to abort the first request and process the second request so we don't leak requests on the client while the user clicks around with search filters.

Excellent suggestion.  I've added abort handling, _and_ retained the debounce.  Both seem useful for minimizing server load.

otchet-broad (approved):
Nice work on this one!

PR #3255, opened 2026-01-09
[DT-2706] Convert `storage.js` to typescript

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2706

### Summary

This PR migrates `storage.js` and related files to Typescript.

----

- Verify all tests go green


src/components/data_search/DatasetSearchTableDisplay.tsx
```diff
@@ -54,7 +54,7 @@ const storageDatasetSearchSort = 'storageDatasetSearchSort'
 // Get the sort configuration for the active tab
 const getSortForTab = (tabKey: string): Sort => {
```
fboulnois, line 55: The typing `as Sort | undefined` is very unusual. I would recommend something like this instead:

```
const getSortForTab = (tabKey: string): Sort => {
  const storageKey = `${storageDatasetSearchSort}_${tabKey}`
  const savedSort = Storage.getCurrentUserSettings(storageKey)

  if (savedSort && typeof savedSort === 'object' && savedSort !== null) {
    const candidate = savedSort as Record<string, unknown>
    if (
      typeof candidate.colIndex === 'number'
      && typeof candidate.dir === 'number'
    ) {
      return {
        colIndex: candidate.colIndex,
        dir: candidate.dir,
      }
    }
  }

  return { dir: 1 } as Sort
}
```

src/components/institution_table/InstitutionTableUtils.tsx
```diff
@@ -11,12 +11,17 @@ export interface SortType {
...
+interface StorageSort {
+  field: string
+  dir: number
+}
+
 // Sort functionality for the institution table.
 export const storageInstitutionSort = 'storageInstitutionSort'
 
 // Sort functionality for the institution table.
 export const getInitialSort = (columns: string[] = []): SortType => {
```
fboulnois, line 23: The typing `StorageSort | undefined` is very unusual. I would recommend something like this instead:

```
export const getInitialSort = (columns: string[] = []): SortType => {
  const savedSort = Storage.getCurrentUserSettings(storageInstitutionSort)

  let sort: StorageSort = { field: 'name', dir: 1 }

  if (savedSort && typeof savedSort === 'object' && savedSort !== null) {
    const candidate = savedSort as Record<string, unknown>
    if (
      typeof candidate.field === 'string'
      && typeof candidate.dir === 'number'
    ) {
      sort = { field: candidate.field, dir: candidate.dir }
    }
  }

  const sortIndex = columns.indexOf(sort.field)
  if (sortIndex !== -1) {
    return { colIndex: sortIndex, dir: sort.dir }
  }
  else {
    return { colIndex: 0, dir: 1 }
  }
}
```

src/libs/storage.ts
```diff
@@ -0,0 +1,140 @@
...
+  setCurrentUser: (data: DuosUser): void => {
+    localStorage.setItem(CurrentUser, JSON.stringify(data))
+  },
+
+  getCurrentUser: (): DuosUser => {
+    const user = localStorage.getItem(CurrentUser)
+    return user ? JSON.parse(user) : DEFAULT_DUOS_USER
+  },
+
+  getCurrentUserSettings: (key: string): unknown => {
```
fboulnois: A better solution would be to not return an `unknown` type here, and use it instead.

fboulnois (commented):
A few suggestions:

cypress/component/libs/storage.spec.ts
```diff
@@ -0,0 +1,262 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
```
fboulnois: Also, is this required?

src/components/dac_bot/DACBotComponent.tsx
```diff
@@ -35,9 +36,9 @@ export const DACBotComponent = (props: DACBotComponentProps) => {
   const { dacId, 'data-cy': dataCy } = props
   const [DACbotRules, setDACbotRules] = useState<Array<DACbotRule>>([])
   const [isLoading, setIsLoading] = useState(true)
-  const userIsChair = Storage.getCurrentUser().roles.filter((r: { dacId: number, name: string }) => {
-    return r.dacId == dacId && r.name == 'Chairperson'
-  }).length > 0
+  const userIsChair = Storage.getCurrentUser().roles.filter((r: UserRole) => {
+    return r.name === 'Chairperson'
```
rushtong: I think this simplification loses some required information. We need to know not only if the user is a chair, but are they a chair in **THIS** data access committee. It is possible for a user to be a member of one DAC, but a chair in a different DAC.

src/components/data_search/DatasetSearchTableDisplay.tsx
kevinmarete: I think refactoring getCurrentUserSettings to use generic typing and this refactor makes this easier:

```return Storage.getCurrentUserSettings<Sort>(storageKey) ?? { dir: 1 } as Sort```

File: src/components/institution_table/InstitutionTableUtils.tsx (hunk 2 above)
Comment (kevinmarete) [reply src/components/institution_table/InstitutionTableUtils.tsx:23]: Refactoring `getCurrentUserSettings` now handles generic typing:

```Storage.getCurrentUserSettings<StorageSort>(storageInstitutionSort)```

**Review by rushtong — APPROVED** [review]
👍🏽

File: src/libs/storage.ts [hunk 6]
```diff
@@ -0,0 +1,140 @@
+import get from 'lodash/fp/get'
```
fboulnois: Avoid the use of `lodash/fp`, as it does not generally play well with Typescript

fboulnois (commented):
Another potential issue:

fboulnois (approved):
👍

PR #3257, opened 2026-01-12
DT-2548: Migrate App to Typescript

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2548

### Summary
Migrates main App component to Typescript

----

- Verify all tests go green


package.json
```diff
@@ -52,6 +52,7 @@
     "@types/node": "24.10.4",
     "@types/react": "18.3.12",
     "@types/react-dom": "18.3.1",
+    "@types/react-modal": "3.16.3",
```
rushtong: Added to support `import Modal from 'react-modal'` in App.tsx

otchet-broad (approved):
👍

src/App.tsx
```diff
@@ -35,7 +35,10 @@ function App() {
   const [isLoading, setIsLoading] = useState(false)
 
   useEffect(() => {
-    Modal.setAppElement(document.getElementById('modal-root'))
+    const modalRoot = document.getElementById('modal-root')
+    if (modalRoot) {
+      Modal.setAppElement(modalRoot)
```
fboulnois: It looks like the supported way to do this now is without using `Modal`:

https://github.com/vitejs/vite/tree/main/packages/create-vite/template-react-ts/src

fboulnois (commented):
See alternate approach:

src/App.tsx
fboulnois: Also, I notice we're using `getElementById` to replace the root element in both `index.tsx` and `App.jsx`, which is likely causing an additional rerender.

rushtong: Tested locally after removing it altogether and core modal functionality still works - thanks for the heads up on this 👍🏽

otchet-broad (commented):
👍
Changes look good.
@fboulnois - Good catch!

fboulnois (approved):
👍

PR #3259, opened 2026-01-13
[DT-2781] Fix Cypress for React 18

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2781

### Summary

See the commits for the changes in question.

1. Ensure that we are only looking at the cypress directory for tests (which excludes `node_modules`, which may contain tens of thousands of files, thus substantially slowing initialization)
2. We are using an unsupported mechanism to run our tests with React 18, update to use the recommended approach instead.
3. Change all the test files to use `cy.mount` instead of `mount`.

----

- Verify all tests go green

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -13,7 +13,7 @@ export default defineConfig({
     },
   },
   component: {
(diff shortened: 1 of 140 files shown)


rushtong (approved):
👍🏽

otchet-broad (approved):
👍

PR #3262, opened 2026-01-13
[DT-2594] Convert `NavigationTabsComponent.jsx` to TypeScript

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2594

### Summary

This PR migrates `NavigationTabsComponent.jsx` and related files to Typescript.

----

- Verify all tests go green


src/components/NavigationTabsComponent.tsx
```diff
@@ -60,10 +101,11 @@ export const NavigationTabsComponent = (props) => {
     tabs, initialTab, initialSubTab,
     onSubtabChange, showProfileLinks, profileState,
   } = props
-  const [selectedMenuTab, setSelectedMenuTab] = useState(false)
-  const [selectedSubTab, setSelectedSubTab] = useState(false)
+  const [selectedMenuTab, setSelectedMenuTab] = useState<number | false>(false)
+  const [selectedSubTab, setSelectedSubTab] = useState<number | false>(false)
```
fboulnois: The typing looks unusual to me here, should this be a number?

src/components/NavigationTabsComponent.tsx
```diff
@@ -60,10 +101,11 @@ export const NavigationTabsComponent = (props) => {
...
 
   useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect
```
fboulnois: Is this currently a warn or error? I thought this was a warn, does this disable the warning?

fboulnois (commented):
A few comments:

src/components/NavigationTabsComponent.tsx
kevinmarete: Yes it is a warning and this disables the warning!

fboulnois: Should we suppress the warning? In this case, it is reporting a real error that we have in our codebase (and that we need to fix). Previously, the `react-hooks` plugin was not running correctly and not reporting anything. When I made the recent `eslint` changes, the `react-hooks` plugin started working and reported numerous errors with our codebase. I switched these errors to warnings so that we are able to continue to iterate without blocking progress on the entire codebase. I fear that suppressing this error will make it more difficult to address this issue in the future.

Thoughts?

fboulnois (approved):
One additional thought:

src/components/NavigationTabsComponent.tsx
```diff
@@ -6,7 +6,48 @@ import React, { useEffect, useState } from 'react'
...
+  initialSubTab: number
+  onSubtabChange: (event: React.SyntheticEvent, newValue: number) => void
+  showProfileLinks: () => void
+  profileState: boolean
+}
+
+const styles: Record<string, React.CSSProperties> = {
   mainTab: {
     padding: '0 25px',
     fontSize: '16px',
```
rushtong, line 37: There is something off about the tab alignment that is not due to this PR - I see it in dev too - that I noticed while testing. Moving from a wider window to a narrower window (or vice versa), the top level tabs shift slightly to the right. There is a range of widths - somewhere between 1400 and 1500 px - where the transition takes place.

src/components/NavigationTabsComponent.tsx
rushtong: Agreed that we should look at trying to fix this rather than ignoring it. It does look like fixing this might be a relatively larger refactor.

rushtong (approved):
👍🏽

PR #3265, opened 2026-01-14
[DT-2654] Refs should not be accessed during render

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2654

### Summary

This PR reduces the number of eslint error rules we need to ignore. Refs should not be accessed during a render. Instead:
1. `useCallback` defers execution
2. `searchRef` is passed at invocation time
3. `searchRef` is correctly excluded from dependencies
4. This is functionally equivalent to the previous approach, but with better performance.

----

- Verify all tests go green


rushtong (approved):
👍🏽

eweitz (approved):

PR #3266, opened 2026-01-14
[DT-2654] Fix incomplete dependency arrays

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2654

### Summary

We have incomplete dependency arrays in our `useCallback`s and `useEffect`s. The missing dependencies are often functions that get recreated on every render:

```
function Component() {
  const handleClick = () => { /* ... */ }  // New function every render
  
  useEffect(() => {
    // This runs on EVERY render because handleClick is always "new"
  }, [handleClick])  // ⚠️ handleClick changes every render
}
```

This pattern can cause infinite loops. The fix requires two steps:

1. Wrap functions in `useCallback` to memoize them to create a stable reference
2. Add them to the dependency arrays (now safe because they're stable)

Once memoized with `useCallback`, these functions have stable identities and only recreate when their own dependencies change, breaking the loop.

----

- Verify all tests go green

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -67,7 +67,6 @@ export default defineConfig([
       '@stylistic/jsx-one-expression-per-line': ['off'],
       // TODO: these issues should be fixed
       'react-hooks/static-components': 'warn',
-      'react-hooks/preserve-manual-memoization': 'warn',
       'react-hooks/immutability': 'warn',
       'react-hooks/set-state-in-effect': 'warn',
     },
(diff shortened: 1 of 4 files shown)


otchet-broad (approved):

rushtong (approved):
👍🏽

rushtong (approved):
👍🏽

kevinmarete (approved):

PR #3268, opened 2026-01-15
[DT-2790] Allow DAC Rules Page to enforce mutual exclusivity rules

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2790

### Summary

This PR enforces mutual exclusivity on the rules page to prevent invalid rule combinations.

https://github.com/user-attachments/assets/f03ec295-393d-41a1-bdf8-430e9ee41702

----

- Verify all tests go green


rushtong (approved):
This looks reasonable to me. Two observations:
1. This will hit production - I believe that is OK since the additional rule isn't deployed to production yet, but haven't tested with a local branch of consent that doesn't have the rule applied.
2. The new toggle effect is jarring. On dev, only the checkboxes look like they're re-rendering. In this branch, the whole page looks like it's re-rendering. If there's a way to maintain the current UX, that would be preferable.

kevinmarete:
> This looks reasonable to me. Two observations:
> 
> 1. This will hit production - I believe that is OK since the additional rule isn't deployed to production yet, but haven't tested with a local branch of consent that doesn't have the rule applied.
> 2. The new toggle effect is jarring. On dev, only the checkboxes look like they're re-rendering. In this branch, the whole page looks like it's re-rendering. If there's a way to maintain the current UX, that would be preferable.

As agreed, this PR will not be merged until we are ready. Also, I have updated the UX to work as advised.

rushtong (approved):
👍🏽

src/components/dac_bot/DACBotComponent.tsx
```diff
@@ -27,9 +27,14 @@ export type DACbotRule = {
...
-export type DACbotChangeResult = {
-  ruleId: number
-  isRuleEnabled: boolean
+export type ParsedDACbotRule = DACbotRule & {
+  exclusiveRuleType?: string
+  isDisabled: boolean
+}
+
+const MUTUALLY_EXCLUSIVE_RULES: { [key: string]: string } = {
+  REQUIRE_SO_DAR_APPROVAL: 'AUTO_OPEN_DAR_FOR_ALL_MEMBERS',
```
otchet-broad: @kevinmarete - Given the current state, are you proposing removing both of these entries from this object as the change to "leave the code in place"?

kevinmarete: Yes!

otchet-broad (approved):
👍

PR #3279, opened 2026-01-21
[DT-2704] Convert `utils.js` to TypeScript

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2704

### Summary

This PR migrates `utils.js` and related files to Typescript.

----

- Verify all tests go green


cypress/component/DAC/editDac.spec.tsx
```diff
@@ -38,7 +39,7 @@ describe('EditDAC Tests', () => {
   it('Admins can create a DAC', () => {
     cy.viewport(600, 600)
     Storage.clearStorage()
-    setUserRoleStatuses(admin, Storage)
+    setUserRoleStatuses(admin as DuosUser, Storage)
```
fboulnois: I would recommend ensuring that the `admin` object itself is defined as a `DuosUser`, instead of coercing the object at the test level.

cypress/component/DAC/editDac.spec.tsx
```diff
@@ -69,7 +70,7 @@ describe('EditDAC Tests', () => {
   it('Chairs cannot create a DAC', () => {
     cy.viewport(600, 600)
     Storage.clearStorage()
-    setUserRoleStatuses(chair, Storage)
+    setUserRoleStatuses(chair as DuosUser, Storage)
```
fboulnois: ditto here.

cypress/component/DAC/manageEditDac.spec.tsx
```diff
@@ -19,7 +20,7 @@ describe('ManageEditDAC Tests', () => {
   Cypress._.each([admin, chair], (user) => {
     it('Manage Edit DAC page should load for ' + user.displayName, () => {
       cy.viewport(600, 600)
-      setUserRoleStatuses(user, Storage)
+      setUserRoleStatuses(user as DuosUser, Storage)
```
fboulnois: ditto in this test.

cypress/component/DAC/manageRadar.spec.tsx
```diff
@@ -28,7 +29,7 @@ describe('ManageRadar Component Tests', () => {
 
   describe('Basic Rendering', () => {
     it('should render the ManageRadar component with loading state', () => {
-      setUserRoleStatuses(admin, Storage)
+      setUserRoleStatuses(admin as DuosUser, Storage)
```
fboulnois: ditto in this test.

cypress/component/libs/utils.spec.ts
```diff
@@ -132,31 +219,16 @@ describe('LC Search Filter', () => {
...
     expect(filteredList.length).equals(sampleLCList.length)
 
-    const term = formatDate(originalCard.createDate)
+    const term = formatDate(originalCard.createDate as unknown as number)
     filteredList = cardSearchFn(term, sampleLCList)
     expect(filteredList.length).equals(1)
     const filteredCard = filteredList[0]
     forEach(originalCard, (value, key) => {
-      expect(filteredCard[key]).equals(value)
+      expect((filteredCard as unknown as Record<string, unknown>)[key]).equals(value)
```
fboulnois: I think we should avoid `as unknown as xxx` as it essentially disables typechecking.

cypress/component/libs/utils.spec.ts
```diff
@@ -167,7 +239,7 @@ describe('LC Search Filter', () => {
...
```
fboulnois: Ditto in the rest of this file.

src/libs/utils.ts
```diff
@@ -0,0 +1,854 @@
...
+          || includes(loweredTerm, toLower(institution.updateDate?.toString() || ''))
+          || includes(loweredTerm, toLower(institution.createDate))
+          || includes(loweredTerm, toLower(domains))
+          || includes(loweredTerm, toLower(soStrings || ''))
+        )
+      }, targetList),
+  }
+}
+
+type ListTypeForModel<T extends keyof SearchFilterFunctions>
```
fboulnois, line 559: This looks unusual, can you explain the purpose?

fboulnois (commented):
A few suggestions:

src/libs/utils.ts
```diff
@@ -0,0 +1,854 @@
...
+  first,
+  forEach as lodashFPForEach,
+  get,
+  includes,
+  isEmpty,
+  isNil,
+  join,
+  map,
+  toLower,
+} from 'lodash/fp'
```
fboulnois: This is the one file I haven't converted over to `lodash` from `lodash/fp`, I'm not sure if you would like to do it in your PR or if I should do it in a separate PR after you've converted this over to Typescript.

eweitz (approved):
Looks good, contingent on Florian's suggestions!

src/libs/utils.ts
kevinmarete: This type alias maps a string key (from the `SearchFilterFunctions` interface) to the correct array type for that model. It ensures that when you use a key like `'libraryCard'` or `'darCollections'`, you get the corresponding list type (e.g., `LibraryCard[]`, `DarCollection[]`). If the key does not match any known model, it defaults to `unknown[]`. This helps maintain type safety in generic table or search utilities.

cypress/component/libs/utils.spec.ts
```diff
@@ -263,16 +326,14 @@ describe('Researcher Search Filter (SO Console)', () => {
...
-    })
+    expectResearcherMatch(filteredResearcher, originalResearcher)
   })
 })
 
 describe('processElectionStatus utils - tests', () => {
   it('Returns Unreviewed when election has a null status', () => {
-    const election = { status: null }
-    const status = processElectionStatus(election, null, false)
+    const election = {}
```
fboulnois:
```suggestion
    const election = {} as Election
```

cypress/component/libs/utils.spec.ts
```diff
@@ -263,16 +326,14 @@ describe('Researcher Search Filter (SO Console)', () => {
...
+    const status = processElectionStatus(election as Election, null, false)
```
fboulnois:
```suggestion
    const status = processElectionStatus(election, null, false)
```

Ditto for the other test items.

src/components/library_card_table/LibraryCardTable.tsx
```diff
@@ -218,7 +218,7 @@ const LibraryCardTable: React.FC<LibraryCardTableProps> = (props) => {
       const searchTerms = searchRef.current.value
       let filteredList = libraryCards
       if (!isEmpty(searchTerms)) {
-        filteredList = lcFilterFunction(searchRef, libraryCards)
+        filteredList = lcFilterFunction(searchRef as unknown as string, libraryCards)
```
fboulnois: I think you can set the `searchRef` to have type `<string>`

fboulnois (commented):
Overall looks good, a few additional suggestions:

fboulnois (commented):
Also the SonarQube warnings seem reasonable to address.

src/libs/utils.ts
```diff
@@ -0,0 +1,861 @@
...
+  LibraryCard,
+  DuosUser as Researcher,
+  Dataset,
+  DatasetTerm,
+  InstitutionInterface,
+  DataAccessRequest,
+  DacObject,
+  UserRole,
+  UserProperty,
+} from '../types/model'
```
fboulnois: (don't want to hold up this PR) worth switching to the `src/...` paths at some point.

fboulnois (approved):
👍

PR #3288, opened 2026-01-27
[DT-2848] Set-state-in-effect part 1: Add init functions to address state updates

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2848

### Summary

The goal of this set of PRs is to decrease the number of warnings from ESLint to zero.

When you call state setters or async functions that update state directly in useEffect without wrapping them in an async function, it can cause React to warn about setting state during render or cause race conditions. By wrapping the logic in an async init function and calling it, the state updates happen properly after the effect runs, preventing these warnings and ensuring correct async behavior. This is the documented way to approach this problem: https://react.dev/reference/react/useEffect#fetching-data-with-effects

This decreases the number of ESLint warnings from 24 to 17 (-7).

----

- Verify all tests go green

### DIFF
--- a/src/components/dar_dataset_table/DarDatasetTable.jsx
+++ b/src/components/dar_dataset_table/DarDatasetTable.jsx
@@ -124,12 +124,15 @@ export const DarDatasetTable = (props) => {
   }, [collection, isUnfilteredView, user])
 
   useEffect(() => {
-    try {
-      init()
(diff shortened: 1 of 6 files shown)


otchet-broad (approved):
👍

kevinmarete (approved):

PR #3290, opened 2026-01-27
[DT-2849] Read feature flags from Consent

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2849

### Summary

Adds the ability for code to read feature flags from Consent.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/FeatureFlag.ts
+++ b/src/libs/ajax/FeatureFlag.ts
@@ -1,0 +1,21 @@
+import { Config } from 'src/libs/config'
+import { fetchGet } from 'src/libs/ajax/fetchAdapter'
+
+type FeatureFlagValue = string
+
+export async function getAllFeatureFlags(): Promise<Record<string, FeatureFlagValue> | FeatureFlagValue[]> {
+  const url = `${await Config.getApiUrl()}/api/featureFlags`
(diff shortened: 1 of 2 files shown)


src/libs/ajax/FeatureFlag.ts
```diff
@@ -0,0 +1,21 @@
+import { Config } from 'src/libs/config'
+import { fetchGet } from 'src/libs/ajax/fetchAdapter'
+
+type FeatureFlagValue = string
```
kevinmarete, line 4: This could be, once you decide the flags:
```suggestion
type FeatureFlagValue = 'on' | 'off'
```

kevinmarete (approved):
I have a nit suggestion, but looks good!

src/libs/ajax/FeatureFlag.ts
fboulnois: The flags themselves are not necessarily boolean, in the future I could see us having more than two states for a specific value (e.g. per environment, A/B/n testing, multivalue feature, etc.).

fboulnois: In fact for the ElasticSearch feature, I can imagine having a `elasticsearch5`, `elasticsearch9`, `opensearch` feature flag.

rushtong (approved):
👍🏽

PR #3294, opened 2026-01-28
[DT-2832] Add feature flag for newer versions of ElasticSearch

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2832

### Summary

Add a feature flag for newer versions of ElasticSearch.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/FeatureFlag.ts
+++ b/src/libs/ajax/FeatureFlag.ts
@@ -4,13 +4,13 @@ import { fetchGet } from 'src/libs/ajax/fetchAdapter'
 type FeatureFlagValue = string
 
 export async function getAllFeatureFlags(): Promise<Record<string, FeatureFlagValue> | FeatureFlagValue[]> {
-  const url = `${await Config.getApiUrl()}/api/featureFlags`
+  const url = `${await Config.getApiUrl()}/feature`
   const res = await fetchGet<Record<string, FeatureFlagValue> | FeatureFlagValue[]>(url, Config.authOpts())
   return res.data
 }
 
 export async function getFeatureFlag(key: string): Promise<FeatureFlagValue | undefined> {
-  const url = `${await Config.getApiUrl()}/api/featureFlags/${encodeURIComponent(key)}`
+  const url = `${await Config.getApiUrl()}/feature/${encodeURIComponent(key)}`
   try {
     const res = await fetchGet<FeatureFlagValue>(url, Config.authOpts())
     return res.data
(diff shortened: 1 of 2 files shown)


otchet-broad (approved):
👍 when the test issue is resolved.

src/libs/ajax/FeatureFlag.ts
```diff
@@ -4,13 +4,13 @@ import { fetchGet } from 'src/libs/ajax/fetchAdapter'
 type FeatureFlagValue = string
 
 export async function getAllFeatureFlags(): Promise<Record<string, FeatureFlagValue> | FeatureFlagValue[]> {
-  const url = `${await Config.getApiUrl()}/api/featureFlags`
+  const url = `${await Config.getApiUrl()}/feature`
```
fboulnois, line 7: The endpoint changed to have an unauthenticated `GET`.

src/libs/ajax/FeatureFlag.ts
```diff
@@ -19,3 +19,22 @@ export async function getFeatureFlag(key: string): Promise<FeatureFlagValue | un
     return undefined
   }
 }
+
+// Cache the ES index key name to avoid repeated feature flag lookups
+let esIndexKeyNamePromise: Promise<string> | undefined = undefined
```
fboulnois, line 24: Maybe we should have a separate file for specific feature flags? But also it makes it convenient to have everything in a single file...

src/libs/ajax/FeatureFlag.ts
```diff
@@ -19,3 +19,22 @@ export async function getFeatureFlag(key: string): Promise<FeatureFlagValue | un
...
+
+export const getFlagEsIndexKeyName = (): Promise<string> => {
+  if (esIndexKeyNamePromise === undefined) {
+    esIndexKeyNamePromise = getFeatureFlag('ES_TYPE_TO_INDEX_ENABLED').then((flag) => {
+      return flag === 'true' ? '_index' : '_type'
+    }).catch(() => {
+      return '_type'
```
fboulnois, line 31: This is how I would recommend using the flags: always provide a default if there is an error or if the flag is not defined.

kevinmarete (approved):

src/libs/ajax/FeatureFlag.ts
rushtong: Agreed with keeping it here for now. If it does become unwieldy, we can move it to a utility.

rushtong (approved):

PR #3302, opened 2026-02-03
[DT-2846] Fix "Controlled" facet not working in prod

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2846

### Summary

The "Controlled" facet currently fails in production while functioning correctly in dev. This PR investigates and resolves a probable dataset indexing discrepancy between environments.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/libs/ajax/FeatureFlag.spec.ts
+++ b/cypress/component/libs/ajax/FeatureFlag.spec.ts
@@ -57,99 +57,68 @@ describe('FeatureFlag ajax', () => {
   })
 })
 
-describe('FeatureFlag tests for ES_TYPE_TO_INDEX_ENABLED flag', () => {
-  let fetchStub: ReturnType<typeof cy.stub>
+const createFlagTestSuite = <T>(
+  flagName: string,
+  getFlagFn: () => Promise<T>,
+  resetFn: () => void,
+  mockResponseValue: string,
+  expectedValue: T,
+  expectedErrorValue: T | undefined = undefined,
+) => {
+  describe(`FeatureFlag tests for ${flagName} flag`, () => {
+    let fetchStub: ReturnType<typeof cy.stub>
 
-  beforeEach(() => {
-    resetEsIndexKeyNamePromise()
-    cy.initApplicationConfig()
-    cy.stub(Config, 'getApiUrl').resolves('')
-    cy.window().then((win) => {
-      fetchStub = cy.stub(win, 'fetch')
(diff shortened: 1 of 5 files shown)


cypress/component/libs/ajax/FeatureFlag.spec.ts
```diff
@@ -51,45 +57,68 @@ describe('FeatureFlag ajax', () => {
...
+          expect(fetchStub.callCount).to.equal(1)
+        })
+      })
     })
   })
-})
+}
+
+createFlagTestSuite('ES_TYPE_TO_INDEX_ENABLED', getFlagEsIndexKeyName, resetEsIndexKeyNamePromise, 'true', '_index', '_type')
+createFlagTestSuite('NHGRI_RESTRICTED_DAC', getFlagNhgriDacId, resetNhgriDacIdPromise, 'dac-id', 'dac-id')
```
otchet-broad, line 124: 👍
Nice pattern!

src/libs/ajax/FeatureFlag.ts
```diff
@@ -38,3 +38,16 @@ export const getFlagEsIndexKeyName = (): Promise<string> => {
 export const resetEsIndexKeyNamePromise = () => {
   esIndexKeyNamePromise = undefined
 }
+
+// Cache the NHGRI DAC ID to avoid repeated feature flag lookups
+let nhgriDacIdPromise: Promise<string | undefined> | undefined = undefined
```
otchet-broad, line 43: I'm wondering if this should be generic and we cache all of the feature flags for some period of time (e.g. 10 minutes) and also keep the function to clear the cache so we can do this as needed as well.

otchet-broad (approved):
👍
Nice work!  I've left one thought about caching all of the feature flags for a period of time that could be implemented in a different ticket.  Did you have any thoughts on that?

eweitz (approved):

src/libs/ajax/FeatureFlag.ts
kevinmarete: I think that makes sense, but I would like to hear from @fboulnois if that would work with his original vision of the feature flags.

fboulnois: Right now as defined, this caches the flag for the duration of the entire client session, which seems simplest to me. I don't anticipate we'll be modifying feature flags so frequently that we need to do it mid user-session, but it's something to think about!

PR #3304, opened 2026-02-04
Revert "[DT-2846] Fix "Controlled" facet not working in prod"

kevinmarete:
Reverts DataBiosphere/duos-ui#3302


kevinmarete:
Implemented a fix here: https://github.com/DataBiosphere/duos-ui/pull/3305

PR #3306, opened 2026-02-05
[DT-1689] Allow Signing Officials to Create Users in Same Institution

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1689

### Summary

This PR enables Signing Officials (SOs) to create a new user and assign them a Library Card. It also enforces that the new user's email domains map to the SO’s institution, in line with NIH requirements.

After

Screenshot 2026-02-05 at 5 04 30 PM

----

- Verify all tests go green


src/libs/ajax/User.ts
```diff
@@ -32,13 +32,8 @@ export const User = {
...
   create: async (request: CreateDuosUserRequest): Promise<DuosUser | false | undefined> => {
     const url = `${await Config.getApiUrl()}/api/user/create`
-    try {
-      const res = await fetchPost<DuosUser, CreateDuosUserRequest>(url, request, Config.authOpts())
-      return res.data
-    }
-    catch {
-      return false
-    }
+    const res = await fetchPost<DuosUser, CreateDuosUserRequest>(url, request, Config.authOpts())
```
kevinmarete, line 35: Allow errors to be thrown for displaying to the user.

src/libs/storage.ts
```diff
@@ -68,7 +69,9 @@ export const Storage = {
     localStorage.setItem(OidcUser, JSON.stringify(DEFAULT_OIDC_USER))
   },
 
-  setCurrentUser: (data: DuosUser): void => {
+  setCurrentUser: async (data: DuosUser): Promise<void> => {
+    // Fetch institution data
+    data.institution = await Institution.getById(data.institutionId)
```
kevinmarete: This allows the institution object to be available in the user object.

src/components/modals/AddUserModal.jsx
```diff
@@ -54,15 +54,24 @@ export const AddUserModal = (props) => {
...
 
     setState({
       ...state,
-      emailValid: createdUser,
+      emailValid: !!createUser,
     })
     event.preventDefault()
 
-    if (state.emailValid !== false) {
+    if (createUser) {
```
kevinmarete, line 74: This change is unrelated to this PR, since the `createUserRequest` does not trap errors and return `false` anymore. We must handle this downstream where called.

src/components/forms/formValidation.js
```diff
@@ -25,6 +29,27 @@ export const emailValidator = {
...
+
+    return institutionDomains.includes(newUserDomain)
+  },
+  get msg() {
+    const domains = cachedInstitution?.domains?.join(', ') || ''
+    const baseMessage = 'Please enter an email that matches your organization domains'
+    return domains ? `${baseMessage}: ${domains}` : baseMessage
+  },
+}
+
```
kevinmarete, line 52: This is the new validator that matches emails to institutional domains.

src/components/forms/formValidation.js
```diff
@@ -25,6 +29,27 @@ export const emailValidator = {
   msg: 'Please enter a valid email address (e.g., person@example.com)',
 }
 
+export const emailDomainValidator = {
+  id: 'emailDomain',
+  isValid: async (newUserEmail) => {
+    const institutionId = Storage.getCurrentUser().institutionId
+
+    if (!cachedInstitution) {
+      cachedInstitution = await Institution.getById(institutionId)
```
rushtong, line 38: This should work as expected. I just want to call out a back end detail that we have [some role-based logic to determine which institution fields are populated](https://github.com/DataBiosphere/consent/blob/4e3c41f556b3bddf954a3a12d1c04fa36655784d/src/main/java/org/broadinstitute/consent/http/util/InstitutionUtil.java#L36). Since we're looking at domains here, it is important that those are populated, which they should be for all roles.

rushtong (approved):
Looks good, thank you 👍🏽

otchet-broad (approved):
Nice work!

PR #3326, opened 2026-02-13
[DT-2950] Fix Remaining ESLint Warnings

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2950

### Summary

This PR fixes the remaining 17 ESLint warnings on the error below:
```
Error: Calling setState synchronously within an effect can trigger cascading renders
```

----

- Verify all tests go green

### DIFF
--- a/src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
+++ b/src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react'
+import React, { useMemo, useState } from 'react'
 import { isEmpty, isNil } from 'lodash'
 import CollectionVoteYesButton from './CollectionVoteYesButton'
 import CollectionVoteNoButton from './CollectionVoteNoButton'
@@ -151,7 +151,7 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props) 
     return props.isDisabled || (isFinal && submitted) || adminPage
   }, [props.isDisabled, isFinal, submitted, adminPage])
 
-  useEffect(() => {
+  React.useEffect(() => {
     if (!isEmpty(votes)) {
       const prevVote = votes[0]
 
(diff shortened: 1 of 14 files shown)


src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
```diff
@@ -151,7 +151,7 @@ const CollectionSubmitVoteBox: React.FC<CollectionSubmitVoteBoxProps> = (props)
     return props.isDisabled || (isFinal && submitted) || adminPage
   }, [props.isDisabled, isFinal, submitted, adminPage])
 
-  useEffect(() => {
+  React.useEffect(() => {
```
rushtong, line 154: What is the reason for this change? IOW, is there a specific lint or sonar warning for this and how is it different from the original code? If it is necessary, why not also do this for `useMemo` or `useState`?

kevinmarete: This is the error this change is fixing:

`Error: Calling setState synchronously within an effect can trigger cascading renders`

The change to `React.useEffect` fixes the error because the original `useEffect` callback was not being scheduled correctly, and `setState` was effectively running synchronously, triggering the cascading render error.

Using `React.useEffect` ensures we call the hook directly from the `react` module, restoring the correct lifecycle behavior.

We don’t need to do the same for `useMemo` or `useState` unless there’s a similar risk for them. There is no behavioral difference between `useX` and `React.useX`.

otchet-broad (approved):
👍

src/components/collection_vote_box/CollectionSubmitVoteBox.tsx
rushtong: I have a hard time grokking that difference. Other [references](https://stackoverflow.com/questions/78431141/is-there-a-difference-between-e-g-useeffect-react-useeffect) seem to disagree with this.

rushtong (approved):
👍🏽

PR #3328, opened 2026-02-17
[DT-2459] Add new data library filters

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2459

### Summary

Adds the data library filters to the new data library

Screenshot 2026-02-19 at 10 26 24 AM

### DIFF
--- a/src/components/data_library/LibraryFilters.tsx
+++ b/src/components/data_library/LibraryFilters.tsx
@@ -1,0 +1,245 @@
+import React from 'react'
+import {
+  Box,
+  Button,
+  Checkbox,
+  TextField,
+  Typography,
+  Accordion,
+  AccordionSummary,
+  AccordionDetails,
+  FormGroup,
+  FormControlLabel,
+  Skeleton,
+} from '@mui/material'
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
+import { LibraryFiltersProps } from 'src/types/library'
+
+export const LibraryFilters: React.FC<LibraryFiltersProps> = ({
+  filters,
+  onChange,
+  onClear,
+  availableFilters,
+  loading = false,
+}) => {
+  const handleFilterToggle = (category: keyof typeof filters, value: string) => {
+    const currentValues = filters[category] as string[]
+    const newValues = currentValues.includes(value)
+      ? currentValues.filter(v => v !== value)
+      : [...currentValues, value]
+
+    onChange({
+      ...filters,
+      [category]: newValues,
+    })
+  }
+
+  const handleParticipantChange = (type: 'min' | 'max', value: string) => {
+    const numValue = value === '' ? undefined : Number.parseInt(value)
+    onChange({
+      ...filters,
+      participantCount: {
+        ...filters.participantCount,
+        [type]: numValue,
+      },
+    })
+  }
+
+  const hasActiveFilters
+    = filters.accessManagement.length > 0
+      || filters.dataUse.length > 0
+      || filters.dataType.length > 0
+      || filters.dac.length > 0
+      || filters.participantCount.min !== undefined
+      || filters.participantCount.max !== undefined
+
+  return (
+    <Box sx={{ width: '100%' }}>
+      <Box
+        sx={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          alignItems: 'center',
+          mb: 2,
+        }}
+      >
+        <Typography variant="h6">Filters</Typography>
+        {hasActiveFilters && (
+          <Button size="small" onClick={onClear}>
+            Clear
+          </Button>
+        )}
+      </Box>
+
+      {loading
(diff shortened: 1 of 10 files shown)


src/types/library.ts
```diff
@@ -8,9 +8,56 @@ export enum AssetType {
   MODELS = 'models',
 }
 
+export interface LibraryVersionNew {
+  key: string
+  query: unknown | null
+  icon: string | null
```
kevinmarete: These can be optional

```suggestion
  query?: any
  icon?: string
```

src/pages/DataLibrary.tsx
```diff
@@ -27,22 +29,80 @@ export const DataLibrary: React.FC = () => {
 
   const searchRef = useRef<HTMLInputElement>(null)
 
+  const libraryConfig: LibraryVersionNew = {
+    key: 'duos',
+    query: null,
+    icon: null,
```
kevinmarete: Do not need to be `null`

src/hooks/useLibraryData.ts
```diff
@@ -0,0 +1,25 @@
...
+  filters: FilterState,
+) => {
+  return useQuery({
+    queryKey: [
+      'library-data',
+      libraryConfig.key,
+      assetType,
+      filters,
+    ],
+    staleTime: 5 * 60 * 1000,
```
kevinmarete: Is this `300 miliseconds`?  Maybe add a //comment

src/hooks/useLibraryUrlState.ts
```diff
@@ -1,17 +1,75 @@
...
+    dataType: searchParams.get('dataType')?.split(',').filter(Boolean) || [],
+    dac: searchParams.get('dac')?.split(',').filter(Boolean) || [],
+    participantCount: {
+      min: searchParams.get('minParticipants')
+        ? Number.parseInt(searchParams.get('minParticipants')!)
+        : undefined,
+      max: searchParams.get('maxParticipants')
+        ? Number.parseInt(searchParams.get('maxParticipants')!)
+        : undefined,
+    },
```
kevinmarete, line 20: You can use a spread operator to extract the searchParams into variables expecially if you need to keep referencing them multiple times.


```suggestion
const { access, dataUse, dataType, dac, minParticipants, maxParticipants} = searchParams
    accessManagement: access?.split(',').filter(Boolean) || [],
    dataUse: dataUse?.split(',').filter(Boolean) || [],
    dataType: dataType?.split(',').filter(Boolean) || [],
    dac: dac?.split(',').filter(Boolean) || [],
    participantCount: {
      min: Number.parseInt(minParticipants!) || undefined,
      max: Number.parseInt(maxParticipants!) || undefined,
    },
```

src/components/data_library/LibraryFilters.tsx
```diff
@@ -0,0 +1,245 @@
...
+import { LibraryFiltersProps } from 'src/types/library'
+
+export const LibraryFilters: React.FC<LibraryFiltersProps> = ({
+  filters,
+  onChange,
+  onClear,
+  availableFilters,
+  loading = false,
+}) => {
+  const handleFilterToggle = (category: keyof typeof filters, value: string) => {
```
kevinmarete, line 25: It would be nice to add an explicit category type for re-use and testing instead of `keyof typeoff filters`

src/components/data_library/LibraryFilters.tsx
```diff
@@ -0,0 +1,245 @@
...
+      ? currentValues.filter(v => v !== value)
+      : [...currentValues, value]
+
+    onChange({
+      ...filters,
+      [category]: newValues,
+    })
+  }
+
+  const handleParticipantChange = (type: 'min' | 'max', value: string) => {
```
kevinmarete, line 37: This one too!

kevinmarete (approved):
I left a few comments, mostly optional, but great job on this!

src/hooks/useLibraryUrlState.ts
fboulnois: This doesn't seem to work as expected:

```
const searchParams = new URLSearchParams("hello=world&foo=bar");
const { hello, foo } = searchParams;
console.log(hello)
console.log(foo)
```

Both are `undefined`.

src/components/data_library/LibraryFilters.tsx
fboulnois: I will hold off on these for now. This is the only location where `keyof typeof filters` is used. Adding a `type FilterStateKey = keyof FilterState` in the library types creates extra indirection.

src/components/data_library/LibraryFilters.tsx
fboulnois: Ditto here.

rushtong (approved):
Nice 👍🏽

PR #3331, opened 2026-02-19
[DT-2459] New data library grid

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2459

### Summary

Creates an empty data library grid for the new data library.

Screenshot 2026-02-19 at 10 28 10 AM

### DIFF
--- a/src/components/data_library/columns/datasetColumns.tsx
+++ b/src/components/data_library/columns/datasetColumns.tsx
@@ -1,0 +1,99 @@
+import React from 'react'
+import { GridColDef } from '@mui/x-data-grid'
+import { Link, Chip, Box, Tooltip } from '@mui/material'
+import { DatasetTerm, getAccessManagementSummary } from 'src/types/model'
+
+/**
+ * Column definitions for dataset view
+ */
+export const makeDatasetColumns = (): GridColDef<DatasetTerm>[] => [
+  {
+    field: 'datasetName',
+    headerName: 'Dataset Name',
+    flex: 1.5,
+    minWidth: 200,
+    renderCell: params => (
+      <Link href={`/dataset/${params.row.datasetId}`} underline="hover">
+        {params.value}
+      </Link>
+    ),
+  },
+  {
+    field: 'studyName',
+    headerName: 'Study Name',
+    flex: 1,
+    minWidth: 150,
+    valueGetter: (_value, row) => row.study?.studyName || '',
+    renderCell: params => (
+      <Link href={`/studies/${params.row.study?.studyId}`} underline="hover">
+        {params.value}
+      </Link>
+    ),
+  },
+  {
+    field: 'participantCount',
+    headerName: 'Participants',
+    width: 120,
+    type: 'number',
+    align: 'right',
+    headerAlign: 'right',
+  },
+  {
+    field: 'dataUse',
+    headerName: 'Data Use',
+    width: 150,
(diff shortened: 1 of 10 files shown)


src/types/library.ts
```diff
@@ -58,6 +58,10 @@ export interface LibraryUrlState {
   library: string
   tab: AssetType
   filters: FilterState
+  page: number
+  pageSize: number
+  sortField?: string
+  sortOrder?: 'asc' | 'desc'
```
kevinmarete: I would suggest creating a SortType with 'asc' and 'desc' since there is repetition in defining 'asc' and 'desc' on this page.

src/hooks/useLibraryUrlState.ts
```diff
@@ -83,6 +83,10 @@ export const useLibraryUrlState = () => {
     library: searchParams.get('library') || 'duos',
     tab: (searchParams.get('tab') as AssetType) || AssetType.STUDIES,
     filters: parseFiltersFromUrl(searchParams),
+    page: Number.parseInt(searchParams.get('page') || '0'),
+    pageSize: Number.parseInt(searchParams.get('pageSize') || '25'),
```
kevinmarete, line 87: I think the defaults should be integers 
```suggestion
    page: Number.parseInt(searchParams.get('page') || 0),
    pageSize: Number.parseInt(searchParams.get('pageSize') || 25),
```

src/hooks/useLibraryUrlState.ts
```diff
@@ -83,6 +83,10 @@ export const useLibraryUrlState = () => {
...
+    sortField: searchParams.get('sort') || undefined,
+    sortOrder: (searchParams.get('order') as 'asc' | 'desc') || undefined,
```
kevinmarete: Shouldn't we have default sorting instead of undefined, like the ID and DESC?

src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -0,0 +1,99 @@
...
+      const summary = getAccessManagementSummary(params.value)
+      return (
+        <Tooltip title={summary.description}>
+          <Chip
+            label={summary.name}
+            size="small"
+            color={
+              params.value === 'controlled'
+                ? 'primary'
+                : params.value === 'open'
```
kevinmarete: Do we have types for `controlled` | `open`?

kevinmarete (approved):
Some nit comments, 👍🏾

src/hooks/useLibraryUrlState.ts
fboulnois: `parseInt` parses a string argument, but instead I think we want: `Number.parseInt(searchParams.get('page')) || 0`

fboulnois: Actually, `get` can return `null` which is why it's done this way.

src/hooks/useLibraryUrlState.ts
fboulnois: This is linked to the URL state, so it can be undefined.

src/components/data_library/columns/datasetColumns.tsx
fboulnois: No, I will add that.

rushtong (approved):
👍🏽

PR #3337, opened 2026-02-20
DT-2918: Get Terms of Service state from user instead of separate API call

rushtong:
### Addresses
Front end work related to https://broadworkbench.atlassian.net/browse/DT-2918

### Summary
~This PR requires https://github.com/DataBiosphere/consent/pull/2817~ (merged)

In the back end PR, the user status from Sam is now stored on the user object returned from the `GET /api/user/me` call which allows us to skip the API call we were making in `ToS.getStatus`

Note in this network tab that the extra diagnostics call seen in https://github.com/DataBiosphere/consent/pull/2817 is no longer made. Note also that a user who has not accepted ToS is redirected to the ToS screen instead of their original intended location.

Screenshot 2026-02-20 at 4 22 05 PM

Screenshot 2026-02-20 at 4 23 49 PM

----

- Verify all tests go green


kevinmarete (approved):

otchet-broad (approved):
👍

PR #3340, opened 2026-02-24
[DT-2975] Add "apply for access" footer and asset selection

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2975

### Summary

Adds the "apply for access" library footer with tests.

Screenshot 2026-02-24 at 10 30 16 AM

----

- Verify all tests go green

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -828,11 +828,13 @@ a[disabled], .cell-button[disabled], [disabled] .cell-button, [disabled] button,
   text-overflow: ellipsis;
 }
 
-input[type=checkbox] {
+.checkbox input[type=checkbox],
+.checkbox-inline input[type=checkbox] {
   display: none;
 }
 
-input[type=checkbox]:checked+label:before {
+.checkbox input[type=checkbox]:checked+label:before,
+.checkbox-inline input[type=checkbox]:checked+label:before {
   content: "\2713";
   text-shadow: 1px 1px 1px rgba(0, 0, 0, .2);
   color: #ffffff;
--- a/src/pages/DataLibrary.tsx
+++ b/src/pages/DataLibrary.tsx
@@ -5,7 +5,8 @@ import LibraryTabs from 'src/components/data_library/LibraryTabs'
 import SearchBar from 'src/components/SearchBar'
 import TableHeaderSection from 'src/components/TableHeaderSection'
 import { useLibraryUrlState } from 'src/hooks/useLibraryUrlState'
-import { AssetType, AvailableFilters, LibraryVersionNew, SortOrder, TabConfig } from 'src/types/library'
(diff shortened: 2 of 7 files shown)


src/index.css
```diff
@@ -828,11 +828,13 @@ a[disabled], .cell-button[disabled], [disabled] .cell-button, [disabled] button,
   text-overflow: ellipsis;
 }
 
-input[type=checkbox] {
+.checkbox input[type=checkbox],
```
fboulnois, line 831: These rules completely broke Material UI checkboxes.

src/pages/DataLibrary.tsx
```diff
@@ -103,6 +108,32 @@ export const DataLibrary: React.FC = () => {
...
+            studyIds.add(dataset.study.studyId)
+          }
+          break
+        }
+        default:
+          throw new Error('Unknown asset type')
+      }
+    })
+    return Array.from(studyIds)
+  }, [data, selectedDatasetIds, urlState.tab])
```
otchet-broad, line 136: @fboulnois Can urlState change from having `.tab` to not having `.tab`?  What are the drawbacks of just using urlState?

fboulnois: The drawbacks are:
* the memoized value would be recomputed whenever any property of `urlState` changes (page, pagesize, sortField, sortOrder, query, filters)
* the computation loops through `data.items`, which can be expensive if the list and/or items are large.
* downstream components that consume `selectStudyIds` would also be recomputed unnecessarily.

rushtong (approved):
Looks good, thank you 👍🏽

src/pages/DataLibrary.tsx
```diff
@@ -160,6 +191,11 @@ export const DataLibrary: React.FC = () => {
     setSelectedDatasetIds(datasetIds)
   }
 
+  const handleApplyForAccess = () => {
+    const datasetIdsParam = selectedDatasetIds.join(',')
+    navigate(`/dar_application?datasetIds=${datasetIdsParam}`)
```
otchet-broad: Is this what we did before?  I vaguely remember we would actually post to submit a draft so that we would hit logic that would deny someone without a library card before they'd get to the DAR application form.

fboulnois: Yes good point. We have a tooltip that prevents applying for access when you don't have a library card.

otchet-broad: It looks like we post to /[draft](https://consent.dsde-dev.broadinstitute.org/api/dar/v2/draft) and then handle the display of errors before we navigate to the new page.

Do we need to still do that today?

fboulnois: Yes, I'll restore this functionality.

otchet-broad (approved):
👍

PR #3343, opened 2026-02-24
DT-2957: Update Dataset Statistics page to use new API and clearly show expiration

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2957

### Summary

* Requires back end PR here: https://github.com/DataBiosphere/consent/pull/2822
* Minor update to the page so that it pulls from the authenticated dataset metrics API
* Changes the update date to show if the DAR is expired or not

Screenshot 2026-02-24 at 4 34 00 PM

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/DatasetMetrics.js
+++ b/src/libs/ajax/DatasetMetrics.js
@@ -3,7 +3,7 @@ import { fetchGet } from 'src/libs/ajax/fetchAdapter'
 
 export const DatasetMetrics = {
   getDatasetStats: async (datasetId) => {
-    const url = `${await Config.getApiUrl()}/metrics/dataset/${datasetId}`
+    const url = `${await Config.getApiUrl()}/api/metrics/dar-summaries/${datasetId}`
     const res = await fetchGet(url, Config.authOpts())
     return res.data
   },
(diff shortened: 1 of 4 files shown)


src/libs/ajax/DatasetMetrics.js
```diff
@@ -3,7 +3,7 @@ import { fetchGet } from 'src/libs/ajax/fetchAdapter'
 
 export const DatasetMetrics = {
   getDatasetStats: async (datasetId) => {
-    const url = `${await Config.getApiUrl()}/metrics/dataset/${datasetId}`
+    const url = `${await Config.getApiUrl()}/api/metrics/dar-summaries/${datasetId}`
```
fboulnois, line 6: Shouldn't this already have `/api/`?

fboulnois (commented):
Looks good, one question:

src/libs/ajax/DatasetMetrics.js
otchet-broad: I think this pairs with [https://github.com/DataBiosphere/consent/pull/2822](https://github.com/DataBiosphere/consent/pull/2822)

otchet-broad (approved):
👍

src/libs/ajax/DatasetMetrics.js
otchet-broad: Config.getApiUrl() returns the base, but we tack on /api/ to it in all of our fetch requests.  The name of the get function is somewhat misleading.  Maybe we should rename it from `Config.getapiUrl` to `Config.getConsentBaseUri`

rushtong: There are other unauthenticated calls we make to Consent (the previous metrics call, status, terms of service, etc.) which is why the base url doesn't have `/api` by default.

fboulnois (approved):
👍

eweitz (approved):

PR #3348, opened 2026-02-25
DT-2918: Remove awaits for sign-in related reporting calls

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-2918

### Summary
This PR removes the `await` calls to Metrics and Stackdriver reporter as the results of those calls are ignored. The result is a bit more improved in that the destination path of the UI begins to paint more quickly (i.e. the spinner loads faster) given that the sign-in flow no longer blocks on the metrics calls. The overall page load is the same, however.

----

- Verify all tests go green

### DIFF
--- a/src/components/SignInButton.tsx
+++ b/src/components/SignInButton.tsx
@@ -72,7 +72,8 @@ export const SignInButton = () => {
         if (!duosUser.roles) {
           await StackdriverReporter.report('roles not found for user: ' + duosUser.email)
         }
-        await syncSignInOrRegistrationEvent(eventList.userSignIn)
+        // noinspection ES6MissingAwait
+        syncSignInOrRegistrationEvent(eventList.userSignIn)
         await checkToSAndRedirect(shouldRedirect ? redirectTo : null)
       }
       else {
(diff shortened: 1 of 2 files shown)


src/components/SignInButton.tsx
```diff
@@ -103,7 +104,8 @@ export const SignInButton = () => {
   const registerAndRedirectNewUser = async (redirectTo: string, shouldRedirect: boolean) => {
     const registeredUser: DuosUser = await User.registerUser()
     setUserRoleStatuses(registeredUser, Storage)
-    await syncSignInOrRegistrationEvent(eventList.userRegister)
+    // noinspection ES6MissingAwait
+    syncSignInOrRegistrationEvent(eventList.userRegister)
```
fboulnois, line 106: It looks like the underlying methods in `syncSignInOrRegistrationEvent` already ignore the `await` issue, so should we instead switch it to a synchronous method?

fboulnois (approved):
One suggestion:

otchet-broad (approved):
👍

PR #3362, opened 2026-03-02
[DT-3012] Data models tab

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3012

### Summary

Refactors the asset code to make adding new assets more scalable, and adds the AI models tab:

Screenshot 2026-03-02 at 9 57 19 AM

----

- Verify all tests go green

### DIFF
--- a/src/components/data_library/columns/modelColumns.tsx
+++ b/src/components/data_library/columns/modelColumns.tsx
@@ -1,0 +1,152 @@
+import React from 'react'
+import { GridColDef } from '@mui/x-data-grid'
+import { Link, Chip, Box, Tooltip } from '@mui/material'
+import { ModelRow } from 'src/types/library'
+
+/**
+ * Column definitions for AI model view
+ */
+export const makeModelColumns = (): GridColDef<ModelRow>[] => [
+  {
+    field: 'name',
+    headerName: 'Model Name',
+    flex: 1.5,
+    minWidth: 200,
+    renderCell: (params) => {
+      const text = params.value || ''
+      return (
+        <Tooltip title={text} placement="top">
+          <Box
+            sx={{
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+              whiteSpace: 'nowrap',
+            }}
+          >
+            {text}
+          </Box>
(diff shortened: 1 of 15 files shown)


kevinmarete (approved):
This is great. Love the simplification from conditionals to asset definitions, which provides a scalable solution for other assets. Why do we need to define `ModelRow` when it seems to duplicate the existing `AiModel` type?

fboulnois:
> This is great. Love the simplification from conditionals to asset definitions, which provides a scalable solution for other assets. Why do we need to define `ModelRow` when it seems to duplicate the existing `AiModel` type?

One is supposed to be the Elasticsearch representation, and one is the representation in as a row in the data library. But in this case we can probably simplify.

src/components/data_library/columns/modelColumns.tsx
```diff
@@ -0,0 +1,152 @@
...
+    minWidth: 150,
+    sortable: false,
+    valueGetter: (_value, row) => (row.tags || []).join(', '),
+    renderCell: (params) => {
+      const tags = params.row.tags || []
+      if (tags.length === 0) return null
+      return (
+        <Box sx={{ display: 'flex', gap: 0.5, flexWrap: 'wrap' }}>
+          {tags.slice(0, 3).map((tag, i) => (
+            <Chip key={i} label={tag} size="small" variant="outlined" />
```
otchet-broad, line 141: Is there a way to do this without the key being the index of the item in the array? We're getting a sonar/react warning.

otchet-broad (approved):
Looks great!  Tested locally.

I left you one quick question on the columnModel that can be addressed now or later.

src/components/data_library/columns/modelColumns.tsx
fboulnois: Yes, Sonarqube is warning us but I think we should skip the warning. The array keys are guaranteed to be unique, the `map` index is a [well defined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) parameter, and there is no additional information on the "error" in question.

PR #3374, opened 2026-03-05
[DT-3067] Add workspaces asset tab to new data library

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3067

### Summary

Add workspaces to the new data library with tests.

Screenshot 2026-03-05 at 1 47 32 PM

----

- Verify all tests go green

### DIFF
--- a/cypress/component/data_library/assets/workspaceAsset.spec.ts
+++ b/cypress/component/data_library/assets/workspaceAsset.spec.ts
@@ -1,0 +1,392 @@
+/**
+ * Unit tests for workspaceAsset — the Workspaces AssetDefinition.
+ *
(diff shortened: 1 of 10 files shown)


rushtong (approved):
Looks great, thank you 👍🏽 
I do have one question that could apply to multiple tab types. Would it make sense to remove the checkbox entirely from tabs where selection is not possible?

kevinmarete (approved):

fboulnois:
> Looks great, thank you 👍🏽 I do have one question that could apply to multiple tab types. Would it make sense to remove the checkbox entirely from tabs where selection is not possible?

I think our current implementation has a tooltip on non-selectable checkboxes. We can probably investigate whether this is possible on the new data library.

PR #3382, opened 2026-03-10
[DT-2953] Use study tags for branded data library filtering

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2953

### Summary

Branded data libraries currently use custom queries that filter datasets based on study description text. 
This PR change updates the queries to also use study tags. 

----

- Verify all tests go green

### DIFF
--- a/src/libs/libraryVersions.ts
+++ b/src/libs/libraryVersions.ts
@@ -230,7 +230,7 @@ export const getLibraryVersions = (
             },
             {
               terms: {
-                'study.data.tags': ['Single Cell Portal'],
+                'study.data.tags': ['Platform: Single Cell Portal'],
               },
             },
           ],
--- a/src/libs/libraryVersions.ts
+++ b/src/libs/libraryVersions.ts
@@ -38,9 +38,15 @@ interface TermQuery {
   }
 }
 
-interface BoolQuery {
+interface TermsQuery {
+  terms: {
+    [key: string]: string[]
+  }
+}
+
+export interface BoolQuery {
   bool: {
-    should: Array<MatchPhraseQuery | TermQuery>
+    should: Array<MatchPhraseQuery | TermQuery | TermsQuery>
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/libs/libraryVersions.ts
```diff
@@ -166,8 +221,19 @@ export const getLibraryVersions = (
...
+          should: [
+            {
+              match_phrase: {
+                'study.description': 'Single Cell Portal',
+              },
+            },
+            {
+              terms: {
+                'study.data.tags': ['Single Cell Portal'],
+              },
```
rushtong, line 234: I think this is likely outside the scope of this PR, but the dev environment tag for SCP looks like this:
```
"data": {
    "tags": [
        "(Platform: Single Cell Portal)"
    ]
}
```

rushtong (approved):
This looks reasonable assuming tags all follow the same pattern as the current study description 👍🏽

otchet-broad (approved):
The queries look good.  Since they're or'ed together with the existing terms we shouldn't have issues.
One thing I was left thinking about with this is if we should make the tags queries case insensitive.  Not necessary for this PR, but link to the concept is here:
[https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-term-query#:~:text=the%20relevance%20score.-,case_insensitive,-Added%20in%207.10.0](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-term-query#:~:text=the%20relevance%20score.-,case_insensitive,-Added%20in%207.10.0)

kevinmarete:
> The queries look good. Since they're or'ed together with the existing terms we shouldn't have issues. One thing I was left thinking about with this is if we should make the tags queries case insensitive. Not necessary for this PR, but link to the concept is here: https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-term-query#:~:text=the%20relevance%20score.-,case_insensitive,-Added%20in%207.10.0

I agree we should for any use of the `term` query. However, for this case, we are using `match_phrase` and `terms`, which do not have the `case_insensitive` parameter.

PR #3385, opened 2026-03-11
[DT-2919]  Prevent Data Library hard-refresh logout

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2919

### Summary

This PR fixes a Data Library issue in which a hard refresh could log users out due to multiple background 401 responses. 

We centralized 401 skip logic in `src/utils/AuthRedirectUtils.ts` and updated `src/libs/ajax/fetchAdapter.ts` to use it. 

Now, when users are on Data Library routes, 401s from the TDR enrichment GET call `/api/repository/v1/snapshots` no longer trigger forced logout redirects, while existing auth behavior remains unchanged for other endpoints.

----

- Verify all tests go green


src/utils/AuthRedirectUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ *     by the auth layer and should not trigger a forced logout.
+ *  2. Data Library background endpoints — may return 401 due to missing resource
+ *     permissions rather than an expired session, so the user should not be
+ *     logged out while browsing the library.
+ *
+ * All non-GET requests fall through immediately: a 401 on a mutating request
+ * almost always means the session is invalid and logout is the correct response.
+ */
+export const shouldSkip401Redirect = (url: string, method: string): boolean => {
+  if (method !== 'GET') return false
```
otchet-broad, line 15: I'm wondering if there's a different approach here and it might get to why we're seeing some premature logouts.

We should only consider 401s as logouts when they're from DUOS.  If the url's hostname isn't the Config.getApiUrl() hostname, we should also skip the redirect - because the 401 isn't coming from **consent** (e.g. it could be the TDR enrichment call - /api/repository/v1/snapshots).

Thoughts?

src/libs/ajax/fetchAdapter.ts
```diff
@@ -69,9 +71,9 @@ async function handleResponse<T>(
   responseType: ResponseType,
   method: Method = 'GET',
 ): Promise<FetchData<T>> {
-  const isMeCheck = url.endsWith('/api/user/me') && method.toLowerCase() === 'get'
   if (!res.ok) {
-    if (res.status === 401 && !isMeCheck) {
+    const apiUrl = await Config.getApiUrl()
+    if (res.status === 401 && !shouldSkip401Redirect(url, method, apiUrl)) {
```
fboulnois, line 76: Why did this trigger with TDR beforehand?

src/utils/AuthRedirectUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+ * This is used to prevent infinite redirect loops when the auth probe endpoint returns a 401 response, which can happen if the user's session has expired or if they are not authenticated.
+ *
+ * @param url The URL of the request that resulted in a 401 response.
+ * @param method The HTTP method of the request that resulted in a 401 response.
+ * @param apiUrl The base URL of the DUOS API, used to determine if the request was made to the DUOS API.
+ * @returns A boolean indicating whether to skip the 401 redirect.
+ */
+export const shouldSkip401Redirect = (
+  url: string, method: string, apiUrl: string): boolean => {
+  if (method !== 'GET') return false
```
fboulnois, line 15: What if it is a POST that returns 401, shouldn't we log out in that case too?

fboulnois (commented):
A few questions:

src/utils/AuthRedirectUtils.ts
eweitz: I think this does that.  (POST isn't GET, so we do not skip redirect, i.e. we do redirect / sign-out on POST 401.)

kevinmarete: Yes @eweitz is correct, all non-GET requests would be redirect/sign-out on POST 401

eweitz (approved):
Looks good!

IIUC, we cannot reliably reproduce users getting logged out prematurely.  My forthcoming PR will add some metrics / analytics to evidence a fix (or not), but your updates here seem like a plausible solution for the problem.

src/libs/ajax/fetchAdapter.ts
kevinmarete: I may not fully understand your question, but this could be triggered by a race condition that occurs occasionally when there is a delay between DUOS and TDR authentication.

fboulnois:
> Looks good!
> 
> IIUC, we cannot reliably reproduce users getting logged out prematurely. My forthcoming PR will add some metrics / analytics to evidence a fix (or not), but your updates here seem like a plausible solution for the problem.

Should we wait until your PR is in before merging this PR, to measure the difference?

fboulnois (approved):
👍

PR #3389, opened 2026-03-16
[DT-2805] Add Mixpanel metrics on logout

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2805

### Summary
Users have sometimes gotten logged out very soon after logging in.  We have not been able to reliably and realistically reproduce this premature sign-out issue.  That said, we have implemented a plausible solution in #3385.

This adds analytics to corroborate our individual observations on whether our solution indeed fixes the problem.  The new `user:autoLogout401` is logged to Bard / Mixpanel upon applicable 401 redirects.

Here's how the event looks in Mixpanel.  (You can observe it [in the "Terra Analytics Dev" Mixpanel project](https://mixpanel.com/project/2085496/view/19055/app/events#ezv5HC2FzXYV) now, and "Terra Analytics Prod" after deployment.). The most relevant property is `time_until_expires`.  If it's 0 or negative, then the user's auth session has expired.  

If the rate of these requests events increase, or these rich metrics reveal we're seeing `user:autoLogout401` where `time_until_expires > 0`, then we'll have new causal insight on the problem.  (By default, we anticipate the metrics will simply corroborate the fix.)  The new data here also provides a baseline for monitoring future premature logout issues.

Screenshot 2026-03-17 at 10 32 44 AM

----

- Verify all tests go green

### DIFF
--- a/cypress/component/libs/ajax/fetchAdapter.spec.ts
+++ b/cypress/component/libs/ajax/fetchAdapter.spec.ts
@@ -629,6 +629,7 @@ describe('fetchAdapter - 401 Bard metric logging', () => {
           expect(event).to.equal(eventList.userAutoLogout401)
           expect(details).to.have.property('expires_on', mockExpTime)
           expect(details).to.have.property('current_time').that.is.a('number')
+          expect(details).to.have.property('time_until_expires').that.is.a('number')
           expect(details).to.have.property('endpoint_url', 'https://consent.example.org/api/something')
         },
       )
@@ -694,6 +695,7 @@ describe('fetchAdapter - 401 Bard metric logging', () => {
           expect(captureEventStub.calledOnce).to.equal(true)
           const [, details] = captureEventStub.firstCall.args
           expect(details).to.have.property('expires_on', null)
+          expect(details).to.have.property('time_until_expires', null)
         },
       )
     })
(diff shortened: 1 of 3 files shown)


cypress/component/libs/ajax/fetchAdapter.spec.ts
```diff
@@ -574,3 +579,144 @@ describe('fetchAdapter - Fetch methods', () => {
     })
   })
 })
+
+describe('fetchAdapter - 401 Bard metric logging', () => {
+  let fetchStub: ReturnType<typeof cy.stub>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let captureEventStub: any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let signOutStub: any
```
kevinmarete: You can use 'never'  instead of 'any' to avoid using eslint-disable comments.

kevinmarete (approved):

cypress/component/libs/ajax/fetchAdapter.spec.ts
fboulnois: I think in this case the typing could be more specific without resorting to `never`, `any` or `unknown`.

fboulnois (approved):
👍

cypress/component/libs/ajax/fetchAdapter.spec.ts
eweitz: Done!  9b73a31

PR #3391, opened 2026-03-17
[DT-3093] Fix broken search box in Console pages

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3093

### Summary
This fixes the search box on DAR tables for admin, chairs, and researchers.  Previously, entering text in the bug would never return results in the table, even when the rows had matching content.  Now, entering text returns matching rows as expected.

The cause was that `filterDarCollections` searched `DarCollection` fields, but the API returns `DarCollectionSummary` objects.  Every search term matched against [undefined, "0", "- -", "- -"] for every row, so all rows disappeared.  Fixing the subtle type error updates the filter to search the actual summary fields, and returns expected results.

Before and after:

https://github.com/user-attachments/assets/0e860af4-b859-4503-a710-224db603c404

----

- Verify all tests go green


fboulnois (approved):
👍

kevinmarete (approved):

PR #3418, opened 2026-03-31
[DT-2886] Add external profile fields and validation to SO role request

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2886

### Summary

This PR updates the “Request a New Role” page to support identity verification for SO role requests. When the SO checkbox is selected, four new input fields are displayed for external profile URLs (LinkedIn, ORCID, Through.Bio, and Institutional). The system validates that all entered values are well-formed URLs and requires at least one valid URL before submission. Submitted URLs are saved to the user’s profile for review by admins and DACs.

After

----

- Verify all tests go green


package.json
```diff
@@ -27,7 +27,7 @@
     "ajv-formats": "3.0.1",
     "dayjs": "1.11.20",
     "dompurify": "3.3.3",
-    "lodash": "4.17.23",
+    "lodash": "4.18.1",
```
rushtong: Shouldn't this be in a separate PR? If there are regressions in any lodash functions, all of the current functionality will need to be reverted alongside the library update.

src/pages/user_profile/RequestForm.tsx
```diff
@@ -69,12 +71,14 @@ export default function RequestForm(): JSX.Element {
...
   const [isSubmitting, setIsSubmitting] = useState<boolean>(false)
   const [supportRequests, setSupportRequests] = useState<SupportRequests>(supportRequestsCond)
+  const [showExternalProfileUrls, setShowExternalProfileUrls] = useState<boolean>(false)
 
   const handleSupportRequestsChange = ({ key, value }: HandleSupportRequestsChangeArg) => {
     const newSupportRequests = { ...supportRequests, [key]: value }
     setSupportRequests(newSupportRequests)
     const hasAnyRequests = possibleSupportRequests.some(request => newSupportRequests[request.key])
     setHasSupportRequests(hasAnyRequests)
+    setShowExternalProfileUrls(newSupportRequests.checkSOPermissions)
```
rushtong, line 87: One thing that's odd about this page is that the user can request SO even if they already have that role. This is out of scope of the PR, but I would recommend we disable that checkbox if the user has the SO role.

rushtong (commented):
See comments inline. Overall, this looks like it fulfills the goal of the ticket. However, I'm confused about the requirements. The additional fields added when the checkbox is selected are already available on the user profile page. This seems confusing that we're conditionally rendering something in one place, but always rendering it in the parent location.

src/pages/user_profile/RequestForm.tsx
kevinmarete: Fixed here: https://github.com/DataBiosphere/duos-ui/pull/3418/commits/5f280730205ef1e05836ec068d1daca4856d2b45

kevinmarete:
> See comments inline. Overall, this looks like it fulfills the goal of the ticket. However, I'm confused about the requirements. The additional fields added when the checkbox is selected are already available on the user profile page. This seems confusing that we're conditionally rendering something in one place, but always rendering it in the parent location.

This was discussed and agreed to keep the feature as-is during the standup parking lot!

rushtong (approved):
Thanks for the updates 👍🏽

otchet-broad (approved):
Looks good!

PR #3448, opened 2026-04-22
[DT-3117] Create FileStorageObjects.ts API client for new endpoints

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3117

### Summary
This adds an API client for new Consent FSO endpoints at https://consent.dsde-dev.broadinstitute.org/#/Document.  Those were implemented in https://github.com/DataBiosphere/consent/pull/2867 and https://github.com/DataBiosphere/consent/pull/2878.

Here's an example of a few of the new endpoints demonstrating the integration.  (The screenshot uses a transient manual test harness; it will be callable as an internal library and not via `window`.)

Screenshot 2026-04-24 at 3 31 09 PM

----

- Verify all tests go green


kevinmarete (approved):

src/libs/ajax/FileStorageObject.ts
```diff
@@ -0,0 +1,99 @@
...
+  entity: EntityType,
+  entityId: string,
+  file: File,
+  category: FileCategory,
+): Promise<FileStorageObject> {
+  const url = await basePath(entity, entityId)
+  const formData = new FormData()
+  formData.append('file', file)
+  formData.append('category', category)
+  const res = await fetchMultipart<FileStorageObject>(url, formData, Config.multiPartOpts())
```
otchet-broad: Is this always a POST and never a PUT?
What about error handling if the backend returns an error?

eweitz: > Is this always a POST and never a PUT?

Yup!  This `uploadDocument` function calls https://consent.dsde-dev.broadinstitute.org/#/Document/uploadDocumentByEntity, which is only POST.

> What about error handling if the backend returns an error?

Insightful question.  The version you reviewed would throw a generic error `"Request to ${url} failed with status 502"`, regardless of any specific error message returned from the backend.  That's because the underlying `fetchMultipart` function defaulted `returnError` to `false`.  I've now updated `returnError` to `true`, so the original backend error message is preserved.  

That's a good incidental refinement.

kevinmarete:
```suggestion
  const res = await fetchMultipart<FileStorageObject>(url, formData, Config.multiPartOpts(), 'POST', true)
```

I think you should not change the behavior of fetchAdapter without changing how it has been called. The best solution was to specify the method, i.e., 'POST', and the requestError flag in the uploadDocument function call for fetchMultipart.

kevinmarete (changes requested):

src/libs/ajax/fetchAdapter.ts
```diff
@@ -283,5 +283,5 @@ export const fetchMultipart = <T>(
   formData: FormData,
   config: FetchMultipartConfig = {},
   method: Exclude<Method, 'GET' | 'DELETE'> = 'POST',
-  returnError: boolean = false,
+  returnError: boolean = true,
```
kevinmarete:
```suggestion
  returnError: boolean = false,
```

src/libs/ajax/fetchAdapter.ts
```diff
@@ -203,7 +203,7 @@ async function fetchMultipartRequest<T>(
     params,
     headers = {},
     credentials,
-    returnError = false,
+    returnError = true,
```
kevinmarete:
```suggestion
    returnError = false,
```

src/libs/ajax/FileStorageObject.ts
eweitz: Reasonable!  Thanks for discussing.  Per chat, looking through `fetchMultipart` callers I don't see a specific case where defaulting `returnError` to `true` actually harms any caller.

But I take the point that a more holistic solution is preferable, along the lines of [DT-3181](https://broadworkbench.atlassian.net/browse/DT-3181).

kevinmarete (approved):

otchet-broad (approved):

PR #3449, opened 2026-04-23
DT-2808: Balance left and right padding on study details page

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2808

## Summary
Minor alignment update so long study descriptions are balanced.

### Examples
Screenshot 2026-04-23 at 9 46 30 AM

Screenshot 2026-04-23 at 9 46 40 AM

Screenshot 2026-04-23 at 9 46 47 AM

----

- Verify all tests go green


kevinmarete (approved):

src/components/study_details/StudyDetails.tsx
```diff
@@ -154,7 +154,7 @@ export const StudyDetails = () => {
         <div style={{ fontSize: 20, fontWeight: 600, paddingTop: 20 }}>
           {study?.studyName}
         </div>
-        <Section>
+        <Section style={{ paddingRight: 100 }}>
```
fboulnois: This suggests that in the CSS component hierarchy something is setting `paddingLeft: 100`. Would it make sense to set `paddingLeft: 100, paddingRight: 100` on that same parent element, instead of specifically setting it here?

fboulnois (approved):
One suggestion:

src/components/study_details/StudyDetails.tsx
rushtong: The `Section` component already sets the `paddingLeft` property, but at a value of `20`. That value, combined with the surrounding left margin, is roughly equivalent to the `paddingRight` set here. `Section` is only used in this component, so let me see if I can make that sub component flexible enough so this particular instance of `Section` doesn't need special handling.

PR #3466, opened 2026-04-29
DT-3198: Favor google auth over jwt

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3198

## Summary
This PR looks for the Google OAuth2 bearer token as the preferred token to use for all API requests. It will only fall back to the current usage of the encoded JWT stored in `id_token` if we don't have a google one. 

This approach completely bypasses the Sam error:
```
"Cannot update azureB2cId for user 123 because user does not exist or the azureB2cId has already been set for this user"
```

## Testing
I have tested this locally with a user who hits the above Sam error through normal DUOS usage and through RAS account linking.

----

- Verify all tests go green

### DIFF
--- a/cypress/component/libs/config.spec.ts
+++ b/cypress/component/libs/config.spec.ts
@@ -196,7 +196,23 @@ describe('Config', () => {
   })
 
   describe('Token', () => {
-    it('should get token from storage when available', () => {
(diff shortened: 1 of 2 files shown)


otchet-broad (approved):
👍

kevinmarete (approved):

PR #3485, opened 2026-05-06
[DT-3188] Dynamic Asset-Specific Filters for DUOS Data Library

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3188

### Summary

This PR implements configuration-driven dynamic filters for Data Library study assets so each asset tab only displays relevant filters based on the metadata collected during Study Registration. Hidden filters must not affect Elasticsearch queries, unsupported filters should be cleared on tab changes, and the architecture should remain scalable for future asset types and filter additions.

After

----

- Verify all tests go green


fboulnois:
Sometimes the filters flicker:

https://github.com/user-attachments/assets/7e535bbc-1752-4e14-93e6-8ab217828c75

Also, the `Data Use` for Biospecimens is empty.

kevinmarete:
> Sometimes the filters flicker:
> 
>  Screen.Recording.2026-05-07.at.1.38.44.PM.mov 
> Also, the `Data Use` for Biospecimens is empty.

Fixed: Flickering 

https://github.com/user-attachments/assets/458a2715-26ca-43de-b967-b482a248c92e

`Data Use` is an optional open-text field for `Biospecimens`; it does not have any value for the specimens available, hence no filters are present. I have added a label for "No Filters Available".

No Filters

rushtong (approved):
Some of the filter choices are counter-intuitive to me. They do match the spreadsheet linked in the ticket, but, for example, Biospecimens has a participant count filter but the Biospecimens asset doesn't have that property. Publications has zero filters (which matches the spreadsheet), but I would expect at least a published date and author filter. For AI Models, in a similar vein, I would expect License to be a filter. Again, those choices meet the spec, but I think these should be reviewed in future tickets. 👍🏽

fboulnois (approved):
I agree with @rushtong 's comment. Should we review this in standup?

kevinmarete:
> Some of the filter choices are counter-intuitive to me. They do match the spreadsheet linked in the ticket, but, for example, Biospecimens has a participant count filter but the Biospecimens asset doesn't have that property. Publications has zero filters (which matches the spreadsheet), but I would expect at least a published date and author filter. For AI Models, in a similar vein, I would expect License to be a filter. Again, those choices meet the spec, but I think these should be reviewed in future tickets. 👍🏽 

The participant count filter in Biospecimen is filtering at the study level not at the Biospecimen level.

For the additional features, maybe @solideoglori can advise if we add them now or in future tickets.

kevinmarete:
> I agree with @rushtong 's comment. Should we review this in standup?

I have shared a response but happy to discuss at standup.

PR #3492, opened 2026-05-12
[DT-3259] Better sanitize user provided URLs

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3259

### Summary

Better sanitizes user provided URLs to avoid certain classes of vulnerabilities.

One major change: if the URL is not considered a safe URL, it will no longer be shown as a link.

----

- Verify all tests go green

### DIFF
--- a/src/pages/user_profile/ExternalProfile.tsx
+++ b/src/pages/user_profile/ExternalProfile.tsx
@@ -5,7 +5,7 @@ import IconButton from '@mui/material/IconButton'
 import { User } from 'src/libs/ajax/User'
 import { Notifications } from 'src/libs/utils'
 import { ExternalProfiles } from 'src/types/model'
-import { getSafeHttpUrl } from 'src/utils/UrlUtils'
+import { validateHttpUrl } from 'src/utils/UrlUtils'
 
 interface ExternalProfileProps {
   readonly userId?: number
@@ -191,13 +191,13 @@ export default function ExternalProfile(props: ExternalProfileProps) {
   }
 
   const getUrlLink = (url: string) => {
-    const safeUrl = getSafeHttpUrl(url)
-    if (!safeUrl) {
+    const validUrl = validateHttpUrl(url)
+    if (!validUrl) {
       return <span>{url}</span>
     }
 
     return (
-      <a href={safeUrl} target="_blank" rel="noopener noreferrer">
+      <a href={validUrl} target="_blank" rel="noopener noreferrer">
         {url}
       </a>
     )
--- a/src/pages/user_profile/ExternalProfile.tsx
+++ b/src/pages/user_profile/ExternalProfile.tsx
@@ -5,6 +5,7 @@ import IconButton from '@mui/material/IconButton'
 import { User } from 'src/libs/ajax/User'
 import { Notifications } from 'src/libs/utils'
 import { ExternalProfiles } from 'src/types/model'
+import { getSafeHttpUrl } from 'src/utils/UrlUtils'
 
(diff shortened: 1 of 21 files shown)


src/utils/RenderUtils.tsx
```diff
@@ -8,9 +9,11 @@ export function renderMaintainer(value: unknown): React.ReactNode {
 }
 
 export function renderUrl(value: unknown): React.ReactNode {
-  return typeof value === 'string' && value
-    ? <a href={value} target="_blank" rel="noreferrer">{value}</a>
-    : '—'
+  if (typeof value !== 'string' || !value) return '—'
+  const href = getSafeHttpUrl(value)
+  return href
+    ? <a href={href} target="_blank" rel="noreferrer">{value}</a>
```
otchet-broad, line 15: should this rel also include "noopener"?

src/pages/user_profile/ExternalProfile.tsx
```diff
@@ -190,8 +191,13 @@ export default function ExternalProfile(props: ExternalProfileProps) {
...
 
   const getUrlLink = (url: string) => {
+    const safeUrl = getSafeHttpUrl(url)
+    if (!safeUrl) {
+      return <span>{url}</span>
+    }
+
     return (
-      <a href={url} target="_blank" rel="noopener noreferrer">
+      <a href={safeUrl} target="_blank" rel="noopener noreferrer">
```
otchet-broad: I'm wondering if this refactor should update all of the links to use a component that consistently applies the `rel="noopener noreferrer"` because I see cases where we're not and likely should be.  Did you have any thoughts on that @fboulnois ?

src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,22 @@
+const SAFE_HTTP_PROTOCOLS = new Set(['http:', 'https:'])
+
+export const getSafeHttpUrl = (value: unknown): string | undefined => {
```
otchet-broad: What does `safe` mean in this context?

fboulnois: Allowed protocols (`http` and `https`) + validates as URL

src/pages/user_profile/ExternalProfile.tsx
fboulnois: If there are a lot of call sites to update I might push it into another PR.

src/utils/RenderUtils.tsx
fboulnois: See below.

rushtong (approved):
👍🏽

src/utils/UrlUtils.ts
otchet-broad: Would you consider refactoring to something like `getExternalHttpUrl` instead of `getSafeHttpUrl`?  To me, the the word safe gives me a sense of confidence.  It turns out that we shouldn't have that when using these urls because we don't actually know that it's safe.  The point of the setup that I think you're doing is to ensure that we treat it as if it isn't safe.  We won't provide referrer information and we set the noopener bit too.

Just a thought for consideration.

otchet-broad (approved):
Thanks for the conversation on the elements I've pointed out.  Feel free to address my remaining comments as you see fit.

src/pages/user_profile/ExternalProfile.tsx
fboulnois: Pushing to a follow up PR due to number of call sites.

otchet-broad (approved):
👍
Thanks for the updates!

PR #3495, opened 2026-05-12
[DT-3255] Upload offline

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3255](https://broadworkbench.atlassian.net/browse/DT-3255)

### Summary

At the core, this change does two things.  It:
1) Uses the legacy DAA API to upload new files for a DAC
2) Doesn't load a list of files for a case when the user has expressed to upload a new file by clicking on a button.

Why this fixes the issue:
When displaying the DAA file list, the call being made was based on the DAC ID.  This behavior was incorrect because the DAAs have DAA IDs that should be used for the entity association in the service, not the DAC.  The encapsulating page also displays the proper file list for the user.

Additional code was added to ensure the elements on the page are reloaded after a file is uploaded so that we can be sure the value was persisted properly.  During testing we identified cases where the file would appear to change to a different value after page reload.

----

- Verify all tests go green


rushtong (approved):

kevinmarete (approved):
👍🏾

PR #3499, opened 2026-05-13
DT-3250: Convert DUOS to run as a NodeJS server

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3250

## Summary
This PR updates the base application to run as a nodejs server. Content is still served from a static build, but this effort sets us up to add more complex behavior related to MCP server functionality in future work.

Requires a change to configuration that mounts `config.json` in a new location. See https://github.com/broadinstitute/terra-helmfile/pull/6437 for that.

### Running Locally
```
docker build . -t duos
docker compose up -d
```
Note that the platform warning is just that. It can be suppressed by adding `--platform linux/amd64`. Dev notes is also updated to reflect this.

Visit https://local.dsde-dev.broadinstitute.org/ to see the app running.

`npm start` works normally.

----

- Verify all tests go green

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,8 @@ RUN npm config set update-notifier false
 RUN npm install --loglevel verbose
 RUN npm run build
 
-FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
+# Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
+FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:7bb73493171d6c0b1bf00018915266cf8e80910b172d14bf249dcd01af8f3aa9
 ENV NODE_ENV=production
 WORKDIR /usr/src/app
 COPY --from=builder /usr/src/app/build ./build
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,10 +28,13 @@ RUN cd server && npm install && npm run build && npm prune --omit=dev
 
 # Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
 FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:7bb73493171d6c0b1bf00018915266cf8e80910b172d14bf249dcd01af8f3aa9
-ENV NODE_ENV=production
+ARG NODE_ENV=production
+ARG PORT=8080
+ENV NODE_ENV=${NODE_ENV}
+ENV PORT=${PORT}
 WORKDIR /usr/src/app
 COPY --chmod=550 --chown=node:node --from=builder /usr/src/app/build ./build
 COPY --chmod=550 --chown=node:node --from=builder /usr/src/app/server ./server
 USER node
-EXPOSE 8080
+EXPOSE ${PORT}
 CMD ["node", "server/dist/index.js"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,13 +22,16 @@ RUN npm config set update-notifier false
 RUN npm install --loglevel verbose
 RUN npm run build
 
+# build the server
+COPY server /usr/src/app/server
+RUN cd server && npm install && npm run build && npm prune --omit=dev
+
 # Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
 FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:7bb73493171d6c0b1bf00018915266cf8e80910b172d14bf249dcd01af8f3aa9
 ENV NODE_ENV=production
 WORKDIR /usr/src/app
 COPY --chown=node:node --from=builder /usr/src/app/build ./build
-COPY --chown=node:node server /usr/src/app/server
-RUN cd server && npm install --omit=dev
+COPY --chown=node:node --from=builder /usr/src/app/server ./server
 USER node
 EXPOSE 8080
-CMD ["node", "server/src/index.js"]
+CMD ["node", "server/dist/index.js"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,9 @@ RUN npm run build
 FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:7bb73493171d6c0b1bf00018915266cf8e80910b172d14bf249dcd01af8f3aa9
 ENV NODE_ENV=production
 WORKDIR /usr/src/app
-COPY --from=builder /usr/src/app/build ./build
-COPY server /usr/src/app/server
+COPY --chown=node:node --from=builder /usr/src/app/build ./build
+COPY --chown=node:node server /usr/src/app/server
 RUN cd server && npm install --omit=dev
+USER node
 EXPOSE 8080
 CMD ["node", "server/src/index.js"]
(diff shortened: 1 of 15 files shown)


Dockerfile
```diff
@@ -22,8 +22,11 @@ RUN npm config set update-notifier false
...
-RUN rm -rf /etc/nginx/conf.d
-COPY conf /etc/nginx
-CMD ["nginx", "-g", "daemon off;"]
+FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
+ENV NODE_ENV=production
+WORKDIR /usr/src/app
+COPY --from=builder /usr/src/app/build ./build
+COPY server /usr/src/app/server
+RUN cd server && npm install --omit=dev
+EXPOSE 8080
```
kevinmarete: I think you should allow the ENV and PORT to be configurable from ENVIRONMENT VARIABLES

kevinmarete (approved):
Works well, just had an issue setting up, but I used these commands:

`docker build . -t duos --platform linux/amd64`
`docker-compose up `

Left a comment on environment configurability. 
Also, is it an easy lift to convert the server files to Typescript?

Dockerfile
```diff
@@ -22,8 +22,11 @@ RUN npm config set update-notifier false
 RUN npm install --loglevel verbose
 RUN npm run build
 
-FROM us.gcr.io/broad-dsp-gcr-public/base/nginx:mainline-alpine
-COPY --from=builder /usr/src/app/build /usr/share/nginx/html
-RUN rm -rf /etc/nginx/conf.d
-COPY conf /etc/nginx
-CMD ["nginx", "-g", "daemon off;"]
+FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
```
otchet-broad: Should we pin this to a hash?

server/src/index.ts
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const express = require('express')
+const path = require('node:path')
+
+const app = express()
```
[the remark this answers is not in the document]
otchet-broad: Do we need to consider using something like helmet here?  [https://expressjs.com/en/advanced/best-practice-security.html#use-helmet](https://expressjs.com/en/advanced/best-practice-security.html#use-helmet)

Dockerfile
rushtong: Will look into those options, thank you.

Dockerfile
rushtong: Got that working, thanks

server/src/index.ts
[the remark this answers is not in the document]
rushtong: We run the complete app behind an apache proxy. You can see all of the current header settings here: https://github.com/broadinstitute/terra-helmfile/blob/master/charts/duos/templates/_site.conf.tpl
I'm open to updating them, but prefer not to co-mingle those configurations here.

Dockerfile
rushtong: I was able to add these as config options. They now have corresponding settings in https://github.com/broadinstitute/terra-helmfile/pull/6437.

server/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "type": "commonjs",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc",
+    "start": "node dist/index.js"
+  },
+  "dependencies": {
+    "express": "4.22.2"
```
fboulnois: If it is possible, I would suggest using one of the other alternate frameworks such as `fastify`, which offers better performance (= uses less resources in the same envelope)

server/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "duos-server",
+  "version": "1.0.0",
+  "private": true,
+  "type": "commonjs",
```
fboulnois: We likely do not want to build the server as `commonjs`, because it brings in deprecated functionality like `require`.

fboulnois (commented):
Some suggestions mainly around the package choices:

server/package.json
rushtong: Will make this change for the present. Future work will require a proxy and embedded MCP server which was the reason for the original choice.

server/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "start": "node dist/index.js"
+  },
+  "dependencies": {
+    "express": "4.22.2"
+  },
+  "devDependencies": {
+    "@types/express": "4.17.23",
+    "@types/node": "24.12.3",
+    "typescript": "5.9.3"
+  }
```
kevinmarete, line 18: Will these dependencies also be managed by our current dependabot configuration, or will you need to add a section for this?

server/package.json
rushtong: Thank you, this was great feedback.

eslint.config.ts
```diff
@@ -67,4 +67,11 @@ export default defineConfig([
       '@stylistic/jsx-one-expression-per-line': ['off'],
     },
   },
+  // Server is an explicit CommonJS package — allow require() there
+  {
+    files: ['server/**/*.js'],
+    rules: {
+      '@typescript-eslint/no-require-imports': 'off',
+    },
+  },
```
fboulnois: I think this can be removed, once using common ESM.

server/src/index.ts
```diff
@@ -0,0 +1,38 @@
+import Fastify, { FastifyError } from 'fastify'
+import fastifyStatic from '@fastify/static'
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const PORT = Number(process.env.PORT) || 8080
+const BUILD_DIR = path.join(fileURLToPath(new URL('.', import.meta.url)), '..', '..', 'build')
```
fboulnois: I think there's a Node project dir that can be used to derive the `build` dir (instead of using relative paths with `.` and `..`)

server/src/config.ts
```diff
@@ -0,0 +1,45 @@
+import fs from 'node:fs'
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+// The runtime config.json is mounted into build/ at deploy time.
+// Respect CONFIG_PATH so tests or local dev can point elsewhere.
+const CONFIG_PATH
+  = process.env.CONFIG_PATH
+    || path.join(fileURLToPath(new URL('.', import.meta.url)), '..', '..', 'build', 'config.json')
```
fboulnois: ditto here

fboulnois (approved):
A few additional comments, otherwise lgtm

otchet-broad (approved):
👍

server/package.json
kevinmarete: I think if you add this to [here](https://github.com/DataBiosphere/duos-ui/blob/38c81733e514c534ba59c869611ca0d33fd72858/.github/dependabot.yml#L24), we should be good about the updates.


```
   directories:
      - "/"
      - "/server/"
```

https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files

kevinmarete (approved):
Small suggestion. All good.

server/package.json
rushtong: Great point ... this did cross my mind earlier and I quickly forgot about it.

PR #3515, opened 2026-05-15
[DT-3261] Convert to pnpm.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3261](https://broadworkbench.atlassian.net/browse/DT-3261)

### Summary

----

- Verify all tests go green


.github/workflows/integration-tests.yml
```diff
@@ -13,15 +13,21 @@ jobs:
...
+          version: 11.1.2
       - uses: actions/setup-node@v6
         with:
           node-version: ${{ env.NODE_VERSION }}
-      - name: NPM Install
-        run: npm ci
+          cache: pnpm
+          cache-dependency-path: pnpm-lock.yaml
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
fboulnois: I think there's an equivalent `pnpm ci` that could be used.

.github/workflows/audit.yml
```diff
@@ -1,18 +1,24 @@
...
     steps:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
       - run: echo "NODE_VERSION=$(sed -n 's/FROM node:\([0-9.]*\).*/\1/p' Dockerfile)" >> $GITHUB_ENV
+      - name: Install pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093
+        with:
+          version: 11.1.2
       - uses: actions/setup-node@v6
```
fboulnois, line 16: I think this action also installs `pnpm`, should we use the version from this action instead? Otherwise, I think the `actions/setup-node` needs to come before `pnpm/action-setup`.

fboulnois (commented):
A few suggestions:

.github/workflows/integration-tests.yml
otchet-broad: I'm letting copilot spin on this a bit.  I'm curious to see where it goes.

.github/workflows/audit.yml
otchet-broad: I would have thought that as well, but it seems like pnpm needs to be installed first.

.github/workflows/integration-tests.yml
otchet-broad: Looks like we've got a good build this way now!

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+allowBuilds:
+  cypress: true
+  esbuild: true
```
fboulnois, line 3: Is this needed? Cypress and Esbuild work fine without them, and this is another vector for compromise.

fboulnois (approved):
One more suggestion:

pnpm-workspace.yaml
otchet-broad: It is needed.  Otherwise there's interactive prompting that takes place for running these each time.  As a result, builds fail in CI without this.

rushtong (approved):
Finally got this to work locally after much dev container reloading 👍🏽

kevinmarete (approved):
👍🏾

Dockerfile
```diff
@@ -13,21 +13,25 @@ ENV PATH=/usr/src/app/node_modules/.bin:$PATH
...
 # build the server
 COPY server /usr/src/app/server
-RUN cd server && npm install && npm run build && npm prune --omit=dev
+RUN pnpm --dir /usr/src/app/server --ignore-workspace ci \
+	&& pnpm --dir /usr/src/app/server --ignore-workspace run build \
+	&& CI=true pnpm --dir /usr/src/app/server --ignore-workspace prune --prod --loglevel warn
 
 # Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-alpine
-FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:7bb73493171d6c0b1bf00018915266cf8e80910b172d14bf249dcd01af8f3aa9
+FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-debian@sha256:9383ff586d7d17d76ec2c31fc6c0559d625beff285c024f6d88e93ed5b746fa1
```
rushtong: Copilot feedback seems reasonable. If you do that, the comment can change:
`# Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-debian`

rushtong (approved):
Looks good, one optional comment inline 👍🏽

PR #3529, opened 2026-05-19
[DT-400] `nodash` improvements

fboulnois:
### Addresses

### Summary

Removes `lodash` from codebase, reducing the bundle size from 4.96MB to 4.72MB (-5%)

----

- Verify all tests go green


src/utils/NodashUtil.ts
```diff
@@ -1,30 +1,122 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+export type Dictionary<T> = Record<string, T>
```
otchet-broad, line 3: It looks like we only have one use of the Dictionary and I'm not opposed to this landing as a straight 1:1 conversion.

I'd personally rather have us fully type the record than to use Pythonic conventions here.  I imagine the tooling will like us to do that as well.  We can address this later.

src/utils/NodashUtil.ts
```diff
@@ -36,20 +128,61 @@ export { numberIsNaN as isNaN }
...
-export const values = Object.values
+export const values = (obj: any) => obj == null ? [] : Object.values(obj)
+
+export const clone = <T>(obj: T): T => {
+  if (Array.isArray(obj)) return [...obj] as T
+  if (obj != null && typeof obj === 'object') return { ...obj }
+  return obj
+}
+
+const cloneDeepValue = <T>(obj: T, seen: WeakMap<object, any>): T => {
```
otchet-broad, line 148: Could we just call structuredClone?

src/utils/DarCollectionUtils.ts
```diff
@@ -1,4 +1,4 @@
-import { cloneDeep } from 'lodash'
+import { cloneDeep } from 'src/utils/NodashUtil'
```
rushtong: Suggestion: merge this into line 6

rushtong (approved):
Really nice, thank you 👍🏽

otchet-broad (approved):
Nice work.  Left some comments/questions for consideration but they don't need to block this.

src/utils/NodashUtil.ts
fboulnois: Long term I am hoping we can totally deprecated `nodash` as well.

src/utils/NodashUtil.ts
fboulnois: This is how the code was originally defined, but `structuredClone` doesn't work on functions apparently.

PR #3534, opened 2026-05-20
[DT-3296] Updated anvil data library query

ncalvanese1:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3296

### Summary

Some datasets from the anvil14 release are not being pulled into the AnVIL Data Library in DUOS because the tags we are using are different than those the Data Library configuration is looking for. This addresses that. 

Changes the AnVIL Data Library to explicitly look for the "Platform: AnVIL" study tag rather than the generic "anvil" term in the study description OR tags.

----

- Verify all tests go green


kevinmarete (approved):

fboulnois (approved):
Nice work!

PR #3537, opened 2026-05-21
[DT-315] Vitest testing configuration

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-315

### Summary

Adds the basic scaffolding for vitest testing. Includes one of each type of test:

* Basic unit test
* Basic `render` test
* More complicated `renderWithRouter` test

----

- Verify all tests go green


vite.config.ts
```diff
@@ -16,17 +15,6 @@ const defaultOptions = {
...
-function aliases_from_tsconfig() {
-  const paths: Record<string, string[]> = tsconfig.compilerOptions.paths
-  const aliases: Record<string, string> = {}
-  for (const [key, value] of Object.entries(paths)) {
-    const aliasKey = key.replace(/\/\*$/, '')
-    const aliasPath = resolve(__dirname, value[0].replace(/\/\*$/, ''))
-    aliases[aliasKey] = aliasPath
-  }
-  return aliases
-}
```
fboulnois, line 28: This is pulled into a separate module that can be used by the vitest config.

rushtong (approved):
This look very nice, thank you. 👍🏽 
Two thoughts.
- What's your opinion on putting tests next to their component directly in `/src`? That seems to be an industry standard but I'm open to different ideas. I think that does two things, directories get very large and it's also immediately clear if a component is missing a test.
- Should we look at a github action for these? I have an example in [a draft PR](https://github.com/DataBiosphere/duos-ui/pull/3536). It's not 100% accurate but offers a template.

fboulnois:
> What's your opinion on putting tests next to their component directly in `/src`? That seems to be an industry standard but I'm open to different ideas. I think that does two things, directories get very large and it's also immediately clear if a component is missing a test.

As you mention, it is clear if a component is missing a test, but perhaps this can be achieved by reporting test coverage instead (something we can now do with `vitest`).

> Should we look at a github action for these? I have an example in [a draft PR](https://github.com/DataBiosphere/duos-ui/pull/3536). It's not 100% accurate but offers a template.

Yes definitely. I didn't want to block this PR on setting up the GitHub Action. I think instead of a separate GitHub Action, we could have a separate job that runs in parallel to the other tests.

kevinmarete (approved):
Very nice 🔥

PR #3538, opened 2026-05-22
[DT-315] Add simple Vitest action

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-315

### Summary

Adds a simple Vitest action that can run in parallel to Cypress tests.

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/component-tests.yml
+++ b/.github/workflows/component-tests.yml
@@ -4,6 +4,27 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 jobs:
+  vitest-run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - run: echo "NODE_VERSION=$(sed -n 's/FROM node:\([0-9.]*\).*/\1/p' Dockerfile)" >> $GITHUB_ENV
+      - name: Install pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093
+        with:
+          version: 11.1.2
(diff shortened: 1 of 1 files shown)


.github/workflows/component-tests.yml
```diff
@@ -4,6 +4,27 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 jobs:
+  vitest-run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
rushtong, line 11: Should we pin this action too?

.github/workflows/component-tests.yml
```diff
@@ -4,6 +4,27 @@ concurrency:
...
+        with:
+          persist-credentials: false
+      - run: echo "NODE_VERSION=$(sed -n 's/FROM node:\([0-9.]*\).*/\1/p' Dockerfile)" >> $GITHUB_ENV
+      - name: Install pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093
+        with:
+          version: 11.1.2
+      - uses: actions/setup-node@v6
```
rushtong, line 19: Same question here.

rushtong (approved):
Looks good. Address pin comments as you see fit.

.github/workflows/component-tests.yml
fboulnois: I think we previously discussed using version numbers for official GitHub Actions, and pinning non-official GitHub Actions to specific hashes.

.github/workflows/component-tests.yml
fboulnois: Ditto.

kevinmarete (approved):

PR #3554, opened 2026-06-08
DT-1842, DT-2557: Revise DAR warning messages for old profile settings

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1842
https://broadworkbench.atlassian.net/browse/DT-2557

## Summary
In previous iterations of the DAR form, there were values pulled from the user's profile institution (things like institution id, preferred signing official). That is no longer true, so both the logical checks and the messaging needed updating. Additionally, the page would flash a warning for a brief moment about a missing library card until the component was able to determine that one existed, then it would disappear. This PR fixes that incorrect flash.

This PR also moves the test to Vite and converts the component to Typescript

### Screens
New Library Card Message
Screenshot 2026-06-08 at 9 44 37 AM

Old
Screenshot 2026-06-08 at 9 36 27 AM

Removed
Screenshot 2026-06-08 at 12 50 49 PM

----

- Verify all tests go green


rushtong:
The cypress failure here consistently passes locally but fails in CI. I'm going to rewrite this in Vite and see if I can boost stability.

test/pages/DataAccessRequest/ResearcherInfo.spec.jsx
```diff
@@ -0,0 +1,335 @@
+import React, { useEffect, useState } from 'react'
+import { describe, it, expect, beforeEach, vi } from 'vitest'
+import { screen, waitFor, within } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import ResearcherInfo from 'src/pages/dar_application/ResearcherInfo'
+import { User } from 'src/libs/ajax/User.js'
```
kevinmarete: Nit: Can we avoid including the .js extension to make the transition to TypeScript easier later?

test/pages/DataAccessRequest/ResearcherInfo.spec.jsx
```diff
@@ -0,0 +1,335 @@
...
+import { renderWithRouter } from '../../test-utils'
+
+vi.mock('src/libs/ajax/User.js', () => ({
```
kevinmarete: Here too!

kevinmarete (approved):
👍🏾  All good, just a nit comment!

test/pages/DataAccessRequest/ResearcherInfo.spec.jsx
rushtong: Likely, yes, good catch, I missed that in the test transition.

src/pages/dar_application/ResearcherInfo.tsx
```diff
@@ -62,39 +78,40 @@ export default function ResearcherInfo(props) {
...
   const accountLink = nihAccountInstructions()
 
-  const formatSOString = (name, email) => {
+  const formatSOString = (name?: string, email?: string) => {
     if (isEmpty(name)) {
       return ''
     }
     const nameString = `${name}`
-    const emailString = !isNil(email) ? ` (${email})` : ''
+    const emailString = isNil(email) ? '' : ` (${email})`
```
otchet-broad, line 86: I realize this comment is beyond the scope of what you're going after here, but should we really accommodate a nil email address for a signing official?

otchet-broad (approved):
Looks good.  My question may be in or out of scope for this PR.

src/pages/dar_application/ResearcherInfo.tsx
rushtong: I think we might need to accommodate older DARs. Signing officials are fairly recent in the history of DUOS.

kevinmarete (approved):

PR #3560, opened 2026-06-10
[DT-2330] Update Sign In button outline styling

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2330

### Summary

This PR achieves the following:
- Changed the Sign In button to use white text, a white outline, and transparent fill.
- Added hover/focus styling so the clickable area is clearer.
- Added component test coverage for the updated button colors.

**Before**
Before

**After**
After

----

- Verify all tests go green


src/components/SignInButton.tsx
```diff
@@ -14,6 +14,7 @@ import { OidcUser } from 'src/libs/auth/oidcBroker'
 import { DuosUser } from 'src/types/model'
 import { ServiceStatus } from 'src/libs/ajax/ServiceStatus'
 import 'src/styles/tooltip.css'
+import './SignInButton.css'
```
fboulnois: I tend to like the inline `style={{ ... }}` a little better than these global styles, because it makes it harder to see how the button is styled out of the box.

fboulnois (commented):
I wonder what the team's opinions on inline vs external styles are.

rushtong (approved):
👍🏽

otchet-broad (approved):
👍

PR #3562, opened 2026-06-11
[DT-2776] Use date pickers for study registration dates

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2776

### Summary

This PR does the following:
- Replaced free-text date inputs with DUOS date picker controls in V2 study registration and study asset forms.
- Updated date picker styling to match existing form fields, including disabled/read-only states.
- Added/updated component tests for date picker selection and existing value display.

**Before**
Before

**After**
After

----

- Verify all tests go green

### DIFF
--- a/src/components/forms/forms.jsx
+++ b/src/components/forms/forms.jsx
@@ -32,7 +32,6 @@ import {
   requiredValidator,
   urlValidator,
 } from './formValidation'
-import dayjs from 'dayjs'
 
 // ----------------------------------------------------------------------------------------------------- //
 // ======                                  MAIN FORM FIELD TYPES                                  ====== //
@@ -184,7 +183,7 @@ export const FormFieldTypes = {
     ],
   },
   CALENDAR: {
-    defaultValue: dayjs(),
+    defaultValue: null,
     component: FormDatePicker,
     requiredProps: [],
     optionalProps: ['readOnly'],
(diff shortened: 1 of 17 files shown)


src/components/forms/forms.jsx
```diff
@@ -184,7 +183,7 @@ export const FormFieldTypes = {
     ],
   },
   CALENDAR: {
-    defaultValue: dayjs(),
+    defaultValue: null,
```
fboulnois, line 186: Can you explain this change?

fboulnois (approved):
👍

src/components/forms/forms.jsx
kevinmarete: I changed this from `dayjs()` to `null` because the shared calendar field added a default date when the caller has not provided one. With `dayjs()`, optional/edit-mode date fields rendered as today even when the underlying study asset had no saved date, which made empty data look populated.

Callers that intentionally want today as the default can still pass `defaultValue={dayjs()}` explicitly. This keeps the shared default empty and makes date defaults opt-in at the form level.

rushtong (approved):
The tests here feel to me like we're over-testing. IMO, components that use this form field should really only test that it exists. Any interactions with the component should instead be fully tested separately. In this case, the form field of type CALENDAR is where the interactions should be tested wrt the new changes. Even still more coverage is better than less. 👍🏽

PR #3573, opened 2026-06-12
[DT-2773] update data submissions table style to match data library

jlaw-codes:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2773](https://broadworkbench.atlassian.net/browse/DT-2773)

### Summary

----

- Verify all tests go green


kevinmarete (changes requested):
I was able to test this PR and confirm it works after running `eslint --fix` to resolve the linting issues. However, it looks like one file may be missing from the PR:

```ts
import { makeSubmissionColumns } from 'src/components/data_library/columns/submissionColumns'
```

Because that file is not included, the action buttons are not visible. Could you commit that file as well?

Image

src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -123,6 +115,21 @@ export const makeDatasetColumns = (
...
+    renderCell: (params) => {
+      const exportableSnapshots = exportableDatasets[params.row.datasetIdentifier] || []
+      if (exportableSnapshots.length === 0) return null
+      return (
+        <Box sx={{ display: 'flex', gap: 0.5, alignItems: 'center' }}>
+          <DatasetExportButton snapshots={exportableSnapshots} />
+        </Box>
+      )
+    },
+  },
```
fboulnois: This will also change the order of columns in the data library. Is this expected? We typically expect the actions column to be the last one.

src/hooks/useLibraryUrlState.ts
```diff
@@ -294,8 +294,8 @@ const applySortOrderUpdate = (updates: Partial<LibraryUrlState>, searchParams: U
 
 const applyHideFiltersUpdate = (updates: Partial<LibraryUrlState>, searchParams: URLSearchParams) => {
   if (hasOwnUpdate(updates, 'hideFilters')) {
-    if (updates.hideFilters) {
-      searchParams.set('hideFilters', 'true')
+    if (updates.hideFilters === false) {
+      searchParams.set('hideFilters', 'false')
```
fboulnois: I think this will hide filters by default in the data library. Is this expected?

fboulnois (commented):
A few comments:

src/hooks/useLibraryUrlState.ts
jlaw-codes: addressed, thanks.

src/components/data_library/columns/datasetColumns.tsx
jlaw-codes: addressed, thanks.

src/pages/researcher_console/DatasetSubmissions.jsx
fboulnois, line 1: This file looks like it replicates much of the data library and hardcodes a number of values, for example the submissions tabs, the available filters, the data use, and the handlers.

src/hooks/useLibraryUrlState.ts
```diff
@@ -294,7 +294,7 @@ const applySortOrderUpdate = (updates: Partial<LibraryUrlState>, searchParams: U
 
 const applyHideFiltersUpdate = (updates: Partial<LibraryUrlState>, searchParams: URLSearchParams) => {
   if (hasOwnUpdate(updates, 'hideFilters')) {
-    if (updates.hideFilters) {
+    if (updates.hideFilters === true) {
```
fboulnois:
```suggestion
    if (updates.hideFilters) {
```

I think we want to leave this as is.

fboulnois (changes requested):
See below

kevinmarete (approved):
The new design looks excellent. One small suggestion: could we make the datatable’s min-height dynamic so it adjusts to the content? Right now, the horizontal scrollbar can sometimes be hidden below the visible area, so users may need to scroll down before realizing it’s available.

Image

otchet-broad:
@kevinmarete - the downside of what you suggest is that on the last page of the table, if the results have fewer records, you'll have to move the mouse further away to navigate between pages.  I've often received user complaints over this, so I'm sensitive to the request.  That being said, I'm happy to implement it across our tables if the team decides.  I'd recommend doing that outside of this PR.

src/pages/researcher_console/DatasetSubmissions.jsx
```diff
@@ -1,118 +1,356 @@
...
-  ), [terms])
+  const handleDeleteConfirm = useCallback(async () => {
+    const { term } = deleteDialog
+    if (!term) return
+    setDeleteDialog({ open: false, term: null })
+    try {
+      await DataSet.deleteDataset(term.datasetId)
+      Notifications.showSuccess({ text: `Removed dataset '${term.datasetName}' successfully.` })
+      queryClient.invalidateQueries({ queryKey: ['library-data', 'submissions'] })
+    }
```
rushtong, line 215: I think the logic is here. I deleted a dataset but it remained in the table until a full page refresh.

rushtong (commented):
It looks like `DatasetSubmissionsTable.jsx` and `DatasetTerms.module.css` are now unused

rushtong (approved):
Looks very nice. 👍🏽 pending the delete fix

fboulnois (approved):

PR #3595, opened 2026-06-16
DT-3606: Register new plugins and incorporate into test reporting

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-3606

## Summary
Requires: https://github.com/broadinstitute/terra-helmfile/pull/6458

This PR introduces no functional changes and adds plugin configuration and integration into the current CI.

- Registers `@fastify/postgres`, `@fastify/cookie`, and `@fastify/session` in `server/src/index.ts` in the order required by the BFF session infrastructure design: `postgres` first (so `app.pg` is available to the session store when it's wired in the next set of tasks), then `cookie`, then `session`. 
  - Cookie is configured `HttpOnly`, `Secure` (prod only), `SameSite=Strict`, 8-hour rolling expiry, `saveUninitialized`: false.
  - Session store is left as the default in-memory store pending future session work.
- Extracts plugin registration into an exported `buildApp()` factory and guards the top-level `listen()` call with an ESM entry-point check, making the module importable by tests without starting the server.
- Adds a Vitest suite under `server/test/` covering the `/health` route, error handler (generic response, no internal detail leakage), and plugin registration order. 
  - Log level is set to silent in `vitest.config.ts` so tests run quietly locally; production defaults to info.
- Adds server tests and combined coverage to the `vitest-run` CI job: installs server deps via `pnpm -C server ci`, runs both suites with `--coverage`, merges the two Istanbul JSON outputs via `scripts/merge-coverage.mjs`, and passes the merged files to the existing `davelosert/vitest-coverage-report-action` step so a single report is posted to the PR.

----

- Verify all tests go green


kevinmarete (approved):
Excellent convergence 👍🏾

server/src/index.ts
```diff
@@ -1,13 +1,41 @@
-import Fastify, { FastifyError } from 'fastify'
+import Fastify, { FastifyError, FastifyInstance } from 'fastify'
 import fastifyStatic from '@fastify/static'
+import fastifyPostgres from '@fastify/postgres'
```
otchet-broad, line 3: Do we need the new container from appsec to land this work?  This makes an outbound connection to the database.

server/src/index.ts
```diff
@@ -25,13 +53,18 @@ async function main(): Promise<void> {
...
-}
-catch (err: unknown) {
-  console.error(err)
-  process.exit(1)
+// Only start the server when this file is the direct entry point, not when imported by tests
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
+  const PORT = Number(process.env.PORT) || 8080
+  try {
+    const fastify = await buildApp()
+    await fastify.listen({ port: PORT, host: '0.0.0.0' })
```
otchet-broad, line 64: Should we configure cookie signing?

server/src/index.ts
```diff
@@ -1,13 +1,41 @@
-import Fastify, { FastifyError } from 'fastify'
+import Fastify, { FastifyError, FastifyInstance } from 'fastify'
 import fastifyStatic from '@fastify/static'
+import fastifyPostgres from '@fastify/postgres'
+import fastifyCookie from '@fastify/cookie'
+import fastifySession from '@fastify/session'
```
otchet-broad, line 5: Have you considered @fastify/secure-session?

server/src/index.ts
rushtong: Not yet, no. This doesn't make any connections yet, just adds the plugins. The next batch of work will actually make the connection and that will need the FIPS image.

server/src/index.ts
rushtong: > Should we configure cookie signing?

The session cookie is already signed, just not by @fastify/cookie. @fastify/session creates its own signer from DUOS_SESSION_SECRET and signs/unsigns the session ID on every request.

server/src/index.ts
rushtong: > Have you considered @fastify/secure-session?

Not until now, no, but after looking into it, it doesn't meet our needs. @fastify/secure-session is stateless: session data is encrypted and stored entirely in the cookie, with no server-side store. We're storing our sessions server side so that library doesn't really buy us anything.

server/src/index.ts
```diff
@@ -1,13 +1,41 @@
...
+  const fastify = Fastify({ logger: { level: process.env.FASTIFY_LOG_LEVEL ?? 'info' } })
+
+  // 1. DB pool — must be registered before session so app.pg is available to the store
+  await fastify.register(fastifyPostgres, {
+    host: process.env.DUOS_DB_HOST,
+    database: process.env.DUOS_DB_NAME,
+    port: Number.parseInt(process.env.DUOS_DB_PORT ?? '5432', 10),
+    user: process.env.DUOS_DB_USER,
+    password: process.env.DUOS_DB_PASSWORD,
+    ssl: { rejectUnauthorized: true },
```
otchet-broad, line 21: Isn't this making the connection?  I see the SSL option here.

rushtong: No, it does not make an actual connection at registration time. fastify.register(fastifyPostgres, {...}) initializes a connection pool (via pg.Pool) using the provided config, but the pool doesn't open any real TCP connections to the database until the first query is made.

rushtong: This question was helpful, I missed some necessary infra code in the current repo's docker compose file.

rushtong:
@kevinmarete - I did make a couple more changes to the docker compose and devnotes files, PTAL, thank you.

server/src/index.ts
```diff
@@ -1,13 +1,41 @@
...
+
+  // 2. Cookie + session — store: createPgSessionStore(fastify.pg) wired in next story
+  await fastify.register(fastifyCookie)
+  await fastify.register(fastifySession, {
+    secret: process.env.DUOS_SESSION_SECRET!,
+    cookie: {
+      httpOnly: true,
+      secure: process.env.NODE_ENV === 'production',
+      sameSite: 'strict',
+      maxAge: 8 * 60 * 60 * 1000,
```
kevinmarete: Maybe we should allow maxAge to be configurable from env and have this as default?

kevinmarete (approved):
Small comment, but looks good

otchet-broad (approved):

PR #3617, opened 2026-06-22
[DT-2708] Convert `spinner-service.js` to TypeScript with vitest tests

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2708

### Summary
Convert `spinner-service.js` to TypeScript with tests

----

- Verify all tests go green

### DIFF
--- a/src/libs/spinner-service.ts
+++ b/src/libs/spinner-service.ts
@@ -25,11 +25,11 @@ export class SpinnerService {
   }
 
   showAll(): void {
-    this.spinnerCache.forEach((spinner) => (spinner.show = true))
+    this.spinnerCache.forEach(spinner => (spinner.show = true))
   }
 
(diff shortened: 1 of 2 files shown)


src/libs/spinner-service.ts
```diff
@@ -1,29 +1,26 @@
-// TODO: Delete this class
```
otchet-broad, line 1: Was the TODO stale?

rushtong: Yes, this code is in active use by `SpinnerComponent.jsx`. It's likely that we can do some additional cleanup/refactoring. I'll add a note to the ticket for that one so this doesn't get lost.

kevinmarete (approved):

otchet-broad (approved):
👍

PR #3621, opened 2026-06-23
DT-3001: Add a retry with backoff pattern to fetchAdapter and Metrics calls

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3001

## Summary
Add retry logic to fetch adapter and migrate Metrics calls

Adds `retryFetchPost` to `fetchAdapter`, which wraps `fetchPost` with up to 3 retries (delays: 500 ms → 1 s → 2 s, plus up to 200 ms jitter). Retries are attempted on 5xx responses and network-level failures; 4xx errors are not retried. An `AbortSignal` can cancel a pending retry sleep immediately.

`Metrics.captureEvent`, `syncProfile`, and `identify` all switch from `fetchPost` to `retryFetchPost`, making Bard metric calls more resilient to transient server errors.

The `fetchAdapter` Cypress component test is replaced with a Vitest unit test covering the full surface of the module, including the new retry behavior.

----

- Verify all tests go green


otchet-broad (approved):
I think this is a reasonable first implementation.  If we need to come back to this, it might make sense to put these requests into a worker's queue.

kevinmarete (approved):

PR #3367, opened 2026-03-04
[DT-3001] Simplify Research Use Statement review UI

eweitz:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3044

### Summary
This simplifies the Research Use Statement section in the vote tab of DAR review page.
* Removed vote section on “Was the Research Use Statement (Narrative) accurately converted to a structured format?”
* Show narrative -- i.e. the prose Research Use Statement -- by default.
* Move modifiers to the right
* Trim some vertical padding for easier glanceability

Jonathan reviewed via Slack and gave a Product approval.

### Note on code reviewability
Apologies for the large diff.  For maintainability, I renamed:
* research_proposal_vote_slab.spec.jsx -> research_proposal_slab.spec.jsx
* ResearchProposalVoteSlab.jsx -> ResearchProposalSlab.jsx

But GitHub shows it as a deletion and creation.  I tried a few techniques -- e.g. `git mv` with a rename commit, and a separate content change commit -- to no avail.

#### Example 1
https://duos-k8s.dsde-dev.broadinstitute.org/admin_review_collection/789

Before

https://github.com/user-attachments/assets/f28a1c2d-692d-4328-a0a6-b75978bf04bc

After

https://github.com/user-attachments/assets/7a4b0a2b-cfc0-4a2e-a4bf-88ace92710e6

#### Example 2
https://duos-k8s.dsde-dev.broadinstitute.org/admin_review_collection/791

Before

https://github.com/user-attachments/assets/8f990af2-0915-464d-b72f-cf71983c94aa

After

https://github.com/user-attachments/assets/cb17d2e8-428b-4742-b686-294c9ca7d8a7

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #3632, opened 2026-06-25
[DT-2576] Convert `ResearchProposalVoteSlab.jsx` to TypeScript and `research_proposal_vote_slab.spec.jsx` to Vitest

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2576

### Summary

Convert `ResearchProposalVoteSlab.jsx` to TypeScript and `research_proposal_vote_slab.spec.jsx` to vitest tests

----

- Verify all tests go green


src/pages/dar_collection_review/MultiDatasetVotingTab.tsx
```diff
@@ -105,7 +105,6 @@ export default function MultiDatasetVotingTab({
   const missingLibraryCardMessage = 'The Researcher must have a Library Card before data access can be granted.\n'
     + (adminPage ? '' : 'You can still deny this request and/or vote on the Structured Research Purpose.')
 
-  const rpBucket = find(buckets, bucket => get(bucket, 'isRP')) ?? {} as Bucket
```
rushtong, line 108: This could use a follow-on task to review `BucketUtils` construction of the `isRP` bucket. I believe this would be the only place that would be used.

rushtong (approved):
👍🏽

src/pages/dar_collection_review/MultiDatasetVotingTab.tsx
kevinmarete: I am not sure I understand your comment here: `isRP` is used elsewhere. In this case, I removed it since it was being added as a prop but not being used.

rushtong: > I am not sure I understand your comment here: `isRP` is used elsewhere. In this case, I removed it since it was being added as a prop but not being used.

Yes, I took a look and there are some other components using it. However, we are no longer using RP votes - they are essentially dead ever since we removed them from the voting panels. I think the follow-on task is to finish the RP removal that was started in https://github.com/DataBiosphere/duos-ui/pull/3367

fboulnois (approved):
👍

PR #3651, opened 2026-06-27
[DT-2657] Convert `formComponents.jsx` to TypeScript

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2657

### Summary

Convert `formComponents.jsx` to TypeScript with vitest tests.

----

- Verify all tests go green


rushtong (approved):
👍🏽

src/components/forms/formComponents.tsx
```diff
@@ -0,0 +1,666 @@
...
+}
+
+export interface RadioOption {
+  id?: string
+  name: string
+  text: string
+  [key: string]: unknown
+}
+
+export interface FormComponentConfig {
```
otchet-broad: This should probably be decomposed into a smaller interface tailored to each form component because one of the most confusing parts of this is that not every component will take every config option.  That might be in or out of scope for this ticket.  If out, I'd make a new ticket so we can tackle it later.

otchet-broad (approved):
👍

src/components/forms/formComponents.tsx
kevinmarete: Done!

PR #3655, opened 2026-06-29
[DT-3711] Integrate @fastify/vite for single-origin dev server

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3711

Branch deployed at https://duos.grushton-dt-3711.bee.envs-terra.bio/

## Summary

- Integrates `@fastify/vite` so Fastify server code (sessions, cookies, future auth callbacks and API proxy routes) can be exercised during local development
- Supports **two parallel dev modes**:
  - `pnpm start` → standalone `vite` dev server for fast native HMR (frontend-only work)
  - `pnpm run start:server` → `tsx watch` Fastify with `@fastify/vite` middleware embedded, for exercising server code (slower reload via tsx)
- `pnpm run serve` (new) serves the static build via `vite preview` on port 3000 for CI integration tests
- Declares `server/` as a proper pnpm workspace member; removes the now-stale `server/pnpm-lock.yaml`
- Conditional HTTPS in dev (matches old Vite behaviour): uses `server.key`/`server.crt` when present, plain HTTP in CI/Cypress — applied in both `vite.config.ts` (for `pnpm start`/`serve`) and `server/src/index.ts` (for `pnpm run start:server`)
- Fixes Docker production build: a post-build script generates `build/vite.config.json` (with `root`, `build`, and `fastify.outDirs` fields) and the prod image receives root `package.json` — both required by `@fastify/vite`'s production config resolver

## Changed files

| File | What changed |
|---|---|
| `server/src/index.ts` | Replace `@fastify/static` + SPA fallback with `@fastify/vite`; add HTTPS, browser-open, dev/prod port split |
| `vite.config.ts` | Restore shared `server`/`preview` options block (host, port, CI-conditional HTTPS from cert files, open, allowedHosts). Standalone/preview mode applies all of it; in `@fastify/vite` middleware mode host/port/https/open are ignored (Fastify owns the socket + TLS) while allowedHosts is still enforced |
| `package.json` | `start` → standalone `vite` (fast HMR); add `start:server` (tsx watch Fastify); add `serve` (vite preview); add `tsx`, `open` deps; update `build` to generate `vite.config.json` |
| `pnpm-workspace.yaml` | Declare `server` as workspace package |
| `server/pnpm-lock.yaml` | Deleted — now managed by root workspace lockfile |
| `scripts/write-vite-config-json.mjs` | New: post-build script that writes `build/vite.config.json` with `root` (captured from `process.cwd()` at build time), `build.outDir`, and `fastify.outDirs.client` — fields `@fastify/vite` needs to locate static assets in production without the `viteFastify` Vite plugin (which would move output to `build/client/`) |
| `Dockerfile` | Rewrite: correct layer ordering, split vite/server builds, run `write-vite-config-json.mjs` after `vite build`, `pnpm deploy --prod --legacy`, copy `package.json` into prod image |
| `server/test/index.test.ts` | Replace `@fastify/static` mock with `@fastify/vite` mock using `Symbol.for('skip-override')` so decorations reach the parent instance |
| `.github/workflows/component-tests.yml` | Remove stale `server/pnpm-lock.yaml` cache path and redundant install step |
| `.github/workflows/integration-tests.yml` | Switch `start: pnpm start` → `start: pnpm run serve` |
| `DEVNOTES.md` | Document both dev modes (`pnpm start` vs `pnpm run start:server`) |

## Test plan

- [x] `pnpm start` boots the standalone Vite dev server with native HMR on https://local.dsde-dev.broadinstitute.org:3000, browser opens automatically
- [x] `pnpm run start:server` boots Fastify + Vite middleware on the same URL, exercising server code
- [x] `pnpm build` compiles both Vite bundle and server TypeScript, and writes `build/vite.config.json`
- [x] `pnpm run serve` serves the built app on port 3000 (used by CI integration tests)
- [x] `pnpm test` and `pnpm --filter duos-server test` pass
- [x] `pnpm type-check` and server `tsc --noEmit` pass
- [x] Docker image builds and serves the app correctly (verified locally)
- [x] CI component-tests and integration-tests workflows pass


otchet-broad (approved):
👍
Nice!

rushtong:
> 👍 Nice!

Thank you! Still working on this as the docker-compose isn't working yet, but we're close.

kevinmarete (approved):
👍🏾

package.json
```diff
@@ -5,9 +5,10 @@
   "packageManager": "pnpm@11.1.2",
   "scripts": {
     "analyze": "VISUALIZER_PLUGIN=true vite build",
-    "start": "vite",
-    "build": "vite build",
+    "start": "tsx watch --env-file=.env.local server/src/index.ts",
+    "build": "vite build && node scripts/write-vite-config-json.mjs && pnpm --filter duos-server build",
     "preview": "vite preview",
+    "serve": "vite preview --port 3000",
```
fboulnois, line 12: I think we probably want to keep both mechanisms in parallel:

- The usual `"start": "vite"` for local debugging
- An additional `"start:server": "tsx watch ..."` or something like that?

The TSX approach will be slower for hot-reloading compared to the built-in vite approach.

vite.config.ts
```diff
@@ -3,18 +3,6 @@ import react from '@vitejs/plugin-react'
...
-const defaultOptions = {
-  host: 'local.dsde-dev.broadinstitute.org',
-  port: 3000,
-  https: (process.env.CI || process.env.CYPRESS)
-    ? undefined
-    : {
-        key: 'server.key',
-        cert: 'server.crt',
-      },
-  open: true,
```
fboulnois, line 15: These default options were required so that running locally loaded in the certificates, and running in the CI runs without the certificates.

What happens now?

fboulnois (commented):
Some suggestions, otherwise seems reasonable:

vite.config.ts
rushtong: In the current scheme, we can't load them externally like this, we have to read them directly into `index.ts`.

package.json
rushtong: Let me dig in a bit. As I understood the process, the two approaches are somewhat exclusive, but I'm not 100% sure so will take a look.

otchet-broad:
Would it make sense to have two different Dockerfile files?  One for dev and the other for staging/production?

package.json
rushtong: This is now resolved, both mechanisms are now possible in parallel.

vite.config.ts
rushtong: This is now resolved, we maintain the original structure when running locally via `pnpm start`

rushtong:
> Would it make sense to have two different Dockerfile files? One for dev and the other for staging/production?

@otchet-broad  - I'm hesitant on taking this approach. It would mean non-trivial changes to both CI and terra-helmfile to both create and reference multiple images in GAR.

fboulnois (approved):
thank you for keeping the earlier mode, looks great 👍

PR #3662, opened 2026-06-30
[DT-2589] Convert `ConfirmationDialog_new.jsx` to TypeScript

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2589

### Summary

Convert `ConfirmationDialog_new.jsx` to TypeScript with vitest tests. Renamed `ConfirmationDialog_new` to `AsyncConfirmationDialog`

----

- Verify all tests go green


rushtong (approved):
👍🏽  This PR begs the question - what are doing with the old `ConfirmationDialog`? That component has been migrated, but having both seems like tech-debt.

kevinmarete:
> 👍🏽 This PR begs the question - what are doing with the old `ConfirmationDialog`? That component has been migrated, but having both seems like tech-debt.

These aren't simple duplicates — they have meaningfully different APIs and underlying libraries. `modals/ConfirmationDialog` is MUI-based with a flat `close()`/`action()` interface for simple yes/no confirmations. `ConfirmationDialog_new` uses `react-modal` with a curried async handler `(confirmed: boolean) => () => Promise<void>`, an `AsyncSpinnerButton` for the confirm action, inline alert support, and accepts arbitrary `children` instead of a plain description string. Consolidating them means picking a winner, porting the missing features across, and refactoring `DataAccessRequestApplication`'s handler pattern — which is a meaningful change to a large, unrelated file. Worth doing, but as a focused follow-up PR.

src/components/AsyncConfirmationDialog.tsx
otchet-broad, line 1: Given the discussion about the name hiding the meaningful differences between this and the other dialog, could we possibly rename this to something that better describes the key difference in this component?

otchet-broad (approved):
👍
Looks good!  See comment for optional file/component name change.

PR #3663, opened 2026-06-30
[DT-2591] Convert `DownloadLink.jsx` to TypeScript

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2590

### Summary

Convert `DownloadLink.jsx` to TypeScript with vitest tests.

----

- Verify all tests go green


rushtong (approved):
👍🏽

cypress/component/pages/manage_dac/EditDac.spec.tsx
fboulnois, line 1: Are we planning to move this test in a future PR?

fboulnois (approved):
One question otherwise lgtm 👍

cypress/component/pages/manage_dac/EditDac.spec.tsx
kevinmarete: Yes, moving `EditDac.spec.tsx` to vitest is planned but intentionally deferred to a follow-up PR. The `a → button` change on line 647 is a required fix to keep the existing Cypress test passing after `DownloadLink` switched its rendered element — without it, the test would break.

PR #3687, opened 2026-06-30
[DT-2570, DT-3440] Convert `Home.jsx` to TypeScript, and `home.spec.jsx` to Typescript/Vitest

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2570
https://broadworkbench.atlassian.net/browse/DT-3440

### Summary

Convert `Home.jsx` to TypeScript and migrate `home.spec.jsx` to Vitest tests.

----

- Verify all tests go green


test/pages/Home.spec.tsx
```diff
@@ -0,0 +1,176 @@
+import '@testing-library/jest-dom/vitest'
```
fboulnois, line 1: This test is missing the equivalent of many of the `cypress` tests, including checking that the page renders correctly on desktop, tablet, and mobile.

fboulnois (commented):
One comment:

rushtong (approved):
Pending Florian's comment about window-sized tests, 👍🏽

test/pages/Home.spec.tsx
kevinmarete: Thanks for flagging this — I audited `test/pages/Home.spec.tsx` against the deleted `cypress/component/Home/home.spec.jsx` and added the missing equivalents.                                                     
                                                                                                                                                                                                                    
  ## Added                                                                                                                                                                                                          
                                                                                                                                                                                                                    
  - **`navigates directly without calling handleSignIn when logged in`** — asserts `handleSignIn` isn't called when the user is already logged in.                                                                  
  - **`renders one card per featured library`** (both logged-in and logged-out) — equivalent of cypress's `.logo-card` count assertion, scaled to the mocked library data (3 featured) instead of the real dataset  
  (22), since the vitest suite intentionally mocks `getLibraryVersions` for isolation rather than depending on live data.                                                                                           
  - **`calls handleSignIn with the target library path when a card is clicked`** — equivalent of the logged-out click-interaction test, checking the correct path is passed through.                                
  - **Responsive layout block**:                                                                                                                                                                                    
    - `renders the logo grid as a centered, wrapping flex container`                                                                                                                                                
    - `defines a tablet breakpoint that shrinks the logo cards`                                                                                                                                                     
    - `defines a mobile breakpoint that makes logo cards full width`                                                                                                                                                
                                                                                                                                                                                                                    
  ## Gap I couldn't port 1:1 (with rationale)                                                                                                                                                                       
                                                                                                                                                                                                                    
  The cypress desktop/tablet/mobile tests worked by calling `cy.viewport()` in a **real browser** and reading computed CSS after layout/media-query recalculation. `jsdom` (used by vitest/RTL here) has no layout  
  engine and doesn't evaluate `@media` queries against `window.innerWidth`, so there's no way to genuinely simulate "resize to mobile and check the recomputed style."                                              
                                                                                                                                                                                                                    
  Instead, I split the original intent into what's actually unit-testable:                                                                                                                                          
                                                                                                                                                                                                                    
  1. The **static inline flex properties** on `.logo-grid` (`display: flex`, `flexWrap: wrap`, `justifyContent: center`) — this is what cypress's desktop/mobile assertions were actually checking, and they're     
  viewport-independent already.                                                                                                                                                                                     
  2. The **`@media` rules themselves** (`max-width: 768px` / `max-width: 480px`), asserted directly against the injected `<style>` tag content, since those rules are what actually drive the responsive card       
  sizing.                                                                                                                                                                                                           
                                                                                                                                                                                                                    
  If we want true "rendered at different viewport sizes" coverage (visually verifying the breakpoints kick in), that would need a real-browser tool (Cypress/Playwright), not jsdom

otchet-broad:
The old tests used to adjust the viewport to confirm we were getting the desired behaviors from the design.  I don't see the equivalent tests in the new vite version.  I'd expect to see us bring in something like vite+playwright, use an import like `import { page, render } from '@vitest/browser/context';`, and await a viewport like `await page.viewport(<width>,<height>)` and run it with browser.headless=true in CI.  That might expand the scope of this ticket a bit, but I think it matches @fboulnois' comment.

Maybe we should split this into jus the typescript and then do the vitest migration?

kevinmarete:
> The old tests used to adjust the viewport to confirm we were getting the desired behaviors from the design. I don't see the equivalent tests in the new vite version. I'd expect to see us bring in something like vite+playwright, use an import like `import { page, render } from '@vitest/browser/context';`, and await a viewport like `await page.viewport(<width>,<height>)` and run it with browser.headless=true in CI. That might expand the scope of this ticket a bit, but I think it matches @fboulnois' comment.
> 
> Maybe we should split this into jus the typescript and then do the vitest migration?

  Good catch — you're right that the jsdom-based vitest tests can't actually validate `flex-direction`/`flex-wrap` reflow at a given viewport, since jsdom doesn't evaluate media queries or compute real layout. Splitting scope as you 
  suggested:                                                                                                                                                                                                                             
                                                                                                                                                                                                                                         
  **This PR (TS conversion):** kept the `test/pages/Home.spec.tsx` responsive coverage that jsdom *can* validate — the base flex properties that aren't behind a media query, plus a regex check that the `<style>` tag's source text    
  contains the correct `@media (max-width: 768px/480px)` breakpoint rules. That still catches someone deleting/mistyping a breakpoint, just can't prove the browser actually reflows.                                                    
                                                                                                                                                                                                                                         
  To close the real gap without blocking this PR on new test infra (this repo has no `@vitest/browser`/Playwright dependency yet — it'd need a new project/config plus CI changes to install headless browsers), I restored a trimmed    
  `cypress/component/Home/home.spec.jsx` with just the 3 viewport-dependent tests from the original spec (desktop/mobile/tablet `.logo-grid` reflow checks). Everything else from that file is dropped since it's now covered            
  equivalently in the vitest suite — no point running duplicate assertions in two frameworks.                                                                                                                                            
                                                                                                                                                                                                                                         
  Filed [DT-3771](https://broadworkbench.atlassian.net/browse/DT-3771) to add `@vitest/browser` + Playwright as proper project infra, with Home as the first real consumer; that cypress file has a TODO marking it for removal once that lands.    

[DT-3771]: https://broadworkbench.atlassian.net/browse/DT-3771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

fboulnois (approved):
👍

otchet-broad (approved):
👍
Plan to adjust scope and new ticket sounds great.  Thank you!

PR #3703, opened 2026-07-01
[DT-3629] Enable Postgres, Session, and Cookie plugins

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3629

## Summary
Phase 1 of the BFF migration: server-side, PostgreSQL-backed session infrastructure for the Fastify server. No user-visible changes — the infrastructure is inert until users are routed to it by the Phase 2 auth routes.

The DB/cookie/session plugins register only when the deployment provides the database configuration (`DUOS_DB_HOST` etc.), so the legacy client-side auth flow is completely unaffected until an environment is configured for the BFF — and every pod of a deployment behaves identically, with no network dependency at boot. 

## Changes

**Session infrastructure (`server/src/`)**
- `index.ts`: registers `@fastify/postgres`, `@fastify/cookie`, and `@fastify/session` only when `DUOS_DB_HOST` is set. Validates `DUOS_SESSION_SECRET` (32+ chars) with an actionable error. Sets `trustProxy: 1` so `request.protocol` honors the reverse proxy's `X-Forwarded-Proto` — without it, secure cookies silently never persist behind TLS-terminating proxies.
- `session/pgStore.ts`: a thin typed `SessionStore` (get/set/destroy) against the `user_sessions` table. Expiry is computed in SQL from the same clock `get` filters with, so app/DB clock drift can't shorten or extend sessions. No `rolling` expiry — sessions get a fixed `maxAge`; Phase 2 adds throttled sliding expiry instead of a DB write per request.
- `types/session.ts`: module augmentation typing the session payload (tokens live only server-side, never sent to the browser).
- Postgres TLS is secure-by-default: `DUOS_DB_SSL` must be explicitly set to `false` for loopback transports (Cloud SQL Proxy sidecar, local compose), and doing so logs a warning.

**Client config serving**
- `config.ts`: `/config.json` is now served by the server (via an `onRequest` hook, to avoid colliding with `@fastify/vite`'s static route for the same file), with `DUOS_API_URL` overriding the static file's `apiUrl`. Deployed environments already carry the matching value via terra-helmfile; locally there's now one place to point at a different consent API.

**Local development**
- `docker-compose.yaml` is split into a shared base plus two overlays: `docker-compose.override.yaml` (standalone — bundled Postgres stands in for the consent DB, auto-merged by Compose) and `docker-compose.consent.yaml` (shares a local `consent` compose stack's database, avoiding port collisions with its `sqlproxy` and `app`).
- New `.env.example` template; both compose modes and `pnpm start:server` read `.env.local`.
- `DEVNOTES.md`: full walkthrough of both modes, including the consent-DB sharing setup.

## Deployment notes
- Requires https://github.com/broadinstitute/terra-helmfile/pull/6462 (`DUOS_DB_SSL: "false"` for the Cloud SQL Proxy sidecar). The X-Forwarded-Proto + `duos_api_url` changes already landed in https://github.com/broadinstitute/terra-helmfile/pull/6460.
- This branch (with the helmfile updates) is deployed to a BEE: https://duos.gr-dt-3629.bee.envs-terra.bio/

----

- Verify all tests go green


server/src/session/pgStore.ts
```diff
@@ -0,0 +1,60 @@
...
+ * against `app.pg` — no Express compatibility shim, and no `touch`:
+ * `@fastify/session` re-saves via `set` when `rolling` is enabled and never
+ * calls `touch`.
+ *
+ * Correctness never depends on the `pg_cron` cleanup job: `get` filters on
+ * `expire > NOW()`, so expired rows are invisible even before they are purged.
+ */
+export function createPgSessionStore(pg: PostgresDb): SessionStore {
+  const expiryFromSession = (session: Session) =>
+    new Date(Date.now() + (session.cookie.maxAge ?? EIGHT_HOURS_MS))
```
otchet-broad: Why do we need to consider maxAge from the cookie?  Why not just the EIGHT_HOURS_MS value?

Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN pnpm --filter duos-server run build
 RUN pnpm --filter duos-server deploy --prod --legacy /tmp/server-deploy
 
 # Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:24-debian
-FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:f0e79759bb8cea65c59ed276e3b5c0f19188698cf3a1f2fd67a598c05ca2d902
+FROM us.gcr.io/broad-dsp-gcr-public/base/nodejs@sha256:f4a51d4a89698ce6b37d5840f483c5f537ebec6ba0afa498e70e302d1d576c8f
```
otchet-broad: I'm guessing you're going to take @kevinmarete 's change here?

server/src/session/pgStore.ts
rushtong: The cookie's max age is the actual value that is set from `index.ts` - 8 hours is just a fallback value and we should never really be hitting that condition. `@fastify/session` types `cookie.maxAge` as nullable so we need it for typing. `pgStore` keeps the user's cookie session in sync with the database so we're always saving an actual value, not the max value.

server/src/index.ts
```diff
@@ -18,45 +22,91 @@ const useHttps = isDev && !isCI
...
+      host: process.env.DUOS_DB_HOST,
+      database: process.env.DUOS_DB_NAME,
+      port: Number.parseInt(process.env.DUOS_DB_PORT ?? '5432', 10),
+      user: process.env.DUOS_DB_USER,
+      password: process.env.DUOS_DB_PASSWORD,
+      // App-level TLS is only used for a direct connection to Cloud SQL. When the
+      // BFF reaches Postgres over localhost (a Cloud SQL Proxy sidecar in k8s, or
+      // the bundled `db` container in docker-compose) the transport is already
+      // plaintext-on-loopback, so SSL must be off or the connection is rejected.
+      ssl: process.env.DUOS_DB_SSL === 'true' ? { rejectUnauthorized: true } : false,
```
fboulnois: Inverts the previous always-on `ssl: { rejectUnauthorized: true }`. Any value that isn't exactly `'true'` — unset, `'TRUE'`, `'1'`, a typo, or a blank secret — connects in **plaintext with no cert verification and no warning**. Safe for the documented Cloud SQL Proxy sidecar (loopback), but a deployment that ever connects *directly* to Cloud SQL and forgets the var downgrades credentials + full session payloads to an unauthenticated channel. The old code made that impossible.

**Fix:** secure-by-default with an explicit, documented opt-out only for the loopback/Cloud-SQL-Proxy case; parse the boolean through one coercion helper rather than an exact string compare.

server/src/index.ts
```diff
@@ -18,45 +22,91 @@ const useHttps = isDev && !isCI
...
+      store: createPgSessionStore(fastify.pg),
+      cookie: {
+        httpOnly: true,
+        secure: process.env.NODE_ENV === 'production',
+        sameSite: 'strict',
+        maxAge: Number(process.env.DUOS_SESSION_MAX_AGE_MS) || 8 * 60 * 60 * 1000,
+        path: '/',
+      },
+      saveUninitialized: false,
+      rolling: true,
```
fboulnois: On `develop` the session store was unwired (`set` was a no-op), so `rolling` was free. This branch wires `createPgSessionStore`, so `@fastify/session` issues an `INSERT … ON CONFLICT UPDATE` on **every session-bearing request** just to bump `expire`. A page firing N parallel API calls does N `SELECT`+`UPSERT` round-trips against one `sid` row (WAL/index churn, row contention).

**Fix:** gate the re-save to when the session is near expiry (e.g. within the last 10% of TTL), or drop `rolling`.

server/src/session/pgStore.ts
fboulnois: Relatedly: `set` uses the Node container's `Date.now()`; `get` filters on Postgres's `NOW()`. Clock drift (or a non-UTC container TZ against a `timestamp`-without-tz column) can expire sessions early/late. Also `?? EIGHT_HOURS_MS` only catches null/undefined — a `cookie.maxAge` of `0` would write an already-expired row. Not reachable with the current fixed 8h config, but fragile.

server/src/featureFlags.ts
```diff
@@ -0,0 +1,42 @@
+import type { FastifyBaseLogger } from 'fastify'
+
+export interface FeatureFlag {
```
fboulnois: This interface is byte-identical to the one in `src/libs/ajax/FeatureFlag.ts:4`, and these could drift independently.

server/src/session/pgStore.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  return {
+    async get(sid, callback) {
```
fboulnois: The three methods differ only in SQL and whether a value is returned. A single `run(cb, work)` helper could collapse each to one line and make error handling uniform by construction.

fboulnois (commented):
A few suggestions. Claude has also flagged that the way we do feature flags isn't optimal. This is a preexisting issue, but worthwhile to think about:

`await isBffEnabled(...)` runs once in `buildApp()` and is never re-checked. A transient consent blip or >5s response during a rolling deploy makes a pod come up with **no DB/session plugins for its entire lifetime**, silently serving legacy auth even after consent recovers and flipping the flag on requires restarting every pod. During rollout this yields per-pod inconsistency visible only in logs.

rushtong:
Great feedback @fboulnois - I'm going to revisit the plan and make some fundamental changes and hope to address all of the feedback in one go.

server/src/index.ts
rushtong: Good catch — fixed in 1fcde3b6. TLS is now secure-by-default: `DUOS_DB_SSL` defaults to on, and only an explicit, recognized "off" value (`false`/`0`/`no`/`off`, case-insensitive, via an `envBool` coercion helper) disables it — a typo or blank secret keeps the secure branch. The loopback opt-out is documented at the call site and logs a warning when active. The helmfile side (`DUOS_DB_SSL: "false"` next to the sqlproxy sidecar) is going in as a terra-helmfile follow-up.

server/src/index.ts
rushtong: Dropped `rolling` in 1fcde3b6 — sessions now get a fixed maxAge from creation, so there's no per-request UPSERT. Since `saveUninitialized: false` and no auth routes exist yet, nothing writes sessions in this phase anyway. Sliding expiry returns in Phase 2 as the throttled variant you suggested: re-save only when the session is within the last ~10% of its TTL (noted in the epic-2 plan).

server/src/session/pgStore.ts
rushtong: Fixed in 1fcde3b6 — expiry is now computed in SQL (`NOW() + $3 * interval '1 millisecond'`), so `set` and `get` share Postgres's clock and container drift/TZ can't shorten or extend sessions. The fallback also guards non-positive values now (`maxAge > 0 ? maxAge : EIGHT_HOURS_MS`), so a zero maxAge can't write an already-expired row.

server/src/featureFlags.ts
rushtong: Moot as of 1fcde3b6 — the server no longer fetches the feature flag at all (see the summary comment), so `server/src/featureFlags.ts` and its copy of the interface are gone.

server/src/session/pgStore.ts
rushtong: Done in 1fcde3b6 — the three methods now go through a single `run(callback, work)` bridge, so error handling is uniform by construction and each method body is just its query.

Dockerfile
rushtong: Resolved by merging develop — Kevin's `24-debian-fips` base image change (#3722) came in with the merge, so this branch no longer touches the Dockerfile at all.

rushtong:
@fboulnois thanks for the review — the feature-flag concern was the important one, and it exposed a deeper problem: the plan defines `BFF_ENABLED` as a client-side rollout gate, so the server checking it at boot was a divergence, with exactly the failure modes you described (a consent blip permanently downgrades a pod; flag flips need restarts).

**1fcde3b6 removes the boot-time flag check entirely.** The server now registers its DB/cookie/session plugins whenever the deployment provides the database config (`DUOS_DB_HOST`), so every pod of a deployment behaves identically with no network dependency at startup — a half-configured pod (DB host set, session secret missing) fails loud instead of silently booting legacy. The infrastructure is inert until users are routed to it (`saveUninitialized: false`, no auth routes yet). `BFF_ENABLED` goes back to its planned role: the client-side flag deciding which sign-in flow users get, flippable with no redeploys. The two-switch model is documented in `docs/plans/BFF_Overview.md`.

Also in that commit, per the inline threads: secure-by-default Postgres TLS with an explicit coerced opt-out, `rolling` dropped (throttled sliding expiry planned for Phase 2), session expiry computed on Postgres's clock with a non-positive-maxAge guard, and the store methods collapsed through a single `run()` helper.

kevinmarete (approved):

fboulnois (approved):
👍

PR #3709, opened 2026-07-02
[DT-2555, DT-3382, DT-3384, DT-3385] Convert `DataAccessRequestApplication.jsx` to TypeScript , and its tests to Vitest

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2555

https://broadworkbench.atlassian.net/browse/DT-3382
https://broadworkbench.atlassian.net/browse/DT-3384
https://broadworkbench.atlassian.net/browse/DT-3385

### Summary

Convert `DataAccessRequestApplication.jsx` to TypeScript and migrate all related Cypress tests to Vitest.

----

- Verify all tests go green


src/pages/dar_application/DataAccessRequestApplication.tsx
```diff
@@ -271,9 +314,9 @@ const DataAccessRequestApplication = (props) => {
   useEffect(() => {
     const fetchData = async () => {
       try {
-        if (existingDarsReadOnlyMode) {
+        if (existingDarsReadOnlyMode && collectionId) {
           const { createUser } = await getDarCollection(collectionId)
-          setResearcher(createUser)
+          setResearcher(createUser ?? {})
```
otchet-broad, line 320: Should we throw an exception if createUser doesn't exist instead of using an empty object?

otchet-broad (approved):
👍
Tested locally.  Looks good!

src/pages/dar_application/DataAccessRequestApplication.tsx
kevinmarete: I would keep the fallback instead of throwing. `createUser` is typed as optional on `DarCollection`, so a missing value here is expected (e.g., an old collection whose user can no longer be resolved), not an error condition. We already handle this gracefully further down with the `isResearcherEmpty` check, which renders blank fields instead of crashing. Throwing would get caught by the outer `try/catch` and hide the entire read-only DAR view just because one nested field didn't hydrate — worse UX for a non-critical gap.

fboulnois (approved):
👍

PR #3711, opened 2026-07-02
[DT-3752] Shift LC page to Researcher Status page

jlaw-codes:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3752](https://broadworkbench.atlassian.net/browse/DT-3752)

### Summary

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

src/components/dac_bot/DACBotComponent.tsx
```diff
@@ -114,7 +114,7 @@ export const DACBotComponent = (props: DACBotComponentProps) => {
   const { dacId, 'data-cy': dataCy, mutuallyExclusiveRules = DEFAULT_MUTUALLY_EXCLUSIVE_RULES } = props
   const [DACbotRules, setDACbotRules] = useState<Array<DACbotRule>>([])
   const [isLoading, setIsLoading] = useState(true)
-  const userIsChair = Storage.getCurrentUser().roles.some((r: UserRole) => r.dacId == dacId && r.name == 'Chairperson')
+  const userIsChair = (Storage.getCurrentUser().roles ?? []).some((r: UserRole) => r.dacId == dacId && r.name == 'Chairperson')
```
rushtong: `UserRoles` is a required property on `DuosUser`, so I think this conditional is unnecessary. Here and in other files as well.

src/pages/manage_dac/ManageDac.tsx
```diff
@@ -103,7 +103,7 @@ export const ManageDac = function ManageDac() {
   const roles = useMemo(() => currentUser.roles?.map((r: { name: string }) => r.name) ?? [], [currentUser])
   const userRole: ManageDacRole = roles.includes(ADMIN) ? ADMIN : CHAIR
   const chairDACIds = useMemo(() => new Set(
-    currentUser.roles
+    (currentUser.roles ?? [])
```
rushtong: See previous comment about roles

src/components/manage_users_table/ManageUsersTableCellData.tsx
```diff
@@ -84,7 +84,7 @@ export function emailCellData({ userId, email, label = 'email' }: EmailCellDataP
 
 export function permissionsCellData({ userId, roles, libraryCard, label = 'permissions' }: PermissionsCellDataParams): CellData {
   const hasLibraryCard = !isNil(libraryCard)
-  const roleNames = roles.map((role): string => role.name).filter(name => name !== 'Researcher')
+  const roleNames = (roles ?? []).map((role): string => role.name).filter(name => name !== 'Researcher')
```
rushtong, line 87: See previous comment about roles

src/pages/signing_official_console/DataCustodianTable.tsx
```diff
@@ -163,7 +163,7 @@ const SubmitterCell = ({
 }
 
 const roleCell = (roles: DuosUserWithInstitutionId['roles'], id: TableRowId): TableCell => {
-  const roleString = chain(roles.map(role => role.name))
+  const roleString = chain((roles ?? []).map(role => role.name))
```
rushtong: See previous comment about roles

src/pages/signing_official_console/SigningOfficialTable.tsx
```diff
@@ -157,55 +90,70 @@ const LibraryCardCell = ({
...
-        style={{
-          display: 'flex',
-          justifyContent: 'left',
-        }}
-        key={`lc-action-cell-${id}`}
+        style={{ display: 'flex', alignItems: 'center', gap: '0.5rem' }}
+        key={`lc-status-cell-${id}`}
       >
-        {button}
+        <Switch
```
rushtong, line 133: Love the toggle 👍🏽

src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -1,7 +1,9 @@
 import React, { useEffect, useState } from 'react'
+import { Link } from 'react-router-dom'
 import { Institution as InstitutionAPI } from '../../libs/ajax/Institution'
+import { User } from '../../libs/ajax/User'
```
rushtong: nit: Imports should reference from src: 
```suggestion
import { User } from 'src/libs/ajax/User'
```

src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -1,7 +1,9 @@
...
 import { Notifications } from '../../libs/utils'
-import { DuosUser, InstitutionInterface } from '../../types/model'
+import { DuosUser, InstitutionInterface, SigningOfficialUserWithData } from '../../types/model'
```
rushtong: Same thing here. Generally, when we touch these files, we try to clean these up as well, so lines 3 & 5 could use the same fix.

src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -11,19 +13,24 @@ export default function AffiliationAndRole(props: AffiliationAndRoleProps) {
...
           if (institution) {
             setInstitution(institution)
           }
+          const soUsers = await User.getSOsForInstitution(user.institutionId)
+          setSigningOfficialUsers(soUsers)
+        }
+        else if (user) {
+          const soUsers = await User.getSOsForCurrentUser()
+          setSigningOfficialUsers(soUsers)
         }
```
rushtong: This logic is confusing to me. If the user doesn't have an institution id, we're looking up the Signing Officials for the user (`User.getSOsForCurrentUser()`) which should return no users. 

I think the if condition for institution id should only set the institution name.

Separately, you cal just call 
```
const soUsers = await User.getSOsForCurrentUser()
setSigningOfficialUsers(soUsers)
```
outside of any if condition. There has to be a user here or the page won't load.

src/pages/user_profile/ResearcherStatus.tsx
```diff
@@ -118,40 +132,7 @@ const ResearcherStatus: React.FC<ResearcherStatusProps> = (props) => {
               daas={daaObjects}
             />
           )
```
rushtong, line 134: Something broke here and I'm not clear on what. In my case, I have a library card and DAAs, the APis to get the DAAs are being hit, but this block is not showing up. I'm also not seeing the else conditional for no library card found.

Image

src/pages/user_profile/ResearcherStatus.tsx
```diff
@@ -80,11 +76,29 @@ const ResearcherStatus: React.FC<ResearcherStatusProps> = (props) => {
...
       <div style={{ marginTop: '20px' }} />
+      <p style={subheadStyle}>Status</p>
+      <p style={{
+        fontFamily: 'Montserrat',
+        fontSize: '16px',
+        fontWeight: 600,
+        color: hasCard ? '#00928A' : 'rgb(128, 128, 128)',
+      }}
+      >
+        {hasCard ? 'Active' : 'Inactive'}
```
rushtong, line 94: Something is borked here, likely the same logic that's preventing the DAAs from showing up:

Image

rushtong (commented):
Some initial feedback.

src/components/manage_users_table/ManageUsersTableCellData.tsx
kevinmarete: This is necessary, as some users may not have roles.

src/pages/user_profile/AffiliationAndRoles.tsx
```diff
@@ -25,14 +26,35 @@ export default function AffiliationAndRole(props: AffiliationAndRoleProps) {
             setInstitution(institution)
           }
         }
+        const soUsers = await User.getSOsForCurrentUser()
+        setSigningOfficialUsers(soUsers)
```
kevinmarete, line 30: Good catch! This was the fix, the bug identified in `ResearcherStatus.tsx`.

Image

src/components/manage_users_table/ManageUsersTableCellData.tsx
rushtong: I'm not clear on the case where users have no roles - can you elaborate on that? `DuosUser.roles` is currently a required field. If it is the case that users exist without roles, then `DuosUser` needs to be updated and that change percolated throughout.

rushtong: See https://github.com/DataBiosphere/consent/blob/4bfd3fdae225cdf3ca3730b26bdd614a04aae81a/src/main/java/org/broadinstitute/consent/http/models/CreateDuosUserRequest.java#L33 for why I think roles always exist.

rushtong (approved):
Nice fixes, thank you. One outstanding question that shouldn't hold this up 👍🏽

PR #3715, opened 2026-07-04
[DT-3766] add counts to asset tabs in data library

jlaw-codes:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3766](https://broadworkbench.atlassian.net/browse/DT-3766)

### Summary

----

- Verify all tests go green


jlaw-codes:
Screenshot 2026-07-03 at 10 36 39 PM

src/hooks/useLibraryData.ts
```diff
@@ -123,6 +123,41 @@ export const useLibraryData = (
...
+  filters: FilterState,
+  queryTerm: string,
+): Partial<Record<AssetType, number>> => {
+  const results = useQueries({
+    queries: assetTypes.map((assetType) => {
+      const sanitized = sanitizeFiltersForAsset(assetType, filters)
+      return {
+        queryKey: ['library-tab-count', libraryConfig.key, assetType, sanitized, queryTerm],
+        queryFn: async () => {
+          const query = buildElasticsearchQuery(libraryConfig, assetType, sanitized, queryTerm, { page: 0, pageSize: 0 })
```
otchet-broad, line 142: Should this filter out publicVisibility: no or controlled datasets/consent groups that aren't approved yet?

kevinmarete: The tab count query at line 142 uses the same `buildElasticsearchQuery` path as the main data query, so whatever filtering applies to the grid also applies to the count.                                                                
                                                                                                                                                                                                                                         Controlled/unapproved datasets — filtered correctly. `datasetAsset.buildQuery` applies the `bool.should` / `dacApproval: true` gate unconditionally; `pageSize: 0` doesn't bypass it.                                                          
                                                                                                                                                                                                                                         
`publicVisibility: no` — neither the tab count nor the main grid query filters on this field today. If it needs enforcing, that's a pre-existing gap, not something introduced here.                                                     
                                                                                                                                                                                                                                         
Studies — the study asset doesn't apply the approval gate (counts all studies regardless of dataset approval status), but this is consistent with how the main studies grid already behaves.

kevinmarete (approved):

rushtong:
I think this PR has overlap with https://github.com/DataBiosphere/duos-ui/pull/3720 where there are also asset counts in the tabs. If that one lands, this one can likely be closed.

src/hooks/useLibraryData.ts
otchet-broad: I observed the following query run for each tab:
`{"size":0,"query":{"bool":{"must":[{"exists":{"field":"study"}}]}},"aggs":{"studies":{"terms":{"field":"study.studyId","size":10000},"aggs":{"study_details":{"top_hits":{"size":1,"_source":["study.*"]}}}}}}`

Moreover, this same query seems to run for a number of asset tabs.  That means a nearly 10X increase on the number of queries we're going to hit the backend with just to render one tab.

Image

Interestingly some queries **do** have the query filter as you described, but a number of them do not.

kevinmarete: Agreed, maybe this whole approach needs to be refactored.

PR #3716, opened 2026-07-04
[DT-3753] access column style change

jlaw-codes:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3753
https://broadworkbench.atlassian.net/browse/DT-3783

### Summary

**Before**

Before

Before-duos

**After**

After

After-duos

----

- Verify all tests go green


src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -46,41 +46,25 @@ export const makeDatasetColumns = (
...
+    width: 140,
     renderCell: (params) => {
-      const summary = getAccessManagementSummary(params.value)
       const isRadarEnabled = radarEnabledDatasetIds.has(params.row.datasetId)
+      const label = (() => {
+        switch (params.value) {
+          case 'open': return 'Open Access'
+          case 'controlled': return 'via DUOS'
+          case 'external': return 'External to DUOS'
+          default: return params.value
```
fboulnois, line 59: This removes the chip colors, which I found useful. Is this expected?

src/hooks/useLibraryData.ts
```diff
@@ -123,6 +123,41 @@ export const useLibraryData = (
   })
 }
 
+/**
+ * Custom hook for fetching the total count of assets for each tab type in parallel.
+ * Uses pageSize:0 so only the total is returned, not full row data.
+ */
+export const useLibraryTabCounts = (
```
fboulnois: This seems like an unrelated fix. Can you describe why this is necessary?

fboulnois (changes requested):
See below:

rushtong (commented):
This branch is fairly out of date. @jlaw-codes - can you document the goals behind these changes? Are they still a priority? We can run with it with a little more background.

src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -48,41 +48,41 @@ export const makeDatasetColumns = (
...
-              }
-            />
-          </Tooltip>
+        <Box sx={{ display: 'flex', alignItems: 'center', height: '100%' }}>
+          <Chip
+            label={(
+              <Box sx={{ display: 'flex', alignItems: 'center', gap: 0.5 }}>
+                {label}
+                {isRadarEnabled && (
+                  <Tooltip title="Automatic request approvals available for datasets clearly within the data use terms.">
```
otchet-broad: Does it make sense to invert this?  Put the tooltip outside of the chip so that the entire hit area of the chip is covered by the tooltip?  We could craft a tooltip suitable for all cases and then display that.  Might be nice for screen readers.  Doing so would also close [this](https://broadworkbench.atlassian.net/browse/DT-3783) ticket.

fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3720, opened 2026-07-05
[DT-3775] move request location and add now button

jlaw-codes:
## Addresses
[https://broadworkbench.atlassian.net/browse/DT-3775](https://broadworkbench.atlassian.net/browse/DT-3775)

## Summary
Reworks how users request access to datasets from the Data Library, and adds result counts to the asset tabs.

**"Request Path" column with a "Request Now" button**

- Moves the request-location column up next to the "Access" column and renames it "Request Path".
- The column now renders by access type:
  - **Controlled (via DUOS):** a new `RequestAccessButton` ("Request Now") that creates a DAR draft for that single dataset and navigates to the application — the same flow as the footer's "Apply for Access", via the shared `applyForAccess` helper. Disabled with an explanatory tooltip when the user has no Library Card.
  - **Open:** a "-" (no request needed), ignoring any request location.
  - **External:** a link to the external request location, as before.
- **Selection mode takes over:** when one or more datasets are selected, all row-level "Request Now" buttons are disabled with a tooltip pointing to the footer's "Apply for Access" button. This avoids two competing request paths with different outcomes (requesting one dataset vs. the whole selection) being active at once.

**Asset tab counts**
This has been removed from the PR but screenshots still show them.

### Example Screens
Default view for Controlled Access datasets
Screenshot 2026-07-07 at 8 44 51 AM

When checkbox-selecting datasets:
Screenshot 2026-07-07 at 12 24 48 PM

Default view for Open Access datasets
Screenshot 2026-07-07 at 8 44 57 AM

Default view for External Access Datasets
Screenshot 2026-07-07 at 8 45 07 AM

----

- Verify all tests go green


src/components/data_search/DatasetExportButton.tsx
```diff
@@ -19,27 +20,10 @@ export const DatasetExportButton = ({ snapshots }: DatasetExportButtonProps) =>
...
-        rel="noopener noreferrer"
-        title={title}
-        aria-label={title}
-      >
-        Export
-      </Link>
-    )
-  }
+  const snapshot = snapshots[0]
+  const terraLink = `${terraUrl}/#import-data?snapshotId=${snapshot.id}&format=tdrexport&tdrSyncPermissions=false`
```
rushtong: Collapsing these means the multi-snapshot menu loses its per-snapshot items. Where a user could previously pick which of several snapshots to export, they now only get the first one in the list; the rest are unreachable. The first one they get is unknowable - it could be the oldest from TDR, the newest, or some other one depending on the order TDR chooses to report the snapshots.

jlaw-codes: We do not want to lost that for now, so should avoid this change for the time being.

rushtong: Gotcha, will revert those changes.

otchet-broad:
@rushtong - Did you see my comments [here](https://github.com/DataBiosphere/duos-ui/pull/3715/changes#r3531294354)?  I think this branch is likely to have the same problem.

In addition to the issue there, I think we have a race condition that causes the cache to miss since a number of the queries fire at the same time.

rushtong:
> @rushtong - Did you see my comments [here](https://github.com/DataBiosphere/duos-ui/pull/3715/changes#r3531294354)? I think this branch is likely to have the same problem.
> 
> In addition to the issue there, I think we have a race condition that causes the cache to miss since a number of the queries fire at the same time.

@otchet-broad - thanks for the pointer, I didn't see that. I'll pull those changes from this PR. I've had to do this already since there were unintended changes to the export feature.

otchet-broad (approved):
👍
Looks good, pending merge conflict cleanup.

cypress/component/data_library/columns/datasetColumns.spec.tsx
```diff
@@ -89,3 +89,44 @@ describe('datasetColumns — Access Management chip', () => {
...
+  it('shows a link to the requestLocation for external datasets', () => {
+    renderGrid({ accessManagement: 'external', requestLocation: 'https://example.com/request' })
+    cy.get('[data-field="requestLocation"] a').should('have.attr', 'href', 'https://example.com/request')
+  })
+
+  it('disables the "Request Now" button when datasets are selected elsewhere on the page', () => {
+    renderGrid({ accessManagement: 'controlled' }, true)
+    cy.get('[data-field="requestLocation"]').contains('button', 'Request Now').should('be.disabled')
+  })
+})
```
kevinmarete: Could you make this a vitest test, since this file has been migrated?

kevinmarete (approved):
Looks good, I would suggest moving that tests in `datasetColumns.spec` to vitest since the entitre file has been migrated.

PR #3729, opened 2026-07-07
[DT-3572] Migrate `StudyAssets/` Cypress component tests to Vitest 

kevinmarete:
### Addresses

### Summary

This PR migrates 10 spec files from `cypress/component/StudyAssets/` to Vitest + React Testing Library under test/.                                                                                                                      
                                                                                                                                                                                                                                         
  - `ai_model_list.spec`                                                                                                                                                                                                                   
  - `biospecimen_list.spec`                                                                                                                                                                                                                
  - `clinical_trial_list.spec`                                                                                                                                                                                                             
  - `consent_group_list.spec`                                                                                                                                                                                                              
  - `funding_resource_list.spec`                                                                                                                                                                                                           
  - `intellectual_property_list.spec`                                                                                                                                                                                                      
  - `presentation_list.spec`                                                                                                                                                                                                               
  - `publication_list.spec`                                                                                                                                                                                                                
  - `workspace_list.spec`                                                                                                                                                                                                                  
  - `testUtils` (removed, no longer needed)    

----

- Verify all tests go green


rushtong (approved):
👍🏽

fboulnois (commented):
Systemic finding (all 9 List specs):

* The delete-via-modal test lost 2 of its 3 post-delete assertions everywhere. The original testDeleteViaModal helper checked: (a) item text gone ✅ kept everywhere, (b) the confirmation modal unmounts ❌ dropped everywhere, (c) .collaborator-summary-card count is 0 ❌ dropped everywhere. So delete is now verified only by the item's text disappearing — a weaker signal than confirming the card was structurally removed and the modal closed. .collaborator-summary-card is a real, rendered class (StudyAssetSummary.tsx:39), so (c) was meaningful coverage. Most specs also downgraded the pre-click modal be.visible check to mere presence.

Secondary drops:

* consent_group: its shared clickSaveButton helper dropped the not.toBeDisabled() pre-click assertion, affecting ~6 save/edit tests.
* clinical_trial and funding_resource: AddEdit dropped a "save button not disabled" pre-click assertion — but these components never actually disable the save button (validation only gates the internal save() handler), so this was a no-op even in the original. Low value; the test name over-promises in both versions.
* All types: date-picker interactions are now mocked (plain input + fireEvent.change) rather than driving the real calendar. Reasonable jsdom isolation; no date-value assertions were lost. (publication's List view-mode test still uses the real picker, so #publishedDate disabled+value coverage is preserved there.)

fboulnois (approved):
One small suggestion for `presentation`, otherwise lgtm:

Because this spec mocks ModalWrapper to render its children inline (not via react-modal's portal), the class .ReactModal__Content is never produced in this file — so the newly-added expect(document.querySelector('.ReactModal__Content')).not.toBeInTheDocument() is always true (it would pass whether or not the modal actually closed). The pre-click check correctly uses .delete-modal-primary-button instead, confirming the mock wrapper is in play. So the modal-teardown line is decorative rather than functional. A stricter version would assert .delete-modal-primary-button is gone after delete.

fboulnois (approved):
👍

PR #3733, opened 2026-07-07
[DT-3572] Migrate `routing/` Cypress component tests to Vitest 

kevinmarete:
### Addresses

### Summary

This PR migrates 3 spec files from `cypress/component/routing/` to Vitest + React Testing Library under `test/`.                                                                                                                        
                                                                                                                                                                                                                                         
  - `AppRoutes.spec`                                                                                                                                                                                                                   
  - `EnvRoute.spec`
  - `NotFound.spec`

----

- Verify all tests go green


test/routing/AppRoutes.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const LocationSpy = ({ onLocationChange }: { onLocationChange: (loc: string) => void }) => {
+  const location = useLocation()
+  React.useEffect(() => {
+    onLocationChange(location.pathname + location.search)
+  }, [location, onLocationChange])
+  return null
+}
+
+const roleBACRoutes: string[] = [
```
rushtong, line 29: I remember writing these originally and longed for a way to have these routes be statically referenced.

rushtong (approved):
👍🏽  Inline comment shouldn't hold this up.

test/routing/AppRoutes.spec.tsx
fboulnois, line 1: * One consistent gap: every test dropped the NotFound page's absence check (Cypress verified [data-cy="not-found"] does not exist alongside the redirect; Vitest only checks the redirect spy was called). Minor but genuine — restorable via container.querySelector('[data-cy="not-found"]') since the app uses data-cy, not data-testid.

fboulnois (commented):
See below:

test/routing/AppRoutes.spec.tsx
kevinmarete: Agreed — the list will always need to be manually synced with the JSX route definitions. A true single source of truth would require consuming path constants directly in AppRoutes.tsx, which is a larger refactor outside the scope of this migration.

fboulnois (approved):
👍

PR #3734, opened 2026-07-08
[DT-3572] Migrate `data_library/` Cypress component tests to Vitest 

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3410
https://broadworkbench.atlassian.net/browse/DT-3411
https://broadworkbench.atlassian.net/browse/DT-3412
https://broadworkbench.atlassian.net/browse/DT-3407

### Summary

This PR migrates 4 spec files from `cypress/component/data_library/` to Vitest + React Testing Library under `test/`.                                                                                                                        
                                                                                                                                                                                                                                         
  - `LibraryDataGrid.spec`                                                                                                                                                                                                                   
  - `LibraryFilters.spec`
  - `LibraryFooter.spec`
  - `LibraryTabs.spec`

----

- Verify all tests go green


rushtong (approved):
👍🏽

test/components/data_library/LibraryDataGrid.spec.tsx
fboulnois, line 1: This migration has lost the bulk of the original Cypress coverage:

* Entire pagination test dropped (calls onPaginationChange when page changes).
* Dropped participant/dataset-count column assertions in both study and dataset row tests.
* Dropped the chip color-class assertions (MuiChip-colorPrimary/colorSuccess/colorSecondary for Controlled/Open/External) — this directly tested the access-management→color mapping switch in datasetColumns.tsx; now only the text label is checked, so a bug that scrambles the color mapping would go uncaught.
* Dropped per-row verification that Export links appear on the correct row (now just checks a count).
* Dropped the third row's Bolt-icon-absence check.

test/components/data_library/LibraryFooter.spec.tsx
fboulnois, line 1: This migration has lost the bulk of the original Cypress coverage:

* entire describe('tooltip', ...) block dropped (2 tests): the "Library Card required" tooltip visible on hover, and its absence when the user has a card. This is a full MUI Tooltip feature (with custom styling) left with zero coverage.

test/components/data_library/LibraryFilters.spec.tsx
fboulnois, line 1: * One regression: an exact object-equality check (onChange called with the entire filter state) was loosened to objectContaining, so a bug that mutates unrelated filter keys would no longer be caught.

test/components/data_library/LibraryTabs.spec.tsx
fboulnois, line 1: * Dropped the overflow/scroll-buttons test entirely (viewport-driven, hard to reproduce in jsdom without mocking offsetWidth/scrollWidth — not attempted).
* The "active tab" test now checks MUI's internal Mui-selected class instead of the component's actual fontWeight styling logic — tests a different mechanism, would pass even if the app's own bold/normal logic broke.

fboulnois (commented):
See below:

fboulnois (approved):
👍

PR #3738, opened 2026-07-08
[DT-3572] Migrate `pages/` Cypress component tests to Vitest 

kevinmarete:
### Addresses

### Summary

This PR migrates 5 spec files from `pages/` to Vitest + React Testing Library under `test/`.                                                                                        
                                                                                                                                                                                                                                         
  - `manage_dac/EditDac.spec`                                                                                                                                                                                           
  - `manage_dac/ManageDac.spec`                                                                                                                                                                                                     
  - `manage_dac/DacUsers.spec`                                                                                                                                                                                                              
  - `user_profile/RequestForm.spec`                                                                                                                                                                                                  
  - `data_submission/v2/v2-common-functions.spec` 

----

- Verify all tests go green


rushtong (approved):
👍🏽

test/pages/manage_dac/DacUsers.spec.tsx
fboulnois, line 1: * One test dropped entirely ("displays all chairpersons and members separately"), and one pre-click "row not yet grey" style assertion dropped from the pending-removal test.

test/pages/manage_dac/EditDac.spec.tsx
fboulnois, line 1: * One concrete gap: the download-button tests now mock DaaTabs entirely, so clicking download no longer goes through DAA.getDaaFileById — the assertion is tautological (checks the test's own mock was called by the test's own mock). That wiring has zero coverage anywhere in the repo now.

test/pages/manage_dac/ManageDac.spec.tsx
fboulnois, line 1: This migration has lost the bulk of the original Cypress coverage: The real ManageDacTable is now fully mocked, which means:

* 2 navigation tests dropped entirely (DAC-name click → profile page, edit-icon click → edit page) — the mock uses plain `<a>` tags instead of real router links, so these couldn't be ported as-is.
* The approved/unapproved dataset filtering logic (filter(datasets, {dacApproval: true})) is no longer exercised — the mock just pre-supplies already-filtered data, and the DAC.datasets call-argument assertion is gone.
* Loading-spinner-cleared check dropped.
* Post-delete "DAC actually disappears from the DOM" assertion dropped (only mock-call assertions remain).

test/pages/user_profile/RequestForm.spec.tsx
fboulnois, line 1: * Two regressions: the "non-URL text still blocks submission" edge case test was dropped, and the multi-checkbox test lost its assertion that the submitted ticket description actually contains both selected labels (now only checks the button became enabled).

fboulnois (commented):
See below:

fboulnois (approved):
👍

PR #3758, opened 2026-07-10
[DT-3766] Add counts, harmonize behavior across tabs

otchet-broad:
### Addresses

[DT-3766
](https://broadworkbench.atlassian.net/browse/DT-3766)
### Summary

Reworks the Data Library so filters, search, and item counts behave consistently across every asset tab.

#### Tab count badges
- Every tab now shows a count badge. Counts for all tabs are fetched in a single size: 0 Elasticsearch request (new useLibraryTabCounts hook + libraryCounts.ts), independent of the active tab, pagination, and sort — fetched once and reused as the user pages, sorts, and switches tabs (previous counts stay visible while refetching to avoid flicker).
- Counts derive from one shared aggregation: total_studies (cardinality) → Studies, a datasets_count filter agg → Datasets, and a shared studies agg fed through each asset's own transformResponse, so a tab's count always matches what its grid renders.
- STUDY_ASSET_TABS is derived from the AssetType enum, so newly added asset types are counted automatically.
- Datasets count reuses the extracted APPROVED_CONTROLLED_SHOULD clause shared with datasetAsset.buildQuery, so badge and grid can't diverge.
- Removed the standalone total-count text from LibraryPageShell (counts now live in the tabs). New shared COUNT_BADGE_SX pill style; new FilterAggregation elastic type.

#### Filters harmonized across tabs
- Filters are no longer dropped when switching tabs. buildActiveFilterClauses applies all active filters regardless of which tab set them, so they combine like facets across tabs.
- Filters owned by another tab appear as removable "Filters from other views" chips (getExternalActiveFilters + removeFilterValue), so users can drop them without switching tabs. New ActiveFilterChip type wired through useLibraryPageState → LibraryPageShell → LibraryFilters.
- Active-filter indicator counts by category (stable across tabs) and shows a count badge on the collapsed panel.

#### Search across tabs
- New ALL_SEARCH_FIELDS (union of every asset's search fields); search now matches the same fields on every tab, keeping grids consistent with the count badges.

#### Publications "Datasets Cited"
- Added a distinct publicationsDatasetsCited boolean filter (separate from presentations' datasetsCited), with its own filter definition, URL param handling, and disambiguated labels ("...(Publications)?" vs. "...(Presentations)?"). Publications tab now exposes this filter.
- PublicationAddEdit gains a required yes/no "Did you cite the dataset(s) used in this publication?" field, with validation in darFormUtils.

#### Tests
- Added/updated specs across LibraryFilters, LibraryTabs, filterRegistry, useLibraryData, useLibraryPageState, useLibraryUrlState, the new useLibraryTabCounts, publicationAsset, and darFormUtils.
----

- Verify all tests go green


test/pages/researcher_console/DatasetSubmissions.spec.tsx
rushtong, line 1: I confirmed this Claude feedback on the **My Data Submissions** page:

> "On My Data Submissions, a submitter deletes a dataset: the grid refetches and drops the row, but the Datasets tab badge (rendered via useLibraryPageState → LibraryPageShell) keeps showing the pre-delete count for up to 5 minutes — the exact badge/grid divergence the shared-count design is meant to prevent. Fix: also invalidate [LIBRARY_TAB_COUNTS_QUERY_KEY, libraryConfig.key]."

src/components/data_library/filterRegistry.ts
```diff
@@ -404,6 +399,18 @@ const FILTER_DEFINITIONS: Record<FilterKey, FilterDefinition> = {
...
   },
+  publicationsDatasetsCited: {
+    label: 'Datasets Cited (Publications)?',
+    buildClause: (filters) => {
+      if (filters.publicationsDatasetsCited === undefined) {
+        return undefined
+      }
+
+      return {
+        term: { 'study.assets.publications.citation': filters.publicationsDatasetsCited },
```
rushtong: I confirmed this Claude finding:

> "Publications indexed without the citation field (any pre-Oct-2025 records or non-form ingestion paths): selecting 'No' excludes their studies at the query level, yet the grid displays them as citation=No under 'Any' — and a citation-less publication stays visible if a sibling publication has an explicit false. The pattern mirrors the pre-existing presentations filter, but this PR adds a second instance of it; consider an exists-based clause (bool should: term false OR must_not exists) if legacy documents lack the field."

src/hooks/libraryCounts.ts
```diff
@@ -0,0 +1,94 @@
...
+  total_studies: {
+    cardinality: { field: 'study.studyId' },
+  },
+  datasets_count: {
+    filter: datasetsCountClause(showAllControlled),
+  },
+  studies: {
+    terms: { field: 'study.studyId', size: 10000 },
+    aggs: {
+      study_details: { top_hits: { size: 1, _source: ['study.*'] } },
```
rushtong: I confirmed this - I am seeing two (different) queries for every search request. Not sure if this is fixable, but Claude found it:

> "Every filter change and debounced search edit fires two heavyweight ES requests, each returning the entire matching study corpus, just so counts can be computed client-side — double server load and network transfer per interaction. Cheaper: derive the active tab's count from the data query's response (transformResponse(...).total is already computed) or fold the two cheap extra aggs into one shared request; at minimum narrow _source to the fields the count predicates read."

src/hooks/libraryCounts.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Their counts can therefore all be derived from a single shared `studies`
+ * aggregation. That aggregation is fetched once by the dedicated `size: 0`
+ * tab-counts query (see `useLibraryTabCounts`), so every one of these tabs is
+ * counted from that one shared response rather than from its own data query.
+ */
+// Derived from the AssetType enum rather than hand-listed: every asset other
+// than Studies and Datasets is rendered from the shared study aggregation, so a
+// newly added asset type is counted automatically and can't be silently omitted
+// from the tab badges.
+export const STUDY_ASSET_TABS: AssetType[] = Object.values(AssetType).filter(
```
rushtong, line 23: Unverified Claude finding:
> buildCountAggregations (lines 63-68) is a ninth byte-identical copy of the studies terms+top_hits aggregation defined independently in all eight study-asset buildQuery implementations, and STUDY_ASSET_TABS hardcodes 'every non-STUDIES/DATASETS asset uses exactly this aggregation' — silently breaking the asset registry's documented one-new-file extension contract (definition.ts/index.ts).

> If any asset changes its aggregation (size, _source, nested aggs), its grid renders correctly while its badge — computed by running that asset's transformResponse against the counts query's diverged copy — silently disagrees, with no error. Fix: export one shared STUDIES_AGG constant used by every asset and by buildCountAggregations, or add a count capability to AssetDefinition.

src/components/data_library/filterRegistry.ts
```diff
@@ -464,44 +471,143 @@ export const getFilterSectionsForAsset = (
...
+    return `From ${start}`
+  }
+  return `Until ${end}`
+}
+
+/**
+ * Describe the active value of an object (range/date) filter for display in a
+ * removable chip. Returns `undefined` when the filter holds no active value.
+ */
+const describeObjectFilter = (key: typeof OBJECT_FILTER_KEYS[number], filters: FilterState): string | undefined => {
```
rushtong: Unverified Claude finding:

> describeObjectFilter re-implements the per-key 'is this filter active' predicates that already exist as hasSectionValue in LibraryFilters.tsx (lines 106-128), and the inverted clinicalTrialDates (startDate > endDate) guard now exists in three places (both of those plus buildClause at line 336), with a comment acknowledging the manual sync.

> The implementations currently agree on every key, but adding or changing an object filter requires editing both files or the collapsed-panel active-filter count and the external filter chips silently disagree. Fix: one exported isFilterActive(key, filters) helper in filterRegistry that hasSectionValue and getExternalActiveFilters both call.

src/hooks/useLibraryUrlState.ts
rushtong, line 1: This looks unrelated to this PR, but looks valid:

> useLibraryUrlState builds a fresh state object with filters: parseFiltersFromUrl(searchParams) on every render (no useMemo), so urlState.filters has a new identity each render, defeating the React.memo wrapper on LibraryFilters and forcing the externalFilters useMemo/useCallback in useLibraryPageState (lines 115-148) to recompute every render.

> Every parent render (debounced search commit, pagination, selection change) re-renders the ~500-line memoized filter accordion tree for no state change. Pre-existing pattern, but this PR touches all four involved files and adds more consumers. Fix: wrap the parse in useMemo keyed on searchParams (react-router memoizes it per location).

src/components/data_library/filterRegistry.ts
```diff
@@ -393,7 +388,7 @@ const FILTER_DEFINITIONS: Record<FilterKey, FilterDefinition> = {
         : undefined,
   },
   datasetsCited: {
-    label: 'Datasets Cited?',
+    label: 'Datasets Cited (Presentations)?',
```
rushtong, line 479: I confirmed this Claude finding, the UX is worth cleaning up at some point:

> A user sets 'No' on the Presentations tab and switches to Studies: studies with no presentations at all (the common case) vanish, though 'No' reads as 'did not cite'. The semantics predate this PR and the filter is visible/removable via external chips, but the PR widens its blast radius from one tab to all tabs — worth a product-level double-check that 'No' should require a presentation to exist.

rushtong (approved):
Address feedback as you see fit, either here or in separate tickets/PRs. I think this is still worth moving forward, great work 👍🏽

kevinmarete (approved):
Excellent work 👍🏾

src/components/data_library/filterRegistry.ts
otchet-broad: I think I'll hold off on this one to make sure Jonathan has time to think about this.

kevinmarete (approved):
Excellent 👍🏾

src/utils/darFormUtils.ts
```diff
@@ -390,6 +390,9 @@ export const calcPublicationErrors = (newPublication: Partial<Publication>): Pub
     }
   }
   if (isStringEmpty(newPublication?.datasetCitation)) validation.datasetCitation = requiredError
+  if (newPublication?.citation === undefined || newPublication?.citation === null) {
+    validation.citation = requiredError
+  }
```
rushtong, line 395: Unverified finding:
> Bug: Editing any pre-existing publication now silently fails validation — the new required "Datasets Cited" yes/no field has no backfill for legacy records with citation === undefined, blocking saves on unrelated edits (darFormUtils.ts ~L393, PublicationAddEdit.tsx L164).

src/hooks/useLibraryUrlState.ts
```diff
@@ -212,7 +242,8 @@ const parseFiltersFromUrl = (searchParams: URLSearchParams): FilterState => {
     ...parseArrayFilters(searchParams),
     ...parseRangeFilters(searchParams),
     ...parseDateFilters(searchParams),
-    datasetsCited: parseDatasetsCited(searchParams),
+    datasetsCited: parseBooleanParam(searchParams, ['datasetsCited', 'presentationsDatasetsCited']),
+    publicationsDatasetsCited: parseBooleanParam(searchParams, ['publicationsDatasetsCited']),
```
rushtong, line 246: Unverified finding:
> Bug/regression: A legacy bookmarked URL using the old datasetsCited/presentationsDatasetsCited alias now resolves to a different, distinct filter field (publicationsDatasetsCited) than before, silently changing query semantics (useLibraryUrlState.ts ~L245-253).

rushtong (approved):
This looks good to me. I'd be fine with addressing potential findings separately as a more focused task and have this PR move forward. 👍🏽

fboulnois (approved):
👍

PR #3769, opened 2026-07-15
[DT-3607, DT-3792] Add single B2C OIDC client factory + PKCE helpers

rushtong:
## Addresses
Fully addresses https://broadworkbench.atlassian.net/browse/DT-3792

## Summary
Implements Phase 2, story 2-B of the BFF migration plan: a single Azure B2C OIDC client factory for the server-side OAuth flow, plus the PKCE helper functions the upcoming `/auth/login` and `/auth/callback` routes will use.

- `getOidcConfig()` performs B2C discovery once and caches the result for the life of the process (mirrors the existing cache pattern in `config.ts`'s `readConfig()`). A failed discovery is not cached, so a transient network blip heals on the next call instead of poisoning the process.
- `pkce.verifier()` / `pkce.challenge()` / `pkce.state()` wrap openid-client's PKCE/state primitives (RFC 7636 S256).
- `index.ts` warms the discovery cache at startup (non-fatal on failure) so the first real login doesn't pay the discovery round-trip.
- Adds `--enable-fips` when starting node

This is infrastructure only — no routes are wired up yet. `/auth/login`, `/auth/callback`, etc. land in later stories of Phase 2.

### Build/test infrastructure changes
Separate from the OIDC work above: while adding tests for the new `oidcClient.ts`, I found `server/tsconfig.json` only covered `src/`, so no test file was ever part of the TypeScript project — `tsc --noEmit` silently skipped all of `server/test/`. This was catching real errors in editors (e.g. process reported as undefined) that the CI typecheck would never see.

Fixed by splitting the config:
- `tsconfig.json` is now the type-checking config: noEmit: true, includes src + test + vitest.config.ts.
- `tsconfig.build.json` is new — extends the above, restores outDir/rootDir/noEmit: false, scoped to `src` only. This is what actually produces `dist/`.
- `package.json`: build now points at `tsconfig.build.json`; added a typecheck script (`tsc --noEmit`) for CI/local use.
- `Dockerfile`: copies `tsconfig.build.json` alongside `tsconfig.json` in the build stage.

Bringing test files into the project surfaced pre-existing issues unrelated to the OIDC work, fixed here since typecheck needs to be clean:

- Missing .js extensions on relative imports in `config.test.ts`, `pgStore.test.ts`, `session.test.ts` (required under NodeNext module resolution — tsc doesn't rewrite import paths, so a `.ts` file must already import the `.js` specifier that will exist post-compile).
- In `pgStore.test.ts`, the missing extension was silently degrading the imported `createPgSessionStore` type to any, masking two implicit-any parameter errors in a test helper.

No behavior change — `dist/` output is identical, and the full test suite (44 tests) still passes.

----

- Verify all tests go green

### DIFF
--- a/server/tsconfig.build.json
+++ b/server/tsconfig.build.json
@@ -1,0 +1,9 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": false,
+    "outDir": "dist",
+    "rootDir": "src"
+  },
+  "include": ["src"]
+}
--- a/server/tsconfig.build.json
+++ b/server/tsconfig.build.json
@@ -1,4 +1,13 @@
 {
+  // Separate from tsconfig.json because the two configs need different
+  // `include` scopes, and `include` can only be set in a config file (not
+  // overridden via the tsc CLI). tsconfig.json includes test/ so tsc --noEmit
+  // (and editors, which auto-discover tsconfig.json) typecheck it too — but
+  // building with rootDir: "src" while test/ is included fails with TS6059
+  // ("File is not under 'rootDir'"), even under noEmit. Dropping rootDir
+  // instead would infer it as the common ancestor of src/ and test/, changing
+  // dist/index.js to dist/src/index.js (breaking package.json's "main" and
+  // the Dockerfile/compose CMD) and compiling test files into the prod image.
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "noEmit": false,
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN pnpm exec vite build && node scripts/write-vite-config-json.mjs
 
 # Build server
 COPY server/src ./server/src
-COPY server/tsconfig.json ./server/tsconfig.json
+COPY server/tsconfig.json server/tsconfig.build.json ./server/
 RUN pnpm --filter duos-server run build
 
 # Create a self-contained prod-only server bundle (no devDeps, no workspace symlinks)
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,4 +45,4 @@ COPY --chmod=550 --chown=node:node --from=builder /tmp/server-deploy ./server
 COPY --chmod=444 --chown=node:node --from=builder /usr/src/app/package.json ./package.json
 USER node
 EXPOSE ${PORT}
-CMD ["node", "server/dist/index.js"]
+CMD ["node", "--enable-fips", "server/dist/index.js"]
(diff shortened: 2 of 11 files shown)


otchet-broad (approved):
👍
Nice work!

server/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
```
fboulnois, line 1: Does this need to be a separate file from `server/tsconfig.json`?

fboulnois (approved):
👍

kevinmarete (approved):

server/tsconfig.build.json
rushtong: That's a good question - let me see if they can be consolidated

rushtong: Confirmed with Claude. I will add comments around this question in the new config

---

Why it can't just be one file: the typecheck config needs include: `["src", "test", "vitest.config.ts"]` (so `tsc --noEmit` — and editors — actually check test files), while the build needs to emit only `src/` into dist/, mirroring src/'s own structure. A single `tsconfig.json` can't serve both, because:

1. If you keep `rootDir: "src"` and include `test/` too — TypeScript throws immediately: error TS6059: File '.../server/test/config.test.ts' is not under 'rootDir' '.../server/src'. I confirmed this happens even with `noEmit: true` — it's not just a build-time restriction.

2. If you drop `rootDir` and let both directories build — TypeScript infers rootDir as the closest common ancestor of everything in include, which is `server/` itself. I tested this too; the output becomes:
```
    dist/src/index.js       (not dist/index.js)
    dist/test/config.test.js
    dist/vitest.config.js
```
which breaks `package.json's` `"main": "dist/index.js"` and the `Dockerfile`/`docker-compose.yaml` node `server/dist/index.js` command unless those are all repointed to `dist/src/index.js` — and it ships compiled test files into the production image for no reason.

So the split is a real TypeScript constraint, not just an organizational preference — include can only be set inside a config file, never overridden from the CLI, so two different include scopes require two files.

One thing I can simplify: right now `tsconfig.build.json` extends `tsconfig.json`. That's the right direction specifically because `tsconfig.json` is the filename editors/tsserver auto-discover — keeping the broad (src+test) config as the default one is what fixed the original "editor doesn't see process in test files" problem a few turns back. If we flipped it (build config as the default, a `tsconfig.test.json` extending it), we'd regress that fix. So I'd keep the current direction, but I can tighten `tsconfig.build.json` to just the 3 overridden lines if you want it leaner — it already is:

```
{
  "extends": "./tsconfig.json",
  "compilerOptions": { "noEmit": false, "outDir": "dist", "rootDir": "src" },
  "include": ["src"]
}
```

PR #3774, opened 2026-07-16
[DT-3607] Add POST /auth/login endpoint (BFF Phase 2, 2-C)

rushtong:
### Addresses

### Summary
- Implements `POST /auth/login` per BFF Migration Plan v2 epic-2, section 2-C: generates PKCE parameters, stashes them in the session, and returns the B2C authorization URL for the client to redirect to.
- Adds `safeReturnTo()`, an open-redirect guard for the `returnTo` query param — accepts only same-origin absolute paths (leading /, parsed against a fixed base) and falls back to /, catching absolute URLs, protocol-relative URLs, backslash tricks, and ../ traversal in one origin check. The validated value is saved to session.returnTo to restore the user's destination after callback.
- Exports `requireEnv` from `oidcClient.ts` so `login.ts` fails the same way (named env var in the error) when `DUOS_OAUTH_REDIRECT_URI` is unset.
- Scope string includes `DUOS_AZURE_CLIENT_ID` alongside the standard OIDC scopes, per the B2C quirk noted in `oidcClient.ts` — B2C omits the `access_token` from the response if the client ID isn't in the scope.

----

- Verify all tests go green

### DIFF
--- a/server/src/auth/login.ts
+++ b/server/src/auth/login.ts
@@ -1,0 +1,55 @@
+import * as oidc from 'openid-client'
+import type { FastifyReply, FastifyRequest } from 'fastify'
+import { getOidcConfig, pkce, requireEnv } from './oidcClient.js'
+
+/**
+ * Open-redirect guard for the post-login redirect target: accept only
+ * same-origin absolute paths, falling back to '/'. Requiring a leading '/'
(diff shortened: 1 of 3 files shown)


kevinmarete (approved):

fboulnois (approved):
👍

PR #3778, opened 2026-07-17
[DT-3607] Add POST /auth/logout and GET /auth/me endpoints (BFF Phase 2, 2-E, 2-F)

rushtong:
### Addresses

### Summary
Implements BFF Migration Plan v2 epic-2, sections 2-E and 2-F — the remaining two `/auth/*` handlers before route registration (2-G).

- **`POST /auth/logout`** (`handleLogout`): stamps `end_reason = 'logout'` on the `user_session_audit` row for the current session (table/triggers already landed in Epic 1, DT-3606), destroys the session, and clears the `sessionId` cookie. Token revocation is attempted only when B2C's discovery document exposes a `revocation_endpoint` and only for tokens actually present in the session — B2C doesn't reliably expose one, so session destruction remains the primary logout control, and a revocation failure never blocks logout from completing.
- **`GET /auth/me`** (`getMe`): confirms the session is authenticated by calling the upstream Consent API's `/api/user/me` with the session's access token. Returns `{ authenticated: false }` (401) immediately if there's no access token, destroys the session and returns 401 if the upstream rejects the token (expired/revoked), or returns `{ authenticated: true, user, idp }` on success. Never exposes tokens.

Both handlers are new files only — neither is wired into Fastify routing yet. Route registration for all of `/auth/*`, gated on `bffEnabled`, lands in story 2-G.

----

- Verify all tests go green

### DIFF
--- a/server/src/auth/logout.ts
+++ b/server/src/auth/logout.ts
@@ -9,18 +9,27 @@ import { getOidcConfig } from './oidcClient.js'
  * best-effort on top of it.
  */
 export async function handleLogout(request: FastifyRequest, reply: FastifyReply): Promise<void> {
-  const config = await getOidcConfig()
+  // Revocation is best-effort on top of session destruction (the primary
+  // logout control below) — a discovery failure here (network blip, missing
+  // Azure env var) must not stop logout from completing.
+  try {
+    const config = await getOidcConfig()
 
-  // B2C does not reliably expose a revocation_endpoint in its discovery
-  // document, so revocation is typically skipped here. v6: the endpoint is
-  // read from config.serverMetadata(), and revocation is tokenRevocation().
-  if (config.serverMetadata().revocation_endpoint) {
-    if (request.session.accessToken) {
-      await oidc.tokenRevocation(config, request.session.accessToken).catch(() => {})
+    // B2C does not reliably expose a revocation_endpoint in its discovery
+    // document, so revocation is typically skipped here. v6: the endpoint is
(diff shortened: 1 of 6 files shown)


fboulnois (approved):
👍

fboulnois (approved):
Medium:

Logout audit UPDATE can block session destruction. In logout.ts, the pg.query audit stamp is awaited outside any try/catch and before session.destroy(). A transient DB error on the audit write (the least critical step) rejects the handler, so the session isn't destroyed and the cookie isn't cleared and the user stays logged in. This contradicts the handler's own comment ("session destruction is the primary logout control"). Make the audit stamp best-effort (or stamp after destroy), and add a test where pg.query rejects asserting destroy()/clearCookie() still fire.

Minor:

me.ts 401 branch destroys the session but doesn't clear the cookie — inconsistent with logout.ts; leaves an orphaned sessionId cookie. Consider reply.clearCookie('sessionId') here too.
me.ts comment vs. code — docstring says "returns safe user-profile fields," but the entire upstream body is forwarded verbatim. Curate the fields or soften the comment.
No timeout on the upstream fetch in me.ts — a hung Consent API hangs /me. Add AbortSignal.timeout(...) so it degrades to the existing 502 path.

Nits:

Token revocations could run via Promise.all (negligible).
Test gaps: pg.query rejection in logout (see above), and fetch rejection / res.json() failure in /me.

kevinmarete (approved):
👍🏾

PR #3787, opened 2026-07-22
[DT-3799] Surface SO authorization model on Data Library

jlaw-codes:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3799

### Summary

Reads two fields from the dataset search index that [DT-3888](https://broadworkbench.atlassian.net/browse/DT-3888) resolves server-side — `soApprovalModel` and `instantApprovalEligible` — and stops the Data Library from resolving DAC rules itself. It was issuing one `/api/dac/{id}/rules` request per unique DAC per page and re-deriving the rules in JS.

**Surfacing the SO authorization model**

- Banner on every tab, opt-in via `showSoApprovalReminder` so it doesn't leak into Researcher Console → My Data Submissions, which shares the same shell.
- "SO Approval" column on the Datasets tab: "Per-Request Approval" / "Pre-Authorized Researchers" chips, each with an explanatory tooltip.
- Three states, not two: unresolved (document predates the backfill, unrecognised value) renders **nothing**. Showing the wrong approval process is worse than showing none. A dataset with no DAC resolves to pre-authorized.

**Instant approval now comes from the index**

`getRadarEnabledDatasetsWithRules` — async, N requests, its own copy of the rule logic — becomes `getRadarEnabledDatasetIds`, a filter over what the index already resolved. `DataLibrary` loses its state and effect for a `useMemo`; `DatasetStatistics` loses its effect entirely and reads the flag off the document it had already fetched. The `Set<number>` prop shape is unchanged, so `LibraryDataGrid`, `datasetColumns` and `InstantApprovalBadge` are untouched — only the producer moves.

`useDacRules` and the injectable `DacRulesFetcher` are removed. Both existed only to make per-DAC fetching cheaper, and there is no per-DAC fetching left. The `useAsyncCacheFetch` fix for caching rejected promises stays — it permanently poisoned a key for the session, and the DAR application page is still a consumer.

**This changes which datasets can show the instant-approval badge — measured impact on prod: 3 datasets.** `isOnlyGRUorHMB` rejected four data use modifiers (`IRB, COL, GSO, NPU`); the backend's `hasNoModifiers` rejects roughly twenty, so a dataset carrying a publication moratorium (`MOR`), a geographic restriction (`GS`), or required publication of results (`PUB`) could advertise instant approval that no DAR against it would ever receive.

Measured against prod: **3 datasets lose the badge** — DUOS-000151, DUOS-000157 and DUOS-000266, all in DAC 1, all via `secondaryOther` (free text, rendered as secondary code `OTHER`, which the old frontend check never looked at). Dev showed 0, so this is confined to one DAC. **No approval outcome changes**: `hasNoModifiers` already rejects a non-empty `secondaryOther`, so `compare()` returns false for all three today and the engine has never auto-approved them — the UI simply stops claiming otherwise.

One of the three carries a genuine restriction ("Possible restriction on gamete research"). The other two carry an audit note rather than a restriction, and are being excluded from auto-approval today by that stray text — a pre-existing data issue this change surfaces rather than causes, raised separately with @ncalvanese1 and not a blocker here.

`isOnlyGRUorHMB` is deleted with its last caller — leaving a second expression of the policy here is how the two drifted apart to begin with.

**Filters**

An SO approval checkbox and an instant-approval boolean on the Datasets tab, now that both fields are filterable server-side. Both require the field to `exist`, so selecting "No" cannot silently claim every document that predates the backfill — unlike the citation filters, where a missing field genuinely means "No".

**Deploy order matters — this can no longer ship early.** DataBiosphere/consent#3014 must be deployed and `POST /api/dataset/index` run first. The SO Approval column degrades safely on un-reindexed documents (empty cell rather than a wrong label), but the instant-approval badge is an *existing* feature, so shipping this first would make it vanish rather than degrade, and the filters would exclude everything not yet backfilled.

After

----

- Verify all tests go green

[DT-3888]: https://broadworkbench.atlassian.net/browse/DT-3888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -3,18 +3,50 @@ import { GridColDef } from '@mui/x-data-grid'
...
+  headerName: 'SO Approval',
+  // Wide enough for the full 'Pre-Authorized Researchers' chip, which truncates below ~240
+  width: 250,
+  sortable: false,
+  renderCell: (params) => {
+    const model = soApprovalModelByDatasetId.get(params.row.datasetId)
+    // Blank beats a guess when the DAC's rules couldn't be loaded
+    if (model === undefined || model === 'unknown') return null
+    const isPerDarApproval = model === 'per-dar'
+    const label = isPerDarApproval ? 'Per-Request Approval' : 'Pre-Authorized Researchers'
```
otchet-broad: @kevinmarete - this is where we call it 'per-request' already.  Seems to make sense to switch to per-request in code as well.

src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -72,7 +104,10 @@ export const makeDatasetColumns = (
         : ''
       return (
         <Box sx={{ display: 'flex', alignItems: 'center', height: '100%' }}>
-          <Tooltip title={tooltipTitle}>
+          {/* describeChild keeps the access label as the chip's accessible name — without it MUI
+              sets aria-label to the tooltip text, so screen readers announce the instant-approval
+              explanation instead of the access type */}
```
rushtong: Duplicative comments here. It could be condensed, maybe something like:
```
describeChild forces the aria-label to use the access label text
```

src/components/data_library/columns/datasetColumns.tsx
```diff
@@ -147,6 +182,9 @@ export const makeDatasetColumns = (
     width: 150,
     valueGetter: (_value, row) => row.dac?.dacName || '',
   },
+  // Omitted entirely when the page supplies no authorization-model data, rather than rendering
+  // a column that could never populate
+  ...(soApprovalModelByDatasetId ? [makeSoApprovalColumn(soApprovalModelByDatasetId)] : []),
```
rushtong: IMO, the code makes this clear - the comment is overly explanatory.

src/components/data_library/filterRegistry.ts
```diff
@@ -255,7 +266,48 @@ const citationClause = (field: string, cited: boolean): QueryClause => {
...
+ * "No" would hide datasets whose state is simply unknown rather than false. Both sides therefore
+ * match only what the index actually asserts: "Yes" an explicit true, "No" an explicit false.
+ * Datasets carrying neither are excluded from both, matching the grid, which renders no badge
+ * and no chip for them rather than guessing.
+ */
+const indexedFlagClause = (field: string, value: boolean): QueryClause => ({
+  bool: {
+    must: [{ term: { [field]: value } }, { exists: { field } }],
+  },
+} as unknown as QueryClause)
```
rushtong: Is this necessary? Looking at the consent side, I think we're reindexing everything. The precedent we're setting here is confusing to me. Perhaps I'm not understanding it clearly, but it seems like we're querying on *when* something is indexed as opposed to *what* is in the index.

src/types/library.ts
```diff
@@ -187,6 +191,19 @@ export interface LibraryTabsProps {
...
+/**
+ * Which Signing Official authorization model a dataset's DAC uses, as resolved by the search
+ * index (DT-3888). 'per-request' means the SO named in each access request must approve that
+ * request before the DAC reviews it; 'pre-authorized' means the SO authorizes researchers in
+ * advance instead.
+ *
+ * 'unknown' means the index supplied no usable model — the document was written before
+ * DT-3888 and predates the backfill reindex, the backend could not resolve the DAC's rules, or
+ * the value is one this client does not recognise. Callers must render nothing in that case
+ * rather than defaulting to a model.
```
rushtong: This comment (and code logic) may be out of date (see [comment here](https://github.com/DataBiosphere/consent/pull/3014#discussion_r3769554708)) if we are re-indexing everything.

src/utils/DatasetUtils.ts
```diff
@@ -29,58 +28,47 @@ export const firstNonEmptyPropertyValue = (dataset: Partial<Dataset>, propertyNa
...
-  const secondaryCodes = dataUse.secondary?.map(s => s.code)
+/**
+ * Datasets whose DAC has an automation rule enabled that would auto-approve a matching request,
+ * as resolved server-side by the search index (DT-3888).
+ *
+ * Previously derived here by fetching `/api/dac/{id}/rules` per unique DAC and re-implementing
+ * the rule check in JS. That re-derivation could not see most of the modifiers the approval
+ * engine rejects on, so it marked datasets eligible that would never actually be auto-approved.
+ * A dataset whose document predates the backfill carries no flag and is treated as ineligible,
+ * which matches the pre-existing behaviour of showing no badge.
```
rushtong: I find Claude's commentary on how things used to be done unhelpful. It's more important going forward to know how something is intended to work. I do try to prune these kinds of comments as much as possible.

rushtong (commented):
Code generally looks good, some questions inline:
- The pre/post index date logic looks suspicious to me and may be out of date/incorrect
- Code comments are duplicative and not always helpful

src/components/data_library/columns/datasetColumns.tsx
kevinmarete: Done — condensed to your wording. The SO Approval chip carried a near-identical block; that one is a one-liner too now.

src/components/data_library/columns/datasetColumns.tsx
kevinmarete: Removed.

src/components/data_library/filterRegistry.ts
kevinmarete: Not necessary — and you were right that it read wrong. `exists` was redundant on both sides: an Elasticsearch `term` query only matches documents that carry the field, so `term: { instantApprovalEligible: false }` already cannot claim a document missing the flag. `indexedFlagClause` is gone; the filter is a bare `term`.

`citationClause` keeps its `exists` check, but for the opposite reason — there an absent field genuinely means "No", so the clause has to reach documents that lack it.

src/types/library.ts
kevinmarete: Right, out of date. Reworded to the state that survives a full reindex — rules unresolvable, or a value this client does not recognise — with the pre-backfill framing dropped. Same for the matching comments in `model.ts`, `DatasetUtils.ts` and the specs.

src/utils/DatasetUtils.ts
kevinmarete: Fair — dropped that paragraph, and swept the rest of the PR the same way: no "previously", no requirement-number narration in the specs, and nothing restating what the line below already says. Net 62 lines off, almost all comment.

kevinmarete (approved):

rushtong (approved):
Thank you for the updates 👍🏽

src/components/data_library/LibraryFilters.tsx
```diff
@@ -39,13 +39,27 @@ const CHECKBOX_FILTER_KEYS = [
...
 
 const isCheckboxFilterKey = (key: FilterKey): key is CheckboxFilterKey =>
   (CHECKBOX_FILTER_KEYS as readonly string[]).includes(key)
 
+// Yes/No/Any radio groups. A registered key claimed by neither this list nor
+// CHECKBOX_FILTER_KEYS renders nothing at all.
+const BOOLEAN_FILTER_KEYS = [
+  'datasetsCited',
+  'publicationsDatasetsCited',
+  'instantApproval',
```
otchet-broad, line 55: Nice!

src/libs/dataLibraryFilterConfig.ts
```diff
@@ -5,7 +5,15 @@ export const assetFilterRegistry: Record<AssetType, AssetFilterConfig> = {
...
   },
   [AssetType.DATASETS]: {
-    visibleFilters: ['accessManagement', 'dataUse', 'dataType', 'dac', 'participantCount'],
+    visibleFilters: [
+      'accessManagement',
+      'dataUse',
+      'dataType',
+      'dac',
+      'participantCount',
+      'soApprovalModel',
```
otchet-broad, line 14: Very cool to see both of these in here now.

src/pages/DataLibrary.tsx
```diff
@@ -93,29 +107,6 @@ export const DataLibrary: React.FC = () => {
     applyForAccess(selectedDatasetIds, navigate)
   }
 
-  useEffect(() => {
```
otchet-broad, line 96: Woohoo!

otchet-broad (approved):
Nice work!

fboulnois (approved):
👍

PR #3789, opened 2026-07-22
[DT-3805] Use Data Library grid on study details page

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3805

### Summary

- Replace the legacy dataset table with the shared Data Library grid
- Use the modern selection and access-request footer, preserving selections across pages
- Add server-side pagination and single-column sorting for study datasets
- Preserve dataset exports and prevent stale results when navigating between studies
- Improve loading, empty, and error states and update grid and hook tests

**Before**

Before

**After**

After

----

- Verify all tests go green

### DIFF
--- a/src/components/data_library/LibraryDataGrid.tsx
+++ b/src/components/data_library/LibraryDataGrid.tsx
@@ -27,8 +27,6 @@ const LoadingOverlay = () => (
 interface LibraryDataGridExtendedProps extends LibraryDataGridProps {
   extraColumns?: GridColDef[]
   checkboxSelection?: boolean
-  paginationMode?: 'client' | 'server'
-  sortingMode?: 'client' | 'server'
 }
 
 export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
@@ -46,8 +44,6 @@ export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
   radarEnabledDatasetIds = EMPTY_RADAR_IDS,
   extraColumns,
   checkboxSelection = true,
-  paginationMode = 'server',
-  sortingMode,
 }) => {
   const asset = assetRegistry[assetType]
 
@@ -127,13 +123,13 @@ export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
       <DataGrid
         rows={data as LibraryRow[]}
         columns={columns}
-        rowCount={paginationMode === 'server' ? total : undefined}
+        rowCount={total}
         loading={loading}
         pageSizeOptions={[25, 50, 100]}
         paginationModel={paginationModel}
-        paginationMode={paginationMode}
+        paginationMode="server"
         onPaginationModelChange={onPaginationChange}
-        sortingMode={sortingMode ?? asset.sortingMode}
+        sortingMode={asset.sortingMode}
         sortModel={sortModel}
         onSortModelChange={(model) => {
           onSortChange(model.map(item => ({
@@ -143,6 +139,7 @@ export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
         }}
         checkboxSelection={checkboxSelection}
         disableRowSelectionOnClick
+        keepNonExistentRowsSelected
         rowSelectionModel={rowSelectionModel}
         onRowSelectionModelChange={handleSelectionChange}
         isRowSelectable={isRowSelectable}
--- a/src/components/data_library/LibraryDataGrid.tsx
+++ b/src/components/data_library/LibraryDataGrid.tsx
@@ -27,6 +27,8 @@ const LoadingOverlay = () => (
 interface LibraryDataGridExtendedProps extends LibraryDataGridProps {
   extraColumns?: GridColDef[]
   checkboxSelection?: boolean
+  paginationMode?: 'client' | 'server'
+  sortingMode?: 'client' | 'server'
 }
 
 export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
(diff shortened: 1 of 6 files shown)


rushtong (approved):
Looks great, thank you 👍🏽

otchet-broad (approved):
👍

src/components/data_library/LibraryDataGrid.tsx
```diff
@@ -123,13 +127,13 @@ export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
...
-        rowCount={total}
+        rowCount={paginationMode === 'server' ? total : undefined}
         loading={loading}
         pageSizeOptions={[25, 50, 100]}
         paginationModel={paginationModel}
-        paginationMode="server"
+        paginationMode={paginationMode}
         onPaginationModelChange={onPaginationChange}
-        sortingMode={asset.sortingMode}
+        sortingMode={sortingMode ?? asset.sortingMode}
```
fboulnois: I think this page should sort similarly to the data library grid ("server" mode), otherwise it makes assumptions that we have previously seen to be buggy:
* It implies that no study will ever have more than 10000 datasets (because the query caps to returning 10000 items)
* Sortable dataset columns backed by nested fields (Study Name, DAC) sort correctly under client mode only because they define a `valueGetter`; any future sortable nested column added with just a SORT_FIELD_MAP entry and no `valueGetter` will sort server-side but silently do nothing on the study page's client sort.

fboulnois (changes requested):
See below:

src/components/study_details/StudyDetails.tsx
```diff
@@ -21,70 +20,42 @@ interface SectionProps {
...
+  const [exportableState, setExportableState] = useState<{ studyId?: string, data: ExportableDatasets }>({ data: EMPTY_EXPORTABLE_DATASETS })
+  const [selectionState, setSelectionState] = useState<{ studyId?: string, datasetIds: number[] }>({ datasetIds: [] })
+  const [paginationState, setPaginationState] = useState<{ studyId?: string, model: typeof INITIAL_PAGINATION }>({ model: INITIAL_PAGINATION })
+  const [sortState, setSortState] = useState<{ studyId?: string, model: Array<{ field: string, sort: SortOrder | null }> }>({ model: [] })
 
-  const study: StudyTerm | undefined = datasets.length > 0 ? datasets[0].study : undefined
-  const headers = makeDatasetTableHeader(datasets, selectedDatasets, setSelectedDatasets, exportableDatasets)
-  const rowData = makeDatasetTableRows(datasets, headers)
+  const loading = datasetState.studyId !== studyId
+  const error = loading ? undefined : datasetState.error
```
fboulnois: Separately, the use of `react-query` here could simplify this component substantially. React Query keys each response by its `queryKey`, so a response for a previous `studyId` resolves into a different cache slot; the component only ever reads the entry for the current key.

The dependent-query pattern (`enabled: datasets.length > 0`) would replace the second `useEffect` entirely, and the previous-study snapshot bug the current tests guard against can't occur because that response lands in the ['…', 'previous-study'] cache entry, which nothing reads. You also get caching between visits for free, matching how the library page already behaves.

fboulnois (commented):
And an additional idea:

src/components/study_details/StudyDetails.tsx
kevinmarete: Thanks for the feedback—addressed both points. The study page now uses the v2 endpoint with server-side pagination and the existing `datasetAsset` sort mappings, removing the 10,000-dataset assumption and client-side nested-field sorting risk. I also moved dataset and snapshot loading into React Query hooks keyed by study, pagination, sort, and dataset identifiers. Added dedicated hook and component tests for the new behavior.

src/components/study_details/StudyDetails.tsx
```diff
@@ -158,7 +76,7 @@ export const StudyDetails = () => {
         <Section>
           {study?.description}
         </Section>
-        {!Number.isNaN(participantCount) && (
+        {participantCount !== undefined && !Number.isNaN(participantCount) && (
```
fboulnois: I am noticing a lot of extra defensive checks. Many of these can be removed, especially if we make the following change:

If we create a default destructure as follows, then we can remove a lot of the `?./??` in `data.items` and `data.total`:

```
const EMPTY_PAGE = { items: [], total: 0 }
const { data = EMPTY_PAGE, isFetching: loading, error } = useStudyDatasets(studyId, paginationModel, sort)
```

Additionally, `!Number.isNaN(participantCount)` is not possible, it's either a real number or `undefined`. Similarly `!Number.isNaN(numericStudyId)` is not possible.

src/components/study_details/StudyDetails.tsx
```diff
@@ -21,118 +15,42 @@ interface SectionProps {
...
-  }
+const StudyDetailsContent = ({ studyId }: StudyDetailsContentProps) => {
+  const navigate = useNavigate()
+  const [selectedDatasets, setSelectedDatasets] = useState<number[]>([])
+  const [paginationModel, setPaginationModel] = useState(INITIAL_PAGINATION)
+  const [sortModel, setSortModel] = useState<StudySortModel>([])
+  const handleSortChange = (model: StudySortModel) => setSortModel(model.slice(0, 1))
+  const sort: SortState | undefined = sortModel[0]?.sort
+    ? { field: sortModel[0].field, order: sortModel[0].sort }
+    : undefined
```
fboulnois: There's a more declarative way to do the same thing. MUI exposes `disableMultipleColumnsSorting` on DataGrid, which prevents multi-sort issues at the source.

This would also simplify the code in the Data Library.

fboulnois (approved):
Two more suggestions:

src/components/study_details/StudyDetails.tsx
kevinmarete: Thanks for the suggestions. I’ll use a default empty page to simplify the data access and derive the selected study ID from the loaded study. For sorting, we use MUI’s community DataGrid, which already forces single-column sorting and does not expose `disableMultipleColumnsSorting` as a configurable prop, so I’ll remove the redundant page-level truncation.

PR #3790, opened 2026-07-22
[DT-3867] Add fetchBlob helper and consolidate blob downloads

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3867

### Summary

The octet-stream "download a file as a Blob" request (responseType: 'blob' + Accept header) was duplicated manually in five places: DAR.downloadDARDocument, DAR.getDARDocumentAsBlob, DAA.getDaaFileById, DAA.getDaaFileBlob, and FileStorageObject.getDocumentFile. Introduce fetchBlob in fetchAdapter to centralize this type of request; callers now pass only the URL and Config.authOpts().

Behaviour is unchanged except that the redundant Content-Type: application/octet-stream override on these bodyless GETs is dropped, fetchBlob inherits the adapter's standard default like every other GET, and DAR/DAA no longer diverge from FileStorageObject (which already omitted it).

This has been flagged a number of times by the `/improve-codebase` skill.

----

- Verify all tests go green

### DIFF
--- a/src/libs/ajax/DAA.ts
+++ b/src/libs/ajax/DAA.ts
@@ -2,6 +2,7 @@ import { fileDownload } from 'src/utils/FileDownload'
 import { Config } from 'src/libs/config'
 import { isFileEmpty } from 'src/libs/utils'
 import {
+  fetchBlob,
   fetchGet,
   fetchPost,
   fetchPut,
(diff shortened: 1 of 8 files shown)


rushtong (approved):
Nice cleanup 👍🏽

otchet-broad (approved):
👍

kevinmarete (approved):

PR #3799, opened 2026-07-29
[DT-3607] Fix `ERR_HTTP_HEADERS_SENT` crash on BFF auth routes

rushtong:
### Addresses
Minor fixes in support of https://broadworkbench.atlassian.net/browse/DT-3607

### Summary
The BFF auth routes (`POST /auth/login`, `GET /auth/callback`) could crash the
server process with an uncaught `ERR_HTTP_HEADERS_SENT`, restarting the pod and
failing the request. It reproduced reliably on any auth request that carried an
existing session cookie (e.g. retrying login after a first attempt).

**Root cause:** a double `reply.send()` race between the async route handler and
`@fastify/session`'s async `onSend` save hook:

- `reply.sent` in Fastify 5 is `raw.writableEnded`, which stays `false` while an
  async `onSend` hook is in flight.
- `@fastify/session` defaults `rolling` to `true`, and we never overrode it — so
  its `onSend` hook fired an async DB write on every session-bearing request even
  when the session was unmodified.
- That async `onSend` left `reply.sent === false` when the async handler's promise
  resolved, so Fastify's `wrapThenable` fired a **second** `reply.send()`. The
  later save's `writeHead` ran after headers were already sent → uncaught throw →
  process exit.

**Fix:**
- Set `rolling: false` explicitly on the session registration (`server/src/index.ts`).
  Combined with the existing `saveUninitialized: false` and the pre-response
  `request.session.save()` in the auth handlers, `onSend` now finds the session
  unmodified and skips its write synchronously — so the reply is fully sent before
  the handler promise resolves and no second send occurs. This also matches the
  intended session semantics (fixed maxAge from creation; sliding expiry comes
  later in Phase 2).
- Persist the session before responding in `login.ts` and `callback.ts`
  (`await request.session.save()`), which is required for the synchronous-skip
  path above.
- Updated `login.test.ts` / `callback.test.ts` to assert the session is saved
  before `reply.send`.

### Test plan
- `pnpm --filter duos-server test` — 26 auth tests green; `tsc --noEmit` clean.
- Rebuilt and redeployed the server locally, then hit `POST /auth/login` from the
  browser console repeatedly (including after a session cookie was set) — no
  crash, `redirectUrl` returned each time. Verified the full login → callback
  redirect round-trip completes without a process restart.

----

- Verify all tests go green

### DIFF
--- a/server/src/auth/callback.ts
+++ b/server/src/auth/callback.ts
@@ -44,6 +44,16 @@ export async function handleCallback(request: FastifyRequest, reply: FastifyRepl
   delete request.session.pkceVerifier
   delete request.session.pkceState
 
+  // Persist the session BEFORE responding. Otherwise, @fastify/session saves it
+  // in an async onSend hook (pgStore.set is a DB round-trip); since this handler
+  // is async, its promise resolves while that save is still in flight, so
+  // reply.sent is still false and Fastify's wrapThenable fires a SECOND
+  // reply.send() — a duplicate onSend/session-save whose late writeHead throws
+  // ERR_HTTP_HEADERS_SENT and crashes the process. Saving here resets the
+  // session's modified-hash, so onSend finds it unmodified, skips the async
+  // store write, and completes synchronously — no second send.
+  await request.session.save()
+
   // returnTo was validated to a same-origin path by safeReturnTo() at login —
   // only sanitized values ever reach the session.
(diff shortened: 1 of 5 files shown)


fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3801, opened 2026-07-29
[DT-2405] Replace `window.location` usages

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2405

### Summary

- Replace `window.location` references with `globalThis.location`
- Preserve redirect parameters updated outside React Router
- Add regression coverage for sign-in redirects
- Preserve existing OIDC and logout behavior

----

- Verify all tests go green


rushtong (approved):
👍🏽

fboulnois (approved):
Now that we have tests for the behavior, can any of these be switched to `navigate`? Otherwise, looks good.

kevinmarete:
> Now that we have tests for the behavior, can any of these be switched to `navigate`? Otherwise, looks good.

Yes—`handleSignIn` and its callers now use React Router’s navigate. The remaining `globalThis.location` usages are outside router components or require browser-level OIDC/logout behavior.

PR #3803, opened 2026-07-30
[DT-3608] BFF Phase 3: ADR-004 — API proxy layer decision (story 3-A)

rushtong:
## Addresses

[DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) — Phase 3 story 3-A: decide between `@fastify/http-proxy` and a hand-rolled `fetch` proxy for the BFF API proxy layer.

Security risk: no (documentation only).

## Summary
Documentation only — no runtime code changes. Records the proxy-layer decision before the implementation stories (3-B … 3-H) build on it, and marks Phase 2 complete in the phase table.

**Decision: `@fastify/reply-from` inside a BFF-declared route**, rather than `@fastify/http-proxy` or the hand-rolled `fetch` proxy the migration plan sketched. `@fastify/http-proxy` is `reply-from` plus prefix-mounting plus WebSocket support (a `ws` dependency DUOS has no use for); declaring the route ourselves keeps CSRF enforcement, token refresh, and upstream-401 handling as ordinary Fastify hooks instead of plugin options.

Four findings from the current client drove it, and three of them change the stories that follow:

- **Streaming is real.** Five multipart upload paths (`fetchMultipart` in `FileStorageObject`, `DataSet` ×3, `ProgressReport`), a binary POST to `/support/upload`, and blob downloads in `DAA`, `DAR`, `FileStorageObject`, and `DataSet`. The `fetch` sketch buffers every upload and download fully in memory, and copies `content-encoding`/`content-length` through verbatim while undici has already decompressed the body.
- **Fastify parses nothing useful for a proxy.** `multipart/form-data` and `application/binary` 415 before any handler runs; JSON arrives pre-parsed, so passing `request.body` to `fetch` sends `"[object Object]"`. Needs a proxy-scoped wildcard content-type parser so bodies stay unread streams.
- **9 of the ~113 upstream paths are not under `/api`** (`/status`, `/feature`, `/tos/text/duos`, `/ontology/*`, `/oauth2/configuration`, `/support/*`, `/api-docs/*`), so an `/api/*` wildcard under-covers and stripping the `/api` prefix breaks the other 95. Use one `/duos-api` prefix mapped to the upstream root — which also makes the Phase 4 client change a one-line `getApiUrl()` edit.
- **5 endpoints are called unauthenticated today**, so the proxy needs an explicit allowlist or cutover breaks the signed-out status page and Contact Us form.

### Test plan

No code changes, so nothing to exercise at runtime. Reviewers should sanity-check:

- The relative links in `ADR-004-api-proxy-layer.md` resolve from `docs/plans/bff_adrs/` (the `../../src/libs/ajax/...` citations and the `BFF_Overview.md` → `bff_adrs/ADR-004-api-proxy-layer.md` pointer).
- The decision itself, especially the `/duos-api` prefix — it becomes public API surface between client and BFF, and changing it later means changing `getApiUrl()` and the route together.

----

- Verify all tests go green

### DIFF
--- a/docs/plans/BFF_Overview.md
+++ b/docs/plans/BFF_Overview.md
@@ -52,11 +52,11 @@ The work is split into seven phases, delivered in order.
 |---|---|---|---|---|
 | 0 | Environment Configuration | Provision OAuth clients/app registrations for each identity provider and deliver server-side credentials and configuration (client secrets, session secret, database and issuer config) to the deployment environments. | [DT-3605](https://broadworkbench.atlassian.net/browse/DT-3605) | ✅ |
 | 1 | Server Foundation & Session Infrastructure | Add session middleware to the Fastify server with a PostgreSQL-backed session store, automated expired-session cleanup, and a metadata-only session audit trail. No user-visible changes. | [DT-3606](https://broadworkbench.atlassian.net/browse/DT-3606) | ✅ |
-| 2 | Server-Side OAuth Flow | Implement the BFF auth routes (`/auth/login`, `/auth/callback`, `/auth/logout`, `/auth/me`) using `openid-client` against the single Azure B2C client, with PKCE; the callback extracts the sub-provider from the B2C `id_token`. Runs alongside the legacy client-side flow during rollout. | [DT-3607](https://broadworkbench.atlassian.net/browse/DT-3607) | |
-| 3 | API Proxy Layer | Add a reverse proxy so the client calls relative `/api/*` URLs; the server injects the Bearer token from the session and proactively refreshes tokens before expiry. | [DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) | |
-| 4 | Client Refactor | Remove all token handling from the React client: drop `oidc-client-ts` and localStorage token storage, switch the fetch layer to relative URLs, and replace the popup sign-in with a full-page redirect to the B2C login page (which presents the Google/Microsoft choice, unchanged). | [DT-3609](https://broadworkbench.atlassian.net/browse/DT-3609) | |
-| 5 | Security Hardening | Layer additional defenses: strict Content Security Policy, end-to-end verification of CSRF coverage (enforcement lands with the endpoints in Phases 2–4), session-fixation protection (session ID regeneration), token revocation on logout, SRI/third-party script audit, and rate limiting on auth endpoints. | [DT-3610](https://broadworkbench.atlassian.net/browse/DT-3610) | |
-| 6 | Testing, Observability & Rollout | E2E test coverage, auth/session metrics and alerting, and a config-driven (`bffEnabled` in `config.json`) per-environment cutover. The legacy flow is removed only after the new flow is stable in production. | [DT-3611](https://broadworkbench.atlassian.net/browse/DT-3611) | |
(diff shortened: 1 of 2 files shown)


kevinmarete (commented):
Looks good so far, just a few comments:

- Blocking: The * content-type parser does not override Fastify’s built-in JSON/text parsers. Those bodies will still be buffered and limited to 1 MiB. Register explicit raw parsers for application/json and text/plain too.
- Broken links: The source links need ../../../src/..., not ../../src/.... The RefreshFailedError link also targets a nonexistent file.
- Inconsistency: BFF_Overview.md still says /api/*; update it to the chosen /duos-api/* prefix.

rushtong:
> Looks good so far, just a few comments:
> 
> * Blocking: The * content-type parser does not override Fastify’s built-in JSON/text parsers. Those bodies will still be buffered and limited to 1 MiB. Register explicit raw parsers for application/json and text/plain too.
> * Broken links: The source links need ../../../src/..., not ../../src/.... The RefreshFailedError link also targets a nonexistent file.
> * Inconsistency: BFF_Overview.md still says /api/_; update it to the chosen /duos-api/_ prefix.

Feedback applied:

Point 1:
Content-type parser (blocking) — correct, and worse than stated
ContentTypeParser.getParser (lib/content-type-parser.js:119) resolves the exact content type, then the media type, then the regex list, and only reaches customParsers.get('') — the '*' entry — last. So the built-in application/json and text/plain parsers keep winning: bodies buffered, capped by bodyLimit, handed to the handler pre-parsed. Since nearly every proxied mutation posts JSON, that was the common path, not an edge case — the ADR's sketch would have worked only for multipart.

Fixed at ADR-004 sub-decision (c) with removeAllContentTypeParsers() before the wildcard, plus the precedence reasoning so the next reader doesn't re-derive it.

Two refinements to their suggestion:

Overriding application/json directly does work, contrary to what you'd expect from FST_ERR_CTP_ALREADY_PRESENT. existingParser (line 105) special-cases those two types and returns false while the registered fn is still the default — Fastify explicitly permits replacing its own. So their remedy is viable. I chose removeAllContentTypeParsers() anyway: equivalent today, but not outflankable by a parser some future plugin registers in scope.
Encapsulation holds, which their comment didn't address and which the fix depends on. plugin-override.js:46 rebuilds the parser table per plugin instance, so clearing inside the proxy scope leaves /auth/* JSON parsing intact. Without that, removeAllContentTypeParsers() would have broken every auth route.

Point 2:
Broken links — correct, caused by a file move
The links were written when the ADR sat at docs/plans/; from docs/plans/bff_adrs/ they need ../../../. All seven now resolve (verified on disk). The RefreshFailedError path was worse than a wrong depth — it pointed at server/src/auth/refresh.ts, which doesn't exist on this branch at all, since it lands in 3-B. Replaced with a forward reference to the story rather than a path.

Point 3:
/api/* inconsistency — correct
Phase 3's row in the phase table now says /duos-api/*. I also found the same staleness one line deeper, in the authenticated-request sequence diagram at line 182 — GET /api/user/me from the browser. Now /duos-api/api/user/me, with the BFF→upstream hop at line 193 left as /api/user/me, which makes the diagram show the prefix mapping instead of contradicting it. The two remaining /api/* mentions in the ADR are deliberate — they name the rejected option.

kevinmarete (approved):
Thanks for the updates

otchet-broad (approved):
Good plan.  Should keep memory usage at bay!

PR #3805, opened 2026-07-30
[DT-3608] BFF Phase 3: single-flight token refresh (story 3-B)

rushtong:
### Addresses
 [DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) — Phase 3 story 3-B: `refreshAccessToken`.

Security risk: low — handles refresh tokens and session destruction, but registers no routes and is unreachable until story 3-C wires the proxy.

### Summary
Adds `server/src/auth/refresh.ts` and its tests. Nothing calls it yet: the `expiresIn < 60` trigger and the 401/502 mapping land with `apiProxy` in 3-C, so this PR is additive and inert in every environment.

Proactively refreshes the session's B2C access token via `openid-client`'s `refreshTokenGrant`, handling optional refresh-token rotation. The interesting part is concurrency — a SPA fires many API calls at once, and inside the refresh window every one of them would otherwise redeem the same refresh token. Since B2C may rotate on each refresh, a rotated-then-reused token comes back `invalid_grant`, so a losing request would destroy the session a winning request had just refreshed. Refresh is therefore single-flight per `sessionId`.

Three deliberate departures from the pseudocode in the Phase 3 plan, each of which prevents signing out a healthy session:

- **Only `invalid_grant` is fatal.** The plan destroys the session on any `ResponseBodyError`. But `invalid_client` means the client secret is wrong and `server_error` means B2C is unwell — properties of the deployment, not of one session — so destroying on those turns a single misconfiguration into a fleet-wide logout. Those propagate as transient; `RefreshFailedError` marks only the dead-token case. This is what lets 3-C return 502 rather than 401 for a network blip.
- **The flight spans the store write, not just the token exchange.** Releasing it when B2C responds leaves a window where the rotated token exists upstream but not in the store; a request arriving there redeems the stale token, gets `invalid_grant`, re-reads the store to check for a cross-pod winner, and can still see the pre-rotation token. Holding the flight through persistence keeps the store strictly ahead of the map, so that re-read always finds the winner. (Caught in review by Codex.)
- **Every caller applies the shared result to its own session.** Each concurrent request holds its own `Session` instance hydrated at request start, so a single mutation inside the flight would be invisible to all but the leader. Followers that proceed on a stale token risk an upstream 401 — which 3-E turns into a session destroy. Cost is N idempotent UPSERTs of identical data, roughly once per token lifetime per active session.

Also: a session with no refresh token is treated as terminal rather than retried, and the session `save()` is explicit rather than left to `@fastify/session`'s `onSend` hook — the `ERR_HTTP_HEADERS_SENT` lesson from 25a71a81, which matters more once 3-C streams replies.

### Test plan

22 unit tests in `server/test/refresh.test.ts`, mocking `refreshTokenGrant` while keeping the real `ResponseBodyError` so the `instanceof` branches are genuinely exercised:

```bash
pnpm --filter duos-server test
```

Covered: token/expiry update and persistence; rotation stored; rotation absent → existing token kept; `id_token` updated or preserved; missing `expires_in` treated as already expired; missing refresh token → destroy without calling B2C; `invalid_grant` → destroy; `invalid_grant` + rotated store → adopt the winner, session survives; `invalid_grant` with unchanged / missing / unreadable store → destroy; `invalid_client`, `server_error`, `temporarily_unavailable` and network errors → propagate, session intact; N concurrent requests → one token-endpoint call, all N sessions updated, one save each; shared rejection fails all N; different sessions don't share a flight; a settled flight doesn't block the next one; and the flight stays open while the leader's store write is pending.

That last test was verified to fail against the pre-fix release point, so it discriminates rather than merely passing.

No manual dev-environment verification is possible in this PR — no code path reaches `refreshAccessToken` until 3-C. It should be exercised end-to-end there.

----

- Verify all tests go green

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
     "name": "Node.js",
-    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:dev-26-trixie",
+    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:26-trixie",
     "postCreateCommand": "./scripts/setup-devcontainer.sh",
     "remoteUser": "node",
     "customizations": {
--- a/.devcontainer/uber/devcontainer.json
+++ b/.devcontainer/uber/devcontainer.json
@@ -1,6 +1,6 @@
 {
     "name": "Combined DUOS",
-    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:dev-26-trixie",
+    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:26-trixie",
     "workspaceFolder": "/workspaces",
     "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/duos-ui,type=bind,consistency=cached",
     "mounts": [
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN pnpm --filter duos-server run build
 RUN pnpm --filter duos-server deploy --prod --legacy /tmp/server-deploy
 
 # Commit hash to us.gcr.io/broad-dsp-gcr-public/base/nodejs:26-debian-fips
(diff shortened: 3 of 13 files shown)


kevinmarete (approved):

fboulnois (approved):
👍

PR #3808, opened 2026-08-03
[DT-3608] BFF Phase 3: API proxy handler (story 3-C)

rushtong:
## Addresses

[DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) — Phase 3 story 3-C: implement the `apiProxy` handler.

Security risk: low. The plugin is not registered on the app in this PR, so it adds no reachable route to any deployment — the diff is purely additive and cannot regress a running environment. It is security-relevant code (it attaches the session's bearer token and strips the session cookie at the trust boundary), and it adds one production dependency, so it warrants review rather than "no".

## Summary

Implements the proxy handler decided in ADR-004 (story 3-A): **`@fastify/reply-from` inside a BFF-declared route**, replacing the hand-rolled `fetch` proxy the migration plan sketched for this story. Stacked on 3-B's single-flight refresh, which has since merged, so this now targets `develop` directly.

`server/src/proxy/apiProxy.ts` is an encapsulated Fastify plugin that:

- **Clears every content-type parser in its own scope** and installs one `'*'` pass-through. A `'*'` parser alone is not enough — it is only the last resort, so Fastify's built-in `application/json` and `text/plain` parsers keep winning. Without this, multipart and binary uploads 415 before the handler runs, JSON arrives pre-parsed (and would be sent upstream as `"[object Object]"`), and every body is buffered against the 1 MB `bodyLimit`. Clearing is scoped, so `/auth/*` keeps normal JSON parsing.
- **Maps one prefix to the upstream root** — `/duos-api/<path>` → `${DUOS_API_URL}/<path>`. Covers all ~113 upstream paths with one rule, including the 9 outside `/api`, and cannot collide with the BFF's own routes. Epic 4 then becomes a one-line change to `getApiUrl()`.
- **Exempts five paths from authentication.** `/status`, `/oauth2/configuration`, `/tos/text/duos`, `/support/request`, `/support/upload` proxy through with no session and — importantly — no injected `Authorization` even when the caller does have one, because that is exactly what the client does today. Without this, cutover breaks the signed-out status page and Contact Us form. Each of the five was re-verified against its call site rather than taken from the ADR.
- **Refreshes proactively** inside a 60s window using 3-B's `refreshAccessToken`. A fatal `RefreshFailedError` returns 401 and clears the session cookie; anything transient returns **502, not 401** — a 401 would sign out every user the moment B2C hiccuped or a client secret was rotated wrong.
- **Rewrites headers at the trust boundary.** Injects `Authorization` and `X-App-ID`; never forwards the session `cookie`, a client-supplied `Authorization`, or `X-CSRF-Token`; appends this pod to the `X-Forwarded-For` chain rather than replacing it.
- **Validates `DUOS_API_URL` as a bare origin at registration.** `reply.from` builds the upstream as `new URL(path, base)`, which discards a base path and then trips its own guard, so a value with a path would 500 every proxied request. Fails at startup naming the variable instead. (No deployment impact until 3-D registers the plugin.)

### One addition beyond the story

reply-from maps `ENOTFOUND` to 503 and its timeouts to 504, but leaves `ECONNREFUSED`, `ECONNRESET`, `UND_ERR_SOCKET` and `UND_ERR_CONNECT_TIMEOUT` at **500**. That means the commonest way for the DUOS API to be down would reach the browser as `index.ts`'s generic "An unexpected error occurred" — indistinguishable from a bug in the BFF. An `onError` normalises those to 502 and leaves genuine gateway-class statuses alone.

### Deliberately not in this PR

- Registering the plugin on the app, and CSRF enforcement on unsafe methods — story 3-D. The module comment records where it has to be registered (inside the `DUOS_DB_HOST` block, under `bffEnabled`) and why: `reply.clearCookie` and `csrfProtection` are decorators from plugins that only exist there.
- Destroying the session when the upstream itself returns 401 — story 3-E.

### Dependency

Adds `@fastify/reply-from@12.6.4` (pinned exact), which brings `undici` 7. Connection pooling, hop-by-hop header handling and chunked encoding come from the library rather than from us. The pool is bounded rather than left at undici's unbounded default; story 3-H load-tests it.

## Test plan

- `server/test/apiProxy.test.ts` — 47 new tests. They run against a **real HTTP upstream** on an ephemeral port rather than a mocked `fetch`, because most of what matters here is what actually goes on the wire.
- Streaming and encoding claims are verified rather than asserted in comments: a 2 MB body proxies through (proving it is never measured against `bodyLimit`), `multipart/form-data` and `application/binary` uploads reach the upstream byte-for-byte, a JSON body survives with its key order and spacing intact (proving it was never parsed), and a gzip response arrives byte-identical with `content-encoding` intact.
- Trust-boundary coverage: session cookie never forwarded, client-supplied `Authorization` replaced, `X-CSRF-Token` dropped, no token injected on allowlisted paths, `Host` set to the upstream.
- Failure paths: 401 with no session, fatal refresh → 401 + cleared cookie, transient refresh → 502, refused connection → 502, `ENOTFOUND` → 503 preserved.
- Full server suite green: **182 tests, 13 files**. `tsc --noEmit`, `oxlint && eslint .`, and `pnpm --filter duos-server build` all clean.
- Not exercised end-to-end in a browser, and cannot be — no route is registered until 3-D. First real-environment verification lands with that PR.

----

- Verify all tests go green

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -183,6 +183,9 @@ importers:
       '@fastify/postgres':
         specifier: 6.1.0
         version: 6.1.0(pg@8.22.0)
+      '@fastify/reply-from':
+        specifier: 12.6.4
+        version: 12.6.4
       '@fastify/session':
         specifier: 11.1.2
         version: 11.1.2
@@ -647,6 +650,9 @@ packages:
 
   '@fastify/proxy-addr@5.1.0':
     resolution: {integrity: sha512-INS+6gh91cLUjB+PVHfu1UqcB76Sqtpyp7bnL+FYojhjygvOPA9ctiD/JDKsyD9Xgu4hUhCSJBPig/w7duNajw==}
+
+  '@fastify/reply-from@12.6.4':
+    resolution: {integrity: sha512-WpHqGNRKzEraRQwBFK64g2HijN0ZuDuiiwWg06r5LY0lv1iu0qYg1BfVM0iYcCMiovrM8kKXsKm5GDH9JH9aPA==}
 
   '@fastify/send@4.1.0':
     resolution: {integrity: sha512-TMYeQLCBSy2TOFmV95hQWkiTYgC/SEx7vMdV+wnZVX4tt8VBLKzmH8vV9OzJehV0+XBfg+WxPMt5wp+JBUKsVw==}
@@ -1628,6 +1634,9 @@ packages:
   emoji-regex@10.6.0:
     resolution: {integrity: sha512-toUI84YS5YmxW219erniWD0CIVOo46xGKColeNQRgOzDorgBi1v4D71/OFzgD9GO2UGKIv1C3Sp8DAn0+j5w7A==}
 
+  end-of-stream@1.4.5:
+    resolution: {integrity: sha512-ooEGc6HP26xXq/N+GCGOT0JKCLDGrq2bQUZrQ7gyrJiZANJ/8YDTxTpQBXGMn+WbIQXNVpyWymm7KYVICQnyOg==}
+
   entities@8.0.0:
     resolution: {integrity: sha512-zwfzJecQ/Uej6tusMqwAqU/6KL2XaB2VZ2Jg54Je6ahNBGNH6Ek6g3jjNCF0fG9EWQKGZNddNjU5F1ZQn/sBnA==}
     engines: {node: '>=20.19.0'}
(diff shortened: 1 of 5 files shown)


kevinmarete (approved):
Looks good, one thing noted:

### Strip upstream `Set-Cookie` response headers

`reply.from()` forwards upstream response headers by default, and the proxy currently does not provide a `rewriteHeaders` callback.

Because the response is served from the UI/BFF origin, any `Set-Cookie` returned by the DUOS API is applied to that origin. An upstream response could therefore overwrite the BFF’s `sessionId` cookie, corrupting or terminating the user’s session. Reproduced with a mock upstream and confirmed that its `Set-Cookie` header was returned unchanged.

Please add a response-header rewrite that removes at least `set-cookie`—and consider `clear-site-data` as well—plus a regression test.

Relevant code: [`server/src/proxy/apiProxy.ts`](https://github.com/DataBiosphere/duos-ui/blob/8f1ecd1f884f40b6c44bde3549c6b7fdd3c9d281/server/src/proxy/apiProxy.ts#L154-L155)

rushtong:
> Please add a response-header rewrite that removes at least `set-cookie`—and consider `clear-site-data` as well—plus a regression test.

Thanks @kevinmarete - this was a great catch. Branch is updated with fixes and additional tests.

kevinmarete (approved):

otchet-broad (approved):
👍
Nice work!

PR #3812, opened 2026-08-04
[DT-3608] BFF Phase 3: register the proxy route with CSRF enforcement (story 3-D)

rushtong:
## Addresses

[DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) — Phase 3 story 3-D: register the `/duos-api` route and enforce CSRF on unsafe methods.

Security risk: **medium**. This is the PR that makes the proxy reachable, and the control it adds is the one standing between a signed-in user's session cookie and any site that wants to drive DUOS API writes with it. No deployment changes behaviour today — the route is gated on `bffEnabled === true`, which no environment sets yet — but the gate is the only thing separating this from a live cookie-authenticated write surface, so it warrants a real read rather than "low".

**Deploy prerequisite:** `DUOS_API_URL` becomes a startup requirement whenever `bffEnabled` is true. terra-helmfile already supplies it, so nothing is pending, but an environment that flips `bffEnabled` without it now fails to boot rather than failing on the first user request.

## Summary

[Story 3-C](https://github.com/DataBiosphere/duos-ui/pull/3808) built `apiProxy` but deliberately left it unregistered. This registers it and adds the CSRF guard the story calls a requirement rather than later hardening.

- **Registered inside the `DUOS_DB_HOST` block and under `bffEnabled`** ([server/src/index.ts](server/src/index.ts)). Both gates are load-bearing. The proxy depends on `@fastify/cookie`, `@fastify/session` and `@fastify/csrf-protection`, and `index.ts` registers all three only inside the `DUOS_DB_HOST` block — registered outside it, the fatal-refresh path's `reply.clearCookie` would raise a `TypeError` instead of returning 401, and there would be no `csrfProtection` to hook. The `bffEnabled` gate then keeps the route dark until cutover, so a deployment running the legacy client-side flow exposes no `/duos-api` surface at all.
- **CSRF on unsafe methods**, as an `onRequest` hook on the proxy route ([server/src/proxy/apiProxy.ts](server/src/proxy/apiProxy.ts)). `GET`/`HEAD`/`OPTIONS` pass through, as do the two signed-out Contact Us POSTs. Both exclusions, and the residual gap they imply, are listed below.
- **One CSRF configuration, in [server/src/auth/csrf.ts](server/src/auth/csrf.ts)** (new), imported by `index.ts` and by every test harness that stands the plugin up. `getToken` is narrowed to `x-csrf-token`, down from the plugin's default of `body._csrf` plus four header spellings. Per ADR-004(c): bodies pass through the proxy unparsed, so a body-borne token could not be read without buffering the upload it is attached to — and one documented spelling is a clearer contract than four, matching what `/auth/csrf-token` tells the client to send. Sharing the object rather than restating it is a review fix; see *Changes since first review*.
- **`apiProxy` refuses to register without `@fastify/csrf-protection`.** The alternative to this check is a route that accepts cookie-authenticated writes from any origin if the plugin order ever changes.
- **`DUOS_API_URL` promoted to a `bffEnabled` startup precondition.** Both `/auth/me` and the proxy forward to it, so without it a cutover is a deployment that boots, passes health checks, and then fails on the first user request.
- **All nine ADRs are now summarised in [docs/plans/BFF_Overview.md](docs/plans/BFF_Overview.md)**, in one continuous sequence with the phase each belongs to. ADR-001–003 and 005–008 had no in-repo record before; 004 and 009 get short entries linking to their full files. This is why `bff_adrs/` holds 004 and 009 rather than 001 and 002 — the numbers belong to the list, not the directory, which was a review question. `BFF_Migration_Plan_v2.md` (outside this repo) was updated to match: it gained the ADR-009 entry whose number it had already allocated, and three stale pointers to the pre-`bff_adrs/` path were corrected.

### Changes since first review

- **Addressed — the header-only `getToken` was untested.** Three different CSRF configurations existed and none was production's, so deleting the narrowing left all 215 tests green. There is now one options object in `server/src/auth/csrf.ts`; `apiProxy.test.ts` and `auth.test.ts` import it instead of restating it, and `index.test.ts`'s mock records the options `buildApp()` registers with, which is the only place production's configuration can be asserted. Deleting `getToken` now fails two tests, one per half of the contract.
- **Addressed — the rejection tests didn't name their path.** The `it.each` no-token cases were exercising the missing-secret path. They now assert the error code, are renamed to match, and a new case covers the gap: a real session that has a secret and sends no token.
- **Does not reproduce — CSRF rejections reaching the browser as a generic server fault.** They arrive with their real `code` and `message`. `index.ts`'s sanitising `setErrorHandler` is registered *after* `register(apiProxy)`, and the proxy is an encapsulated plugin, so its route context captured Fastify's default handler and the later root call never reaches it. That is why the new tests can assert `FST_CSRF_*` codes. The reverse of the finding is a real follow-up: nothing from the proxy scope is sanitised, so a genuine 500 from a proxy-path bug returns its message. Not changed here — it wants its own decision about which shape is correct.
- **Deferred to story 3-E — uploaded documents served from the SPA origin.** Once documents proxy through `/duos-api`, an uploaded `.html`/`.svg` is served from the origin that holds the session, and a top-level navigation reaches it with the Lax cookie attached and a bearer injected. Not reachable in any deployment today (`bffEnabled` is false everywhere and the client does not call `/duos-api` until Phase 4). The mitigation is cheap and provably non-breaking — every document path is consumed via `fetchBlob`, never a navigation, so `X-Content-Type-Options: nosniff` plus `Content-Security-Policy: sandbox` on proxied responses is invisible to `fetch`.
- **Already handled before review** — the CSRF exemption keyed on `UNAUTHENTICATED_PATHS` rather than method+path. Narrowed to `CSRF_EXEMPT_UNSAFE_REQUESTS`; see exclusion 2 below.

### Deviations from the plan

- **The hook lives in `apiProxy.ts`, not `index.ts`.** The plan sketched `app.all('/api/*', {...}, apiProxy)` in `index.ts`; ADR-004 (3-A) made the plugin declare its own route, and the prefix is `/duos-api`, so the guard belongs with the route it guards. `index.ts` just registers the plugin.
- **Callback style, not `await`.** The plan's `await app.csrfProtection(request, reply)` does not work: `csrfProtection` takes a `done` callback and returns undefined, so awaiting it calls `next()` as undefined and throws a `TypeError` on the **passing** path — a failure that only surfaces once a valid token arrives, i.e. not in any test that checks rejection.

### What CSRF enforcement does not cover

Three exclusions, each deliberate. The third follows from the first and is the only one that leaves real risk.

1. **`GET`, `HEAD`, `OPTIONS`** — safe methods. Worth stating explicitly that this is not a choice: a CSRF token *cannot* protect a GET. `SameSite=Lax` is required (the OAuth callback is a top-level cross-site redirect, and `Strict` would withhold the cookie and lose the PKCE verifier), and `Lax` deliberately sends the cookie on top-level GET navigations — which cannot carry a custom header.

2. **`POST /support/request` and `POST /support/upload`** — the signed-out Contact Us form. Enforcing here would break them outright: with no session there is no CSRF secret, so the plugin returns `MissingCSRFSecretError`. There is also nothing to protect — these endpoints carry no credential, so there is no authority for an attacker to borrow.

   Keyed on **method and path together**, in a set holding exactly these two entries, rather than on the five-path unauthenticated allowlist. Narrowed from a path-only check after review feedback: the allowlist also holds read-only endpoints (`/status`, `/oauth2/configuration`, `/tos/text/duos`), so keying on it waived CSRF for unsafe methods against those too. Nothing was exploitable either way — allowlisted paths get no injected `Authorization`, and the caller's own cookie and `authorization` are stripped before forwarding, so a forged write to one of them reaches the upstream unauthenticated, which is a request anyone can already make without a victim. But that safety lives in `rewriteRequestHeaders`, not in the CSRF hook, and a path-keyed exemption would have widened silently if the allowlist or the token logic ever changed. Drift now fails closed: an unauthenticated POST added to the allowlist but not to the exempt set is rejected with `MissingCSRFSecretError`, loudly and in tests, rather than quietly exempted.

3. **Two upstream endpoints that mutate state on `GET`** — a consequence of (1), and the reason this story includes an audit. Every `@GET` method across the 32 `*Resource.java` files was checked; two are real, both reaching `NihService.syncAccount`:
   - `GET /api/nih/sync` — can **delete the user's NIH/eRA account linkage** when ECM answers 404 or 500.
   - `GET /api/user/me` — calls the same mutating chain unconditionally, plus Sam registration for a new user.

   **Neither is excluded from the proxy.** Blocking `/api/user/me` breaks the app outright, since it is how the client learns who the user is; blocking `/api/nih/sync` breaks eRA Commons linking. Requiring a token on these two GETs specifically was also rejected — the client would need a token before its first user fetch, which inverts the `/auth/csrf-token` bootstrap order.

   Residual risk: an attacker page can cause a signed-in victim's browser to issue either GET with its cookie attached. The attacker reads nothing back (the navigation renders under the DUOS origin, and same-origin policy keeps the body out of reach), so this is write-only, partly non-deterministic, and user-recoverable — the victim can re-link their account. Full analysis in [docs/plans/bff_adrs/ADR-009-state-changing-gets.md](docs/plans/bff_adrs/ADR-009-state-changing-gets.md), added in this PR.

   **The fix is upstream and small: make the sync side effect a POST, or move it out of `GET /api/user/me` so reading a profile stops writing one.** That is a Consent API change and needs its own ticket. One BFF-side mitigation — rejecting `Sec-Fetch-Mode: navigate` on the proxy, which JavaScript cannot forge — is documented and deliberately not implemented here: it is a new request-rejection rule that should be decided on its own merits, and is worth considering for Phase 5.

## Test plan

- [x] **29 new tests** — 25 in [server/test/apiProxy.test.ts](server/test/apiProxy.test.ts), 4 in [server/test/index.test.ts](server/test/index.test.ts).
- [x] The proxy test harness now registers **real `@fastify/session` and `@fastify/csrf-protection`** in index.ts's order and with index.ts's own options, instead of faking `request.session`. Stubbing `csrfProtection` would have made every CSRF test below vacuous; tokens are obtained the way a client does, through a mirror of `/auth/csrf-token`.
- [x] CSRF enforcement: `POST`/`PUT`/`PATCH`/`DELETE` rejected with 403 and no upstream call when the caller has no session secret, accepted with a valid token; a session that has a secret but sends no token is rejected; a token that does not verify against the session secret is rejected; a valid token presented **without its session cookie** is rejected (the secret lives in the session, so a token lifted from another browser is worthless); an alternative header spelling is not accepted. Each asserts **which** 403 it got — `FST_CSRF_MISSING_SECRET` or `FST_CSRF_INVALID_TOKEN` — so a case cannot silently drift onto the other path, which is what had happened to the no-token cases.
- [x] The header-only `getToken` is pinned from both ends: `apiProxy.test.ts` proves an alternative spelling is rejected end-to-end, and `index.test.ts` asserts the options `buildApp()` actually registers the plugin with (its mock records them — no other test can see production's configuration). Deleting `getToken` from `auth/csrf.ts` fails exactly those two tests; before this change it failed none.
- [x] `auth.test.ts` now registers the plugin with production's options rather than the bare defaults, so the `/auth/logout` CSRF tests reflect what production accepts. They pass unchanged — they were already sending `x-csrf-token`.
- [x] Exemptions, and their limits: `GET`/`HEAD`/`OPTIONS` need no token, and both Contact Us `POST`s proxy through without one. `POST` to the three read-only allowlisted paths is **still rejected**, and so are `PUT`/`PATCH`/`DELETE /support/request` — the exemption is keyed on method and path together, not on the allowlist. All six of those were checked against the previous path-only condition and fail there, so they pin the narrowing rather than passing vacuously. The exempt set is also pinned as a whole, so adding an entry has to appear as a deliberate edit in review.
- [x] Registration: the plugin refuses to register without `@fastify/csrf-protection`; `/duos-api` answers 401 when `bffEnabled` is true and falls through to the SPA shell when false (asserted via the upstream call, since `setNotFoundHandler` serves the shell at 200 for anything unmatched — status alone cannot distinguish a registered route from an absent one); startup fails naming `DUOS_API_URL` when it is unset.
- [x] Full server suite green: **217 tests, 13 files**. `tsc --noEmit`, `oxlint && eslint .`, and `pnpm --filter duos-server build` all clean.
- [ ] Not yet exercised in a browser: no environment sets `bffEnabled`, so first real-environment verification comes with the Phase 4 cutover that points `getApiUrl()` at `/duos-api`.

----

- Verify all tests go green


fboulnois (approved):
Some suggestions from Claude and filtered by me:

## Findings

### 1. A user-uploaded document not renders on the SPA's own origin (needs triage)

The client serves user-uploaded documents back through:

| Endpoint | Call site |
| --- | --- |
| `GET /api/daa/{id}/file` | `src/libs/ajax/DAA.ts:75` |
| DAR documents | `src/libs/ajax/DAR.ts:122` |
| `FileStorageObject` documents | `src/libs/ajax/FileStorageObject.ts:65-70` |

Once those go through `/duos-api/...`, an uploaded `.html` or `.svg` is served from the BFF origin. A top-level navigation reaches it with the Lax cookie attached and a bearer injected, which is stored XSS on the origin that holds the session, not merely on the Consent API's. This branch is what makes the route reachable at all, which is why it's worth raising here even if the fix belongs in ADR-004's header list or Epic 5.

### 2. CSRF rejections reach the browser indistinguishable from a server fault

The plugin calls `reply.send(new MissingCSRFSecretError())` / `InvalidCSRFTokenError()`, which routes through `server/src/index.ts:281-284`:

    no secret  -> 403 {"error":"An unexpected error occurred."}
    bad token  -> 403 {"error":"An unexpected error occurred."}
    logged     -> FST_CSRF_MISSING_SECRET / FST_CSRF_INVALID_TOKEN, both at log.error

### 3. The header-only `getToken` narrowing is not covered by any test

Three different CSRF configurations now exist, and none of them is the production one:

- `server/test/apiProxy.test.ts:125` re-declares the same lambda.
- `server/test/index.test.ts:27-35` mocks the plugin away with a pass-through.
- `server/test/auth.test.ts:127` registers it with the plugin **default**.

If the `getToken` line from `server/src/index.ts:173` is deleted, all 215/215 tests still pass.

---

## Minor

- The `it.each(['POST', …])('rejects %s with no token')` cases actually exercise the *missing-secret* path, not the missing-token path, since `buildProxyApp(freshSession())` sends no session cookie.
- ADR-009 lands at number 009 in a directory holding only 004.

rushtong:
> Some suggestions from Claude and filtered by me:
> 
> ## Findings
> ### 1. A user-uploaded document not renders on the SPA's own origin (needs triage)
> The client serves user-uploaded documents back through:
> 
> Endpoint	Call site
> `GET /api/daa/{id}/file`	`src/libs/ajax/DAA.ts:75`
> DAR documents	`src/libs/ajax/DAR.ts:122`
> `FileStorageObject` documents	`src/libs/ajax/FileStorageObject.ts:65-70`
> Once those go through `/duos-api/...`, an uploaded `.html` or `.svg` is served from the BFF origin. A top-level navigation reaches it with the Lax cookie attached and a bearer injected, which is stored XSS on the origin that holds the session, not merely on the Consent API's. This branch is what makes the route reachable at all, which is why it's worth raising here even if the fix belongs in ADR-004's header list or Epic 5.

The fix here lives in the response so I'll tackle it in the next story (3-E) which covers response-handling.

> ### 2. CSRF rejections reach the browser indistinguishable from a server fault
> The plugin calls `reply.send(new MissingCSRFSecretError())` / `InvalidCSRFTokenError()`, which routes through `server/src/index.ts:281-284`:
> 
> ```
> no secret  -> 403 {"error":"An unexpected error occurred."}
> bad token  -> 403 {"error":"An unexpected error occurred."}
> logged     -> FST_CSRF_MISSING_SECRET / FST_CSRF_INVALID_TOKEN, both at log.error
> ```

Two things here:
1. this is also in the response handling so I'll tackle updates in the next story (3-E)
2. Locally testing this, Claude found that the browser gets the full error. I'll ensure that we have tests for this as well since it is related to response handling:
```
no cookie : 403 {"statusCode":403,"code":"FST_CSRF_MISSING_SECRET",...}
bad token : 403 {"statusCode":403,"code":"FST_CSRF_INVALID_TOKEN",...}
```

> ### 3. The header-only `getToken` narrowing is not covered by any test
> Three different CSRF configurations now exist, and none of them is the production one:
> 
> * `server/test/apiProxy.test.ts:125` re-declares the same lambda.
> * `server/test/index.test.ts:27-35` mocks the plugin away with a pass-through.
> * `server/test/auth.test.ts:127` registers it with the plugin **default**.
> 
> If the `getToken` line from `server/src/index.ts:173` is deleted, all 215/215 tests still pass.

This is actionable - will add some tests to this PR to cover these cases.

> ## Minor
> * The `it.each(['POST', …])('rejects %s with no token')` cases actually exercise the _missing-secret_ path, not the missing-token path, since `buildProxyApp(freshSession())` sends no session cookie.

Also actionable - will add some tests to cover these cases

> * ADR-009 lands at number 009 in a directory holding only 004.

This is expected, the ADRs are not all defined in the top level document (they're in Jira/Drive) - I'll add them so the numbering doesn't look arbitrary.

kevinmarete (approved):

docs/plans/bff_adrs/ADR-009-state-changing-gets.md
```diff
@@ -0,0 +1,127 @@
...
+
+The attacker cannot read any response: a top-level navigation renders in the
+victim's own tab under the DUOS origin, and the same-origin policy keeps the body
+out of the attacker's reach. So this is a write-only, low-severity, partly
+non-deterministic CSRF, and its worst outcome is user-recoverable — the victim
+can re-link their account.
+
+### The real fix is upstream
+
+It is small: make the sync side effect a `POST`, or move it out of
```
otchet-broad, line 64: Is this proposed fix ticketed up?  It seems like the right answer overall - and not to big of a lift.

otchet-broad (approved):
👍

docs/plans/bff_adrs/ADR-009-state-changing-gets.md
rushtong: Good point, no, I have not done that yet. I will make sure that happens before merging - there are some other bits of PR feedback that need to be incorporated into the current Phase 3 ticket as well.

rushtong: Documentation tasks are now complete. ADR-009 is updated with link to new ticket to cover this detail. [DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) is also updated with new work from findings in this PR.

PR #3815, opened 2026-08-04
[DT-3932] Improve unit and Playwright test stability

rushtong:
### Addresses

Fully addresses [DT-3932](https://broadworkbench.atlassian.net/browse/DT-3932)

### Summary

Removes two sources of test flakiness — one in the unit suite, one in the e2e suite.

**`test/components/presentations_list/PresentationAddEdit.spec.tsx`**
- Replaces `container.querySelector('#id')` lookups with accessible queries (`screen.getByLabelText`, `getByRole`), so the test exercises the form the way a user does and no longer breaks on DOM/id changes.
- Drops the `fireEvent.change` on the date input in favor of `user.type`, keeping all interactions on a single `userEvent` timeline instead of mixing event models.
- Asserts against `vi.fn()` spies (`onPresentationChange`, `closeAction`) wrapped in `waitFor`, rather than a mutable `collected` array read synchronously right after the click — this was the actual race, since the assertion could run before React flushed the state update.
- Adds an assertion that `closeAction` is called, which the previous version didn't cover.
- Drops the now-unused `@testing-library/jest-dom/vitest` import (no jest-dom matchers remain in this file; it's already registered globally by the vitest setup).

**`playwright.config.ts`**
- Adds explicit `timeout: 30s` (test), `expect.timeout: 10s`, and `navigationTimeout: 30s` so slow-but-passing steps get room instead of failing at the default expect timeout.
- Adds `retries: 1` on the chromium project when `process.env.CI` is set; local runs stay at 0 so flakes remain visible during development.

**`test/e2e/support/auth.ts`**
- The `signInAs` fixture now waits for an observable application state after submitting the background sign-in form — `await expect(page).not.toHaveURL(/backgroundsignin/)` — instead of returning as soon as the click resolves. Sign-in redirects client-side via `Navigation.console`, so the destination varies by role; leaving `/backgroundsignin` is the role-agnostic signal, and each spec then asserts its own role's console heading.
- Replaces the deprecated `GoogleAuth#fromJSON` with the credential-type-specific `new JWT({ email, key, scopes })` constructor, and mints the token via `authorize()` instead of an incidental authenticated request to `BASE_URL`. `fromJSON` is deprecated because it accepts any credential shape without validation; these env vars always hold a service account key, so the narrower constructor is both the recommended path and removes the `as JWT` cast.
- Renames the fixture's setup callback from `use` to `provideSignIn`, which drops a `rules-of-hooks` suppression comment and clears the equivalent Sonar S6440 warning — both linters read a call to `use()` as React's `use()` hook. Playwright passes this callback positionally, so the name is behaviorally irrelevant.

### Test plan

- [x] `pnpm type-check` passes
- [x] `pnpm test test/components/presentations_list/PresentationAddEdit.spec.tsx` passes (1/1)
- [x] Full unit suite green: `pnpm test`
- [x] `pnpm test:e2e` green in CI (requires `DUOS_AUTOMATION_*_SA`, so it can't be run locally)
- [x] Re-run the e2e suite a few times (or via CI re-run) to confirm the previously flaky sign-in path is stable across all five roles
- [x] CI green, including the retry path on the chromium project

----

- Verify all tests go green

[DT-3932]: https://broadworkbench.atlassian.net/browse/DT-3932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -3,14 +3,20 @@ import { BASE_URL } from './test/e2e/support/baseUrl'
 
 export default defineConfig({
   testDir: 'test/e2e',
+  timeout: 30_000,
+  expect: {
+    timeout: 10_000,
+  },
   use: {
     baseURL: BASE_URL,
     ignoreHTTPSErrors: true,
+    navigationTimeout: 30_000,
   },
   projects: [
     {
       name: 'chromium',
       use: { ...devices['Desktop Chrome'] },
+      retries: process.env.CI ? 1 : 0,
     },
   ],
   webServer: {
--- a/test/components/presentations_list/PresentationAddEdit.spec.tsx
+++ b/test/components/presentations_list/PresentationAddEdit.spec.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import { describe, it, expect, vi } from 'vitest'
-import '@testing-library/jest-dom/vitest'
-import { render, fireEvent } from '@testing-library/react'
+import { render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 
 vi.mock('src/components/DuosDatePicker', () => ({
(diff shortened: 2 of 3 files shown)


kevinmarete (approved):

fboulnois (approved):
nice test cleanup 👍

PR #3816, opened 2026-08-05
[DT-3933] Consolidate DAR application details into the collection review header

jlaw-codes:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3933

### Summary

Fold the Application Information tab's applicant, collaborator, institution, and cloud-use details into the review header as a four-column facts layout, leaving the tab focused on the Non-Technical Summary and Research Use Statement. Adds researcher external profile links and aligns the Signing Official section's styling with the Researcher section.

After

----

- Verify all tests go green


rushtong (approved):
Looks good. Claude found this, but I only partially confirmed it. The SO lookups refire on the voting history and full dar tab switches:

> Worth confirming with the team: moving the SO lookup + doc-link rendering into the shared header means that API call now fires on every tab view, not just the Application Information tab.

otchet-broad (approved):
👍

kevinmarete:
> Looks good. Claude found this, but I only partially confirmed it. The SO lookups refire on the voting history and full dar tab switches:
> 
> > Worth confirming with the team: moving the SO lookup + doc-link rendering into the shared header means that API call now fires on every tab view, not just the Application Information tab.

Good thing to flag, but it actually goes the other way — this change *reduced* the call count.

`ReviewHeader` is rendered unconditionally in the `review-page-header` div (`DarCollectionReview.tsx:265`), above and outside every `selectedTab === ...` conditional (those start at line 310). So it mounts once per page load and never unmounts on tab switch — the effect doesn't re-fire.

The deps can't retrigger it either: `researcherInstitutionId` (number), `signingOfficialEmail` (string), and the derived `soLookupKey` are all primitives, so the `init()` re-run on the Chair Vote tab produces new object identities without changing dep values.

Previously the lookup lived in `ApplicationInformation`, which *is* conditionally rendered. That unmounted when you left the tab and remounted fresh when you came back, refetching every time.

I measured both branches over the same tab sequence (load → Full DAR → Voting History → back to App Info → Vote), counting `User.getSOsForInstitution` calls:

| | initial | →Full DAR | →Voting History | →back to App Info | →Vote | **total** |
|---|---|---|---|---|---|---|
| `develop` | 1 | 1 | 1 | **2** | 2 | **2** |
| this branch | 1 | 1 | 1 | 1 | 1 | **1** |

The second call on `develop` is the remount when returning to the Application Information tab. On this branch it stays at one call for the life of the page.

**If you want to reproduce this:** the stock fixtures in `DarCollectionReview.spec.tsx` won't exercise the lookup at all. The DAR fixture has no `signingOfficialEmail` and the `researcher` fixture has no `institutionId`, so the effect short-circuits on its first guard and you'll see 0 calls on both branches. You need to supply both — e.g. mock `User.getById` to resolve `{ ...researcher, institutionId: 90210 }` and add `signingOfficialEmail` to the DAR's `data` — before the counts above show up.

PR #3825, opened 2026-08-07
[DT-3608] BFF Phase 3-E: The proxy's response leg

rushtong:
### Addresses
[DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) — BFF Phase 3: API Proxy Layer, story 3-E and 3-F

Security risk: **low** — this story only closes gaps (an XSS vector, an error-message leak, and a relayed-header leak). No new attack surface; the one behavior change visible to a client is the shape of the proxy's CSRF rejection body, which nothing consumes until Phase 4.

### Summary
Everything the proxy does on the way back to the browser. (b) and (c) came out of the peer review of story 3-D (#3812) — they are properties of the proxy's own response handling, not app-wide hardening, so they belong here rather than in Epic 5.

**(a) Destroy the session on an upstream 401.** If the DUOS API rejects the token the proxy injected, the session cannot recover on its own — so it is destroyed and the browser gets a 401 to redirect on, the same verdict `/auth/me` already acts on. Only requests that borrowed the session's authority are signed out; an allowlisted path is proxied with no token, so its 401 is about the upstream and passes through untouched.

**(b) Neuter proxied responses.** DUOS serves user-uploaded documents back through the API. Once proxied they come from the BFF's *own* origin, so a top-level navigation to one carries the `SameSite=Lax` session cookie, the proxy injects the bearer, and the browser renders attacker-uploaded markup on the origin that holds the session. `HttpOnly` does not help — the script needs same-origin requests, not the cookie's value. This is new with the proxy: today the same file comes from the Consent API's origin, where a cookie-less navigation simply 401s. Every proxied response now carries `X-Content-Type-Options: nosniff` and `Content-Security-Policy: sandbox`, overriding an upstream that sends either itself.

**(c) A chosen error shape for the proxy scope — [ADR-010](docs/plans/bff_adrs/ADR-010-proxy-error-shape.md).** The proxy is deliberately encapsulated, and the app sets its error handler after registering it, so the scope kept Fastify's default and returned `err.message` on an unexpected 500 while the rest of the app returned a generic body. The shape being incidental rather than chosen was the defect. The proxy now declares its own: `403 { error: 'csrf_validation_failed', reason }` for CSRF rejections, the app's generic body for everything else. The client needs that code because **403 is ambiguous through this proxy** — an upstream authorization denial is an ordinary proxied response and arrives as a 403 too — so Epic 4 must key its refetch-and-retry on the body, not the status.

ADR-010 records the one-line alternative (moving the error handler above the BFF registrations) and why it was rejected: lossy, not risky.

**Also: the connection-specific response headers were being relayed.** Found while closing out story 3-F's test coverage against this code — the checklist asks for the hop-by-hop exclusion and nothing asserted it. `rewriteHeaders` stripped the three origin-state headers and forwarded the rest, so `keep-alive`, `proxy-authenticate`, `proxy-authorization`, `te`, `trailer` and `upgrade` reached the browser carrying the upstream's values, describing the BFF↔upstream hop to a client that is not on it (RFC 9110 §7.6.1).

Two reasons to fix it here rather than note it:

- **`proxy-authenticate` is `www-authenticate`'s proxy-side twin.** This proxy already strips the latter to stop a `Basic` challenge popping a native credential dialog on the BFF's own origin — and then forwarded the header that does the same thing. An inconsistency, not a decision.
- **`keep-alive` leaked on every response, unprompted.** Node's HTTP server adds `Keep-Alive: timeout=5` by itself on any keep-alive connection, so a Node upstream published its own socket timeout to every browser on every proxied response. Not a contrived header.

----

- Verify all tests go green

### DIFF
--- a/server/src/errors.ts
+++ b/server/src/errors.ts
@@ -1,0 +1,10 @@
+/**
+ * The body the app returns for an error it did not plan for.
+ *
+ * Shared rather than written out twice. Two scopes answer errors — the root
+ * instance and the encapsulated proxy (see ADR-010) — and the whole point of the
+ * proxy having its own handler is that the browser cannot tell which of the two
+ * failed. A string literal in each place drifts, and the drift is invisible:
+ * both still return *a* generic body, just not the same one.
+ */
+export const GENERIC_ERROR_BODY = { error: 'An unexpected error occurred.' }
--- a/server/src/errors.ts
+++ b/server/src/errors.ts
@@ -1,10 +1,5 @@
 /**
- * The body the app returns for an error it did not plan for.
- *
- * Shared rather than written out twice. Two scopes answer errors — the root
- * instance and the encapsulated proxy (see ADR-010) — and the whole point of the
- * proxy having its own handler is that the browser cannot tell which of the two
- * failed. A string literal in each place drifts, and the drift is invisible:
- * both still return *a* generic body, just not the same one.
+ * The generic error body, shared by the root and proxy error handlers (ADR-010).
+ * Shared so the two cannot drift into returning different "generic" bodies.
  */
 export const GENERIC_ERROR_BODY = { error: 'An unexpected error occurred.' }
--- a/server/src/errors.ts
+++ b/server/src/errors.ts
@@ -1,2 +1,0 @@
-/** The generic error body, shared by the root and proxy error handlers (ADR-010). */
-export const GENERIC_ERROR_BODY = { error: 'An unexpected error occurred.' }
--- a/server/src/errors.ts
+++ b/server/src/errors.ts
@@ -1,5 +1,2 @@
-/**
- * The generic error body, shared by the root and proxy error handlers (ADR-010).
- * Shared so the two cannot drift into returning different "generic" bodies.
- */
+/** The generic error body, shared by the root and proxy error handlers (ADR-010). */
 export const GENERIC_ERROR_BODY = { error: 'An unexpected error occurred.' }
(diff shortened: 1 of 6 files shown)


server/src/errors.ts
fboulnois, line 1: Do we anticipate having a number of other errors in here? Otherwise it feels premature.

server/src/proxy/apiProxy.ts
```diff
@@ -409,19 +582,138 @@ function forwardedFor(request: ProxyRequest, inbound: string | string[] | undefi
...
+    reply.send(res.stream)
+    return
+  }
+
+  // The upstream's 401 body is replaced by the reply below, but it still has to
+  // be read: an unconsumed response keeps its undici socket checked out of the
+  // pool. Drained rather than destroyed because draining preserves connection
+  // reuse. The error listener is required, not defensive: an 'error' on a stream
+  // with no listener is an unhandled exception, and reply-from's default path
+  // only gets one because `reply.send` attaches it.
```
fboulnois: The density of comments makes it difficult to follow the code. There are almost 10 times as many comment lines as actual lines of code in this file, for example. I think for some of the tricky behavior, it could make sense to have a multiline comment. From my perspective, most of the code should be self-explanatory or have a short comment above explaining the why if it's capturing more complex state.

fboulnois (commented):
A few comments relating to verbosity:

server/src/proxy/apiProxy.ts
rushtong: I agree. I've needed to start ruthlessly pruning the docs some time ago, thank you for the reminder. Will be making some updates for this branch.

server/src/errors.ts
rushtong: Good point - will refactor it.

otchet-broad (approved):
👍

fboulnois (approved):
👍

PR #3831, opened 2026-08-09
[DT-3952] Replace Library Card wording with Active Researcher Status

jlaw-codes:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3952

### Summary

This PR replaces "Library Card" with "Active Researcher Status" in the three data-access gating components:

- `LibraryFooter.tsx` — tooltip (L41) and disabled state (L57)
- `RequestAccessButton.tsx` — tooltip (L23) and disabled state
- `DatasetSearchFooter.tsx` — same pattern

Also removes the Library Card Agreement notice from the Researcher Console (`ResearcherConsole.tsx`) and its two now-unused PDF asset imports. This is beyond the ticket's written description — the removal was PO-confirmed but had no ticket of its own, so it was folded in here. Both PDFs are still used by `LibraryCardAgreementTermsDownload` and `SigningOfficialDaaAgreementWrapper`, so no assets are orphaned.

No behaviour change: the gate is still `Storage.getCurrentUser()?.libraryCard`. Copy and variable names only. Specs updated in step, including a regression test asserting the notice no longer renders.

After

----

- Verify all tests go green

### DIFF
--- a/src/components/data_library/LibraryFooter.tsx
+++ b/src/components/data_library/LibraryFooter.tsx
@@ -11,7 +11,7 @@ export const LibraryFooter: React.FC<LibraryFooterProps> = ({
   const hasSelection = selectedDatasetIds.length > 0
   const datasetText = selectedDatasetIds.length === 1 ? 'dataset' : 'datasets'
   const studyText = selectedStudyIds.length === 1 ? 'study' : 'studies'
-  const hasLibraryCard = Storage.getCurrentUser()?.libraryCard != null
+  const hasActiveResearcherStatus = Storage.getCurrentUser()?.libraryCard != null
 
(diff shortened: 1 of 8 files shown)


fboulnois (approved):
As soon as tests are fixed

rushtong (approved):
Assuming public facing support documentation includes similar updates, 👍🏽

otchet-broad (approved):
Code looks good.  Thank you @kevinmarete !

Agree with @rushtong that we probably want to make sure our ducks are in order for outreach before this lands.

fboulnois (approved):
Agreed with previous comments, otherwise lgtm

PR #3833, opened 2026-08-10
[DT-3948] Remove `dompurify` and add sanitization tests

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3948

### Summary
Removes the `dompurify` dependency and the two `DOMPurify.sanitize()` calls that used it, and replaces the implicit safety guarantee with explicit test coverage.

- Drops `dompurify` from `package.json` / `pnpm-lock.yaml`.
- Removes `DOMPurify.sanitize()` from `src/components/ScrollableMarkdownContainer.tsx` and `src/libs/TosService.tsx`. In both places `sanitize()` was applied to the **markdown source string** before it was handed to `<ReactMarkdown>` — the wrong layer. react-markdown builds a React element tree (it never touches `innerHTML`), escapes raw HTML by default since no `rehype-raw` is configured anywhere in this repo, and already rewrites `javascript:`/`vbscript:`/`data:` URLs to an empty `href`/absent `src`. The sanitize call was stripping HTML out of markdown *text*, not defending the DOM.
- Adds `test/components/ScrollableMarkdownContainer.sanitization.spec.tsx` exercising the **real** react-markdown — the existing spec mocks it out entirely, so it could not have caught a sanitization regression. 
- Adds a matching `getFormattedText sanitization` block to `test/libs/TosService.spec.tsx` against the ToS body, which is fetched from Sam.

These tests pin behavior we now depend on implicitly: they fail if `rehype-raw`, a custom `urlTransform`, or raw-HTML rendering is ever introduced, or if a react-markdown major bump changes the defaults.

----

- Verify all tests go green

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "@mui/x-date-pickers": "9.2.0",
     "@tanstack/react-query": "5.101.4",
     "dayjs": "1.11.21",
-    "dompurify": "3.4.13",
     "oidc-client-ts": "3.5.0",
     "query-string": "9.4.1",
     "react": "19.2.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -38,9 +38,6 @@ importers:
       dayjs:
         specifier: 1.11.21
         version: 1.11.21
-      dompurify:
-        specifier: 3.4.13
-        version: 3.4.13
       oidc-client-ts:
         specifier: 3.5.0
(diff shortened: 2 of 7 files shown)


kevinmarete (approved):

otchet-broad (approved):
👍

fboulnois (approved):
👍

PR #3839, opened 2026-08-12
[DT-3948] Stabilize ScrollableMarkdownContainer spec

rushtong:
### Addresses
Related to fixes from https://broadworkbench.atlassian.net/browse/DT-3948

### Summary

* Fixes a flaky assertion in `ScrollableMarkdownContainer.spec.tsx` ("shows no stale body and reports loading again when the markdown prop changes") that intermittently failed in CI with `expected last call with ['loaded']` but received `['loading']`.
* Root cause: `onLoadStateChange` fires from a passive `useEffect`, which React flushes after the DOM commit. The test's `waitFor` on the rendered markdown text can resolve on the MutationObserver tick right after the commit, before the effect runs, so the synchronous `toHaveBeenLastCalledWith('loaded')` assertion raced the effect flush.
* Fix: await the callback assertion itself with `waitFor` instead of asserting it synchronously. The same change is applied to the identical latent race at the end of the test (after the second document resolves).
* No component changes; test-only.

### Test plan

* `pnpm vitest run test/components/ScrollableMarkdownContainer.spec.tsx` — 9/9 passing across 20 consecutive runs.

----

- Verify all tests go green


kevinmarete (approved):

otchet-broad (approved):
👍

PR #3846, opened 2026-08-12
[DT-3975] Proxy ECM for RAS/eRA Commons linking

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3975

### Summary
- Phase 4 removes browser-held tokens, but the [BFF proxy covered only the DUOS API](https://broadworkbench.atlassian.net/browse/DT-3608) — the two ECM-bound calls in `AuthenticateNIH.ts` (`authorization-url`, `oauthcode`) would go out unauthenticated, breaking RAS/eRA Commons linking in BFF environments.
- Generalizes `server/src/proxy/apiProxy.ts` into a config-driven `upstreamProxy.ts`. Per-upstream config: prefix, upstream env var, unauthenticated-path allowlist, CSRF exemptions, and upstream-401 semantics. The security behavior — cookie/authorization/x-csrf-token stripping, session-token injection, CSRF on unsafe methods, response hardening — is deliberately **not** configurable, so future upstream proxies (3-J TDR, 3-K Bard) cannot opt out of it. The existing 101 `apiProxy` test cases pass unchanged against the refactor.
- Adds `server/src/proxy/ecmProxy.ts`: `/ecm-api/*` → new `DUOS_ECM_URL` env var, with no unauthenticated paths and no CSRF exemptions. One deliberate divergence from the DUOS API proxy: an upstream 401 passes through **without** destroying the BFF session — ECM authenticates via Sam, so its 401 can mean an audience mismatch or an ECM-side account problem, and ending the DUOS session over it would turn a broken linking flow into a broken everything.
- Registers the ECM proxy in `index.ts` under the same `bffEnabled` + `DUOS_DB_HOST` gates, conditional on `DUOS_ECM_URL`: a BFF deployment whose env predates the terra-helmfile change boots with linking broken and a startup warning naming the fix, rather than crash-looping (unlike `DUOS_API_URL`, ECM serves exactly one feature).

----

- Verify all tests go green


kevinmarete (approved):
Nice work 👍🏾

otchet-broad (commented):
Does this path log the user out if the session timeout is reached but a proxy request comes in?  And by log the user out, I mean perform the BFF session expiration.  I can see the path on the upstream 401, but it seems like generically, we shouldn't allow requests outside of the expiration time to remote systems.

rushtong:
> Does this path log the user out if the session timeout is reached but a proxy request comes in?

@otchet-broad - No, this path does not does not treat ECM 401s as definitive. 401s that come back from Consent **do** destroy the session. That code is in `apiProxy.ts`, where we set the configuration of `destroySessionOnUpstream401: true`. In theory, since both systems use the same proxy, both will be returning 401s at the same time.

rushtong:
> I can see the path on the upstream 401, but it seems like generically, we shouldn't allow requests outside of the expiration time to remote systems.

Here is a more detailed answer, but ~TLDR, an expired request never makes it to the downstream service:

Enforcement is proactive, just outside this diff:

1. Session timeout: the session store's get filters `expire > NOW()` (`pgStore.ts`), so once the 8-hour `maxAge` passes the session comes back empty and `ensureUpstreamAuth` 401s before `reply.from` runs ("returns 401 without calling the upstream…" tests in `apiProxy.test.ts`).
2. Token expiry inside a live session: the same `preHandler` checks `tokenExpiry` against a 60s window before forwarding and refreshes first; a terminal refresh failure destroys the session, clears the cookie, and returns 401 `session_expired`.

The upstream-401 path is a backstop for the upstream rejecting a token the BFF thought was fresh, not the primary gate.

otchet-broad (approved):
👍

PR #3852, opened 2026-08-13
[DT-3327] Remove the Action column from the Admin DAR table

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3327

### Summary

Removes the "Action" column from the Admin DAR table. `ACTIONS` is shared by every console, so admin is excluded via a new `excludedColumns` map rather than dropping the column outright — other consoles are unchanged.                                                                                                                                                                                                          

The page's `cancelCollection` and `openCollection` handlers go with it, since the actions cell was their only trigger. Admins can no longer cancel or reopen a DAR from this table.  

<img width="1785" height="904" alt="After" src="https://github.com/user-attachments/assets/5b11aca6-9c57-4db2-956f-a2b078e2c861" />

----

- Verify all tests go green

### DIFF
--- a/src/pages/AdminManageDarCollections.tsx
+++ b/src/pages/AdminManageDarCollections.tsx
@@ -4,7 +4,7 @@ import { Collections } from 'src/libs/ajax/Collections'
 import { Notifications, searchOnFilteredList, getSearchFilterFunctions, USER_ROLES } from 'src/libs/utils'
 import { Styles } from 'src/libs/theme'
 import { DarCollectionTable } from 'src/components/dar_collection_table/DarCollectionTable'
-import { cancelCollectionFn, consoleTypes, openCollectionFn, updateCollectionFn } from 'src/utils/DarCollectionUtils'
+import { consoleTypes } from 'src/utils/DarCollectionUtils'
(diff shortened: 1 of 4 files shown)


rushtong (approved):
This change has API consequences. If we are removing admin access to open/cancel elections so we will need to tighten up the roles allowed for these endpoints. Separately, this means that these actions are no longer allowable for Admins so the action calculations for Admins will need rework. We should file additional tickets for this.

Since Jonathan wrote this ticket, I think he is aware, but this also means admins can't perform clean up tasks when there are elections that might need the kind of attention he has provided in the past. 

Assuming all risks are understood, this looks good 👍🏽

fboulnois (approved):
👍 following @rushtong 's comments.

kevinmarete:
> This change has API consequences. If we are removing admin access to open/cancel elections so we will need to tighten up the roles allowed for these endpoints. Separately, this means that these actions are no longer allowable for Admins so the action calculations for Admins will need rework. We should file additional tickets for this.
> 
> Since Jonathan wrote this ticket, I think he is aware, but this also means admins can't perform clean up tasks when there are elections that might need the kind of attention he has provided in the past.
> 
> Assuming all risks are understood, this looks good 👍🏽

Follow-up ticket: https://broadworkbench.atlassian.net/browse/DT-3989

PR #3854, opened 2026-08-13
[DT-3579] Keep the data use when access management is chosen afterwards

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3579

### Summary

`onAccessTypeChange` rebuilt the consent group from five fields, dropping the data use along with everything else, then revalidated — so a data use entered before an access strategy was marked required. The radios are uncontrolled, so the selection still looked checked while the state behind it was gone.  

Now only the fields the new strategy stops showing are cleared: the data use for open access, the DAC when leaving controlled. That also stops the handler from discarding `consentGroupId`, `datasetId`, `name`, file types, and the NIH certification file on every access-management change. 

After

----

- Verify all tests go green


rushtong (approved):
👍🏽

fboulnois (approved):
👍

PR #3856, opened 2026-08-13
[DT-3977] Proxy Bard to restore identified metrics

rushtong:
### Addresses
Partially addresses [DT-3608](https://broadworkbench.atlassian.net/browse/DT-3608) (BFF Phase 3 — API proxy, story 3-K)
Fully addresses https://broadworkbench.atlassian.net/browse/DT-3977

### Summary
- Adds `server/src/proxy/bardProxy.ts`: `/bard-api/*` → new `DUOS_BARD_URL` env var, configured exactly like the ECM and TDR proxies — no unauthenticated paths, no CSRF exemptions (all three Bard calls are POSTs from an authenticated client, which holds an `X-CSRF-Token`), and an upstream 401 passes through **without** destroying the BFF session (Bard authenticates via Sam, so its 401 is not authoritative about the session).
- **Only the signed-in legs route through the proxy.** A signed-out `captureEvent` never carried a token, so the client keeps sending it directly to Bard (`bardApiUrl`), exactly as today. That split is what keeps the config free of an unauthenticated allowlist: a sessionless request through `/bard-api` is a client bug and 401s loudly rather than proxying an anonymous event. No silent 401s in the other direction either — signed-in requests carry a freshly-refreshed session token, and any Bard-side rejection passes through visibly instead of being swallowed.
- `DUOS_BARD_URL` joins the optional-proxies loop in `index.ts` (boot + warn + dark route when unset) and is staged in the same terra-helmfile branch as the ECM/TDR vars (dev: `https://terra-bard-dev.appspot.com`).
- Server-side only: un-inerting `Metrics.identify`/`syncProfile` and the identified `captureEvent` path is part of the Phase 4 client refactor (Epic 4), which should also flag the `distinct_id` population shift for the Phase 6 RUM regression checks.
- Also fixes `scripts/render-configs.sh --write_env` for all three optional proxies: the generated `.env.local` never included `DUOS_ECM_URL`/`DUOS_TDR_URL`/`DUOS_BARD_URL` (so a locally-run BFF booted with those routes dark), and a re-run dropped any hand-added lines. All three are now emitted with dev defaults and carried forward like `DUOS_API_URL`.

----

- Verify all tests go green

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -67,3 +67,6 @@ DUOS_ECM_URL=https://externalcreds.dsde-dev.broadinstitute.org
 # TDR, proxied at /tdr-api for snapshot enumeration on dataset pages.
 # Optional, same behavior as DUOS_ECM_URL when unset.
 DUOS_TDR_URL=https://jade.datarepo-dev.broadinstitute.org
+# Bard (Mixpanel gateway), proxied at /bard-api for identified usage metrics.
+# Optional, same behavior as DUOS_ECM_URL when unset.
+DUOS_BARD_URL=https://terra-bard-dev.appspot.com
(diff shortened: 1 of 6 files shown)


fboulnois (approved):
👍

otchet-broad (approved):
👍

kevinmarete (approved):

PR #3858, opened 2026-08-14
[DT-3982] Redesign DAC Data Access Requests table with MUI DataGrid

jlaw-codes:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3982](https://broadworkbench.atlassian.net/browse/DT-3982)

Security risk: no

### Summary

Replaces the bespoke SimpleTable-based `DarCollectionTable` with MUI X DataGrid to match the Data Library's table styling, and recovers the per-data-use-group visibility that the old expand-to-view nested table provided. Shared across all six consoles that use the component (Chair, Member, Admin, Signing Official ×2, Researcher).

- One grid row per (collection, data-use group) pair, with collection-level columns (DAR Code, Title, Researcher, Status, Actions, etc.) visually merged down across a collection's rows via DataGrid's native row spanning.
- Data Use, Datasets and Votes are per-group columns: a data-use pill, a per-group dataset count (hover for dataset names and identifiers), and Chair/Member-only Approve/Deny/Pending vote count pills (hover for member names).
- Pagination and sorting are self-managed so a collection's rows are never split across a page boundary, which would break row spanning. The footer counts collections, not grid rows.
- Action buttons resized to fit a single row height, and a leftover `!important` media-query rule that was silently overriding button sizing under 1024px is gone.

https://github.com/user-attachments/assets/6c736e5b-b13a-4c01-a7fc-b1874eaa97a7

### Where the group data comes from

The table no longer derives its own groups. It reads `dataUseGroups` off the collection summary, served by [consent#3024](https://github.com/DataBiosphere/consent/pull/3024), which removes roughly thirty requests per page view — `useDarCollectionDataUseBuckets` and the per-collection fetching behind it are deleted, along with the loading and error states they needed. **That PR must be merged and deployed before this one comes out of draft**, or Data Use, Datasets and Votes render empty on all six consoles.

Data use is rendered with `orderDataUseCodes`/`dataUseTooltip`, moved out of `datasetColumns.tsx` into `DataUseUtils` so this grid and the Data Library grid share one renderer rather than two copies. This matters: the earlier flat label showed `DS, NPOA, POP-F, POP-PD` where Data Library showed `DS (cancerophobia, diabetes mellitus), NPOA, POP-F, POP-PD` for the same dataset, losing the disease list and the OTH1/OTH2 distinction between a primary and secondary "Other".

### Also worth a reviewer's attention

- **The grid had no resolvable height.** Its container sets only `width` and `Styles.PAGE` sets no height, so MUI X would have rendered it at 0px. Now passes `autoHeight`, matching the Manage Users and SO Researcher Status grids.
- **No keyboard focus indicator.** The inline `sx` suppressed every cell and header outline without restoring a `:focus-visible` ring. Replaced with the same `DATAGRID_SX` shape those grids use, including the in-cell link rule needed because `index.css` resets `a { outline: 0 }` globally.
- Drops an unreachable `datasetCount` sort branch and ignores a persisted sort naming a non-sortable per-group column, which is possible since this reuses the old table's storage key.
- `BucketUtils` is untouched, so the DAR review page and DUC addendum are unaffected by this change.
- The DAR table and the DAR review page it links into present data use differently — codes on a chip here, one pill per term in labelled Permissions/Modifiers columns there. That predates this work; the table has simply moved from a bespoke format onto the Data Library's.

`pnpm run lint`, `pnpm run type-check` and `pnpm test` are all clean (3900 tests). Tests cover the hover tooltips, pagination boundaries and sort adjacency the acceptance criteria call for. Verified end to end in the browser against a local `consent` stack with a seeded database.

----

- Verify all tests go green


otchet-broad (approved):
Nice work!

fboulnois (approved):
👍

PR #3861, opened 2026-08-14
[DT-3609] Add the BFF session probe (BFF Phase 4, story 4-E)

rushtong:
> [!NOTE]
> Part of the BFF Phase 4 stack: #3861 → #3862 → #3863 → #3864 → #3869 (#3855 merged to develop). Bottom of the stack — this PR's diff is against develop; review bottom-up. New to the stack? Start with the [review guide & retrospective](https://github.com/DataBiosphere/duos-ui/blob/claude/epic-4-signout-redirect-cleanup/docs/plans/epic-4-stack-review-guide.md).

### Addresses
https://broadworkbench.atlassian.net/browse/DT-3609 (BFF Phase 4, story 4-E)

### Summary
Adds `src/libs/auth/session.ts` — the `GET /auth/me` session probe that becomes the async replacement for `Storage.userIsLogged()`:

- **Bounded promise cache (BFF branch only)**: the many components that ask "is the user logged in?" during a render share one `/auth/me` round-trip. A 200 or a 401 ("genuinely no session") is cached; transient failures (5xx, network) are not, so the next ask retries. Cached answers expire after 5 minutes — the fixed-lifetime session can die under a long-lived tab, and another tab can sign in/out on the shared cookie — and `revalidateSessionInfo()` (throttled to one probe per burst) lets the UI layer force a fresh look on tab focus (wired up in #3862). A stale "signed in" is UI-only exposure: the first real API call after expiry still 401s into the logout redirect.
- **Coexistence**: with `bffEnabled` off there are no `/auth/*` routes, so the probe falls back to the synchronous oidc-client-ts localStorage check — legacy environments keep working unchanged. That branch is deliberately **never cached**: the popup sign-in flow and the dev-only `/backgroundsignin` page mutate localStorage without a page load, and the read is free (config.json itself is promise-cached).
- **Fail-safe**: 401, 502, network failures, and config-lookup failures all collapse to `{ authenticated: false }` rather than crashing render paths.

**Deviation from the Phase 4 text (4-E "remove token keys from storage.ts")**: the legacy token storage functions (`setOidcUser`/`getOidcUser`/sync `userIsLogged`) are deliberately retained during coexistence — the legacy sign-in flow, the fetch-adapter 401 metric, and the dev-only `/backgroundsignin` page still need them until the Phase 6 cutover cleanup (6-J), consistent with story 4-B's own rationale for leaving legacy storage untouched.

**Stories 4-I and 4-J ship no code anywhere in this stack**: they are delivered by construction through #3855's config-level gating — every ajax module's `${apiUrl}/api/...` URL and the RAS/ECM callback ride the `/duos-api` / `/ecm-api` / `/tdr-api` / `/bard-api` proxy prefixes in BFF mode, with `Authorization` attached server-side. The per-module `authOpts` deletion the phase sketches would regress legacy parity and is Phase 6 cleanup.

- **Server contract fix (first server change in the stack)**: `/auth/me` reports an upstream **401 or 404** from `/api/user/me` as `authenticated: true` with no `user`, so #3863's registration bootstrap is reachable. The DUOS API conflates "no profile for this email" with "bad token" — its auth filter (`DuosUserAuthenticator`) turns the user lookup's `NotFoundException` into a plain 401 (confirmed against a live consent stack during manual testing; a consent regression from DT-3788 — filed as [DT-3997](https://broadworkbench.atlassian.net/browse/DT-3997), in the current sprint), so treating that as a dead session destroyed every new user's session on first probe. Safe because the token is refreshed just before forwarding (below), and a genuinely revoked token also fails registration, which signs the session out. 502 remains for genuine upstream failures; `/auth/me` itself still 401s when there is no session.
- **Legacy identity**: the legacy probe branch reports the stored `CurrentUser` as the session identity, so the identity-reconciliation check in #3863 (session user vs stored user) is trivially satisfied in legacy mode.

### Testing
New `test/libs/auth/session.spec.ts` (10 cases: payload passthrough, 401/502/network/config failures, cache behavior, legacy fallback that never calls `/auth/me`). Nothing consumes the module yet (that's story 4-F, next in the stack), so `bffEnabled: false` behavior is unchanged by construction.

----

- Verify all tests go green

[DT-3997]: https://broadworkbench.atlassian.net/browse/DT-3997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/auth/me.ts
+++ b/server/src/auth/me.ts
@@ -42,6 +42,16 @@ export async function getMe(request: FastifyRequest, reply: FastifyReply): Promi
     return
   }
 
+  if (res.status === 404) {
+    // Authenticated but not yet registered: the session is valid, the DUOS
+    // profile just doesn't exist yet. Report authenticated with no user so
+    // the client can run its post-sign-in registration bootstrap — collapsing
+    // this into a failure would make every new user look signed out and leave
+    // registration unreachable.
+    reply.send({ authenticated: true, idp: request.session.idp })
+    return
+  }
+
   if (!res.ok) {
     // A non-401 failure (5xx, upstream outage) says nothing about whether the
     // token itself is still valid — don't destroy the session or parse an
--- a/server/src/auth/me.ts
+++ b/server/src/auth/me.ts
@@ -61,20 +61,20 @@ export async function getMe(request: FastifyRequest, reply: FastifyReply): Promi
     return
   }
 
-  if (res.status === 401) {
-    // The upstream API is the source of truth for token validity — a session
-    // whose access token it now rejects (expired, revoked) is dead weight.
-    await request.session.destroy()
-    reply.clearCookie('sessionId').status(401).send({ authenticated: false })
-    return
-  }
-
-  if (res.status === 404) {
-    // Authenticated but not yet registered: the session is valid, the DUOS
-    // profile just doesn't exist yet. Report authenticated with no user so
-    // the client can run its post-sign-in registration bootstrap — collapsing
-    // this into a failure would make every new user look signed out and leave
+  if (res.status === 401 || res.status === 404) {
+    // Authenticated but (almost certainly) not yet registered. The DUOS API
+    // conflates "bad token" with "no DUOS profile for this email": its auth
+    // filter (DuosUserAuthenticator) turns the user lookup's NotFoundException
+    // into an empty principal, which Dropwizard reports as 401 — the same
+    // status a rejected token gets. Treating that 401 as a dead session
+    // destroyed every brand-new user's session on their first probe and made
     // registration unreachable.
+    //
+    // Reporting the session honestly here is safe: the access token was
+    // refreshed just above when anywhere near expiry, so a 401 from a fresh
+    // token means "no profile", and in the rare revoked-token case the
+    // client's registration attempt fails too, which signs the session out.
+    // (404 kept for symmetry should the upstream ever disambiguate.)
     reply.send({ authenticated: true, idp: request.session.idp })
     return
   }
--- a/server/src/auth/me.ts
+++ b/server/src/auth/me.ts
@@ -11,6 +11,12 @@ const UPSTREAM_TIMEOUT_MS = 5000
  * tokens themselves, which stay server-side in the session.
  */
 export async function getMe(request: FastifyRequest, reply: FastifyReply): Promise<void> {
+  // The answer is per-session and now gates the whole SPA: no intermediary
+  // (or the browser's heuristic cache) may replay one user's profile to
+  // another, or a stale answer to the same user.
+  reply.header('cache-control', 'no-store')
+  reply.header('vary', 'Cookie')
+
   if (!request.session.accessToken) {
     reply.status(401).send({ authenticated: false })
     return
--- a/server/src/auth/me.ts
+++ b/server/src/auth/me.ts
@@ -1,5 +1,7 @@
 import type { FastifyReply, FastifyRequest } from 'fastify'
 import { requireEnv } from './oidcClient.js'
+import { RefreshFailedError, refreshAccessToken } from './refresh.js'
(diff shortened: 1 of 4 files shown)


kevinmarete (commented):
Looks good, one comment:

**P2 – Successful and unauthenticated session results can remain stale for the entire SPA lifetime.** `sessionPromise` permanently caches both 200 and 401 responses. The assumption that BFF auth changes only after full-page navigation is incomplete: the fixed eight-hour session can expire, and another tab can clear the shared session cookie. Subsequent route probes will still return the resolved cached value, potentially leaving protected UI and previously loaded data visible after session termination—or leaving another tab signed out after authentication. Consider a bounded TTL, invalidation on tab visibility/focus, or caching only concurrent in-flight requests. [session.ts at lines 30–70](https://github.com/DataBiosphere/duos-ui/blob/0bc9721641ac38a1795e185365f4e534a357d8ea/src/libs/auth/session.ts#L30-L70)

rushtong:
> **P2 – Successful and unauthenticated session results can remain stale for the entire SPA lifetime.**

Good catch — the "auth only changes across full-page navigations" assumption did miss both session expiry under a long-lived tab and cross-tab sign-in/out on the shared cookie.

Addressed in 97a8a3ac (this PR) + f3bf4811 (#3862, where the hook lives):

- Cached answers now expire after a 5-minute TTL, and transient failures were already uncached, so only real answers (200 / 401) are ever held.
- New `revalidateSessionInfo()` — throttled so a burst of callers shares one probe — is wired to window focus / visibilitychange in #3862, which covers the cross-tab case as soon as the user returns to the tab.

Worth noting the blast radius was UI-only: a stale "signed in" can't fetch anything — the first real API call after expiry 401s into the existing logout redirect. But agreed the UI shouldn't lie either; hence the TTL + focus revalidation rather than in-flight-only caching (which would have re-probed on literally every navigation for no gain in the idle case).

kevinmarete (approved):

kevinmarete (approved):
👍🏾

kevinmarete (approved):

fboulnois (approved):
👍

otchet-broad (approved):
👍

rushtong:
Re-requesting review due to feedback from #3863 - fyi @kevinmarete, @fboulnois, @otchet-broad

kevinmarete (approved):

server/src/auth/me.ts
```diff
@@ -34,11 +67,21 @@ export async function getMe(request: FastifyRequest, reply: FastifyReply): Promi
...
   }
 
-  if (res.status === 401) {
-    // The upstream API is the source of truth for token validity — a session
-    // whose access token it now rejects (expired, revoked) is dead weight.
-    await request.session.destroy()
-    reply.clearCookie('sessionId').status(401).send({ authenticated: false })
+  if (res.status === 401 || res.status === 404) {
+    // Authenticated but (almost certainly) not yet registered. The DUOS API
+    // conflates "bad token" with "no DUOS profile for this email": its auth
```
otchet-broad, line 72: @rushtong do we have a ticket to engineer a solution to this conflation?

otchet-broad (approved):

server/src/auth/me.ts
rushtong: This could be partly out of date with the fixes that went into https://github.com/DataBiosphere/consent/pull/3021

While testing that code for the new-user flow, I found an additional bug (https://broadworkbench.atlassian.net/browse/DT-4011) and created a follow up to that one: https://broadworkbench.atlassian.net/browse/DT-4012 so I'll bundle this fix into that one because they are very closely related, i.e. handling 200/401/404/409 responses from Consent.

PR #3862, opened 2026-08-14
[DT-3609] Read auth state from the session probe in components (BFF Phase 4, story 4-F)

rushtong:
> [!NOTE]
> Part of the BFF Phase 4 stack: #3861 → #3862 → #3863 → #3864 → #3869 (#3855 merged to develop). This PR's diff is against #3861; review bottom-up. New to the stack? Start with the [review guide & retrospective](https://github.com/DataBiosphere/duos-ui/blob/claude/epic-4-signout-redirect-cleanup/docs/plans/epic-4-stack-review-guide.md).

### Addresses
https://broadworkbench.atlassian.net/browse/DT-3609 (BFF Phase 4, story 4-F)

### Summary
Migrates every component call site off the synchronous `Storage.userIsLogged()` and onto the session probe from #3861:

- **New `src/hooks/useSession.ts`**: `useSessionInfo()` / `useUserIsLogged()` return `undefined` while the probe is in flight so render paths can distinguish "still checking" from "signed out". The hooks re-probe on every navigation and revalidate on window focus / visibilitychange (throttled in session.ts so many mounted hooks share one probe) — a background tab can outlive its session or watch another tab sign in/out on the shared cookie.
- **`Authenticated.tsx`** renders nothing while the probe is in flight rather than bouncing an actually-signed-in user to `/?redirectTo=…` before the answer arrives.
- **`App.tsx`, `DuosHeader`, `SupportRequestModal`, `TermsOfService`** read the hook (`?? false`). The support modal re-derives its name/email prefill when the probe resolves (adjust-state-during-render, not an effect).
- **Sign-out call sites** (`DuosHeader`, both ToS pages) keep their SPA `navigate()` / `queryClient.clear()`: legacy `Auth.signOut()` only clears local state (`OidcBroker.signOut()` does not navigate), so the caller's navigation is what actually moves a legacy user off the page — and it also triggers the hook's re-probe that refreshes the header. In BFF mode `Auth.signOut()` follows up with a full-page reload to the same destination (`DuosHeader` passes `/home` through), which supersedes the SPA navigation harmlessly. The hook re-probes on every navigation — `location` identity, not just pathname — so signing out while already on `/home` still refreshes auth state.

The synchronous check deliberately survives in `auth.ts` (legacy fallback), `Metrics.ts` (legacy identified events), and `storage.ts` until the Epic 6 cutover cleanup — see the deviation note on #3861.

### Testing
Specs updated in lockstep: `Authenticated` (new in-flight-renders-nothing case), `AppRoutes` (session probe mocked; the legacy-DAC-console block now flips the hook, not the storage spy), `DuosHeader`, `SupportRequestModal` (jsdom + browser), `TermsOfService` (new in-flight reject-button case), `TermsOfServiceAcceptance`. With `bffEnabled: false` the probe resolves from legacy storage, so behavior is unchanged.

----

- Verify all tests go green

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,6 +12,7 @@ import { useNavigate, useLocation } from 'react-router'
 import { Storage } from 'src/libs/storage'
 import AppRoutes from 'src/routing/AppRoutes'
 import { Notifications, setUserRoleStatuses } from 'src/libs/utils'
+import { useUserIsLogged } from 'src/hooks/useSession'
 import { extractError } from 'src/utils/ErrorUtils'
 import { Spinner } from 'src/components/Spinner'
 
@@ -27,11 +28,12 @@ const queryClient = new QueryClient({
 })
 
 function App() {
-  const [isLoggedIn, setIsLoggedIn] = useState(false)
   const [env, setEnv] = useState('')
   const navigate = useNavigate()
   const location = useLocation()
   const [isLoading, setIsLoading] = useState(false)
+  // Auth state comes from the BFF session probe (GET /auth/me), not localStorage.
+  const isLoggedIn = useUserIsLogged() ?? false
 
   useEffect(() => {
     const setEnvironment = async () => {
@@ -40,14 +42,6 @@ function App() {
       Storage.setEnv(environment)
     }
     setEnvironment()
-  })
-
-  useEffect(() => {
-    const setUserIsLogged = () => {
-      const isLogged = Storage.userIsLogged()
-      setIsLoggedIn(isLogged)
-    }
-    setUserIsLogged()
   })
 
   /**
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -42,7 +42,9 @@ function App() {
       Storage.setEnv(environment)
     }
     setEnvironment()
-  })
+    // The environment never changes within a page load — look it up once on
+    // mount instead of on every render.
+  }, [])
 
   /**
      * Check for RAS Authentication URL params. If we have a code and state, we will call ECM APIs to get redirect
--- a/src/components/DuosHeader.tsx
+++ b/src/components/DuosHeader.tsx
@@ -18,7 +18,7 @@ import { Notification } from './Notification'
 import { useLocation, useNavigate } from 'react-router'
 import { DuosUser } from 'src/types/model'
 import { useNavigationState } from 'src/contexts/NavigationStateContext'
-import { useQueryClient } from '@tanstack/react-query'
+import { useUserIsLogged } from 'src/hooks/useSession'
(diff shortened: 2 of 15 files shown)


kevinmarete (commented):
One Finding: 

**P1 – The BFF session probe conflates authentication with having an existing DUOS user profile.**

After a successful BFF login, `/auth/me` calls `/api/user/me`. For a newly authenticated user who has not registered yet, that lookup is expected to fail—the legacy sign-in flow catches this and calls `registerUser()`. The BFF handler converts any non-401 upstream failure into a 502, and `getSessionInfo()` converts that into `authenticated: false`.

As a result, this PR’s `useUserIsLogged()` reports the new user as signed out, and #3863’s post-sign-in registration bootstrap never runs because it requires `isLoggedIn === true`.

Could we separate “valid BFF session” from “DUOS profile exists”? For example, `/auth/me` could return `authenticated: true` with an absent user for the expected not-registered response, allowing the post-sign-in bootstrap to fetch/register the profile. We should add an integration test covering: successful OAuth callback → no existing DUOS user → registration flow starts.

Relatedly, please ensure revalidation reconciles the returned user with `Storage.getCurrentUser()`. Otherwise, signing into a different account in another tab can leave the current tab using the previous user’s locally stored roles/profile.

rushtong:
> **P1 – The BFF session probe conflates authentication with having an existing DUOS user profile.**

Confirmed — `/auth/me` turned an unregistered user's expected upstream 404 into a 502, the probe read that as signed out, and the registration bootstrap was unreachable. Fixed across two layers (the contract lives below this PR, the consumer above it):

- **#3861 (48d0cff8)**: `/auth/me` now reports the valid session honestly for the upstream-404 case — `authenticated: true` with no `user` — keeping 502 for genuine upstream failures and 401 for rejected tokens. Server test added for the new shape.
- **#3863 (45ed86c9)**: `App` keys its bootstrap off `useSessionInfo`, so an authenticated session with no profile starts `completeSignIn`, whose `getMe` failure path reaches `registerUser()`. App-level tests cover: unregistered session → registration path starts, plus mismatch/no-op cases.

On the revalidation/reconciliation point: also addressed in #3863 — the bootstrap now fires on identity mismatch between the session's user and `Storage.getCurrentUser()` (and re-arms if focus revalidation detects an account switch after the fact), overwriting the stale profile, clearing the query cache, and re-running the ToS gate. The legacy probe branch reports the stored user as the session identity (#3861), so the check can't mis-fire pre-cutover.

kevinmarete (approved):
👍🏾

kevinmarete (approved):

fboulnois (approved):
👍

PR #3849, opened 2026-08-12
[DT-3609] Rewrite Auth for the BFF session flow (BFF Phase 4, story 4-A)

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3609 (BFF Phase 4, story 4-A)

See also this companion configuration PR: https://github.com/broadinstitute/terra-helmfile/pull/6483

### Summary
Rewrites `Auth.signIn()`, `signOut()`, and `initialize()` for the BFF session flow, gated on the new `Config.isBffEnabled()` (the `bffEnabled` key in `config.json` — the same key the server uses to register the `/auth/*` routes, so both sides agree; missing key = legacy flow).

- **BFF mode**: `signIn` POSTs `/auth/login` and full-page-redirects to the B2C authorization URL (no `idp` parameter — provider selection happens on the B2C page). `signOut` fetches `/auth/csrf-token`, POSTs `/auth/logout` with `X-CSRF-Token` (the server CSRF-guards that route), clears local storage, and redirects. `initialize` purges legacy `oidc-client-ts` localStorage keys left from before the cutover (story 4-B) and skips the OidcBroker. New `Auth.isAuthenticated()` calls `GET /auth/me`.
- **Legacy mode**: the oidc-client-ts popup flow is unchanged, so pre-cutover environments are unaffected.
- Full-page navigations route through an exported `Redirect.to()` seam because jsdom's `window.location` is unforgeable in tests.

## Testing
Note that local testing with `"bffEnabled": true` is not functional yet. Setting it to `"bffEnabled": false` will allow the application to bypass all BFF code and ensure that there are no regressions until phase 4 is complete and we can begin the cutover.

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3863, opened 2026-08-14
[DT-3609] Complete BFF sign-in after the OAuth redirect (BFF Phase 4, story 4-G)

rushtong:
> [!NOTE]
> Part of the BFF Phase 4 stack: #3861 → #3862 → #3863 → #3864 → #3869 (#3855 merged to develop). This PR's diff is against #3862; review bottom-up. New to the stack? Start with the [review guide & retrospective](https://github.com/DataBiosphere/duos-ui/blob/claude/epic-4-signout-redirect-cleanup/docs/plans/epic-4-stack-review-guide.md).

### Addresses
https://broadworkbench.atlassian.net/browse/DT-3609 (BFF Phase 4, story 4-G)

### Summary
Sign-in through the BFF is a full-page redirect (`POST /auth/login` → B2C → `/auth/callback` → back here), so the popup `onSuccess` handler that used to fetch/register the user can no longer live in `SignInButton`:

- **New `src/libs/auth/postSignIn.ts`**: `completeSignIn()` owns everything the popup callbacks used to — `User.getMe()`, persist + role statuses, query-cache reset (pre-sign-in cached results were built with the anonymous user's visibility clauses), sign-in metrics, the ToS gate, registration with 409-means-already-registered, and the AzureB2C-error-from-Sam path (which signs the session back out).
- **`App.tsx`** detects the post-redirect state — a session exists but `CurrentUser` is empty — and runs `completeSignIn` exactly once, keeping the routes behind the spinner until it resolves.
- **`SignInButton.tsx`** shrinks to a single button calling `Auth.signIn(redirectTo?)` — no provider selection (that's the B2C page's job) and no BFF/legacy branching in the component (that already lives inside `Auth.signIn()` from #3849). `ServiceStatus` health gating, the tooltip, and `?redirectTo=` handling are preserved.

- **Bootstrap triggers**: App keys the bootstrap off `useSessionInfo` — it runs for a session with no local user (post-callback), for an authenticated session with no DUOS profile (unregistered — reaches `registerUser()`), and on identity mismatch between the session user and stored `CurrentUser` (cross-tab account switch / re-login after expiry), re-arming if the mismatch appears after completion. A non-409 registration failure signs the session out rather than unlocking the routes with an empty user; `errorStatus()` unwraps `Error.cause` to match the fetch adapter's real wrapped error shape.

- **Session reconciliation (`src/hooks/useSessionReconciler.ts` + [ADR-011](../blob/HEAD/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md))**: every fresh `/auth/me` result is classified once — hydrate (same identity; auth-equivalence covers role `{name, dacId}` pairs, `enabled`/`adminEnabled`, ToS), full bootstrap (no local identity / different user / no profile), or — per ADR-011's scope decision that cross-tab account switching is unsupported — a conflict under an in-flight bootstrap cancels the run (`completeSignIn` polls a cancellation token before every side effect) and hard-reloads the tab. Identity-bearing UI (routes AND header) hides while reconciling.

**Reviewer note (behavior change)**: errors from the moved registration/AzureB2C paths now surface via `Notifications.showError` instead of the button's inline `Alert`, since the button is no longer on screen when they occur.

### Testing
New `test/libs/auth/postSignIn.spec.ts` (happy path, redirect preservation, ToS gate, 409, registration failure, AzureB2C error). `SignInButton.spec.tsx` rewritten for the single-button contract (redirect capture from `/`, `/home`, deep links, `?redirectTo=` preference, error alert). Legacy flow: with `bffEnabled: false`, `Auth.signIn()` still runs the OidcBroker popup, and `completeSignIn` never triggers because the legacy flow persists `CurrentUser` itself before any reload.

----

- Verify all tests go green

### DIFF
--- a/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md
+++ b/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md
@@ -40,7 +40,9 @@ multi-account applications do — by discarding the stale tab's state:
   the reconciliation spinner. This needs no conflict machinery: nothing else
   is in flight.
 - An identity conflict that appears **under an in-flight bootstrap** cancels
-  that run and **hard-reloads the tab** (`Redirect.to(location.href)`). The
+  that run and **hard-reloads the tab** (`Redirect.reload()` — a true
+  `location.reload()`, because assigning `location.href` to a URL with a
+  `#fragment` is a same-document navigation that reloads nothing). The
   fresh page load reconciles from scratch; the reload kills all in-flight
   JavaScript, so no obsolete run can persist a user, emit metrics, navigate,
   or destroy the session after being overtaken.
@@ -72,6 +74,37 @@ withheld for the one render until the fresh profile commits.
   repeatedly), reload more than once. Each reload converges on the current
   session, so there is no livelock — the fresh load either matches storage
   (hydrate) or bootstraps in place.
+
+## Accepted risks
+
+Two residual windows, surfaced by the cross-tab review on #3863
+(2026-08-17), are accepted rather than engineered around:
+
+1. **A registration run can adopt a different unregistered identity.**
+   `/auth/me` cannot distinguish unregistered accounts — every one of them
+   reports `authenticated: true` with no user, so a registration run's
+   target identity is "none". If, during the seconds a registration run is
+   in flight, another tab signs the shared cookie into a *different*
+   unregistered account, a focus revalidation in the first tab joins the
+   run (same "none" target) and the in-flight `registerUser()` registers
+   whoever now owns the cookie. The tab converges on that account rather
+   than reloading. Detecting this needs a stable per-session identifier for
+   unregistered sessions from `/auth/me`; the join rule cannot be tightened
+   without also reloading on the legitimate joins (StrictMode double
+   invocation, focus revalidation) that the run-once machinery exists for.
+   The window is seconds wide and requires two different unregistered
+   accounts switching mid-registration.
+
+2. **A transient upstream 401 for a registered user drives one
+   registration POST.** `/auth/me` reports an upstream 401 as "valid
+   session, no profile" (the upstream conflates the two — see #3861). For
+   a *registered* user this shape only appears when the upstream 401s a
+   freshly refreshed token: a revoked token, or upstream auth outage. The
+   bootstrap then attempts `registerUser()`; the expected `409 Conflict`
+   recovers into a normal sign-in, and any other failure signs the session
+   out — which is the correct end state for a genuinely revoked token. The
+   cost of the window is one spurious POST and, during an upstream auth
+   outage, a sign-out instead of an error banner.
 
 ## Alternative considered
 
--- a/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md
+++ b/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md
@@ -1,0 +1,83 @@
+# ADR-011 — One identity per browser: cross-tab account switching reloads the stale tab
+
+**Status:** Accepted (2026-08-15) &nbsp;|&nbsp; **Phase:** 4, stories 4-E/4-F/4-G
+**Related:** [ADR-004](ADR-004-api-proxy-layer.md) (the session cookie all tabs share)
+**Implemented by:** `src/hooks/useSessionReconciler.ts` (duos-ui #3863)
+
+---
+
+## Context
+
+With the BFF, authentication state lives in one server-side session addressed
+by one cookie, shared by every tab of the browser profile. "Two different
+users in two different tabs" is therefore not a state the system can be in:
+every API call from every tab acts as whoever authenticated most recently.
+The only real phenomenon is **serial account switching with a stale tab** —
+tab A still *displaying* user X while the shared session has become user Y
+(or has been signed out, or belongs to a not-yet-registered account).
+
+The client reconciles the session identity against the locally stored profile
+(`CurrentUser`, also shared across tabs) whenever the session probe
+(`GET /auth/me`) returns a fresh answer — on navigation, focus/visibility
+changes, and cache expiry. Review of the reconciliation logic (ten findings
+across four rounds) kept surfacing new interleavings of the same underlying
+problem: making a stale tab converge **gracefully, in place**, requires
+supersede semantics for in-flight bootstraps, provenance tracking for which
+run wrote which identity to shared storage, and a distinguishable identity
+for unregistered sessions (which `/auth/me` cannot provide — every
+unregistered account looks the same). Each mechanism fixed one interleaving
+and exposed another.
+
+## Decision
+
+**Concurrent different-user usage across tabs is unsupported.** The
+reconciler handles a detected identity conflict the way mainstream
+multi-account applications do — by discarding the stale tab's state:
+
+- An identity conflict that appears **while no reconciliation is running**
+  (a fresh probe naming a different user, or reporting no profile over a
+  stored identity) runs the normal post-sign-in bootstrap in place, behind
+  the reconciliation spinner. This needs no conflict machinery: nothing else
+  is in flight.
+- An identity conflict that appears **under an in-flight bootstrap** cancels
+  that run and **hard-reloads the tab** (`Redirect.to(location.href)`). The
+  fresh page load reconciles from scratch; the reload kills all in-flight
+  JavaScript, so no obsolete run can persist a user, emit metrics, navigate,
+  or destroy the session after being overtaken.
+- Within one tab, `localStorage` consequently has a **single writer** during
+  a reconciliation run. The only join rule that remains is the one the
+  single-writer assumption makes sound: a probe naming the user that a
+  registration run (target = no identity) itself just persisted joins that
+  run; anything else conflicting reloads.
+- A probe turning **unauthenticated**, and hook **unmount**, cancel the
+  in-flight run via the same token (`completeSignIn` polls it before every
+  side-effecting step). These are same-user events (sign-out in another tab,
+  session expiry) and remain fully supported.
+
+What remains supported and graceful: the same user in many tabs, including
+sign-out from any of them; same-user profile changes (roles, DAC
+assignments, ToS state) hydrating on revalidation — with identity-bearing UI
+withheld for the one render until the fresh profile commits.
+
+## Consequences
+
+- The reconciler shrinks: no supersede generations, no storage-provenance
+  tracking, no unregistered-identity disambiguation. The review surface for
+  cross-tab interleavings closes — conflicts have exactly one behavior.
+- A user who switches accounts in another tab loses transient state (e.g. a
+  half-typed form) in stale tabs when they next touch them: the tab reloads.
+  This is the accepted cost, identical to the trade made by mainstream
+  multi-account web applications.
+- The reload can, in a pathological ping-pong (two tabs switching accounts
+  repeatedly), reload more than once. Each reload converges on the current
+  session, so there is no livelock — the fresh load either matches storage
+  (hydrate) or bootstraps in place.
+
+## Alternative considered
+
+Full in-place reconciliation: record the identity each run commits, expose a
+stable session identity for unregistered accounts from `/auth/me`, and
+supersede-with-cancellation across every interleaving. Rejected as
+disproportionate: it re-implements a session manager in the client to
+polish an unsupported workflow, and each review round demonstrated the
+interleaving space is larger than the value of handling it.
--- a/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md
+++ b/docs/plans/bff_adrs/ADR-011-single-identity-per-browser.md
@@ -95,16 +95,18 @@ Two residual windows, surfaced by the cross-tab review on #3863
    The window is seconds wide and requires two different unregistered
    accounts switching mid-registration.
 
-2. **A transient upstream 401 for a registered user drives one
+2. **An upstream 401 on a session's first-ever probe drives one
    registration POST.** `/auth/me` reports an upstream 401 as "valid
-   session, no profile" (the upstream conflates the two — see #3861). For
-   a *registered* user this shape only appears when the upstream 401s a
-   freshly refreshed token: a revoked token, or upstream auth outage. The
-   bootstrap then attempts `registerUser()`; the expected `409 Conflict`
-   recovers into a normal sign-in, and any other failure signs the session
-   out — which is the correct end state for a genuinely revoked token. The
-   cost of the window is one spurious POST and, during an upstream auth
-   outage, a sign-out instead of an error banner.
+   session, no profile" (the upstream conflates the two — see #3861). The
+   session's `profileSeen` flag narrows the exposure: once a session has
+   served a profile, a later 401/404 is treated as a terminal token
+   verdict (revoked mid-lifetime, disabled account) and destroys the
+   session instead of reading as "unregistered". What remains is the
+   first-contact case — a token revoked between login and the session's
+   first probe. The bootstrap then attempts `registerUser()`; the expected
+   `409 Conflict` recovers into a normal sign-in, and any other failure
+   signs the session out — the correct end state for a revoked token. The
+   cost of the residual window is one spurious POST.
 
 ## Alternative considered
 
--- a/server/src/auth/me.ts
+++ b/server/src/auth/me.ts
@@ -5,6 +5,38 @@ import { REFRESH_WINDOW_SECONDS, RefreshFailedError, refreshAccessToken } from '
 const UPSTREAM_TIMEOUT_MS = 5000
 
 /**
+ * Answers an upstream 401/404. The DUOS API conflates "bad token" with "no
+ * DUOS profile for this email" (both 401 — DuosUserAuthenticator turns the
+ * lookup's NotFoundException into an empty principal), so the session's
+ * `profileSeen` flag is the disambiguator this endpoint controls.
+ */
+async function answerNoProfile(request: FastifyRequest, reply: FastifyReply): Promise<void> {
+  if (request.session.profileSeen) {
+    // This session has served a profile before, so "no profile" cannot be
+    // the explanation. The upstream is rejecting the token itself — revoked
+    // mid-lifetime (before refresh-before-forward would touch it) or the
+    // account was disabled. The terminal 401 is the honest answer; leaving
+    // the session alive would send the client into re-registering an
+    // existing user.
+    try {
+      await request.session.destroy()
+    }
+    catch (err: unknown) {
+      request.log.error({ err }, '[auth] upstream rejected a profile-seen session but it could not be destroyed — returning 401 anyway')
+    }
+    reply.clearCookie('sessionId').status(401).send({ authenticated: false })
+    return
+  }
+  // Never seen a profile: authenticated but not yet registered. Treating
+  // this 401 as a dead session destroyed every brand-new user's session on
+  // their first probe and made registration unreachable. In the rare case
+  // of a token revoked before first contact, the client's registration
+  // attempt fails too, which signs the session out. (404 kept deliberately:
+  // DT-3997 restores the upstream's 404 for unregistered users.)
+  reply.send({ authenticated: true, idp: request.session.idp })
+}
+
+/**
  * Confirms the user is authenticated against the upstream Consent API.
  * Forwards the upstream user profile and the active sub-provider — never the
  * tokens themselves, which stay server-side in the session.
@@ -67,34 +99,7 @@ export async function getMe(request: FastifyRequest, reply: FastifyReply): Promi
   }
 
   if (res.status === 401 || res.status === 404) {
-    // The DUOS API conflates "bad token" with "no DUOS profile for this
-    // email": its auth filter (DuosUserAuthenticator) turns the user lookup's
-    // NotFoundException into an empty principal, which Dropwizard reports as
-    // 401 — the same status a rejected token gets. `profileSeen` is the
-    // disambiguator this endpoint controls:
-    if (request.session.profileSeen) {
-      // This session has served a profile before, so "no profile" cannot be
-      // the explanation. The upstream is rejecting the token itself — revoked
-      // mid-lifetime (before refresh-before-forward would touch it) or the
-      // account was disabled. The terminal 401 is the honest answer; leaving
-      // the session alive would send the client into re-registering an
-      // existing user.
-      try {
-        await request.session.destroy()
-      }
-      catch (err: unknown) {
-        request.log.error({ err }, '[auth] upstream rejected a profile-seen session but it could not be destroyed — returning 401 anyway')
-      }
-      reply.clearCookie('sessionId').status(401).send({ authenticated: false })
-      return
-    }
-    // Never seen a profile: authenticated but not yet registered. Treating
-    // this 401 as a dead session destroyed every brand-new user's session on
-    // their first probe and made registration unreachable. In the rare case
-    // of a token revoked before first contact, the client's registration
-    // attempt fails too, which signs the session out. (404 kept deliberately:
-    // DT-3997 restores the upstream's 404 for unregistered users.)
-    reply.send({ authenticated: true, idp: request.session.idp })
+    await answerNoProfile(request, reply)
     return
   }
 
--- a/server/src/auth/me.ts
+++ b/server/src/auth/me.ts
@@ -1,7 +1,6 @@
 import type { FastifyReply, FastifyRequest } from 'fastify'
 import { requireEnv } from './oidcClient.js'
-import { RefreshFailedError, refreshAccessToken } from './refresh.js'
-import { REFRESH_WINDOW_SECONDS } from '../proxy/upstreamProxy.js'
+import { REFRESH_WINDOW_SECONDS, RefreshFailedError, refreshAccessToken } from './refresh.js'
 
 const UPSTREAM_TIMEOUT_MS = 5000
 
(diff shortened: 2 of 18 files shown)


kevinmarete (commented):
Some findings:

- **P1 — New-user bootstrap is unreachable.** `App` starts `completeSignIn()` only when `/auth/me` reports authenticated. But `/auth/me` calls `/api/user/me`; an unregistered user’s expected failure becomes a 502 and then `authenticated: false`. Therefore this PR never reaches `registerUser()` for the users it is intended to register.

- **P1 — Failed registration unlocks the application with an empty user.** Non-409 registration failures only show a notification and resolve. `App` then marks bootstrap complete and renders authenticated routes, while `CurrentUser` remains empty. The bootstrap should reject, remain retryable, or sign out instead.

- **P2 — The production 409 error shape does not match the tests.** `errorStatus()` checks `error.response.status` or `error.status`, but `fetchAdapter` wraps HTTP errors inside `Error.cause`. A real `registerUser()` 409 therefore won’t enter the recovery branch tested with `{ status: 409 }`.

- **P2 — Stale identities bypass bootstrap.** It runs only when stored `userId === 0`. If a session expires or another tab signs in as a different account, the old nonzero `CurrentUser` remains and its profile/roles continue being used.

rushtong:
All four findings confirmed and fixed in 45ed86c9 (this PR), with the enabling server contract change in #3861 (48d0cff8):

> **P1 — New-user bootstrap is unreachable.**

Root cause was `/auth/me` reporting unregistered users as signed out — fixed in #3861 (404 → `authenticated: true`, no `user`). Here, `App` keys the bootstrap off `useSessionInfo`, so that shape starts `completeSignIn` and reaches `registerUser()`. New App-level test: unregistered session → registration path starts.

> **P1 — Failed registration unlocks the application with an empty user.**

A non-409 registration failure now signs the session out after the error notification (consistent with the AzureB2C path) instead of resolving — the routes never render authenticated with an empty `CurrentUser`. Signing in again retries.

> **P2 — The production 409 error shape does not match the tests.**

Confirmed against the adapter: `handleResponse`'s axios-like error gets re-wrapped by the outer catch, so the status lives on `Error.cause`. `errorStatus()` now walks the cause chain, and the specs construct the adapter's real wrapped error rather than hand-built bare shapes.

> **P2 — Stale identities bypass bootstrap.**

The bootstrap also fires on identity mismatch between the session's user and stored `CurrentUser`, and re-arms if focus revalidation detects an account switch after completion — fresh `completeSignIn` overwrites the profile, clears the query cache, and re-runs the ToS gate. The run-once guard is keyed by session identity (StrictMode double-runs stay blocked; genuine re-arms pass). Legacy mode reports the stored user as the session identity (#3861), so this can't mis-fire pre-cutover.

kevinmarete (commented):
One more finding:

**P1 — Successful new-user registration can leave the app stuck on the spinner.**

After `registerUser()` succeeds, `setUserRoleStatuses()` stores the new nonzero user ID. However, `useSessionInfo()` still contains the cached `/auth/me` response for the unregistered user—`authenticated: true` with no `user`.

`App` consequently detects an identity mismatch and resets `signInBootstrapDone` to `false`. The bootstrap effect cannot run again because `signInBootstrapKickedOffFor.current` is already `0`, matching the still-missing session user. This leaves `isBootstrappingSignIn` true indefinitely and prevents the Terms of Service route from rendering.

Please ensure registration updates/revalidates the session identity or adjust the bootstrap state machine so completion for the current bootstrap target cannot immediately re-arm into a run blocked by the run-once guard. A regression test should simulate registration updating `CurrentUser` while `useSessionInfo()` still returns the cached authenticated-without-user response.

rushtong:
> **P1 — Successful new-user registration can leave the app stuck on the spinner.**

Confirmed — exactly the livelock you traced: the persisted nonzero user vs the cached
authenticated-no-user probe read as an identity mismatch, the re-arm reset the bootstrap,
and the once-per-identity guard blocked the re-run. Fixed in e2a29fe8, two parts:

- An identity mismatch now requires the session to *name* a different user
  (`sessionUserId !== undefined && sessionUserId !== storedUserId`) — a no-user session is
  what the cached probe looks like right after registration and proves nothing about the
  stored profile. Cross-tab switch detection still works: once the other tab's bootstrap
  registers its user, this tab's focus revalidation sees a session user that differs.
- `completeSignIn` now calls `resetSessionCache()` whenever it persists an identity
  (registration and the getMe/409 paths), so the next navigation/focus re-probe reports the
  registered user instead of the stale shape.

Added the regression test you suggested: `completeSignIn` persists a nonzero profile while
`useSessionInfo` stays authenticated-without-user — asserts the routes unlock and the
bootstrap runs exactly once.

kevinmarete (commented):
Everything else looks good, just one finding:

**P1 — The BFF registration flow can destroy the session before registering a new user.**

`/auth/me` intentionally treats an upstream `/api/user/me` 401 as an authenticated session without a DUOS profile. The reconciler then calls `completeSignIn()`, which calls `User.getMe()` again through `/duos-api`. The proxy treats that same upstream 401 as `session_expired` and destroys the BFF session. `handleRegistration()` then attempts `registerUser()` without a valid session.

Please avoid the duplicate `User.getMe()` when the session probe already reports no profile, or make the proxy preserve the session for this endpoint. Please also add an integration-style regression test where the upstream returns 401 for `/api/user/me` and verify registration retains the session and completes.

rushtong:
I confirmed the finding and fixed it in 9901b650.

The cause is the one you traced. `completeSignIn` called `User.getMe()` through `/duos-api` before registration. The upstream returned the same 401 that the probe saw. The proxy reads that 401 as proof of a bad token (`destroySessionOnUpstream401`). The proxy then destroyed the session before `registerUser()` ran.

The fix adds one option, `sessionReportsNoProfile`. The reconciler sets it when the probe reports a session without a profile. Only the BFF (backend for frontend) probe can produce that shape — the legacy probe always includes a user. When the option is set, `completeSignIn` does not call the first `User.getMe()`. It goes directly to registration, and the session stays intact.

If the probe answer is stale because the user registered in another tab, `registerUser()` returns 409. The 409 branch then fetches the user again and completes the sign-in. That fetch is safe: for a registered user, `User.getMe()` succeeds.

Tests: a new regression test mocks `User.getMe()` to fail in the same way as the proxy. The test asserts three things: the code does not call `User.getMe()`, registration completes, and no sign-out occurs. A second test covers the stale-probe path through the 409 branch. Two reconciler tests pin the option: true for the BFF probe without a profile, false for the legacy probe.

kevinmarete (approved):
Excellent, thanks for the fixes!

fboulnois (approved):
👍

otchet-broad:
Overall this looks good.  I asked claude to evaluate multiple tabs and ensure that flow works as expected.  Below is the feedback from that exercise:

# Code Review — `claude/epic-4-story-4g-post-signin`

Reviewed `develop...HEAD` (27 files, ~2.4k added lines), with a focus on cross-tab race conditions.

## Verdict

The architecture is sound. ADR-011 is right that with one shared session cookie, "two different users in two tabs" is unreachable, so the only real phenomenon is a stale tab. `server/src/auth/refresh.ts` already has single-flight-per-session plus cross-pod loser-adoption, so the concurrent-`/auth/me`-refresh race that `server/src/auth/me.ts` newly opens is already covered.

Five cross-tab issues need addressing. Two are on ordinary paths, not exotic interleavings.

---

## Must fix

### 1. Cross-tab account switch bypasses `queryClient.clear()` *(cross-tab)*

**`src/hooks/useSessionReconciler.ts:189-201`** (hydrate path); same gap in **`applyPendingHydration` at `:108`**

The hydrate path persists profile and role statuses but never clears the react-query cache. Only the bootstrap paths do (`src/libs/auth/postSignIn.ts:133`, `:193`) and in-tab sign-out (`src/components/DuosHeader.tsx:181`). Confirmed by grep: `queryClient` appears in the reconciler only as an argument to `completeSignIn`.

Repro — no timing window required:

1. Tab B open, signed in as X. Its `QueryClient` holds X's role-scoped library results, tab counts, filter metadata.
2. In tab A the user signs out and signs in as Y. Tab A bootstraps, writes `CurrentUser = Y` to **shared** `localStorage`, clears **its own** cache.
3. User returns to tab B → focus → revalidate → probe reports Y, and `Storage.getCurrentUser().userId` is already Y → `probeNamesStoredUser` true → **hydrate**.
4. Tab B adopts Y with X's cached queries still resident (`staleTime` 5 min, `gcTime` 10 min, `refetchOnWindowFocus: false`), served under Y. Keys like `library-metadata` and `STUDY_DATASETS_QUERY_KEY` are not user-scoped.

Classification compares the probe against *shared storage*, not against what this tab was displaying. "Single writer to `localStorage`" is sound *for `localStorage`* — but the query cache is **per-tab** while the identity record is **shared**.

**Fix:** track the identity this tab's `QueryClient` was populated under (a ref seeded from `Storage.getCurrentUser().userId` at mount, updated on every commit) and clear when it differs.

### 2. Identity-conflict reload is a no-op on fragment URLs → permanent spinner *(cross-tab)*

**`src/hooks/useSessionReconciler.ts:185`**

`Redirect.to(globalThis.location.href)` only reloads when the URL has no fragment. With `#anything` present the browser performs a same-document fragment navigation and nothing reloads.

The conflict branch returns without updating `snapshot.classifiedProbe`, so `unclassifiedIdentityChange` stays true and `authEquivalent` stays false (storage still holds the old user) → **`reconciling` stays true permanently**. `App.tsx` then renders only the spinner: no header, no routes. Self-heals only if the user tabs away and back.

**Fix:** `globalThis.location.reload()`, which always reloads. Same construct at `src/components/SignInButton.tsx:51` for legacy popup completion, where a fragment URL makes a successful sign-in appear to do nothing.

### 3. `probeCoveredByRun` adopts a *different* unregistered identity *(cross-tab)*

**`src/hooks/useSessionReconciler.ts:94`**

`/auth/me` cannot distinguish unregistered accounts — they all return `{ authenticated: true, idp }` — so `targetUserId` is `0` for all of them.

Tab A is mid-registration for unregistered user A (`active.targetUserId = 0`). In tab B, unregistered user B signs in on the shared cookie. Tab A revalidates on focus, computes `targetUserId = 0`, matches `0 === active.targetUserId`, and **joins** the dead run instead of reloading. Tab A's already-issued `User.registerUser()` then rides B's session token: registers B, persists B's profile, navigates tab A to `/tos_acceptance` as B — silently, no reload, no cache clear.

ADR-011 names the missing unregistered identity as a known gap, but the join rule turns "can't detect" into "actively adopts the wrong identity."

### 4. Upstream 401 for a *registered* user drives re-registration, then sign-out

**`server/src/auth/me.ts:64`**

A registered user's `/api/user/me` 401s transiently → `getMe` returns `{ authenticated: true, idp }` with no user. The reconciler sees `sessionUser === undefined`, so `probeNamesStoredUser` fails even though storage names a real user, and it bootstraps with `sessionReportsNoProfile: true` (`useSessionReconciler.ts:223`) — skipping `getMe` and going straight to `User.registerUser()`, a POST for an account that already exists.

Recovery depends entirely on the upstream answering exactly `409`. Any other status hits `handleRegistration`'s catch → error notification → `Auth.signOut()`, signing out a healthy session. `registerUser` is a POST, so `shouldSkip401Redirect` returns false and `redirectOnLogout()` fires concurrently with that `signOut`.

### 5. Transient `/auth/me` failure publishes `authenticated: false`

**`src/libs/auth/session.ts:56`**

`probeBffSession` maps `!res.ok` to `{ authenticated: false }`, contradicting `me.ts`'s comment that 502 means "retry on the next ask." Every mounted `useUserIsLogged` flips false: `Authenticated` renders `<Navigate to="/?redirectTo=…">` and yanks the user off their page (losing form state), `DuosHeader` switches to the signed-out header, and `useSessionReconciler:143` retires any in-flight bootstrap. Triggered by a >5 s upstream (`UPSTREAM_TIMEOUT_MS`), a B2C hiccup during refresh-before-forward, or a store write failure — and the focus-revalidation path `me.ts` calls out as "hot" is exactly the one that hits it.

### 6. No `Cache-Control: no-store` on `/auth/*`

Verified by grep: no cache directives anywhere on the auth routes. `getMe` returns a full DUOS profile with no directives and no validators, keyed only by URL, and this endpoint is now the sole auth gate for the SPA. Set `Cache-Control: no-store` and `Vary: Cookie`.

### 7. Legacy sign-in failures lost their handling

**`src/components/SignInButton.tsx:53`**

The rewrite deletes `onFailure`, so closing the legacy popup now shows "Unexpected error, please contact customer support" instead of the friendly cancel message, and the failure-path `Storage.clearStorage()` is gone. `bffEnabled` defaults to false, so this is the live production path until cutover.

### 8. BFF sign-in metrics post anonymously

**`src/libs/auth/postSignIn.ts:84`**

Verified: `Metrics.captureEventFn` gates identified-vs-anonymous on `Storage.userIsLogged()`, which reads the legacy OIDC token that `purgeLegacyOidcKeys()` deletes in BFF mode. Every BFF `userSignIn`/`userRegister` event posts anonymously to the direct Bard URL with no `Authorization` header — while `Metrics.identify` and `syncProfile` in the same function use the identified URL, so they disagree about who the user is.

---

## Secondary

- **No cross-tab change signal.** No `storage` listener and no `BroadcastChannel` anywhere in `src/` (confirmed by grep). Convergence is bounded only by focus/visibility and the 5-min `SESSION_TTL_MS`. A visible-but-unfocused sibling window can render a stale identity for the full TTL. A `storage` listener calling `resetSessionCache()` would collapse this to near-instant.
- **`src/App.tsx:50`** *(cross-tab)* — adding `[]` to the environment effect removes the per-render repair of `localStorage.env`, which `Storage.clearStorage()` (`localStorage.clear()`) wipes. A legacy sign-out in one tab wipes `env` for every open tab and none repair it; `EnvironmentUtils.checkEnv` then returns false, so `EnvRoute` stops matching and `isDevEnv()` is false for the rest of the session.
- **`src/hooks/useSessionReconciler.ts:251`** — `reconciling` is computed during render from `Storage.getCurrentUser()`, an impure render read of shared mutable state that the hook's own effects mutate. Under concurrent rendering two renders of one commit can disagree on `authEquivalent`. Same shape at `src/components/modals/SupportRequestModal.tsx:104`.
- **`src/libs/auth/session.ts:60`** — the failure-path `resetSessionCache()` nulls whatever is cached, which may be a newer healthy in-flight probe, costing an extra `/auth/me` (and an extra server-side refresh) per occurrence. Guard on still owning the cache slot.
- **`src/hooks/useSession.ts:18`** — instantiated independently in five components, each with its own state and focus/visibilitychange listener pair. `App.tsx:106` unmounts `DuosHeader` on every `reconciling` transition, so the header remounts fresh and flashes signed-out after every bootstrap. Consider one shared source (context or `useSyncExternalStore` over `session.ts`).
- **`src/libs/auth/postSignIn.ts:120`** — `redirectTo` interpolated into a query string without `encodeURIComponent`; a target containing `&` or `#` is truncated. `Auth.signIn` already encodes its `returnTo` (`auth.ts:55`).
- **`src/libs/auth/postSignIn.ts:76`** — `errorStatus` walks the `cause` chain with no cycle guard or depth bound.
- **`server/src/auth/me.ts:4`** — importing `REFRESH_WINDOW_SECONDS` from `../proxy/upstreamProxy.js` pulls the proxy stack (undici, agent construction) onto the auth path for one integer. Move it to a shared module.
- Duplicate `userSignIn` metrics and a surprise navigation in a sibling tab that probes between tab A's cookie landing and tab A persisting `CurrentUser`. Narrow window, converges correctly.

---

## Test gap

`test/hooks/useSessionReconciler.spec.tsx` always sets `localStorage` in-test, so "storage changed under us from another tab" is never modeled — which is why findings 1 and 3 survived four review rounds.

rushtong:
Thank you for the thorough review — six of the eight findings are confirmed and fixed, and two are documented as accepted risks in ADR-011.

**Fixed in #3861** (the PR that owns the code):
- **(5) Transient `/auth/me` failure:** the probe now holds the last authoritative answer (a real 200 or 401) across transient failures. One 502 no longer unmounts the authenticated routes or loses form state. Authoritative answers still win.
- **(6) Cache headers:** every `/auth/me` answer now carries `Cache-Control: no-store` and `Vary: Cookie`.

**Fixed in this PR:**
- **(1) Cross-tab switch bypassed the cache clear:** the reconciler now tracks the identity that this tab's QueryClient was populated under. When a hydrate adopts a different identity, it clears the cache first. Two new tests model "storage changed under us" — the gap you named.
- **(2) Fragment-URL reload no-op:** both sites now call a real `location.reload()` through a new `Redirect.reload()` seam. An href assignment to a `#fragment` URL reloads nothing.
- **(7) Legacy failure handling:** restored. A closed popup shows "Sign in cancelled" again, and the failure path clears storage.
- **(8) Anonymous BFF metrics:** the gate now counts a persisted registered profile as signed in when the BFF is enabled, so events, identify, and syncProfile agree about the user.

**Accepted risks, documented in ADR-011:**
- **(3) Target-0 join:** `/auth/me` cannot distinguish unregistered accounts, so a tighter join rule would also reload on the legitimate joins (StrictMode, focus revalidation). The window needs two different unregistered accounts to switch mid-registration. Documented with the reasoning.
- **(4) Transient 401 for a registered user:** the 409 branch recovers the common case, and a genuinely revoked token must end in sign-out — the alternative path (getMe through the proxy) destroys the session anyway. The cost is one spurious POST. Documented.

rushtong:
Re-requesting review due to feedback - fyi @kevinmarete, @fboulnois, @otchet-broad

kevinmarete (approved):
👍🏾

otchet-broad (commented):
Some notes with the help of claude.

### 1. `probeBffSession` turns a 502 into a sign-out
`src/libs/auth/session.ts:56`

`me.ts` carefully separates 401 ("no session") from 502 ("transient, retry"), but the client's `!res.ok` branch collapses both into `{ authenticated: false }`.

**Failure scenario:** User is on `/datalibrary` and refocuses the tab after 20 minutes. The focus revalidation hits `/auth/me`; B2C returns a 5xx during the refresh, so `me.ts` deliberately answers 502. `probeBffSession` discards that distinction and returns `{ authenticated: false }`; `useSessionInfo` publishes it, `<Authenticated>` immediately renders `<Navigate to="/?redirectTo=/datalibrary">`, and `useSessionReconciler` retires any in-flight bootstrap. Resetting the cache only helps the *next* ask — by then the user has already lost the page they were on.

**Fix:** A 502 needs to surface as "unknown, keep the previous answer", not "signed out". The server's 401-vs-502 separation (`server/src/auth/me.ts:39`, `server/src/proxy/upstreamProxy.ts:301`) currently has no effect at the client boundary.

### 2. Identity-conflict reload is a no-op when the URL has a fragment
`src/hooks/useSessionReconciler.ts:185`

`Redirect.to(globalThis.location.href)` is `globalThis.location.href = url`, which performs a same-document fragment navigation when the URL contains a `#` — no page load happens.

**Failure scenario:** Tab is at `/datalibrary/broad#results` with a bootstrap in flight for user 9; a probe reports user 7. The effect sets `active.cancelled = true` and calls `Redirect.to(location.href)`. Nothing reloads, and the effect returns without clearing `activeBootstrapRef.current`, without setting `bootstrapRunning: false`, and without recording `classifiedProbe` — and `consumedProbeRef` has already consumed this probe, so it can never be reclassified. `reconciling = isLoggedIn && (bootstrapRunning || unclassifiedIdentityChange)` stays true indefinitely: header hidden, routes hidden, spinner forever.

**Fix:** `globalThis.location.reload()` reloads unconditionally. Also worth an in-place fallback that clears the run state.

### 3. Same fragment bug in the legacy sign-in path
`src/components/SignInButton.tsx:51`

**Failure scenario:** Legacy (non-BFF) config; user is on `/#datasets` and clicks Sign In. `Auth.signIn` resolves with the OidcUser, then `Redirect.to(location.href)` does a fragment-only navigation. `handleSignIn` returns with `isLoading` still true (only the catch resets it), so the button stays disabled showing the spinner, and nothing triggers App's post-sign-in bootstrap.

**Fix:** `globalThis.location.reload()`.

### 4. Blanket 401 → "unregistered" also swallows revoked tokens
`server/src/auth/me.ts:64`

**Failure scenario:** A registered user's B2C access token is revoked (admin action, account disabled) 30 minutes before expiry. `/auth/me` sees `secondsRemaining >= 60`, skips the refresh, forwards the revoked token, gets a 401, and answers `{ authenticated: true, idp }` with no user. The client classifies that as unregistered, so `completeSignIn` takes the `sessionReportsNoProfile` branch straight to `User.registerUser()` — a POST that `shouldSkip401Redirect` does not exempt, so the 401 fires `redirectOnLogout()` mid-flight, or `handleRegistration`'s catch shows "There was an error completing your registration. Please try again."

The pre-change behavior — destroy the session, clear the cookie, 401 — was the correct terminal answer.

**Fix:** Disambiguating "no profile" from "bad token" needs a signal other than the status code (e.g. probing as unregistered only when the session has never seen a profile), not a blanket reinterpretation.

### 5. Folding 404 into the same branch masks upstream misconfiguration
`server/src/auth/me.ts:64`

**Failure scenario:** `DUOS_API_URL` points at a host/base path where `/api/user/me` 404s (bad env value, upstream route rename, an ingress serving a 404 page). Every existing user's probe returns `{ authenticated: true }` with no user, so every client runs `completeSignIn` with `sessionReportsNoProfile: true` and POSTs `/api/user` to "register" an account that already exists — instead of the honest 502 `upstream_unavailable` the old `!res.ok` branch produced, which would have made the misconfiguration obvious.

The comment calls this "kept for symmetry should the upstream ever disambiguate", but the upstream does not disambiguate today, so the branch only fires on genuine 404s.

### 6. Anonymous UI renders while the probe is in flight
`src/hooks/useSessionReconciler.ts:253`

`reconciling` is false while `sessionInfo === undefined`.

**Failure scenario:** A signed-in user hard-loads `/home`. `useSessionInfo` returns undefined, so `isLoggedIn` is false and `reconciling` is false — App renders `<DuosHeader />` (Sign In) plus `<AppRoutes isLogged={false} />`. In `src/pages/Home.tsx:517` each featured-library card is a `<Link to='#'>` whose onClick calls `e.preventDefault()` and `handleSignIn(libraryPath, navigate)`, so a user who clicks a library card before `/auth/me` answers is pushed into a sign-in flow instead of the library. The same gap makes `<Authenticated>` render `null` (blank content area, no spinner) for deep links.

**Fix:** The in-flight probe should gate identity-bearing UI the same way an unclassified probe does.

### 7. Removing `onFailure` lost popup-cancellation handling
`src/components/SignInButton.tsx:53`

**Failure scenario:** Legacy popup flow (still the default until Epic 6). The user closes the B2C popup: `OidcBroker.signIn()` resolves null, `Auth.signIn` throws `new Error(Auth.signInError())`, and the single catch renders the danger Alert "Error / Unexpected error, please contact customer support." The deliberate "Sign in cancelled by closing the sign in window" message is gone, and so is the `Storage.clearStorage()` that cleared the half-written oidc keys. Any real login error is also flattened to the same generic string, since the caught error's message is discarded in favor of `Auth.signInError()`.

### 8. BFF sign-out never resets the session cache
`src/libs/auth/auth.ts:78`

**Failure scenario:** User clicks Sign Out in `DuosHeader`. `signOut` calls `navigate('/home')` first, which makes every mounted `useSessionInfo` re-probe — and `getSessionInfo` hands back the still-cached authenticated answer, since nothing invalidated it. `Auth.signOut` then does `Storage.clearStorage()` (CurrentUser reset to the empty default) before `Redirect.to('/home')`. In that window React re-renders with `isLogged === true` and an empty stored user, so `DuosHeader` paints a signed-in header with a blank display name and no role tabs.

**Fix:** `resetSessionCache()` alongside `Storage.clearStorage()` in the BFF branch. `postSignIn` already does this on the registration path for exactly this reason.

### 9. Redirect target interpolated into a query string without encoding
`src/libs/auth/postSignIn.ts:120,127`

**Failure scenario:** A new user signs in from `/datalibrary?filter=a&b=c`. `?redirectTo=${redirectTo}` produces `/tos_acceptance?redirectTo=/datalibrary?filter=a&b=c` — the ToS page reads `redirectTo` as `/datalibrary?filter=a` and the rest becomes sibling params, so accepting the ToS drops the user somewhere else.

**Fix:** `encodeURIComponent(redirectTo)` on both lines; the reading side already uses `URLSearchParams`, which decodes. (Code was moved verbatim into this module, so the bug is pre-existing.)

---

## Cleanups

| # | Finding | Location |
|---|---------|----------|
| 10 | **Duplicated localStorage writes.** `setUserRoleStatuses` (`src/libs/utils.ts:202`) already ends with `Storage.setCurrentUser(user)`, so each explicit call immediately before it costs a second `JSON.stringify` + `localStorage.setItem` for the same object. Dropping the explicit call in all four spots is behavior-identical. | `useSessionReconciler.ts:109-110, 190-191`; `postSignIn.ts:185-186, 206-207` |
| 11 | **Render-phase storage reads.** `authEquivalent` does `Storage.getCurrentUser()` + `JSON.parse` + a role-array sort/join in the render body of the hook used by App — which re-renders on every navigation, `isLoading` flip, and reconciler state change. `SupportRequestModal` repeats it: `resetFormData(isLogged)` builds all six form fields (plus a storage parse) per render to use only `name` and `email`. These impure reads are also why the effect needs `react/react-compiler` escapes; caching the parsed profile in state removes both the cost and the impurity. | `useSessionReconciler.ts:251`; `SupportRequestModal.tsx:104` |
| 12 | **Layering inversion for one constant.** `import { REFRESH_WINDOW_SECONDS } from '../proxy/upstreamProxy.js'` pulls the whole proxy module (`@fastify/reply-from`, pool config, the proxy factory) into the auth handler, and `upstreamProxy` already imports from `../auth/*` — so the two now reference each other in both directions. It is a refresh-policy constant: it belongs next to `refreshAccessToken` in `server/src/auth/refresh.ts`, with `upstreamProxy` re-exporting it for existing consumers. | `server/src/auth/me.ts:4` |
| 13 | **Dead third session probe.** `Auth.isAuthenticated` has no production callers left (only its own unit test) — it duplicates the probe without the caching, TTL, or legacy fallback in `session.ts`, and this diff changed what `res.ok` means: it is now true for an authenticated-but-unregistered session, so anyone who reintroduces it gets "signed in" for a user with no DUOS profile. Delete it, or make it delegate to `userIsLogged()`. | `src/libs/auth/auth.ts:109` |
| 14 | **Misplaced JSDoc.** Lines 99-102 describe `applyPendingHydration` ("persist the freshest profile a joined probe carried… read the live location") but sit directly above `tosExplicitlyRejected`, which has its own doc block right after. Hovers and doc extraction attribute the wrong contract to both functions, and the "read the live location" rationale is detached from the code it explains. | `useSessionReconciler.ts:99` |

rushtong:
Thank you — this caught real gaps. Status by finding: four were already fixed in the Aug 17 round, eight are fixed in 1716212c, one is declined with a reason, and one is deferred.

**Already fixed (Aug 17):**
- **(1)** The probe holds the last authoritative answer (a real 200 or 401) across transient failures — shipped in #3861 (`lastAuthoritativeAnswer`, session.ts). A 502 no longer signs the user out.
- **(2), (3)** Both reload sites call `Redirect.reload()`, a true `location.reload()`. The fragment no-op is gone.
- **(7)** The cancel branch and the `Storage.clearStorage()` are restored. One note: `OidcBroker.signIn` has no catch, so the "Popup closed by user" rejection propagates to that branch — the resolves-null path does not occur. Non-cancel errors show the generic support message by design.

**Fixed in 1716212c:**
- **(4)** The session records `profileSeen` when `/auth/me` first serves a profile. A 401/404 on a profile-seen session destroys the session and answers a terminal 401 — no re-registration of an existing user. A never-seen session still reads as unregistered, which keeps new-user registration reachable. ADR-011's accepted-risk entry narrows to the first-contact window.
- **(6)** `reconciling` is true while the first probe is in flight, so the app shows the spinner instead of anonymous chrome. Cost: one probe round-trip behind the spinner on a BFF hard load; the legacy probe answers synchronously.
- **(8)** The BFF branch of `Auth.signOut` calls `resetSessionCache()` next to `Storage.clearStorage()`.
- **(9)** Both `redirectTo` interpolations use `encodeURIComponent`; a test covers a destination that carries its own query string.
- **(10)** All four duplicate `Storage.setCurrentUser` calls are removed — `setUserRoleStatuses` persists the user itself.
- **(12)** `REFRESH_WINDOW_SECONDS` lives in `auth/refresh.ts` now; the proxy re-exports it for its consumers.
- **(13)** `Auth.isAuthenticated` is deleted — no callers, and its `res.ok` reading became wrong with this diff.
- **(14)** The doc block sits above `applyPendingHydration` again.

**Declined:**
- **(5)** The 404 branch stays. DT-3997 (in sprint) restores the upstream's 404 for unregistered users, and this branch is the forward-compatibility for it. The `profileSeen` gate from (4) also covers the misconfiguration case for existing users: their sessions end in a visible sign-out, not a re-register storm.

**Deferred:**
- **(11)** The render-phase storage reads are real, but the refactor touches review-hardened render logic, and `SupportRequestModal` is outside this stack. Noted as follow-up work.

otchet-broad (approved):
👍
Thanks for addressing the findings!  Nice work!

PR #3867, opened 2026-08-14
[DT-3316] Sortable data grid for the Manage Users table

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3316

### Summary

Replaces SimpleTable and PaginationBar with MUI DataGrid so every column sorts: User Name, Email, Institution and Roles. Rows carry the displayed text for institution and roles, so each column sorts on what is read.

Search filtering and paging are derived rather than driven by effects, with the page clamped so a narrower search cannot land on an empty page. The cell data module only built SimpleTable cells, so its role and institution formatting moves to manageUsersTableUtils.

**Before**

Before

**After**

After

----

- Verify all tests go green


rushtong (approved):
Nice 👍🏽

fboulnois (approved):
One less set of custom controls, nice! 👍

otchet-broad:
I asked claude to do a review and the following issues were identified as the three that should be addressed:

1. **The focus ring you just added is invisible on the link you just made focusable** — `:focus-within { outline: none }` plus the global `a { outline: 0 !important }` in `index.css` cancel each other out. The two commits work against each other here.
2. **Pagination state is never written back** — clearing a search jumps the admin to a stale page, and a new search doesn't reset to page 1.
3. **`'N/A'` / `'None'` are now sort keys** — those columns were `sortable: false` before this PR, so the placeholder strings now scatter empty values into the middle of sorted results.

The rest are cleanup: three duplication findings (focus `sx` copied from `LibraryDataGrid`, search filter as a third copy of `utils.ts` logic that splits on space instead of `/\s|,/`), a `paginationModel` referential-identity re-render, missing `aria-label` on the grid, and two test-quality issues — the specs render without the app's `ThemeProvider`, so they exercise MUI's 16px default rather than the `htmlFontSize: 10` fix production uses.

Given this has landed, feel free to ignore or follow-up.

PR #3871, opened 2026-08-17
[DT-3316] Fix grid pagination write-back and link focus ring

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3316

Security risk: no

Follow-up to #3867, which merged before this review feedback landed.

### Summary

Two fixes to the Manage Users grid, both confined to `ManageUsersTable`.

**Pagination is written back, not just clamped.** The clamp added in #3867 shaped the render but never reached state, so the component rendered a valid page while `paginationModel.page` kept the abandoned one. MUI syncs a controlled `paginationModel` into its internal state without firing `onPaginationModelChange`, so nothing corrected it — clearing a search jumped the admin back to a page they had already been moved off. The page is now adjusted during render, and a new search restarts at the first page. This restores the pre-#3867 behaviour, where `recalculateVisibleTable` wrote the clamped page back through `setCurrentPage`.

**A focused user name link is visible again.** #3867 gave the link its cell's `tabIndex` so it joins the grid's roving focus, which makes the link the focus target inside the cell — but the cell's `:focus-within { outline: none }` and the global `a, input { outline: 0 !important }` in `index.css` between them left it with no indicator at all. The grid now rings a focused link at matching weight, so keyboard users can see what they are about to activate.

### Notes for review

Only this table needed either fix. `SigningOfficialTable` owns its `SearchBar` and already resets pagination in the handler, and has no links in cells; `ManageUsersTable` receives `searchText` as a prop, so it has to react during render instead.

Two other points from the same review were checked and deliberately not actioned:

- The search filter splitting on `' '` rather than `/\s|,/` is not a regression. This table previously used `searchOnFilteredList`, which splits on `' '` too; the `/\s|,/` variant lives in `tableSearchHandler`, which it never called. Behaviour is unchanged, and `filterUsers` lowercases internally.
- The duplicated focus `sx` is real but wider than this PR — the block is copy-pasted across six grids, four of which (`LibraryDataGrid`, `ManageDac`, `DacProfile`, `ControlledAccessGrants`) have no `:focus-visible` ring at all, and `ManageDac`/`DacProfile` carry the same invisible-link-focus bug fixed here. That is worth centralising in its own ticket rather than partially here, ideally before DT-3982 adds a seventh grid.

### Testing

Three tests added, each verified to fail against `develop`: clearing a search does not restore the abandoned page, a new search restarts at page 1, and the link ring is emitted at a weight that survives the global reset. Full suite green (3894 tests), lint and type-check clean.

----

- Verify all tests go green

### DIFF
--- a/src/components/manage_users_table/ManageUsersTable.tsx
+++ b/src/components/manage_users_table/ManageUsersTable.tsx
@@ -19,6 +19,12 @@ const DATAGRID_SX = {
     outline: `2px solid ${Theme.palette.link}`,
     outlineOffset: '-2px',
   },
+  // The link takes focus in its cell's place, and index.css drops link outlines with !important,
+  // so the ring has to be reinstated at the same weight to be seen at all.
+  '& .MuiDataGrid-cell a:focus-visible': {
+    outline: `2px solid ${Theme.palette.link} !important`,
+    outlineOffset: '2px',
+  },
 }
 
 // Left inset matches SearchBar's own margin, negative right mirrors the search and add button row.
@@ -69,6 +75,7 @@ const filterFn = getSearchFilterFunctions().users
 
 export const ManageUsersTable = function ManageUsersTable({ isLoading, userList, searchText }: ManageUsersTableProps) {
   const [paginationModel, setPaginationModel] = useState<GridPaginationModel>({ page: 0, pageSize: PAGE_SIZE_OPTIONS[0] })
+  const [lastSearchText, setLastSearchText] = useState(searchText)
 
   // Filtering is derived, so a keystroke costs one render rather than a cascade of effects.
   const rows = useMemo(() => {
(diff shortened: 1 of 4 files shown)


otchet-broad (approved):
👍

src/components/manage_users_table/ManageUsersTable.tsx
```diff
@@ -19,6 +19,12 @@ const DATAGRID_SX = {
     outline: `2px solid ${Theme.palette.link}`,
     outlineOffset: '-2px',
   },
+  // The link takes focus in its cell's place, and index.css drops link outlines with !important,
+  // so the ring has to be reinstated at the same weight to be seen at all.
+  '& .MuiDataGrid-cell a:focus-visible': {
+    outline: `2px solid ${Theme.palette.link} !important`,
```
fboulnois: `!important` should be avoided and fixed at the source, otherwise we get a chain of `!important` that makes it harder to reason about the code.

fboulnois (approved):
One comment, otherwise lgtm 👍

PR #3872, opened 2026-08-17
[DT-1854] Download a blank study template, then upload, validate, and replace a filled-in one

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1854

Security risk: **low** — the page accepts a user-supplied file but never parses it in the browser; parsing and validation stay server-side, and the endpoint it posts to is not deployed yet, so this adds no reachable state change. It adds a control rather than surface: every generated cell is escaped against spreadsheet formula injection. Consent's error strings render as React children rather than markup, and the new route reuses the existing `RoleBAC` guard.

### Summary

Ticket 4 of the study-template workflow planned in `docs/plans/study-template-validation-workflow.md`. Adds an **UPLOAD TEMPLATE** action beside **ADD DATASET** on My Data Submissions, routed to a new `/data_submission_template` page that downloads a blank v1 template, then uploads, validates, removes, and replaces a filled-in one without leaving the page. **ADD DATASET is untouched**, including its data-submitter-only disabled state.

- **The blank download is a v1 requirement, not a convenience.** The contract forbids JSON, so every structured wire value is a row — a hand-built file means keeping `recordType`, `recordId`, and `parentRecordId` consistent across 51 rows.
- **The field manifest is derivable rather than transcribed.** It follows the declared wire order of `StudyRegistrationRequest` (items 1–27 of 41), `ConsentGroupRequest` (items 2–23 of 25), and both `FileTypeObject` properties, so it can be diffed against those `@JsonPropertyOrder` annotations instead of against prose. It omits every field `StudyTemplateV1Fields.isExcludedFromV1()` rejects by name — the free-form `assets`/`data` maps, the `alternativeDataSharingPlan*` group, file-backed and integration-owned fields — because offering one would produce a template that cannot validate.
- **Every generated cell passes through the formula-prefix guard.** Nothing the manifest currently emits triggers it, which a test pins deliberately, so adding a field that starts with `=`, `+`, `-`, or `@` cannot silently introduce a CSV-injection vector into the producer's spreadsheet.
- **Validation failures are a completed result, not a request failure.** The client returns the discriminated response as data rather than throwing, and the page renders structured errors — with row and column context when Consent supplies it, and never fabricated when it doesn't. A request failure is different in kind and rendered differently: a 401, a 413, a 5xx, or Consent being unreachable becomes a dismissible toast, like every other failure in the app. Both leave the user on the page with the file still selected and VALIDATE still available, so a retry costs nothing.
- **Success requires both `valid: true` and a `StudyDatasetSubmissionV1` draft reference** before routing to `/data_submission_form/draft/study-dataset/:draftId`. A valid response with a missing or different type is treated as a request failure. Registering that route, and the `DataSubmissionFormV2` draft mode that gives it something to render, is [DT-3872](https://broadworkbench.atlassian.net/browse/DT-3872)'s first line of work; adding the route here without the hydration would resolve a validated template to the blank create form, as though the file had been discarded. The success path is unreachable until both that ticket and DT-3871 land — see Rollout below.
- The picker asks for `.csv` and pre-checks the 5 MiB limit in the browser using Consent's exact wording, so a doomed upload never leaves the page and the two layers cannot appear to disagree.
- **A validation result belongs to the file that produced it.** The picker and Remove stay live while a request is in flight, so a response that outlives its file is discarded rather than applied — otherwise A's errors could render beside B's filename, A's request failure could toast against an empty page, or A's draft could navigate away from B. Discarding the stale continuation is preferable to freezing the controls, since the user has already said they abandoned that file.

Worth a reviewer's eye:

- **`AsyncSpinnerButton` needed `hideOnSuccess={false}`.** It defaults to `true`, and a `valid: false` result resolves normally — the default would delete the Validate button at exactly the moment the user needs to retry.
- **The file picker is a plain `<input type="file">` rather than `FormField type={FormFieldTypes.FILE}`.** `FormField` keeps its own `formValue` state this page would duplicate and attaches a second label to the input, muddying the accessible name the ticket's a11y criteria require. `src/components/forms/FileInput.tsx` already establishes the bespoke-input pattern here. Happy to switch if you'd rather it match `IrbDocumentUpload.tsx`.
- **UPLOAD TEMPLATE is enabled for data submitters, chairpersons, and admins**, matching the route's own `RoleBAC` guard, while ADD DATASET stays data-submitter-only. That is per the ticket's acceptance criteria, but it does mean the two adjacent buttons behave differently for chairpersons and admins.
- **The canonical fixtures are duplicated from `consent`** into `test/fixtures/study-template/v1/` (provenance and commit recorded in a sibling README, since the parser rejects any unrecognised row and so cannot carry a comment). vitest can't reach another repository, and the round-trip spec needs real bytes rather than a hand-written approximation.

### What it looks like

**1. The entry point on My Data Submissions.** UPLOAD TEMPLATE sits beside ADD DATASET as a secondary action. ADD DATASET keeps its primary styling and its data-submitter-only disabled state.

After-1

**2. The upload page, idle.** Download blank template renders through the shared `DownloadLink`, the picker states the `.csv` and 5 MiB limits before any upload, VALIDATE is disabled until a file is chosen, and there is a way back to the manual form. Shown here as an admin, one of the three roles the route admits.

After-2

**3. A file selected.** The filename stays visible, Remove is offered alongside it, the picker relabels to "Choose a different file", and VALIDATE becomes available.

After-3

The last two were captured with both dependencies present — [consent#3028](https://github.com/DataBiosphere/consent/pull/3028) running locally and the draft form from [#3875](https://github.com/DataBiosphere/duos-ui/pull/3875) — because that is the only state this page ships in: it cannot be released before either of them. The fixtures are the canonical ones from the `consent` repository.

**4. A template with errors.** `invalid/field-values.csv`, which is the fixture whose expected errors that repository pins in a sibling `.errors.json`. Each error renders with the location Consent supplies and none is invented: the first five carry a row and a column, and the last carries only a row, because no single cell is at fault for it. Note the third — "Unknown nihAnvilUse value" does not quote the offending cell, which is DT-3870's fix for echoing uploaded content back to the page.

validation-errors

**5. A valid template becomes a draft, and the draft becomes the form.** `valid/minimal-valid.csv` validates, Consent creates a `StudyDatasetSubmissionV1` draft owned by the caller, and the page routes to `/data_submission_form/draft/study-dataset/{id}`, which loads it: study name, description, and data types are the template's, and the dataset it described is a study asset. The route and the form mode behind it are [#3875](https://github.com/DataBiosphere/duos-ui/pull/3875)'s work — on this branch alone the same click reaches a route that does not exist yet, which is what the Rollout section below is about.

draft-hydrated-top

draft-hydrated-datasets

<details>
<summary><strong>The blank template it produces</strong> — 51 rows, 52 lines, 2,241 bytes, CRLF, every <code>value</code> cell empty</summary>

```csv
templateVersion,recordType,recordId,parentRecordId,field,value
1,study,study,,studyName,
1,study,study,,studyType,
1,study,study,,studyDescription,
1,study,study,,dataTypes,
1,study,study,,phenotypeIndication,
1,study,study,,species,
1,study,study,,piName,
1,study,study,,piEmail,
1,study,study,,dataCustodianEmail,
1,study,study,,publicVisibility,
1,study,study,,throughBioId,
1,study,study,,nihAnvilUse,
1,study,study,,submittingToAnvil,
1,study,study,,dbGaPPhsID,
1,study,study,,dbGaPStudyRegistrationName,
1,study,study,,embargoReleaseDate,
1,study,study,,sequencingCenter,
1,study,study,,piInstitution,
1,study,study,,nihGrantContractNumber,
1,study,study,,nihICsSupportingStudy,
1,study,study,,nihProgramOfficerName,
1,study,study,,nihInstitutionCenterSubmission,
1,study,study,,nihGenomicProgramAdministratorName,
1,study,study,,multiCenterStudy,
1,study,study,,collaboratingSites,
1,study,study,,controlledAccessRequiredForGenomicSummaryResultsGSR,
1,study,study,,controlledAccessRequiredForGenomicSummaryResultsGSRRequiredExplanation,
1,consentGroup,consentGroup-1,study,consentGroupName,
1,consentGroup,consentGroup-1,study,accessManagement,
1,consentGroup,consentGroup-1,study,generalResearchUse,
1,consentGroup,consentGroup-1,study,hmb,
1,consentGroup,consentGroup-1,study,diseaseSpecificUse,
1,consentGroup,consentGroup-1,study,poa,
1,consentGroup,consentGroup-1,study,otherPrimary,
1,consentGroup,consentGroup-1,study,nmds,
1,consentGroup,consentGroup-1,study,gso,
1,consentGroup,consentGroup-1,study,pub,
1,consentGroup,consentGroup-1,study,col,
1,consentGroup,consentGroup-1,study,irb,
1,consentGroup,consentGroup-1,study,gs,
1,consentGroup,consentGroup-1,study,mor,
1,consentGroup,consentGroup-1,study,morDate,
1,consentGroup,consentGroup-1,study,npu,
1,consentGroup,consentGroup-1,study,otherSecondary,
1,consentGroup,consentGroup-1,study,dataAccessCommitteeId,
1,consentGroup,consentGroup-1,study,dataLocation,
1,consentGroup,consentGroup-1,study,url,
1,consentGroup,consentGroup-1,study,requestLocation,
1,consentGroup,consentGroup-1,study,numberOfParticipants,
1,fileType,fileType-1,consentGroup-1,fileType,
1,fileType,fileType-1,consentGroup-1,functionalEquivalence,
```

</details>

### Rollout

**This must not be released before [DT-3871](https://broadworkbench.atlassian.net/browse/DT-3871).** That endpoint is written and in review at [consent#3028](https://github.com/DataBiosphere/consent/pull/3028) but not deployed, and deployed `DraftResource` is still `{ADMIN, DATASUBMITTER}` — a chairperson would get a 403. That ordering is the plan's staged-rollout decision (deploy Consent first, then expose the UI entry point) and is [DT-3873](https://broadworkbench.atlassian.net/browse/DT-3873)'s to sequence. Merging is safe; the page is simply unreachable value until then.

### Findings handed to other tickets

Three things surfaced while building against the Consent DTOs and the DT-3870 parser. None blocked this ticket, and all three have since been answered by the tickets that owned them — [DT-3870](https://broadworkbench.atlassian.net/browse/DT-3870) is merged and [DT-3871](https://broadworkbench.atlassian.net/browse/DT-3871) is in review at [consent#3028](https://github.com/DataBiosphere/consent/pull/3028):

- **The `draft` table is the only JSON write in this area without the NUL guard the DAR tables already have.** `jsonb` rejects a U+0000 escape, and U+0000 is valid UTF-8 and not whitespace, so it survives the parser's strict decode and its `isBlank()` check — a template Consent reports as valid can still 500 on insert. A known problem class here rather than a new one: a 2025 migration stripped it out of `data_access_request.data` and the go-forward guard now sits at five DAR call sites, but `DraftDAO.insert` and `updateDraftByDraftUUID` never got it. DT-3870 takes parse-time rejection so the producer is told; DT-3871 takes the write-time guard, which matters on update as much as insert since the draft is meant to be modified and saved.
- **Whether the truncation flag reaches the wire.** It does. DT-3871 serializes `truncated` and keeps the trailing message-only error as well, so the notice on this page renders from the flag while a client that ignores it still tells the producer. The type here stays optional, which is the forgiving side of that.
- **Whether oversize and encoding failures stay `valid: false`.** Split, in DT-3871: the 5 MiB limit is enforced by the validator that reads the upload and answered with `413` carrying the same wording this page pre-checks with, so both layers say the same thing in the same place; an empty, non-UTF-8, or malformed file stays a `valid: false` result, because that is what the producer edits and fixes here.

The plan doc keeps a summary table naming the owning ticket and what each was settled as, plus the one design observation no ticket owns — the supported field set being a derivable slice of declared wire order, which DT-3872 will want when mapping a draft back into the UI model. Its response models, examples, and its own claim about size limits now match what DT-3871 implemented.

### Testing

`pnpm run lint` and `pnpm run type-check` clean. `pnpm test` green at **356 files / 3993 tests**, including 95 new ones across five new specs — manifest fidelity, CSV generation and escaping, the round trip, the multipart client, and the page across idle, pending, validation-error, request-error, replacement, success, wrong-draft-type, stale-response, and accessibility states — plus role-visibility and routing additions to two existing specs. The three stale-response tests were each confirmed to fail without the guard.

The template shown above is the real browser download, and it is byte-identical to what the generator emits in tests: 2,241 bytes, 52 lines, every line CRLF.

The round-trip spec does real work rather than restating the generator: filling the generated scaffold's `value` cells from `minimal-valid.csv` reproduces that fixture exactly, so the download provably covers a known-valid file without hand-authored rows. `multi-consent-group-valid.csv` backs field coverage instead of strict equality, because repeated rows for multi-item arrays and a second consent group are row *additions* — the producer edit the contract documents.

`pnpm run test:e2e` is 6 passed / 5 failed locally; all five are `auth.spec.ts` throwing on missing `DUOS_AUTOMATION_*_SA` service-account credentials before any navigation. Pre-existing and environmental — no e2e spec touches this page or route.

----

- Verify all tests go green

[DT-3871]: https://broadworkbench.atlassian.net/browse/DT-3871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DT-3872]: https://broadworkbench.atlassian.net/browse/DT-3872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/data_submission/StudyTemplateUpload.tsx
+++ b/src/pages/data_submission/StudyTemplateUpload.tsx
@@ -30,6 +30,14 @@ export const StudyTemplateUpload = (): React.JSX.Element => {
   const [file, setFile] = useState<File | null>(null)
   const [validationErrors, setValidationErrors] = useState<TemplateValidationError[]>([])
   const [truncated, setTruncated] = useState(false)
+  // The picker and Remove stay live while a validation is in flight, so the handler needs the
+  // current selection at the moment its response arrives, not the one captured when it started.
+  const selectedFileRef = useRef<File | null>(null)
+
+  const selectFile = useCallback((selected: File | null) => {
+    selectedFileRef.current = selected
+    setFile(selected)
+  }, [])
 
   // Failures are toasts, matching the rest of the app. Validation errors are not failures: they are a
   // completed result the user works through while editing their file, so they stay on the page.
@@ -46,38 +54,40 @@ export const StudyTemplateUpload = (): React.JSX.Element => {
     const selected = event.currentTarget.files?.[0] ?? null
     clearResults()
     if (!selected) {
-      setFile(null)
+      selectFile(null)
       return
     }
     // Pre-checks the two limits Consent enforces, so a doomed 5 MiB upload never leaves the browser.
     // Wording matches the server's so the two layers cannot appear to disagree.
     if (!selected.name.toLowerCase().endsWith('.csv')) {
-      setFile(null)
+      selectFile(null)
       Notifications.showError({ text: 'Template file must be a .csv file' })
       return
     }
     if (selected.size > MAX_TEMPLATE_BYTES) {
-      setFile(null)
+      selectFile(null)
       Notifications.showError({ text: MAX_TEMPLATE_SIZE_MESSAGE })
       return
     }
-    setFile(selected)
-  }, [clearResults])
+    selectFile(selected)
+  }, [clearResults, selectFile])
 
   const handleRemove = useCallback(() => {
-    setFile(null)
+    selectFile(null)
     clearResults()
     // Resetting the element lets the user re-select the same filename and still get a change event.
     if (inputRef.current) {
       inputRef.current.value = ''
     }
-  }, [clearResults])
+  }, [clearResults, selectFile])
 
   const handleValidate = useCallback(async () => {
     if (!file) return
+    const validated = file
     clearResults()
     try {
-      const result = await Draft.validateStudyDatasetTemplate(file)
+      const result = await Draft.validateStudyDatasetTemplate(validated)
+      if (selectedFileRef.current !== validated) return
       if (!result.valid) {
         setValidationErrors(result.errors)
         setTruncated(result.truncated === true)
@@ -90,6 +100,7 @@ export const StudyTemplateUpload = (): React.JSX.Element => {
       navigate(`/data_submission_form/draft/study-dataset/${result.draft.id}`)
     }
     catch (error) {
+      if (selectedFileRef.current !== validated) return
       Notifications.showError({
         text: error instanceof Error ? error.message : 'The template could not be validated. Please try again.',
       })
--- a/src/pages/data_submission/StudyTemplateUpload.tsx
+++ b/src/pages/data_submission/StudyTemplateUpload.tsx
@@ -39,6 +39,14 @@ export const StudyTemplateUpload = (): React.JSX.Element => {
     setFile(selected)
   }, [])
 
+  // Resetting the element lets the user re-select the same filename and still get a change event.
+  const clearSelection = useCallback(() => {
+    selectFile(null)
+    if (inputRef.current) {
+      inputRef.current.value = ''
+    }
+  }, [selectFile])
+
   // Failures are toasts, matching the rest of the app. Validation errors are not failures: they are a
   // completed result the user works through while editing their file, so they stay on the page.
   const clearResults = useCallback(() => {
@@ -54,32 +62,28 @@ export const StudyTemplateUpload = (): React.JSX.Element => {
     const selected = event.currentTarget.files?.[0] ?? null
     clearResults()
     if (!selected) {
-      selectFile(null)
+      clearSelection()
       return
     }
     // Pre-checks the two limits Consent enforces, so a doomed 5 MiB upload never leaves the browser.
     // Wording matches the server's so the two layers cannot appear to disagree.
     if (!selected.name.toLowerCase().endsWith('.csv')) {
-      selectFile(null)
+      clearSelection()
       Notifications.showError({ text: 'Template file must be a .csv file' })
       return
     }
     if (selected.size > MAX_TEMPLATE_BYTES) {
-      selectFile(null)
+      clearSelection()
       Notifications.showError({ text: MAX_TEMPLATE_SIZE_MESSAGE })
       return
     }
     selectFile(selected)
-  }, [clearResults, selectFile])
+  }, [clearResults, clearSelection, selectFile])
 
   const handleRemove = useCallback(() => {
-    selectFile(null)
+    clearSelection()
     clearResults()
-    // Resetting the element lets the user re-select the same filename and still get a change event.
-    if (inputRef.current) {
-      inputRef.current.value = ''
-    }
-  }, [clearResults, selectFile])
+  }, [clearResults, clearSelection])
 
   const handleValidate = useCallback(async () => {
     if (!file) return
--- a/src/pages/data_submission/StudyTemplateUpload.tsx
+++ b/src/pages/data_submission/StudyTemplateUpload.tsx
@@ -1,0 +1,221 @@
+import React, { useCallback, useRef, useState } from 'react'
+import { Link, useNavigate } from 'react-router'
+import { Styles, Theme } from 'src/libs/theme'
+import TableHeaderSection from 'src/components/TableHeaderSection'
+import { DownloadLink } from 'src/components/DownloadLink'
+import { AsyncSpinnerButton } from 'src/components/AsyncSpinnerButton'
+import { Draft } from 'src/libs/ajax/Draft'
+import { Notifications } from 'src/libs/utils'
+import { fileDownload } from 'src/utils/FileDownload'
+import {
+  BLANK_TEMPLATE_FILENAME,
+  BLANK_TEMPLATE_MIME,
+  MAX_TEMPLATE_BYTES,
+  MAX_TEMPLATE_SIZE_MESSAGE,
+  buildBlankStudyTemplateV1,
+} from 'src/libs/studyTemplate/studyTemplateV1Csv'
+import { STUDY_DATASET_DRAFT_TYPE, TemplateValidationError } from 'src/types/studyTemplate'
+
+const FILE_INPUT_ID = 'study-template-file'
+
+const describeLocation = (error: TemplateValidationError): string | null => {
+  if (error.row === undefined) return null
+  return error.column === undefined ? `Row ${error.row}` : `Row ${error.row}, column ${error.column}`
+}
+
+export const StudyTemplateUpload = (): React.JSX.Element => {
+  const navigate = useNavigate()
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  const [file, setFile] = useState<File | null>(null)
+  const [validationErrors, setValidationErrors] = useState<TemplateValidationError[]>([])
+  const [truncated, setTruncated] = useState(false)
+
+  // Failures are toasts, matching the rest of the app. Validation errors are not failures: they are a
+  // completed result the user works through while editing their file, so they stay on the page.
+  const clearResults = useCallback(() => {
+    setValidationErrors([])
+    setTruncated(false)
+  }, [])
+
+  const handleDownload = useCallback(() => {
+    fileDownload(buildBlankStudyTemplateV1(), BLANK_TEMPLATE_FILENAME, BLANK_TEMPLATE_MIME)
+  }, [])
+
+  const handleFileChange = useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+    const selected = event.currentTarget.files?.[0] ?? null
+    clearResults()
+    if (!selected) {
+      setFile(null)
+      return
+    }
+    // Pre-checks the two limits Consent enforces, so a doomed 5 MiB upload never leaves the browser.
+    // Wording matches the server's so the two layers cannot appear to disagree.
+    if (!selected.name.toLowerCase().endsWith('.csv')) {
+      setFile(null)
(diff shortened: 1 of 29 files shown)


src/pages/data_submission/StudyTemplateUpload.tsx
```diff
@@ -0,0 +1,221 @@
...
+    setFile(null)
+    clearResults()
+    // Resetting the element lets the user re-select the same filename and still get a change event.
+    if (inputRef.current) {
+      inputRef.current.value = ''
+    }
+  }, [clearResults])
+
+  const handleValidate = useCallback(async () => {
+    if (!file) return
```
otchet-broad, line 85: Two issues codex found in this method:

 The primary successful-validation path navigates to an unregistered route and therefore cannot reach the draft form. The upload page also applies stale asynchronous results after the selected file changes.

  Full review comments:

  - [P1] Register a route for the generated draft URL — /workspaces/duos-ui/src/pages/data_submission/StudyTemplateUpload.tsx:96-96
    Every successful validation navigates to /data_submission_form/draft/study-dataset/:id, but AppRoutes only defines the child :studyId under /data_submission_form. React Router therefore sends this multi-segment URL to the wildcard NotFound route, breaking
    the workflow after every successful upload; add a matching route or navigate to a supported URL.

  - [P2] Ignore validation results after the selected file changes — /workspaces/duos-ui/src/pages/data_submission/StudyTemplateUpload.tsx:85-87
    While this request is pending, the file chooser and Remove control remain active. If the user validates file A and then selects or removes it before the response arrives, this continuation still displays A's errors or navigates to A's draft even though
    the UI now shows file B or no file; disable file changes while validating or discard responses whose submitted file is no longer current.

kevinmarete: Thanks — P2 is a real bug and is fixed in fee511b0; P1 I'd like to push back on, because the route it names is the next ticket's.

**P2 (stale results): fixed.** The picker and Remove deliberately stay live during a request, so a response could outlive the file it belongs to — A's errors rendering beside B's filename, A's request failure toasting against an empty page, or A's draft navigating away from B. `handleValidate` now captures the file it submitted and compares it against a ref holding the current selection before touching state, in both the resolve and the reject path. I discarded the response rather than freezing the controls: the user has already told us they abandoned that file, so re-enabling their own choice after the fact is the worse of the two. Three tests cover it (replace mid-flight, remove mid-flight before success, remove mid-flight before failure), each verified to fail without the guard.

**P1 (unregistered route): intentional, and not reachable.** Registering `/data_submission_form/draft/study-dataset/:draftId` is explicitly DT-3872's first line of work, together with the draft mode in `DataSubmissionFormV2` that gives the URL something to render. Adding the route here without that hydration wouldn't fix the workflow — React Router would resolve it to the blank create form, so a user whose template just validated would land on an empty registration form as though their file had been discarded. That's a quieter failure than a NotFound, not a better one, and DT-3872 would immediately rewrite it.

Nothing can reach either outcome from this branch, either: `valid: true` requires the validation endpoint, which DT-3871 hasn't shipped — the Rollout section above is explicit that this must not be released before it, and DT-3873 owns sequencing the two. The success path is unreachable until both land, and DT-3872 is what makes the destination exist.

Happy to be wrong if you'd rather see the route registered now as a marker, but it would be a route to a form that discards the draft, so my preference is to leave the navigation target as the contract DT-3872 implements against.

otchet-broad (approved):
👍
Thanks for addressing the feedback!  I'll look for the route item in the other PR.

fboulnois (approved):
👍

kevinmarete:
Review of DT-3871 ([consent#3028](https://github.com/DataBiosphere/consent/pull/3028)) moved two things this branch describes, and `2e097ba9` follows them. No client code changes.

**A valid template is now answered 201 with a `Location` for the created draft**, rather than 200. Errors still answer 200 with `valid: false`. Nothing here needs to change to accept that: `fetchMultipart` gates on `res.ok`, so the whole 2xx range reaches the caller as data, and this page already discriminates on `valid` plus the draft's type rather than on the status. The JSDoc on `Draft.validateStudyDatasetTemplate` now says so explicitly, so a later status check that a 201 would fail does not get added by accident.

**The 5 MiB limit is enforced by the validator that reads the upload**, not by the resource ahead of it. The endpoint still answers `413` with the same wording this page pre-checks with, so the toast and the browser-side check are unaffected — only where consent enforces it moved.

The plan doc's sequence diagram, boundary paragraph, and acceptance bullet now match, along with the one bullet in the description above. `tsc --noEmit` and `eslint` are clean; no test changed, since none assert a success status.

PR #3875, opened 2026-08-20
[DT-3872] Load validated templates into the draft study form

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3872

Security risk: **low** — no new endpoint, no new permission. It reads and deletes a draft the caller already owns through the existing generic draft endpoints, under the same `RoleBAC` guard as the rest of the submission workflow, and refuses to map a draft that is not a study/dataset draft. Nothing new is stored; the document a draft holds is one this app already submits.

### Summary

Ticket 5 of the study-template workflow planned in `docs/plans/study-template-validation-workflow.md`. Gives a validated template somewhere to land: `/data_submission_form/draft/study-dataset/:draftId` loads the draft [DT-3871](https://github.com/DataBiosphere/consent/pull/3028) created, maps its registration-shaped document into the study form, and creates the study from what the user edits there.

**This is the other half of [#3872](https://github.com/DataBiosphere/duos-ui/pull/3872).** That PR's success path routes here; until this lands, a valid template creates a draft and then dead-ends on a route that does not exist.

Six commits, each reviewable on its own:

- **the typed client and the response it maps** — `getDraft`/`deleteDraft` beside the validation call, the envelope types, and the guard that reads `meta.draftType`.
- **the reverse mapper** — `datasetSchemaSubmissionToStudy`, placed beside `studyToDatasetSchemaSubmission`, with round-trip specs in both directions.
- **draft mode** — the form had two paths a URL segment distinguished; a draft is neither, so the mode is explicit and the route is registered rather than falling through to the study-id one.
- **submission and cleanup** — the draft is removed once the study exists, and only then.
- **hydration and regression specs** — through the real form sections, and around the two paths draft mode was added beside.

### What it looks like

Both captured against [consent#3028](https://github.com/DataBiosphere/consent/pull/3028) running locally: a real template uploaded through #3872's page, validated by the real endpoint, and the draft it created loaded here.

**1. The draft, loaded.** The title says draft rather than form, and the values are the template's — study name, description, data types.

draft-hydrated-top

**2. The dataset the template described**, as a study asset, editable like any other. `Create Study` sits at the bottom of the same form — a draft has never been submitted, so it creates rather than updates.

draft-hydrated-datasets

Worth a reviewer's eye:

- **A property is written only where the document carries a value.** Consent serializes the draft document with `NON_NULL`, so a field the producer left empty is *absent* rather than `null` — and writing an empty property for it would submit a value they never gave. This is the single most load-bearing decision in the mapper, and the fixture that proves it is captured rather than hand-written: a hand-made fixture full of nulls would let a mapper pass its own tests and fail on a real draft.
- **The draft type is checked before the document is mapped.** The read endpoint is generic across draft types, so the id says nothing about what it holds. A draft of another type, or one that does not say what it is, replaces the form with a recoverable error rather than rendering something submittable.
- **Creation and cleanup are not one operation.** The study is created from what the form holds; only then is the draft removed. A failed creation leaves the draft alone, which is what makes trying again possible, and a failed removal is reported on its own rather than turning a study that exists into a failure.
- **`fetchAdapter` retries 5xx three times for every request**, this one included. That is transport-level and predates this work. What the ticket rules out, and what this does not do, is a retry of the cleanup as a workflow step.
- **Based on `km-dt-1854-upload-template-handling`, not `develop`**, because it extends the same client and types that PR introduces. It needs a rebase once #3872 merges, and it should not merge before consent#3028 is approved, since the document shape is what it is built against.

### Testing

`pnpm run lint` and `pnpm run type-check` clean. `pnpm test` green at **360 files / 4039 tests**, including 41 new ones: the client and its envelope, the mapper in both directions, the draft loader's refusals, the form across all three of its modes, hydration through the real sections, and submission with cleanup.

The ones that would catch a regression rather than restate the code:

- the round trip asserts *key sets*, not just values, so a field the document left out cannot reappear as a null the producer never gave;
- the submitted payload is parsed out of the multipart body and checked to have come through the `Study` model, so a shortcut that posted the loaded draft straight back would fail;
- the file control is asserted empty after hydration and then given a file, which arrives in the submission — a study can carry a sharing plan the template never could;
- the persisted-study path is asserted to still load by study id, still offer Update, and still report a failed load inline, since only the draft path replaces the form.

`pnpm run test:e2e` is unchanged from #3872: the `auth.spec.ts` failures are missing `DUOS_AUTOMATION_*_SA` credentials, environmental and pre-existing. No e2e spec covers this route yet — that is [DT-3873](https://broadworkbench.atlassian.net/browse/DT-3873).

----

- Verify all tests go green

[DT-3871]: https://broadworkbench.atlassian.net/browse/DT-3871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3873]: https://broadworkbench.atlassian.net/browse/DT-3873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routing/AppRoutes.tsx
```diff
@@ -96,6 +96,7 @@ const AppRoutes = (props: AppRoutesProps) => {
           <Route path="/data_submission_form" element={<DataSubmissionFormV2 />}>
             <Route path=":studyId" element={<DataSubmissionFormV2 />} />
           </Route>
+          <Route path="/data_submission_form/draft/study-dataset/:draftId" element={<DataSubmissionFormV2 />} />
```
otchet-broad: Would it be better to indicate the drafId expected is a UUID?

```suggestion
          <Route path="/data_submission_form/draft/study-dataset/:draftUUId" element={<DataSubmissionFormV2 />} />
```

otchet-broad: Or just a UUID?
```suggestion
          <Route path="/data_submission_form/draft/study-dataset/:uuid" element={<DataSubmissionFormV2 />} />
```

otchet-broad (approved):
Looks good overall.  Left you some options to consider for the app routes to help future readers understand this endpoint expects a UUID, not an integer (like most of the other routes expect).

fboulnois:
A few findings:

### 1. Empty submit deletes the validated draft
`src/pages/data_submission/v2/DataSubmissionFormV2.tsx:200`

Draft mode has no loading state, so the blank form and an enabled **Create Study** button render before `loadStudyDatasetDraft` resolves. Clicking in that window submits `{studyName: "", dataTypes: [], consentGroups: []}` to `registerDataset` and, on success, calls `Draft.deleteDraft("abc")`, destroying the validated draft that seeded the form.

### 2. Explicit `false` from a template is dropped on submit
`src/pages/data_submission/v2/v2-common-functions.tsx:203,214,216`

The new inverse mapper `datasetSchemaSubmissionToStudy` (line 266) deliberately preserves `false`, but the forward mapper's `... as boolean || undefined` converts it back to `undefined`. So `multiCenterStudy`, `submittingToAnvil`, and `controlledAccessRequiredForGenomicSummaryResultsGSR` set to `false` in a template vanish from the `registerDataset` payload while the form visibly shows "No".

The new round-trip spec only exercises `true`, so it doesn't catch this.

### 3. Rejected file can't be re-picked
`src/pages/data_submission/StudyTemplateUpload.tsx:62`

On a `.csv`-extension or size pre-check rejection the hidden input's `value` isn't reset — unlike `handleRemove` (line 80), which does. Re-selecting the same filename fires no `change` event, so the stale error stays on screen and the page appears dead.

fboulnois (approved):
👍

src/routing/AppRoutes.tsx
kevinmarete: Agreed — went with `:draftUuid` rather than `:uuid` so the parameter still says *what* the UUID identifies, and it keeps the camelCase of the neighbouring `:studyId`/`:collectionId`. Renamed through the component, `Draft` client and docs.

kevinmarete: Used `:draftUuid` — same idea, but the name keeps the subject as well as the shape.

kevinmarete:
Thanks — all three were real. Fixed in 6734b7f:

**1. Empty submit deletes the validated draft.** Draft mode now renders a spinner until the load settles, so the form and its Create Study button do not exist during that window. The gate is derived (`loadedDraftUuid !== draftUuid`) rather than a flag set inside the effect, which also covers navigating straight from one draft to another. Covered by `offers nothing to submit until the draft has loaded`.

**2. Explicit `false` dropped on submit.** Added `getBooleanStudyPropertyValueByKey`, which returns the boolean as-is. Applied to all five boolean fields — the three you flagged plus `alternativeDataSharingPlan` and `alternativeDataSharingPlanDataReleased`, which had the same `|| undefined`. The round-trip spec now exercises `false`.

**3. Rejected file cannot be re-picked.** Extracted `clearSelection`, which clears the state *and* resets `inputRef.current.value`, and used it on both pre-check rejections; `handleRemove` reuses it. Covered by `resets the input after a rejection so the same file can be re-picked`.

PR #3876, opened 2026-08-20
[DT-3866] Align duos-ui with the canonical Data Use classification

kevinmarete:
### Addresses

[DT-3866](https://broadworkbench.atlassian.net/browse/DT-3866). Depends on [DT-3862](https://broadworkbench.atlassian.net/browse/DT-3862) and [DT-3864](https://broadworkbench.atlassian.net/browse/DT-3864), both merged in Consent.

Security risk: low.

### Summary

Consent persists matches as `v5` and rewrites them on every DAR create and update, but duos-ui gated on `algorithmVersion === 'v3'` — so whenever a bucket's data use tripped the client's own `isOther`/`shouldAbstain` heuristic the match values were dropped and the bucket rendered **N/A**, hiding the ABSTAIN and the rationale `DataUseMatcherV5` writes for DAC members on Other-only, NONE/null, and every MULTIPLE primary shape.

Now `v3`–`v5` are trusted because Consent decides abstention itself, `v1`/`v2` keep the client-side suppression they were written for, and an unknown version, a missing version, or mixed versions within one bucket render "Unable to interpret the system match" — neither silently trusted nor hidden. The heuristic also over-fired on secondary modifiers (`PUB`, `IRB`, `COL`, `GSO`, `MOR`, `POP-*`) that V5 matches normally, so it is now scoped to the legacy path.

Also reconciled:

- Study **update** rendered Consent's newline-joined violations as one unreadable line, then reloaded the study — discarding the user's edits along with the violations they had to act on. Both paths share a renderer; a 400 no longer reloads, a 5xx still does.
- `selectedPrimaryGroup` precedence now follows `DataUsePrimaryCategory` order (POA before DS).
- A primary Other is a permission; secondary stays a modifier. No display change — that `type` field feeds only code-keyed acknowledgement checks.
- `AbstainDataUseCodes` was a numeric enum, so `Object.keys()` leaked reverse-mapped indices (`'0'`, `'1'`, …) into the code list it was compared against. Now a readonly array.
- `ConsentGroup.diseaseSpecificUse` is typed `string[]`, which is what every caller narrows in.

Retained deliberately, documented and tested: the Library grid renders every primary a legacy record carries, and `processDefinedLimitations` shows only the narrowest permission rather than the classifier's full enumeration.

Decisions recorded on the ticket: no operator UI for the curator-assisted legacy process, which stays backend-only. This should land before the legacy-record migration, which recomputes persisted matches under `v5` at scale.

### Testing

`pnpm run lint`, `pnpm run type-check`, and `pnpm test` (355 files, 4044 tests) pass. New coverage: v5 ABSTAIN across all five shapes with Consent's verbatim rationales, the eight secondary codes the heuristic over-fired on, per-version handling, unknown/missing/mixed versions, both registration paths against a simulated 400, primary radio mutual exclusivity, and canonical primary ordering.

[DT-3866]: https://broadworkbench.atlassian.net/browse/DT-3866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3862]: https://broadworkbench.atlassian.net/browse/DT-3862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3864]: https://broadworkbench.atlassian.net/browse/DT-3864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/libs/dataUseTranslation.ts
+++ b/src/libs/dataUseTranslation.ts
@@ -369,8 +369,8 @@ export const processRestrictionStatements = async (
 }
 
 /**
- * Shows only the narrowest permission when a record carries several primaries (DS over HMB, those
- * over GRU). Unlike the classifier's full enumeration: a broad statement beside a narrow one misleads.
+ * Shows only the narrowest permission along the DS > HMB > GRU chain, since a broad statement beside
+ * a narrow one misleads. POA sits outside that chain and still renders alongside DS.
  */
 export const processDefinedLimitations = (
   key: string,
--- a/src/libs/dataUseTranslation.ts
+++ b/src/libs/dataUseTranslation.ts
@@ -368,6 +368,10 @@ export const processRestrictionStatements = async (
   return processDiseaseRestrictionsWithUrls(diseaseValue as string[])
 }
 
+/**
+ * Shows only the narrowest permission of a legacy multi-primary record (DS over HMB, those over
+ * GRU). Unlike the classifier's full enumeration: a broad statement beside a narrow one misleads.
+ */
 export const processDefinedLimitations = (
   key: string,
   dataUse: DataUse,
@@ -412,7 +416,8 @@ const processOtherInDataUse = (
       Promise.resolve({
         code: 'OTH1',
         description: `Primary Other: ${isEmpty(dataUse.other) ? 'Not provided' : dataUse.other}`,
-        type: ControlledAccessType.modifiers,
+        // A permission, like OTHER in the classifier and srpTranslations.other; only OTH2 is not.
+        type: ControlledAccessType.permissions,
       }),
     )
   }
--- a/src/libs/dataUseTranslation.ts
+++ b/src/libs/dataUseTranslation.ts
@@ -369,8 +369,8 @@ export const processRestrictionStatements = async (
 }
 
 /**
- * Shows only the narrowest permission of a legacy multi-primary record (DS over HMB, those over
- * GRU). Unlike the classifier's full enumeration: a broad statement beside a narrow one misleads.
+ * Shows only the narrowest permission when a record carries several primaries (DS over HMB, those
+ * over GRU). Unlike the classifier's full enumeration: a broad statement beside a narrow one misleads.
  */
 export const processDefinedLimitations = (
   key: string,
--- a/src/types/model.ts
+++ b/src/types/model.ts
@@ -25,7 +25,11 @@ export type UserRoleName
     | 'ServiceAccount'
     | 'All'
 
-export enum AbstainDataUseCodes {
+/**
+ * Secondary data use codes that suppress a legacy (v1/v2) system match. Scoped to legacy results
+ * on purpose: from v3 Consent decides abstention itself — see `calculateAlgorithmResultForBucket`.
+ */
+export const AbstainDataUseCodes: readonly string[] = [
   'OTHER',
   'POP-M',
   'POP-F',
@@ -35,7 +39,7 @@ export enum AbstainDataUseCodes {
   'PUB',
   'MOR',
   'POP-PD',
-}
+]
 
 export interface UserRole {
   roleId: number
--- a/src/types/model.ts
+++ b/src/types/model.ts
@@ -29,7 +29,7 @@ export type UserRoleName
  * Secondary data use codes that suppress a legacy (v1/v2) system match. Scoped to legacy results
  * on purpose: from v3 Consent decides abstention itself — see `calculateAlgorithmResultForBucket`.
  */
-export const AbstainDataUseCodes: readonly string[] = [
+export const AbstainDataUseCodes: ReadonlySet<string> = new Set([
   'OTHER',
   'POP-M',
   'POP-F',
@@ -39,7 +39,7 @@ export const AbstainDataUseCodes: readonly string[] = [
   'PUB',
   'MOR',
   'POP-PD',
-]
+])
 
 export interface UserRole {
   roleId: number
--- a/src/components/CollectionAlgorithmDecision.tsx
+++ b/src/components/CollectionAlgorithmDecision.tsx
@@ -24,19 +24,18 @@ const OtherResult = (otherResultProps: Readonly<OtherResultProps>) => {
 }
 
 const getResult = (resultValue: string | undefined): React.ReactElement => {
-  if (resultValue?.toLowerCase().trim() === 'abstain') {
-    return <AbstainResult />
+  const result = resultValue?.trim()
+  switch (result?.toLowerCase()) {
+    case 'abstain':
+      return <AbstainResult />
+    case 'yes':
+      return <YesResult />
+    case 'no':
+      return <NoResult />
+    default:
+      // Anything else is the caller's own wording, such as a match no longer interpretable.
+      return <OtherResult text={result || 'N/A'} />
   }
-  if (resultValue && resultValue.trim().length > 0) {
-    switch (resultValue.toLowerCase().trim()) {
-      case 'yes':
-        return <YesResult />
-      case 'no':
-        return <NoResult />
-      default:
-    }
-  }
-  return <OtherResult text="N/A" />
 }
 
 export default function CollectionAlgorithmDecision(props: Readonly<CollectionAlgorithmDecisionProps>) {
--- a/src/components/consent_group_list/ConsentGroupAddEdit.tsx
+++ b/src/components/consent_group_list/ConsentGroupAddEdit.tsx
@@ -69,10 +69,14 @@ export default function ConsentGroupAddEdit(props: ConsentGroupAddEditProps): Re
   const [showGSText, setShowGSText] = useState(!isNil(consentGroup?.gs))
   const [gsText, setGSText] = useState(consentGroup?.gs || undefined)
 
-  const [showOtherPrimaryText, setShowOtherPrimaryText] = useState(!isNil(consentGroup?.otherPrimary))
+  // Follow the radio `selectedPrimaryGroup` lights: reading each field alone checks one primary
+  // beside another's input.
+  const initialPrimaryGroup = selectedPrimaryGroup(consentGroup as ConsentGroup)
+
+  const [showOtherPrimaryText, setShowOtherPrimaryText] = useState(initialPrimaryGroup === 'otherPrimary')
   const [otherPrimaryText, setOtherPrimaryText] = useState(consentGroup?.otherPrimary || undefined)
 
-  const [showDiseaseSpecificUseSearchbar, setShowDiseaseSpecificUseSearchbar] = useState(!isEmpty(consentGroup?.diseaseSpecificUse))
+  const [showDiseaseSpecificUseSearchbar, setShowDiseaseSpecificUseSearchbar] = useState(initialPrimaryGroup === 'diseaseSpecificUse')
   const [selectedDiseases, setSelectedDiseases] = useState<{ displayText: string, id: string }[]>([])
 
   const diseaseSpecificUse = consentGroup?.diseaseSpecificUse
@@ -146,7 +150,7 @@ export default function ConsentGroupAddEdit(props: ConsentGroupAddEditProps): Re
     setValidation(calcErrors(next))
   }
 
-  const onPrimaryChange = ({ key, value }: { key: string, value: boolean | string | string[] | { displayText: string, id: string }[] }) => {
+  const onPrimaryChange = ({ key, value }: { key: string, value: boolean | string | string[] }) => {
     const next = structuredClone(current)
     next.generalResearchUse = false
     next.hmb = false
@@ -158,14 +162,12 @@ export default function ConsentGroupAddEdit(props: ConsentGroupAddEditProps): Re
 
     setShowDiseaseSpecificUseSearchbar(key === 'diseaseSpecificUse')
     setShowOtherPrimaryText(key === 'otherPrimary')
-    if (!current.otherPrimary) {
+    if (!next.otherPrimary) {
       setOtherPrimaryText(undefined)
     }
-    if (!current.diseaseSpecificUse) {
+    if (isNil(next.diseaseSpecificUse)) {
       setSelectedDiseases([])
     }
-    setShowDiseaseSpecificUseSearchbar(key === 'diseaseSpecificUse')
-    setShowOtherPrimaryText(key === 'otherPrimary')
 
     setValidation(calcErrors(next))
   }
@@ -354,7 +356,7 @@ export default function ConsentGroupAddEdit(props: ConsentGroupAddEditProps): Re
                   onChange={({ value }: { value: string }) => {
                     onPrimaryChange({
                       key: value,
-                      value: selectedDiseases,
+                      value: selectedDiseases.map(disease => disease.id),
                     })
                   }}
                   validation={validation.primaryConsent}
@@ -415,7 +417,7 @@ export default function ConsentGroupAddEdit(props: ConsentGroupAddEditProps): Re
                   toggleText="Other"
                   defaultValue={selectedPrimaryGroup(current as ConsentGroup)}
                   onChange={({ value }: { value: string }) => {
-                    onPrimaryChange({ key: value, value: value })
+                    onPrimaryChange({ key: value, value: otherPrimaryText ?? '' })
                   }}
                   disabled={readOnly}
                   validation={validation.primaryConsent}
--- a/src/pages/data_submission/consent_group/consentGroupUtils.ts
+++ b/src/pages/data_submission/consent_group/consentGroupUtils.ts
@@ -3,10 +3,11 @@ import { DataLocationType } from 'src/pages/data_submission/v2/v2-models'
 import { FileStorageObject } from 'src/types/model'
 import { DatasetDataMetadata } from 'src/libs/data-metadata'
 
+/** The primary data use fields of a `ConsentGroup2`, which is what every caller narrows in. */
 export interface ConsentGroup {
   generalResearchUse?: boolean
   hmb?: boolean
-  diseaseSpecificUse?: boolean
+  diseaseSpecificUse?: string[]
   poa?: boolean
   otherPrimary?: string
 }
@@ -50,6 +51,11 @@ export interface FileType {
   functionalEquivalence: string
 }
 
+/**
+ * Which primary radio renders as selected. Precedence follows the classifier's category order so a
+ * legacy multi-primary record resolves the same way here as in Consent. One deliberate divergence:
+ * an empty `diseaseSpecificUse` counts as selected, keeping the radio lit while diseases are picked.
+ */
 export const selectedPrimaryGroup = (consentGroup: ConsentGroup) => {
   if (isNil(consentGroup)) return undefined
   if (!isNil(consentGroup.generalResearchUse) && consentGroup.generalResearchUse) {
@@ -58,11 +64,11 @@ export const selectedPrimaryGroup = (consentGroup: ConsentGroup) => {
   else if (!isNil(consentGroup.hmb) && consentGroup.hmb) {
     return 'hmb'
   }
-  else if (!isNil(consentGroup.diseaseSpecificUse)) {
-    return 'diseaseSpecificUse'
-  }
   else if (!isNil(consentGroup.poa) && consentGroup.poa) {
     return 'poa'
+  }
+  else if (!isNil(consentGroup.diseaseSpecificUse)) {
+    return 'diseaseSpecificUse'
   }
   else if (!isNil(consentGroup.otherPrimary) && isString(consentGroup.otherPrimary)) {
     return 'otherPrimary'
--- a/src/pages/data_submission/v2/DataSubmissionFormV2.tsx
+++ b/src/pages/data_submission/v2/DataSubmissionFormV2.tsx
@@ -15,6 +15,7 @@ import { studyToDatasetSchemaSubmission, buildConsentGroupsFromStudy, getStudyPr
 import AsyncSpinnerButton from 'src/components/AsyncSpinnerButton'
 import { ConsentGroup2 } from 'src/pages/data_submission/consent_group/consentGroupUtils'
 import { ResponseError } from 'src/types/model'
+import { isEmpty } from 'src/utils/NodashUtil'
 
 export type FileProperty = {
   key: string
@@ -27,8 +28,15 @@ export type DataSubmissionFormV2Props = {
 
 export const ALTERNATIVE_DATA_SHARING_PLAN_FILE = 'alternativeDataSharingPlanFile'
 
-const isValidationRejection = (error: unknown): boolean =>
-  (error as ResponseError)?.response?.status === 400
+/**
+ * Only a 400 carrying Consent's own explanation is a rejection the submitter can act on in the form.
+ * One with no readable body — a malformed multipart, a proxy — carries the generic help-desk line
+ * instead, which is not a violation list and still needs the reload.
+ */
+const isValidationRejection = (error: unknown): boolean => {
+  const response = (error as ResponseError)?.response
+  return response?.status === 400 && !isEmpty(response.data?.message?.trim())
+}
 
 /** A rejection reaches us either as a thrown Error or as the raw response it was built from. */
 const errorMessage = (error: unknown): string => {
--- a/src/pages/data_submission/v2/DataSubmissionFormV2.tsx
+++ b/src/pages/data_submission/v2/DataSubmissionFormV2.tsx
@@ -30,9 +30,15 @@ export const ALTERNATIVE_DATA_SHARING_PLAN_FILE = 'alternativeDataSharingPlanFil
 const isValidationRejection = (error: unknown): boolean =>
   (error as ResponseError)?.response?.status === 400
 
+/** A rejection reaches us either as a thrown Error or as the raw response it was built from. */
+const errorMessage = (error: unknown): string => {
+  if (error instanceof Error) return error.message
+  return (error as ResponseError)?.response?.data?.message ?? String(error)
+}
+
 /** Consent joins Data Use consistency violations with newlines, so each needs its own line here. */
 const renderViolations = (prefix: string, error: unknown): React.ReactElement => {
-  const message = error instanceof Error ? error.message : String(error)
+  const message = errorMessage(error)
   return (
     <>
       {prefix}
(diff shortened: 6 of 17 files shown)


src/libs/dataUseTranslation.ts
```diff
@@ -368,6 +368,10 @@ export const processRestrictionStatements = async (
   return processDiseaseRestrictionsWithUrls(diseaseValue as string[])
 }
 
+/**
+ * Shows only the narrowest permission of a legacy multi-primary record (DS over HMB, those over
```
otchet-broad: Do we have any of these legacy records in production?

src/types/model.ts
```diff
@@ -25,7 +25,11 @@ export type UserRoleName
     | 'ServiceAccount'
     | 'All'
 
-export enum AbstainDataUseCodes {
+/**
+ * Secondary data use codes that suppress a legacy (v1/v2) system match. Scoped to legacy results
```
otchet-broad, line 29: Similar to my earlier comment, it would be good to know if we're dealing with artifacts that are in the production database vs the dev database for handling these legacy results.  If they're in production, seems like a reasonable approach.  If they're not, I'd suggest we remove the 'legacy' data from develop/staging and simplify our lives.

fboulnois (approved):
A few findings:

### 1. Silent file loss on resubmit after a validation 400 — blocking

`src/pages/data_submission/v2/DataSubmissionFormV2.tsx:106` is the new branch that keeps form state instead of reloading on a 400. That's the right call for surfacing Data Use violations, but it collides with `buildMultiPartFormData`, which mutates the live `study` state:

- `DataSubmissionFormV2.tsx:81` — `delete study.alternativeDataSharingPlanFile`
- `DataSubmissionFormV2.tsx:88` — `delete consentGroup.addedNIHInstitutionalCertificationFile`

Only the JSON payload gets a `structuredClone` (`:91`), and it happens *after* the deletes. Meanwhile `NihDataManagement.tsx:158` passes the file as `defaultValue`, so the upload widget is uncontrolled and keeps displaying the attached filename.

Repro: attach an alternative data sharing plan (or an NIH institutional certification), save, get a Data Use consistency 400, fix the codes, save again. The second request carries no file parts. The study saves successfully with the attachment missing, and nothing warns the user.

Fix is small — either clone at the top of `buildMultiPartFormData` and read the files off the original, or drop the two `delete`s and let `studyToDatasetSchemaSubmission` strip the `File` fields from its own clone.

Worth noting this bug pre-exists on the **create** path (`onError` never reloaded either), but this PR extends it to **update**, which is the flow where an already-saved study can be silently degraded.

### 2. Missing `algorithmVersion` now suppresses the system suggestion — needs a data answer

`src/utils/BucketUtils.ts:198` — `algorithmVersion` is optional on `MatchResult`
(`src/types/model.ts:974`). The old code only special-cased `'v3'`; everything else, `undefined` included, fell through to the legacy path and still rendered Yes/No. Now `classifyAlgorithmVersion` returns `'unrecognized'` for a missing version *and* for any bucket holding rows from two different match runs, so the DAC sees "Unable to interpret the system match" instead of a suggestion.

The tests assert this deliberately, so it reads as intended, but can you confirm no production `match_result` rows carry a null `algorithm_version`? If older Consent matches predate that column, DACs lose suggestions on existing collections. Treating `undefined` as `v1`/`v2` would be the conservative alternative.


### 3. "Unable to interpret" renders as a bare N/A — cosmetic

`src/utils/BucketUtils.ts:189` sets `result` to the `UNRECOGNIZED_ALGORITHM_RESULT` sentinel, but `getResult` in `src/components/CollectionAlgorithmDecision.tsx:26-40` only handles `abstain`/`yes`/`no` and falls through to `<OtherResult text="N/A" />`. The Decision line ends up identical to the no-match-recorded case; only the Reason line distinguishes them. Either add a case for the sentinel, or let `getResult` pass unrecognized non-empty values through as-is.

src/libs/dataUseTranslation.ts
kevinmarete: Yes, in production. [DT-3861](https://broadworkbench.atlassian.net/browse/DT-3861) audited all 5,811 persisted production datasets: exactly one is `MULTIPLE(HMB+OTHER)` (controlled), and it carries a DAR reference plus two persisted match rows. Alongside it are 566 `SINGLE(OTHER)` and 509 open-access `NONE`. So this path fires on a real production record, not a dev artifact.

That record is also why it cannot simply be normalized away: the domain review on DT-3861 labelled it **Mixed**, and recorded that mixed cases must not be normalized without an explicit domain-approved disposition. [consent#3029](https://github.com/DataBiosphere/consent/pull/3029) ([DT-3863](https://broadworkbench.atlassian.net/browse/DT-3863)) proposes one and is waiting on the Data Use owners; until it lands, V5 abstains on the record and this UI still has to render it sensibly.

One fix to my own comment either way: `processDefinedLimitations` is not legacy-scoped — `processRestrictionStatements` calls it for every non-`diseaseRestrictions` key, so it runs on current records too. Only the multi-primary collapse is the legacy shape. Reworded in 3f99ad4.

src/types/model.ts
kevinmarete: Production, per the same [DT-3861](https://broadworkbench.atlassian.net/browse/DT-3861) audit. It found 1,085 persisted match rows in production — 669 mapped to a dataset, 416 whose `consent` reference is noncanonical (that reference is [DT-3865](https://broadworkbench.atlassian.net/browse/DT-3865)). Of the mapped rows, **52 across 31 datasets sit on explicit ABSTAIN shapes** (39 involving Other, 13 involving NONE), and 19 of those datasets have DAR references. Those are the rows this client-side suppression is protecting.

On simplifying: agreed as the destination, and it is being approached from the backend rather than here. [consent#3029](https://github.com/DataBiosphere/consent/pull/3029) ([DT-3863](https://broadworkbench.atlassian.net/browse/DT-3863)) recomputes the affected stored matches under V5, so once it runs in production those rows carry v5 and stop taking this path. It cannot cover the 416 rows whose consent reference does not resolve until DT-3865 fixes that join, so deleting the legacy branch now would drop suggestions on collections that still have them. Worth a follow-up ticket to remove it once both have run in production.

Related: fboulnois asked the adjacent question about a *missing* version. Changeset 87.0 backfilled pre-existing rows to `v1`, so 3f99ad4 now reads an absent version as legacy rather than uninterpretable — no collection loses its suggestion.

kevinmarete: Correction to my reply above: the ticket for reconciling the unresolvable `consent` references is [DT-3942](https://broadworkbench.atlassian.net/browse/DT-3942) (New), not DT-3865. I repeated the pointer written in DT-3861's audit comment, but DT-3865 turned out to be the alias-sequence work and is already Completed — so as written my reply reads as though that blocker were cleared. It is not; the legacy branch still cannot be deleted yet.

Follow-ups now tracked, both under DT-3860: [DT-4008](https://broadworkbench.atlassian.net/browse/DT-4008) removes the client-side suppression here once DT-3863's recompute and DT-3942 have run in production, and [DT-4009](https://broadworkbench.atlassian.net/browse/DT-4009) retires the migration endpoints DT-3863 adds on the consent side.

otchet-broad (commented):
Claude feedback:

# Code review — DT-3866: Align duos-ui with the canonical Data Use classification

Verification alongside the review: `tsc --noEmit` clean, `oxlint` clean on all changed `src` files,
full vitest suite passes (5,347 tests, run in chunks — a single full run OOMs in this container,
which is an environment limit rather than a failure).

## Blocking — primary data use path

### 1. `onPrimaryChange` reads the stale `current` instead of `next`

`src/components/consent_group_list/ConsentGroupAddEdit.tsx:161,164`

The clearing logic checks `current` after `next` has already been built, so re-selecting
Disease-Specific writes ontology *option objects* into `diseaseSpecificUse` — directly
contradicting the `string[]` type this diff introduces.

Repro: load a consent group with `diseaseSpecificUse: ['DOID_1']` (so `selectedDiseases` is
`[{displayText: 'Breast Cancer', id: 'DOID_1'}]`), click HMB, click Disease-Specific again, Save.
At line 164 `current.diseaseSpecificUse` is still `['DOID_1']` (truthy), so `setSelectedDiseases([])`
never runs; the DS radio's `onChange` then passes `selectedDiseases` straight through and the group
saves as `diseaseSpecificUse: [{"displayText":"Breast Cancer","id":"DOID_1"}]` where Consent expects
DOID strings. The same staleness leaves `otherPrimaryText` uncleared at line 161.

Reading `next` on both lines fixes both.

### 2. The "Other" radio writes its own field name into the free-text field

`src/components/consent_group_list/ConsentGroupAddEdit.tsx:418`

`onPrimaryChange({ key: value, value: value })` sets `otherPrimary = 'otherPrimary'`.

Repro: fill name/access management/participants, click `#primaryConsent_otherPrimary`, Save without
typing anything. `calcErrors`' check `showOtherPrimaryText && !cg.otherPrimary?.trim()` passes
because the sentinel is a non-empty string, so the literal `"otherPrimary"` is submitted to Consent
as the dataset's primary Other restriction text. `selectedPrimaryGroup`'s `isString(otherPrimary)`
guard papers over the sentinel rather than rejecting it.

Pre-existing, but it is the exact primary-data-use path this PR is aligning, and the new exclusivity
tests always type text so they never hit it.

### 3. `diseaseRestrictions` is loaded with no `isEmpty` guard

`src/pages/data_submission/v2/v2-common-functions.tsx:254`

`buildConsentGroupsFromStudy` assigns `diseaseRestrictions` directly, unlike `gs` and `otherPrimary`
immediately around it, so the diff's new "empty array counts as selected" rule leaks into the load
path and lights the DS radio for records whose real primary is Other or POA.

Repro: a stored dataset serializes `dataUse.diseaseRestrictions: []` (rather than omitting it)
together with `other: 'Not for profit'`. On edit, `consentGroup.diseaseSpecificUse === []`, so
`selectedPrimaryGroup` returns `'diseaseSpecificUse'` (`isNil([])` is false) and the DS radio renders
selected, while `showDiseaseSpecificUseSearchbar` is false (`isEmpty([])`) and `showOtherPrimaryText`
is true — DS selected next to a populated Other text box. `calcErrors` raises nothing, so a save
round-trips both `diseaseSpecificUse: []` and `otherPrimary`, which Consent now rejects as multiple
primaries.

The fix belongs at load (`isEmpty(...) ? undefined : ...`), not in `selectedPrimaryGroup`.

### 4. Moving `poa` ahead of `diseaseSpecificUse` desyncs the radio from the disease searchbar

`src/pages/data_submission/consent_group/consentGroupUtils.ts:67`

`selectedPrimaryGroup` now returns `'poa'` for a legacy `poa: true, diseaseSpecificUse: ['DOID_1']`
record, but `ConsentGroupAddEdit.tsx:75` still initializes `showDiseaseSpecificUseSearchbar` from
`!isEmpty(consentGroup?.diseaseSpecificUse)` → true. The POA radio renders checked while the
Disease-Specific multiselect renders below it populated with 'Breast Cancer'. The curator has no way
to tell which primary will be saved, and `calcErrors` reports nothing.

The searchbar's initial state should be derived from `selectedPrimaryGroup` rather than from the
field independently.

## Should fix — algorithm version handling

### 5. Mixed versions of the same handling class suppress a valid suggestion

`src/utils/BucketUtils.ts:198`

`classifyAlgorithmVersion` demands exactly one distinct version string, so a bucket mixing two
versions that share the same handling class (e.g. v3 + v5) is reported as uninterpretable even though
every version in the mix is interpretable and all matches agree.

Scenario: a bucket holds two datasets with identical data use; dataset A was matched under v3 and
dataset B re-matched under v5, both `match: true`. The old code read
`matchResults[0].algorithmVersion`, saw `'v3'`, and displayed "Yes". Now `versions.length !== 1`
returns `'unrecognized'` and the DAC sees "Unable to interpret the system match", suppressing a valid
suggestion. The rationale is also wrong for this case ("recorded by algorithm version v3, v5, which
DUOS can no longer interpret" — v5 *is* the current version), and `createDate` is taken arbitrarily
from `matchResults[0]`.

Classifying each version and falling back only when the *classes* differ preserves the decision.

### 6. `BACKEND_ABSTAINING_VERSIONS` is a closed allowlist

`src/utils/BucketUtils.ts:184`

The next backend algorithm version turns every system suggestion into an error message until the UI
is redeployed.

Scenario: Consent ships `DataUseMatcherV6` and stamps `algorithmVersion: 'v6'`.
`classifyAlgorithmVersion` falls through both Sets and returns `'unrecognized'`, so every bucket in
every DAC's voting view shows "Unable to interpret the system match" with "which DUOS can no longer
interpret" — even though v6 abstains server-side exactly like v3/v4/v5. The failure is silent (no
build or test break) and affects all DACs until duos-ui releases.

Parsing the numeric suffix and treating `>= 3` as backend-abstaining removes the coupling; only
v1/v2 need enumerating.

## Should fix — error rendering

### 7. `renderViolations` can render `[object Object]`

`src/pages/data_submission/v2/DataSubmissionFormV2.tsx:34`

`isValidationRejection` only tests `error.response.status === 400`, which a plain object
`{response: {status: 400, data: {message: '...'}}}` satisfies. `renderViolations` then takes the
non-`Error` branch and shows "Study update failed:" followed by `[object Object]`, discarding the
violation list the user needs — and because the 400 branch also skips the reload, they are left with
no information at all.

Reading `(error as ResponseError).response?.data?.message` before falling back to `String(error)`
covers both shapes.

### 8. The create path renders server faults in the Data Use violation format

`src/pages/data_submission/v2/DataSubmissionFormV2.tsx:133`

`DataSet.registerDataset` fails with a 500; `fetchAdapter` throws
`Error('Request failed with status 500')`, and `onError` renders it through `renderViolations` as
"Study creation failed:" plus a single bulleted-looking line, styled identically to a list of Data
Use consistency violations. The update path already discriminates with `isValidationRejection` — the
create path should use the same predicate so a server fault is not mistaken for a form problem the
submitter can fix.

### 9. Mixed-match display silently changed from "N/A" to a full sentence

`src/components/CollectionAlgorithmDecision.tsx:36`

Routing unknown results through `OtherResult` verbatim also changes a case the PR does not otherwise
touch. A bucket whose datasets matched both true and false has always produced
`'Unable to determine a system match'`; the old `getResult` fell through to `<OtherResult text="N/A" />`,
so the UI read "Decision: N/A" with the explanation in the Reason line. It now renders the whole
sentence in bold inside a `fontSize: 1.2rem` inline span next to the "Decision:" label, with no
wrapping or width constraint, inside `MultiDatasetVoteSlab`'s fixed-width slab.

If the change is intended it deserves its own test; the added spec only covers the new version string.

## Nits

- **Duplicate state setters** — `ConsentGroupAddEdit.tsx:158-159` and `:167-168` call
  `setShowDiseaseSpecificUseSearchbar` / `setShowOtherPrimaryText` twice with identical arguments.
  Harmless, but it reads as though the second pair compensates for the clearing logic in between,
  inviting a future edit to change one copy and not the other. Delete the second pair.

- **Docstring states an invariant POA breaks** — `src/libs/dataUseTranslation.ts:372` claims
  `processDefinedLimitations` "shows only the narrowest permission", but POA sits outside the
  DS > HMB > GRU chain and renders alongside DS — as the PR's own new test
  ('shows only the narrowest permission when every primary is set') asserts. Narrow the comment to
  the GRU/HMB/DS chain, or add POA to the precedence.

- **Dead `isOther` operand** — `src/utils/BucketUtils.ts:241`:
  `isOther(bucket.dataUse) || shouldAbstain(bucket.dataUse)`. `shouldAbstain`'s first clause is
  already `isOther(dataUse)`, so the left operand is dead.
  `const unmatchable = shouldAbstain(bucket.dataUse)` is equivalent.

- **Vacuous assertion / dead strip** — `src/pages/data_submission/v2/DataSubmissionFormV2.tsx:80`:
  the `alternativeDataSharingPlanFile: undefined` half of `withoutAttachments` is dead.
  `studyToDatasetSchemaSubmission` (`v2-common-functions.tsx:188`) builds
  `DatasetRegistrationSchemaV1` from explicit fields plus `structuredClone(study.assets)` minus
  `consentGroups`; `alternativeDataSharingPlanFile` is a top-level `Study` field and never reaches
  the JSON. So `expect(resubmitted.get('dataset')).not.toContain('alternativeDataSharingPlanFile')`
  passes with the line deleted, giving false confidence that the strip is under test. Only the
  `consentGroups` mapping does real work. `withoutAttachments` also closes over nothing and should
  sit at module scope beside `renderViolations` rather than being rebuilt on every render.

- **`hasBackendAbstention` should use `.some()`** — `src/utils/BucketUtils.ts:294` builds two
  intermediate arrays to answer a boolean question.
  `matchResults.some((m: MatchResult) => m.abstain === true)` is equivalent, allocation-free, and
  short-circuits. The name also claims "backend" abstention while it is still consulted for v1/v2
  buckets, where the diff's own `model.ts` docstring says abstention is decided client-side.

- **`AbstainDataUseCodes` array vs. the Sets this PR introduces** — `src/types/model.ts:32` is a
  `readonly string[]` scanned with `.includes`, while the same PR converted the algorithm-version
  lookups to `ReadonlySet`. `shouldAbstain` runs a linear scan per secondary code per bucket where
  `BACKEND_ABSTAINING_VERSIONS` gets O(1); and because `readonly` is erased at compile time, any
  consumer holding the exported binding can push into the shared array and silently widen abstention
  app-wide. `new Set([...])` typed `ReadonlySet<string>` matches the pattern established two files
  over.

- **Near-identical DAC-facing messages** — `src/utils/BucketUtils.ts:189`: "Unable to interpret the
  system match" sits one word away from the existing "Unable to determine a system match", and both
  render verbatim in the same slot. A DAC member sees both in one review session with no cue that
  they mean different things, and anyone grepping logs or support tickets for one string misses the
  other. Wording that names the cause (e.g. "System match unavailable for this algorithm version")
  separates them.

kevinmarete:
Thanks — all three are in.

1. Fixed in 3f99ad4: `buildMultiPartFormData` no longer mutates the live `study`. `withoutAttachments` clones and both `delete`s are gone, so a resubmit after a 400 still carries the file parts. Covered in `DataSubmissionFormV2.spec.tsx`, now asserting both the alternative data sharing plan and the NIH institutional certification survive.

2. Confirmed, and the conservative alternative is what shipped. Changeset 87.0 backfilled pre-existing rows to `v1`, so `classifyAlgorithmVersion` reads an absent version as legacy rather than uninterpretable and no collection loses its suggestion. Since your review it also stopped demanding a single version string: versions that agree on how the result must be read (v3 + v5) keep the suggestion, and the v3+ check became a floor rather than an allowlist so a future v6 is trusted (6c2d350).

3. `getResult` passes unrecognized non-empty values through as-is (3f99ad4). The sentinel is now "System match unavailable for this algorithm version" rather than "Unable to interpret the system match", so it no longer reads one word off the existing "Unable to determine a system match".

kevinmarete:
Verified all nine findings and the nits against the working tree — all accurate. Fixed in ba2b81a and 6c2d350.

**1, 2, 4 — primary data use form** (ba2b81a). `onPrimaryChange` reads `next`. The Disease-Specific radio hands over `selectedDiseases.map(d => d.id)`, so the `string[]` type holds at runtime. The Other radio passes the retained text (or `''`), so `calcErrors` rejects an empty Other instead of accepting the `"otherPrimary"` sentinel. Both dependent panels initialize from `selectedPrimaryGroup` rather than from their own field, which fixes the POA/searchbar desync at its source. Duplicate setters removed.

Worth calling out: switching primaries now really does clear the disease selections — the stale read is how the option objects survived the switch in the first place. Switching away and back requires re-picking, symmetric with how the Other text already behaved.

**3 — load path** (ba2b81a). Guarded with `isEmpty` in `buildConsentGroupsFromStudy`. One correction: this predates the diff rather than leaking out of it — `isNil([])` was already false when `diseaseSpecificUse` was typed `boolean`. The diff documented the rule, it did not introduce it.

**5, 6 — algorithm versions** (6c2d350). `classifyAlgorithmVersion` classifies each version and falls back only when the *classes* differ, so a v3+v5 bucket keeps its suggestion. The v3+ set became a floor (`>= 3`, parsed off the suffix) rather than an allowlist, so only v1/v2 stay enumerated and a v6 is trusted. `createDate` on the fallback is `undefined` rather than arbitrarily `matchResults[0]`.

**7, 8 — error rendering** (6c2d350). An `errorMessage` helper reads `response.data.message` before falling back to `String(error)`, and the create path discriminates with `isValidationRejection` like the update path. Tests for both shapes.

**9 — mixed match display** (6c2d350). Kept the pass-through and added the missing test. The Decision line is a block `div`, so the sentence wraps rather than overflowing the slab.

**Nits** — all taken: dead `isOther` operand, `hasBackendAbstention` → `hasRecordedAbstention` using `.some()`, `AbstainDataUseCodes` as a `ReadonlySet`, POA added to the `processDefinedLimitations` docstring, and the message renamed to "System match unavailable for this algorithm version" so it no longer sits one word from "Unable to determine a system match".

One correction on the vacuous-assertion nit: right that `alternativeDataSharingPlanFile` never reaches the JSON in production, but the assertion was not passing with the line deleted — the spec mocks `studyToDatasetSchemaSubmission` as identity, so the strip was the only thing keeping the field out. Removing it failed the test. The strip is gone and the test now targets `addedNIHInstitutionalCertificationFile`, which `consentGroups: structuredClone(...)` really does carry into the payload.

`npm run lint` 0 errors, `npm run type-check` 0 errors, 4,057 tests pass. Nine new tests, each checked to fail against the unfixed behavior rather than pass vacuously.

otchet-broad (commented):
Thanks for the updates.  These three are probably the last that need to be addressed:

### 1. Empty disease list shadows a legacy `Other` primary on edit-load

`src/pages/data_submission/consent_group/consentGroupUtils.ts:70`

The reorder moved `diseaseSpecificUse` below `poa` but left it above `otherPrimary`, so an empty disease list still shadows a legacy `Other` primary when an existing study is loaded for edit.

**Failure scenario.** `buildConsentGroupsFromStudy` (`src/pages/data_submission/v2/v2-common-functions.tsx:254`) assigns `dataset.dataUse.diseaseRestrictions` unconditionally, unlike the `isEmpty` guards it applies to `otherPrimary` / `gs` / `morDate`. Load a study whose dataset has `other='Not for profit'` and `diseaseRestrictions=[]`:

- `selectedPrimaryGroup` returns `'diseaseSpecificUse'` (an empty array is not nil), so the Disease-Specific radio renders selected while the Other radio and its text box are hidden.
- But `showDiseaseSpecificUseSearchbar` is initialized with `!isEmpty(...)` (`ConsentGroupAddEdit.tsx:75`), so the disease picker is not shown and `calcErrors` does not flag the missing disease list.
- Saving without touching the radios submits both `otherPrimary` and `diseaseSpecificUse`, which Consent now rejects as multiple primaries; touching the radios silently wipes `otherPrimary`.

**Suggested fix.** Guard at the load boundary (an `isEmpty` guard in `buildConsentGroupsFromStudy`), which preserves the deliberate in-form "empty list keeps the radio lit" behavior.

### 2. Closed algorithm-version allowlist breaks on the next Consent version

`src/utils/BucketUtils.ts:184`

`BACKEND_ABSTAINING_VERSIONS` is a closed allowlist, so the first Consent algorithm version newer than `v5` makes every bucket in the DAC voting UI report "Unable to interpret the system match".

**Failure scenario.** Consent ships `DataUseMatcherV6` and starts stamping `algorithmVersion='v6'`. `classifyAlgorithmVersion` falls through both sets and returns `'unrecognized'`, so every voting slab in every collection shows "Unable to interpret the system match" with no suggestion until duos-ui is released in lockstep. Before this change a `v6` row fell into the `!unmatchable` path and still displayed its match correctly — this is a forward-compatibility regression.

**Suggested fix.** Invert the default: enumerate only the legacy versions needing client suppression (`v1`/`v2`, plus absent) and treat everything else as backend-abstaining.

### 3. Every HTTP 400 treated as a Data Use consistency rejection

`src/pages/data_submission/v2/DataSubmissionFormV2.tsx:30`

`isValidationRejection` treats every HTTP 400 as a Data Use consistency rejection, so unrelated 400s now skip the reload and surface an unsplittable generic message as if it were a violation list.

**Failure scenario.** The multipart PUT is rejected 400 for a non-Data-Use reason (malformed multipart, an oversized/rejected attachment, a proxy 400) or with a non-JSON body. `fetchMultipartRequest` (`src/libs/ajax/fetchAdapter.ts:338`) then builds `message='Request failed with status 400. Please contact the help desk at duos@duos.org.'`, and `renderViolations` presents that single line under "Study update failed:" as the violation. The previously guaranteed reload of the server-side study is also skipped, so the form keeps state the server never accepted.

**Suggested fix.** Narrow the check to the response shape that actually carries consistency violations (e.g. a multi-line message, or a specific error code).

kevinmarete:
Thanks. One of the three was still live; the other two were fixed in ba2b81a and 6c2d350, which were pushed about two hours before this review, so it looks like it read the tree as of 3f99ad4.

**3 — every 400 treated as a Data Use rejection.** Real, fixed in 0b16240. `isValidationRejection` now requires a 400 *and* a non-blank `response.data.message`. Both Consent rejection paths supply one — `DatasetResource#createDatasetRegistration` joins violations under "Please correct the following fields:", and `StudyUpdateRequestValidator` throws single-line `BadRequestException`s such as `DataUsePrimaryValidator.VALIDATION_MESSAGE` — while a malformed multipart, a rejected attachment, a proxy 400, a non-JSON body, and `StudyResource`'s bodiless `Response.status(BAD_REQUEST)` all fall through to `fetchMultipartRequest`'s generic help-desk message with no `data.message`, so those reload exactly as before. Two tests added: a bodiless 400 on update reloads and shows the plain help-desk line, and a whitespace-only message does not render as a violation list.

I did not use the multi-line-message suggestion. The update path's consistency rejection is a single line (`VALIDATION_MESSAGE` has no newlines), so a multi-line test would misclassify the exact rejection this PR exists to surface; only the create path joins with newlines. Presence of a server-authored message is the shape that actually separates the two.

**1 — empty disease list shadowing a legacy `Other` primary.** Fixed in ba2b81a, at the load boundary you suggested: `buildConsentGroupsFromStudy` now assigns `isEmpty(dataset.dataUse.diseaseRestrictions) ? undefined : dataset.dataUse.diseaseRestrictions` (`v2-common-functions.tsx:255`), so `other='Not for profit'` with `diseaseRestrictions=[]` loads with the Other radio lit and the deliberate in-form "empty list keeps the radio lit" behavior untouched. In-form exclusivity is separately guaranteed: `onPrimaryChange` clears `diseaseSpecificUse` and `otherPrimary` before setting the chosen field (`ConsentGroupAddEdit.tsx:157,159`), so no save can carry two primaries.

**2 — closed algorithm-version allowlist.** Fixed in 6c2d350. `BACKEND_ABSTAINING_VERSIONS` is gone; `classifyOneVersion` reads `v(\d+)` and compares against `FIRST_BACKEND_ABSTAINING_VERSION = 3`, so `v6` classifies as backend-abstains with no duos-ui release, and only absent/`v1`/`v2` take the client-suppression path. Anything that is not a `v<major>` string is still `unrecognized`, which is the case that genuinely cannot be interpreted.

`tsc --noEmit` clean, oxlint and eslint clean on both changed files, `DataSubmissionFormV2.spec.tsx` 9/9 passing.

otchet-broad (approved):
Nice work!  Thanks for the fixes!

PR #3879, opened 2026-08-24
[DT-3983] Complete the profile page style update

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3983

Security risk: **low** — presentation-only; the email toggle replaces a radio group over the existing `User.updateSelf` call.

### Summary

Brings Full Name, Researcher Status and Accepted Terms & Policies onto the bordered-card treatment DT-3809 gave the other two sections. All five now share `.user-profile-card`, `.user-profile-section-heading`, `.user-profile-subheading` and `.user-profile-table` from `UserProfile.css`, replacing three separately inline-styled tables.

Also: Accepted Terms & Policies becomes a Document Name / Attestation Date table, "Status" is relabeled "Requestor Status", the email radios become a MUI `Switch`, name and email inputs are a fixed 400px, and the header drops the GA4GH logo, icon and intro block.

Two notes for review: `UserProfile` keeps local state for the toggle, since a `Switch` bound straight to `user.emailPreference` would snap back after each save; and the DAA download moved to `DownloadLink`, which `copilot-instructions.md` requires for all file downloads.

After

----

- Verify all tests go green


rushtong:
One codex finding:

> [P2] Prevent overlapping email-preference updates. The switch remains enabled while User.updateSelf is pending, and each rejection blindly restores !value. Starting off, quickly toggling on then off, with both requests failing in order, leaves the UI on even though the persisted preference remains off. Out-of-order successful requests can similarly leave UI and server state inconsistent. Disable the switch while saving or serialize/version requests and restore the last confirmed value. [UserProfile.tsx lines 65–78](https://github.com/DataBiosphere/duos-ui/blob/63afae5bdf3667fb535926d4e70a2a5b3fb4c3d9/src/pages/user_profile/UserProfile.tsx#L65-L78)

kevinmarete:
Good catch — fixed in 1e48cac2.

The toggle is now disabled while `User.updateSelf` is in flight, so a second request can no longer start before the first resolves, and the error path restores the last confirmed value instead of blindly flipping to `!value`. Added a spec case that holds the update unresolved to assert the toggle is disabled mid-flight, re-enabled afterwards, and that only one request is issued.

rushtong (approved):
Thank you for the updates - I was hoping we would get to this since the profile page contained a mix of newer and older styles. 👍🏽

src/pages/user_profile/ResearcherStatus.css
```diff
@@ -0,0 +1,17 @@
+.researcher-status-value {
```
fboulnois: Instead of custom CSS, can we depend directly on the Material UI components?

fboulnois (approved):
Looks good, one suggestion since @otchet-broad recently added these rules in the codebase:

src/pages/user_profile/ResearcherStatus.css
otchet-broad: +1

src/pages/user_profile/ResearcherStatus.tsx
```diff
@@ -60,46 +62,19 @@ const ResearcherStatus: React.FC<ResearcherStatusProps> = (props) => {
...
-      >
+    <div className="researcher-status">
+      <h1 className="user-profile-section-heading">Researcher Status</h1>
+      <p className="user-profile-subheading">Requestor Status</p>
+      <p className={`researcher-status-value researcher-status-value-${hasCard ? 'active' : 'inactive'}`}>
         {hasCard ? 'Active' : 'Inactive'}
       </p>
       {!hasCard && (
-        <p style={{ marginTop: '10px' }}>
+        <p className="researcher-status-inactive-notice">
```
otchet-broad: @fboulnois I wonder if we should turn these conditionals with almost the same intent into MUI Alerts as well.  Thoughts?

src/pages/user_profile/ResearcherStatus.css
kevinmarete: Done in 8254102b — deleted `ResearcherStatus.css` and let MUI do the styling (`Alert` + `sx`). I checked the rest of the PR: the remaining classes sit on plain `div`/`p` or the legacy `btn-primary` buttons, and the MUI components here (`Switch`/`FormControlLabel` in `UserProfile`, the `Alert` in `SigningOfficialRequest`) are already styled with `sx`, so no CSS class is styling a MUI component now.

src/pages/user_profile/ResearcherStatus.tsx
kevinmarete: Agreed — both conditionals are now one `Alert`, `severity="success"` when active and `severity="info"` with an `AlertTitle` plus the contact-your-SO copy when inactive. Specs assert on `getByRole("alert")`.

otchet-broad (approved):
👍

PR #3880, opened 2026-08-24
[DT-3984] my dac datasets table update

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3984](https://broadworkbench.atlassian.net/browse/DT-3984)

### Summary

Replaced the SimpleTable/PaginationBar rendering in DACDatasetsTable with MUI X DataGrid, following the same pattern already established by the Data Library's LibraryDataGrid/datasetColumns:

New datasetGridColumns.tsx factory builds GridColDefs for all 9 columns (DUOS ID, PHS ID, Dataset Name, Study Name, Data Submitter, Data Custodian, Data Use, NIH Institutional Certification, Status).

Data Use codes now render as MUI Chips (first two codes + a "+N" overflow chip with tooltip), matching the Data Library's Data Use column.

The NIH certification download button and the Approve/Reject/Delete (DACDatasetApprovalStatus) actions were ported into renderCells, unchanged in behavior.

Column sort now uses DataGrid's controlled sortModel, still persisted to the same localStorage key so the chair's last sort choice survives a reload.

Removed the now-dead SimpleTable-only styling (baseStyle/columnStyle in DACDatasetConstants) and the cell-data functions that existed solely to feed SimpleTable rows. SimpleTable/PaginationBar themselves are untouched — still used by ~9 other tables elsewhere in the app.

AFTER:
image

----

- Verify all tests go green


src/components/dac_dataset_table/datasetGridColumns.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { DatasetTerm } from 'src/types/model'
+import { DataSet } from 'src/libs/ajax/DataSet'
+import { DataUseCode, processDataUseCodes } from 'src/utils/DataUseUtils'
+import { Notifications } from 'src/libs/utils'
+import DACDatasetApprovalStatus from 'src/components/dac_dataset_table/DACDatasetApprovalStatus'
+import { DACDatasetTableColumnOptions } from 'src/components/dac_dataset_table/DACDatasetConstants'
+
+const downloadInstitutionalCertification = (dataset: DatasetTerm) => {
+  DataSet.getNIHInstitutionalCertification(dataset.datasetId).catch(() => {
+    Notifications.showError({ text: `Error downloading the NIH Institutional Certification for ${dataset.datasetIdentifier}` })
```
rushtong: Should we also show the user the actual error here? I think this swallows whatever Consent returns.

rushtong (approved):
One question inline, but looks good 👍🏽

kevinmarete (commented):
Looks good, but I have these suggestions:

- Align the table with the header and search box.
- Reuse the Data Library’s Data Use chip implementation instead of introducing a different format.
- Restore a visible `:focus-visible` indicator for keyboard navigation.

kevinmarete (approved):
Thanks for the updates!

PR #3032, opened 2025-09-02
[DT-2093] Speed up DarCollectionReview page

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2093

### Summary

This PR improves performance by making API calls in parallel and reusing collection data instead of re-fetching on tab switches. It also cleans up useEffect usage to avoid unnecessary queries. These changes make the DarCollectionReview page faster and smoother.

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #3881, opened 2026-08-25
[DT-4012] Handle the full consent /api/user/me contract and B2C authorization-response errors

rushtong:
### Addresses

[DT-4012](https://broadworkbench.atlassian.net/browse/DT-4012)

Security risk: low — sign-in error paths change how a refusal is presented; no change to who is admitted.

### Summary

Consent now answers `GET /api/user/me` with distinct statuses ([DT-3997 #3021](https://github.com/DataBiosphere/consent/pull/3021), [DT-4011 #3032](https://github.com/DataBiosphere/consent/pull/3032)): **401** rejected token, **404** authenticated but unregistered, **409** Sam sub-provider conflict with an actionable message. The BFF and client now map that contract instead of the old conflated handling:

- `server/src/auth/me.ts`: 401 destroys the session (the `profileSeen` disambiguator is dead machinery and is removed, along with its once-per-session store write); 404 keeps the "authenticated, no user" answer that routes new users into registration; 409 forwards the upstream message as `{ error: 'provider_conflict', message }` after destroying the session — the account lives under the other provider, so the session cannot become usable.
- `src/libs/auth/session.ts`: the probe treats 409 as an authoritative signed-out answer and shows the message (sign in with the other provider, plus the support link) instead of failing sign-in generically.
- `src/libs/auth/postSignIn.ts`: a 409 from getMe signs the user out with the message instead of attempting a registration that cannot succeed. The legacy `"azureb2c authentication error"` substring check stays for older consent builds.

A second edge case is outside consent entirely: B2C can answer the authorization request itself with an error instead of a code — the user's own cancel, or a B2C-side failure. The observed case is `AADB2C90289`/`invalid_client`: the non-prod B2C tenant's federation client secret to the upstream Microsoft provider has expired, which fails every Microsoft sign-in in those environments (secret rotation is tracked separately). `authorizationCodeGrant` throws `AuthorizationResponseError`, and the resulting 500 answered the browser's top-level `/auth/callback` navigation with raw JSON, stranding the user:

- `server/src/auth/callback.ts`: catch `AuthorizationResponseError`; a cancel (`access_denied`) redirects home silently, anything else logs the B2C error and redirects to `/?signInError=provider`. State-mismatch and token-validation failures still throw.
- `src/App.tsx`: a fixed toast for the `signInError` marker, then the marker is stripped from the URL. The marker is a key, never reflected text, so the toast cannot echo attacker-supplied content.

**Deployment note:** the 401 tightening assumes every environment runs a consent build with both fixes above. An old consent behind this BFF would destroy a brand-new user's session on the first probe and block registration.

### Testing

`pnpm run lint`, `pnpm run type-check`, `pnpm test` (368 files, 4312 tests), and the server suite (327 tests) pass. New coverage: 401 destroys the session (including a failing session store); 409 forwards the upstream message, falls back on an unusable body, and overwrites the probe's held signed-in answer; the bootstrap signs out on a 409 without a registration attempt; the callback's authorization-error redirects run against the real `openid-client` grant (fake B2C), alongside the unchanged security rejections; the App toast renders once and the marker is stripped.

The Microsoft sign-in repro was observed manually and produced the exact error this PR handles.

### 404 User Not Found -> New User Registration Case
Screenshot 2026-08-25 at 12 25 32 PM

### 409 Sam AzureB2C Error Case
Screenshot 2026-08-25 at 11 46 21 AM

### 500 Auth Provider Error 
Screenshot 2026-08-25 at 1 32 59 PM

----

- Verify all tests go green

[DT-4012]: https://broadworkbench.atlassian.net/browse/DT-4012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/auth/callback.ts
+++ b/server/src/auth/callback.ts
@@ -27,11 +27,13 @@ export async function handleCallback(request: FastifyRequest, reply: FastifyRepl
   catch (err: unknown) {
     if (err instanceof oidc.AuthorizationResponseError) {
       // B2C answered the authorization request with an error instead of a
-      // code — the user canceled on the B2C page (access_denied), or chose
-      // an identity the tenant's policy rejects. Land back in the
-      // SPA instead; a cancel is the user's own action and stays silent.
-      // A cancel is routine — info keeps it out of warn-based alerting;
-      // real provider errors stay at warn.
+      // code — the user canceled on the B2C page (access_denied), or B2C
+      // itself failed (error=server_error; observed as AADB2C90289 when a
+      // tenant's federation client secret to the upstream Microsoft provider
+      // expired, which fails EVERY Microsoft sign-in in that environment).
+      // Land back in the SPA instead; a cancel is the user's own action
+      // and stays silent. A cancel is also routine — info keeps it out of
+      // warn-based alerting; real provider errors stay at warn.
       const cancelled = err.error === 'access_denied'
       request.log[cancelled ? 'info' : 'warn']({ error: err.error, description: err.error_description }, '[auth] B2C authorization response is an error')
       reply.redirect(cancelled ? '/' : '/?signInError=provider')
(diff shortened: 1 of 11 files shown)


rushtong:
Moving back to draft temporarily - there are some additional details related to the 500 error seen in MS logins.

kevinmarete (approved):

fboulnois (approved):
👍

PR #3882, opened 2026-08-25
[DT-3950] Add Researcher Status and Data Submitter Status to the Manage Users table

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3950

Security risk: **no** — two read-only columns over data the table already loads.

### Summary

Adds Researcher Status and Data Submitter Status columns to the Manage Users table, and stops the combined Roles column repeating Data Submitter now that it has its own.

Both columns read the same two sources as the Signing Official console's Researcher Status tab — the library card, and `hasDataSubmitterRole`. The `isResearcher`/`isDataSubmitter` flags are not usable here: `setUserRoleStatuses` derives them for the signed-in user only, so the `/api/user/role/Admin` list response carries neither and every row would read No. Deriving Researcher Status from the Researcher role is no better, since that role is granted by default.

After

----

- Verify all tests go green


fboulnois (approved):
👍

rushtong (approved):
👍🏽

PR #3883, opened 2026-08-25
[DT-3751] PI language for CADR on SO pages

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3751](https://broadworkbench.atlassian.net/browse/DT-3751)

### Summary

 - Adds consistent Principal Investigator qualification guidance to the Signing Official DAR approvals and DAA pre-
    authorization pages.

  - Centralizes the PI definition so help text and approval attestations stay synchronized.
  - Extends the shared DAA associations page to support optional help content.
  - Updates tests to verify the new guidance and shared attestation language.
 
Note to reviewers:  Confirmed with PO that we should also do [this ticket](https://broadworkbench.atlassian.net/browse/DT-4026) to standardize how we refer to the PI on forms and in grids/tables.

----

- Verify all tests go green


rushtong (approved):
Nice extraction 👍🏽

src/definitions/definitions-en-us.ts
fboulnois, line 1: Instead of putting this in `libs`, I wonder if we should have a slightly different approach, like a `definitions.ts` file or a specific folder for definitions.

fboulnois (approved):
One suggestion otherwise lgtm 👍

src/definitions/definitions-en-us.ts
otchet-broad: If we had an I18N file for these constants, I would put this in there.  Is that the direction I should really move this in?

fboulnois: That seems like a larger effort. I would do the smallest possible change in that direction.

otchet-broad: How about definitions-en-us.ts?

kevinmarete (approved):

PR #3889, opened 2026-08-26
[DT-4018] Rotate the session ID on the OAuth callback (BFF Phase 5, story 5-C)

rushtong:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-4018 (BFF Phase 5, story 5-C)

### Summary

Rotates the session ID after a successful B2C token exchange, so a session ID planted before login (session fixation) can never become an authenticated session.

Order in `server/src/auth/callback.ts` after `authorizationCodeGrant()`:

1. Capture the pre-auth `sessionId` and the sanitized `returnTo` — `regenerate()` empties the session, so both die with the pre-auth row.
2. `await request.session.regenerate()` — `@fastify/session` creates a new, empty session with a new sid and repoints `request.session`.
3. Write the tokens and claims to the new session.
4. `await request.session.save()` — still required for the ERR_HTTP_HEADERS_SENT double-send fix; once this succeeds, the login has succeeded.
5. `retirePreAuthSession()` — stamp the pre-auth audit row `end_reason='rotated'`, then destroy the pre-auth row through `request.sessionStore` (not `request.session.destroy()`, which now points at the new session). Both steps are best-effort and logged: the old row never held tokens, so replaying its sid yields 401 regardless, and a transient failure is an audit/cleanup problem rather than a fixation attack.
6. Redirect to the captured `returnTo`.

The rotated callback response carries two `sessionId` Set-Cookie headers (a clear of the old cookie, then the new one); the browser keeps the last.

### Audit trail: two rows per login, and a new `end_reason`

Rotation changes the shape of `user_session_audit`. Before, one row per login: `audit_session_start` created it at `/auth/login`, and `audit_session_update` backfilled the email and idp at the callback. Now there are two — the pre-auth row, which stays anonymous, and the authenticated row that `regenerate()`'s INSERT creates and the token-writing UPDATE backfills.

The DELETE in step 5 fires `audit_session_end`, which fills `end_reason` with `COALESCE(end_reason, 'expired')`. Left alone, that files every successful login as an expiry and inflates any count of expired sessions. So the stamp lands before the DELETE, exactly as `logout.ts` stamps `'logout'` — same SQL, same `sid_hash` convention, same best-effort handling.

**`'rotated'` is a new value for `user_session_audit.end_reason`.** It fits `VARCHAR(16)` and the column carries no check constraint, so it needs no Liquibase changeset. The Epic 1 documentation lists only `logout`, `expired`, and `revoked`, and needs the fourth value added.

Both cleanup steps live in `retirePreAuthSession()` rather than inline, which keeps `handleCallback` under SonarCloud's cognitive-complexity limit and puts the two best-effort steps under one rationale.

### Trigger verification

The `user_session_audit` table and its triggers are Liquibase changesets in the Consent repo, so no PG-backed fixture exists in this repo's CI. Per the plan, this PR takes option (b): a tracked deployment-verification step against a BEE, checking that the pre-auth row carries `ended_at` and `end_reason='rotated'`. The security property itself (the old sid is unusable) is proven by the integration tests below.

### Test plan

- `server/test/callback.test.ts` (handler): rotation happens before token writes; save precedes the best-effort cleanup; the audit stamp precedes the destroy; both target the pre-auth sid; an error in either cleanup step logs but the login still succeeds; a failed `regenerate()` writes no tokens, saves nothing, destroys nothing, and does not redirect; the redirect uses the `returnTo` captured before rotation.
- The `regenerate()` mock replaces `request.session` with a new object, as `@fastify/session` does, rather than emptying it in place. Verified by mutation: a handler that writes tokens through a stale pre-rotation reference fails the suite, and passed against the old mock.
- `server/test/auth.test.ts` (integration, real `@fastify/session`): the `sessionId` cookie changes across the callback; `/auth/me` succeeds with the new cookie (tokens survived); replaying the pre-auth cookie gets 401; the pre-auth row is gone from the store; its audit row is stamped `'rotated'`; the double-send lifecycle guard pins the callback to exactly two store writes, both to the rotated sid.
- The story 5-B tests that arrived with the develop merge now build their session through the `authenticate()` helper. They had reused the login cookie, which rotation kills: three failed outright, and the three Fetch Metadata 403 cases had quietly degraded to testing an anonymous session.
- 380 server tests pass; `tsc` and `eslint` are clean; SonarCloud reports 0 open issues on the PR.

----

- Verify all tests go green


otchet-broad (approved):
👍

fboulnois (approved):
The comment density makes the code hard to read, but otherwise lgtm 👍

rushtong:
> The comment density makes the code hard to read, but otherwise lgtm 👍

That is fair - will give it another round of pruning.

PR #3891, opened 2026-08-26
Data Library top level nav item.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3938](https://broadworkbench.atlassian.net/browse/DT-3938)

### Summary

Promotes Data Library to a top level component.
Removes appropriate sub navigation elements for the library.
Keeps SO 'My institution's library' intact.

After:

image

----

- Verify all tests go green


rushtong:
Codex finding:

> **[P2] Unmatched console detail routes now highlight Data Library.** Because Data Library is inserted at index 0, a fresh load or refresh of `/dar_application_review/999` has neither a URL match nor `activeTab` context, so the fallback selects Data Library instead of Researcher Console. The existing test only checks that some tab is selected, masking the regression. Register these detail-route patterns with their console or fall back to the first rendered isConsole tab.

rushtong (approved):
The P2 I found is very low risk. Code looks good otherwise, address as you see fit.

kevinmarete (approved):
👍🏾

fboulnois (approved):
👍

rushtong:
I was able to confirm this (very low) finding locally. 
1. In the Data Library, request access
2. Tab stays Data Library
3. Refresh
4. Tab moves to Researcher
> src/components/DuosHeader.tsx:354 — contextFallbackTab ranks the navigation context above the hidden search registration. A researcher who clicks Apply for Access in the Data Library goes to /dar_application/<id> with activeTab = 0, so Data Library stays highlighted on the application form. A refresh of the same URL resolves through hiddenSearchTab and flips the highlight to Researcher Console. One URL, two highlights. The tests cover only the cold-load direction.

---
I was able to confirm this locally, but it seems like this is intended behavior now. The primary console page doesn't show sub tabs, but navigating to the profile page re-renders them.

> Claude: src/components/NavigationTabsComponent.tsx:247 — The removal of the selectedSubTab >= 0 guard makes the sub-tab bar render on pages that match no tab. Example: an SO opens /profile. resolvedTab is -1, initialSubTab stays -1, and initialTab becomes the last-resort console. renderedSubTabs[-1]?.hideSubTabBar is undefined, so the full SO Console sub-bar renders with no selection. The old guard hid it. The code cannot tell a registered-but-hidden page from an unregistered one.

> Codex: [P2] Unrelated pages now show console sub-navigation" body="Removing the `selectedSubTab >= 0` guard renders the selected console’s entire sub-tab bar whenever no sub-tab matches. `DuosHeader` fallback-selects the first available console for unmatched routes, so an admin or signing official visiting `/profile` now sees Admin/SO console navigation beneath the header—a regression from the previous behavior. Preserve the bar for intentionally hidden console sections only, while keeping it suppressed when the top tab was merely selected as a fallback.

rushtong (approved):
Despite findings, this looks good to me 👍🏽

PR #3892, opened 2026-08-26
[DT-4029] Add the post-BFF AI chat implementation plan

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-4029

Parent epic: https://broadworkbench.atlassian.net/browse/DT-3069

### Summary
Adds `docs/plans/AI_Chatbot_Overview.md`, a high-level plan for an AI chat window in DUOS, plus a supplemental `ai_chatbot_docs/` directory for ADRs and the phase documents the plan expands into.

The chat answers user questions about datasets and data-access requests. A server-side agentic loop in the Fastify BFF drives an LLM and runs read-only tool calls against the Consent API with the user's own session token. The token never reaches the browser, and there is no standalone MCP server.

The plan builds on the BFF migration (DT-3604). It needs three of its stories and no more: 5-F, 5-G and 6-D-pre.

**Security risk: no.** Documentation only. No product code, config, or dependency changes.

### What changed since the last review

Two rounds of review reversed the central infrastructure decision, so this is worth a fresh read rather than a diff skim.

| Round | Raised by | Outcome |
|---|---|---|
| 1 | @otchet-broad | Ollama replaced with llama.cpp, on openness and policy grounds |
| 2 | @fboulnois | The local model removed entirely |

**ADR-001 is now "Use Vertex AI in every environment; run no local model"** (renamed from `ADR-001-local-inference-engine.md`). It was still Proposed and unmerged, so it is rewritten in place with no supersede chain.

The reasoning, in short: a local dependency earns its place when it reproduces production behavior. The compose Postgres does — same engine, same protocol, same semantics, pinned to a patch version. A local 7B model reproduces the shape of the dependency and none of its behavior, and it has three independently churning parts (engine, weights, tool-call dialect) against Postgres's one. June to August broke the choice once already.

What follows from that:

- One backend everywhere: Vertex AI Gemini, one pinned model version. Developers use ADC (`gcloud auth application-default login`); k8s uses Workload Identity. No key files anywhere.
- The stub backend becomes first-class, with recorded fixtures and the failure paths. It carries Chats 1–8.
- `roles/aiplatform.user` on the dev project goes to the same reviewers as the data contract, so the two travel as one ask (Chat 0).
- Chat 0 is slow but no longer stops the world. Sequencing splits on what it gates.
- If Compliance blocks Vertex, self-hosted vLLM is the fallback. That is why the backend interface stays provider-neutral, and why no Vertex-specific code is written before Chat 0 returns.

Rejected alternatives are recorded in the ADR so they are not re-litigated: `llama.app` (right choice had a local engine survived), a Gemini Developer API key from AI Studio, a Vertex Express Mode key, and Ollama.

### What the plan covers

- **Baseline** — the BFF capabilities the chat reuses instead of rebuilding, each verified against `develop`
- **Target architecture** — the token never leaves the server
- **Server work** — `POST /api/chat` with CSRF and SSE, the four hazards the route inherits from the BFF, the bounded agentic loop, the Consent tool client, and three separate limits
- **Rate limits and cost** — a per-user burst bucket copying the pattern from [consent#2976](https://github.com/DataBiosphere/consent/pull/2976), a durable daily turn quota, and a concurrency cap. Records that every tool call spends the same per-user Consent budget as ordinary page traffic
- **Infrastructure** — one backend, two credential paths
- **Client work** — chat panel, SSE transport, `chatEnabled` feature flag, accessibility in the UI story rather than after it
- **Sequencing** — 14 work items, split by what the Compliance ask gates
- **Security notes** and **6 open questions**

Detail stays high level on purpose. Each work item becomes a story after the team approves the plan.

### Test plan
No automated tests. This PR changes two markdown files and no code.

- [ ] Read ADR-001 and confirm the no-local-model decision, above all the Postgres comparison in its Context section
- [ ] Confirm the BFF baseline table in §1 still matches `develop`
- [ ] Confirm the sequencing split: Chats 1–8 must be genuinely buildable against the stub, or Chat 0 becomes a hard gate after all
- [ ] Answer or assign the 6 open questions, above all question 5 — the combined data contract and dev role grant

----

- Verify all tests go green


docs/plans/AI_Chatbot_Overview.md
```diff
@@ -0,0 +1,406 @@
...
+   and a socket timeout that outlasts the turn (open question 2).
+3. Fastify `close()` waits for in-flight requests, so a held stream delays every
+   rolling deploy.
+4. The SPA fallback in `server/src/index.ts` answers a mistyped `/api/*` path
+   with the client shell instead of JSON.
+
+### 3.2 Agentic loop
+
+- One module with two backends behind a common interface.
+- **Local:** Ollama, selected when `OLLAMA_URL` is set. Pin a model tag with
```
otchet-broad: We should probably switch to an open source OpenAI API compliant model inference engine.  For example, llama.cpp, among others.
Ollama has started monetizing some of the tooling.  We probably don't want to slip into cases where we could accidentally use a model that doesn't adhere to our policies.

rushtong: Great point, will look at this suggestion.

rushtong: Agreed, and thanks — this was the right catch. Switched to llama.cpp
`llama-server` for local development, recorded in a new ADR:
`docs/plans/ai_chatbot_docs/ADR-001-local-inference-engine.md`. The plan body,
the sequencing and open question 1 now match it.

Two things I changed on the way through.

**Split the decision in two.** The engine and the weights are separate choices,
and only the second one carries the policy risk you named — a new engine does
not stop someone loading a model with the wrong license. So the ADR also adds a
provenance rule: pin one GGUF repository, one revision, one SHA256 digest, and
record the license beside the pin. Prefer Apache-2.0 or MIT weights (Qwen,
Mistral-Nemo). Worth noting the Llama and Gemma terms are custom licenses that
the OSI has not approved.

**Sharpened the Ollama argument.** The engine itself is still MIT and still free
locally, so the license is not the sharp edge. The hazard is the cloud path:
recent versions serve cloud model tags through the same local API and the same
`/v1` endpoint, so one wrong tag sends DUOS prompts to a third party.
`llama-server` cannot do that — it loads a file from disk and reaches no
registry. That makes the guard structural instead of a rule in a README, which
is now a bullet in the security notes.

Also in the ADR, so we do not re-litigate later:

- vLLM stays on the record as the self-hosted production candidate if
  data-governance sign-off (open question 5) blocks Vertex. No work on it now.
- LM Studio ruled out — proprietary.
- LocalAI rejected — it wraps llama.cpp but its model gallery re-adds the
  provenance problem.
- One caveat accepted: `llama-server` can return `tool_calls.arguments` as a
  JSON object where the OpenAI contract says string (ggml-org/llama.cpp#20198).
  Chat 6 will accept both shapes.
- Docker on Apple Silicon gives the container no Metal access, so the compose
  service is CPU-only. Equal cost for either engine. Documented the host-binary
  escape hatch.

ADR status is Proposed at present - will only accept after final reviews are complete.

docs/plans/ai_chatbot_docs/ADR-001-local-inference-engine.md
```diff
@@ -0,0 +1,92 @@
...
+## Context
+
+The plan needs a local LLM (Large Language Model) backend for development. It
+named Ollama. Production uses Vertex AI Gemini, so this decision covers
+developer machines and the evaluation harness only.
+
+Review feedback asked the team to switch to an open-source, OpenAI-compatible
+inference engine, such as llama.cpp. The comment gave two reasons: Ollama now
+sells some of its tooling, and the team could use a model that breaks policy by
+accident.
```
fboulnois: I strongly agree that we should not use Ollama. Ollama is not really open, reuses work from underlying open-source projects without credit, and sends telemetry to the Ollama folks: https://sleepingrobots.com/dreams/stop-using-ollama/

There is now a better solution: `llama.app` . It is bundled by the real `llama.cpp` developers, is completely open-source, and provides the same easy-of-use as Ollama.

I would take a look at https://llama.app/ and see if it meets our needs.

docs/plans/ai_chatbot_docs/ADR-001-local-inference-engine.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Options considered
+
+| Engine | License | OpenAI tool calls | Dev laptop | Notes |
+|---|---|---|---|---|
+| **llama.cpp `llama-server`** | MIT | Yes, with `--jinja` | Yes, best CPU speed | Small image. No registry. No cloud path. |
+| vLLM | Apache 2.0 | Yes, needs a parser flag | No, needs a GPU | The option if DUOS must self-host production inference. |
+| SGLang | Apache 2.0 | Yes | No, needs a GPU | Same class as vLLM. Smaller community. |
+| LocalAI | MIT | Yes | Yes | Wraps llama.cpp and adds a model gallery. The gallery re-adds the provenance problem. |
+| Ollama (the plan as written) | MIT engine | Yes | Yes | Registry tags and a cloud path. The two items under review. |
```
fboulnois: Only parts of Ollama are open-source (the non-cloud parts).

docs/plans/ai_chatbot_docs/ADR-001-local-inference-engine.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Status:** Proposed (2026-08-27) &nbsp;|&nbsp; **Work item:** Chat 3 (local dev unblock)
+**Related:** [AI_Chatbot_Overview.md](../AI_Chatbot_Overview.md) §2, §3.2, §4.1, open question 1
+**Raised by:** review feedback on duos-ui [#3892](https://github.com/DataBiosphere/duos-ui/pull/3892)
+
+---
+
+## Context
+
+The plan needs a local LLM (Large Language Model) backend for development. It
```
fboulnois: I am wondering if we really *do* need a local LLM for development. Would it be easier to get a Vertex AI key for development, and use that? We can generate keys in https://aistudio.google.com/api-keys

This would make the production and development environments symmetric.

fboulnois (approved):
Overall looks reasonable, some comments:

docs/plans/ai_chatbot_docs/ADR-001-local-inference-engine.md
rushtong: I think this is the more important point and I want to explore this potential path.

rushtong: I have a new approach that should resolve this question entirely in [7609826](https://github.com/DataBiosphere/duos-ui/pull/3892/commits/760982601971bb359546d110ed2097c9682ed8b6)

fboulnois (approved):
👍

docs/plans/ai_chatbot_docs/ADR-001-model-backend.md
```diff
@@ -0,0 +1,127 @@
...
+
+**Round 1** asked for an open-source, OpenAI-compatible engine, and warned that
+the team could use a model that breaks policy by accident. Both points hold.
+The concrete hazard is that recent Ollama versions serve cloud model tags
+through the same local API and the same `/v1` endpoint, so one wrong tag sends
+DUOS prompts to a third party. The engine is MIT, but the cloud parts are not
+open source.
+
+**Round 2** asked a better question: does development need a local model at
+all? It does not.
```
otchet-broad, line 23: 👍

otchet-broad (approved):
This looks good overall.  Some thoughts that can be addressed now or later:

1) Do we need to add a system prompt that includes instructions that get carried with each request? Instructions might include: Don't make external requests, only perform actions supported by duos/consent, don't crypto-mine, etc.?

2) Metrics for the types of usage?

3) How do we know if the model is having the right effect?

rushtong:
> This looks good overall. Some thoughts that can be addressed now or later:
> 
> 1. Do we need to add a system prompt that includes instructions that get carried with each request? Instructions might include: Don't make external requests, only perform actions supported by duos/consent, don't crypto-mine, etc.?
> 2. Metrics for the types of usage?
> 3. How do we know if the model is having the right effect?

Thank you - great questions. I'll add them as open questions so we don't lose track of them and can discuss as a team when we start delving into details.

PR #3893, opened 2026-08-26
Claude suggested flaky test fixes.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4027](https://broadworkbench.atlassian.net/browse/DT-4027)

### Summary

44 tests across 19 files were flaky under CPU load. They looked like many unrelated
bugs but had **one root cause: wall-clock budget exhaustion** — Vitest's default 5 second
`testTimeout` and Testing Library's default 1 second `asyncUtilTimeout`. Fixed by raising
both globally, plus two genuine real-time races and four cascade-amplifiers.

All suites now pass under the same load that previously produced 79 failures.
Coverage is byte-identical.

| Run | Result |
|---|---|
| Baseline, idle | 369 files / 4341 tests — **all pass** |
| Under load, run 1 | **16 files / 37 tests fail** |
| Under load, run 2 | **18 files / 42 tests fail** |

79 failure instances total, 44 distinct tests. Different tests failed on each run —
the signature of a timing problem rather than a logic bug.

---

#### Root cause

All 79 failures trace to one thing: tests running out of wall-clock time, not
tests being wrong.

| Mechanism | Instances | Symptom |
|---|---|---|
| Vitest default **5 s `testTimeout`** | 68 / 79 | `Error: Test timed out in 5000ms` |
| RTL default **1 s `asyncUtilTimeout`** | 10 / 79 | `TestingLibraryElementError: Unable to find an element…` |
| **Cascade** from the above | ~5 / 79 | Scrambled-text uncaught exceptions (below) |

Measured under 10x load *after* the fix, the slowest tests take **4–10.5 s**. Under
the old 5 s ceiling they were guaranteed to fail; the work itself was always legitimate.

### The cascade (the part that hid the root cause)

When a test hits `testTimeout`, Vitest abandons it and starts the next one — but the
dead test's `userEvent` promise chain **keeps typing into the shared jsdom document**.

That produced assertion failures like:

```
AssertionError: expected 'ud woArnkaslpyascies  dWeosrckrsippat…'
              to be 'Analysis Workspace Edited'
```

#### Coverage is unchanged

| | Baseline | After |
|---|---|---|
| Files measured | 514 | 514 |
| **Per-file metric changes** | — | **0** |
| Lines / Statements / Functions / Branches | 89.36 / 88.74 / 86.07 / 81.56 | 89.36 / 88.74 / 86.07 / 81.56 |
| Server | 95.12 / 94.90 / 97.14 / 84.27 | 95.12 / 94.90 / 97.14 / 84.27 |

----

- Verify all tests go green

### DIFF
--- a/server/vitest.config.ts
+++ b/server/vitest.config.ts
@@ -4,6 +4,8 @@ export default defineConfig({
   test: {
     globals: true,
     environment: 'node',
+    testTimeout: 30_000,
+    hookTimeout: 30_000,
     include: ['test/**/*.{spec,test}.{js,ts}'],
     env: {
       FASTIFY_LOG_LEVEL: 'silent',
--- a/test/components/ai_models_list/AiModelAddEdit.spec.tsx
+++ b/test/components/ai_models_list/AiModelAddEdit.spec.tsx
@@ -79,9 +79,7 @@ describe('AiModelAddEdit component', () => {
   it('edits existing model and saves changes', async () => {
     const user = userEvent.setup()
     const models: AiModel[] = [sampleModel]
-    const onAiModelsChange = vi.fn((updated: AiModel[]) => {
-      expect(updated[0].name).toBe('Baseline Model Edited')
-    })
+    const onAiModelsChange = vi.fn()
     const { container } = render(
       <AiModelAddEdit
         id={0}
@@ -94,5 +92,9 @@ describe('AiModelAddEdit component', () => {
     await user.clear(container.querySelector('#name')!)
     await user.type(container.querySelector('#name')!, 'Baseline Model Edited')
     await user.click(container.querySelector('.collaborator-form-add-save-button')!)
+
+    expect(onAiModelsChange).toHaveBeenCalledTimes(1)
+    const [updated] = onAiModelsChange.mock.calls[0] as [AiModel[]]
+    expect(updated[0].name).toBe('Baseline Model Edited')
   })
 })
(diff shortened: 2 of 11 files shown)


rushtong:
Some Claude feedback interleaved with my thoughts.

> Three per-test overrides now cap below the new global. FundingResourceAddEdit.spec.tsx:48, :70, and PresentationAddEdit.spec.tsx:64 keep }, 15000). These are the files someone already marked as slow. The new 30 s global does not reach them. 15 s still exceeds the measured 10.5 s worst case, so nothing fails today. Delete the three overrides and let the global apply.

Note - I'm pretty sure I made those changes so we can likely roll the timeouts back so the new globals take effect.

> Two macrotask sleeps remain in the file the PR corrects. fetchAdapter.spec.ts:604 and :618 still use await new Promise(resolve => setTimeout(resolve, 0)). The new comment at line 1047 calls that pattern unsafe. Both are negative assertions (not.toHaveBeenCalled()), so a short wait gives a false pass rather than a flake. That is worse for correctness, though it is a pre-existing defect.

> The suite loses a speed tripwire. At 5 s, a component that slows from 1 s to 25 s failed the build. At 30 s it passes in silence. This is the real cost of the change. I think it is worth paying, but say so in the ticket.

Note - this one is clear, but I agree - the cost is worth it. We might see some slower components NOT fail when we would want them to. The next point might help ...

> A cheaper fix exists for the tests that type text. userEvent.setup({ delay: null }) removes one real macrotask per keystroke. I measured it on WorkspaceAddEdit.spec.tsx on an idle machine: test time dropped from 874 ms to 620 ms, about 29%. Under load the gap grows. FundingResourceAddEdit already uses it. The other three edit tests do not. This attacks the cost instead of the budget.

Note - none of this is a blocker - this work should move through the pipeline.

rushtong (approved):
👍🏽

kevinmarete (approved):

fboulnois (approved):
👍

PR #3894, opened 2026-08-27
[DT-3994] Update DAC profile page to User profile style

kevinmarete:
### Addresses

[DT-3994](https://broadworkbench.atlassian.net/browse/DT-3994)

Security risk: no

### Summary

Restyles the five DAC profile page sections as cards matching the user profile chrome from DT-3809, replacing the old `<hr>` divider and 2rem navy heading. New `DacProfileSection` (MUI `Paper` + `Typography` via `sx`) is shared by `DacProfile` and `EditDac`'s profile-mode branch.

`EditDac` outside profile mode, the page header, and page width are unchanged.

After

[DT-3994]: https://broadworkbench.atlassian.net/browse/DT-3994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/manage_dac/DacProfileSection.tsx
+++ b/src/pages/manage_dac/DacProfileSection.tsx
@@ -1,0 +1,34 @@
+import React from 'react'
+import { Paper, Typography } from '@mui/material'
+
+// Values mirror .user-profile-card so the DAC and user profile pages share one card style.
+const CARD_SX = {
+  mt: '48px',
+  p: { xs: '20px', sm: '28px 32px 32px' },
+  border: '1px solid #d9e4f2',
+  borderRadius: '16px',
+  boxShadow: '0 10px 24px rgb(34 74 120 / 5%)',
+}
+
+const HEADING_SX = {
+  m: '0 0 24px',
+  color: '#01549f',
+  fontSize: '20px',
+  fontWeight: 600,
+}
+
+interface DacProfileSectionProps {
+  title: string
+  children?: React.ReactNode
+}
+
+export const DacProfileSection: React.FC<DacProfileSectionProps> = ({ title, children }) => {
+  const headingId = `dac-profile-${title.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, '')}`
+
+  return (
+    <Paper component="section" elevation={0} aria-labelledby={headingId} sx={CARD_SX}>
+      <Typography id={headingId} component="h1" sx={HEADING_SX}>{title}</Typography>
+      {children}
+    </Paper>
+  )
+}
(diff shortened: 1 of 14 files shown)


fboulnois (approved):
👍

src/pages/manage_dac/DacProfileSection.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import { Paper, Typography } from '@mui/material'
+
+// Values mirror .user-profile-card so the DAC and user profile pages share one card style.
```
rushtong: Would it make sense to extract the shared style?

src/pages/manage_dac/DacProfileSection.tsx
```diff
@@ -0,0 +1,34 @@
...
+  fontWeight: 600,
+}
+
+interface DacProfileSectionProps {
+  title: string
+  children?: React.ReactNode
+}
+
+export const DacProfileSection: React.FC<DacProfileSectionProps> = ({ title, children }) => {
+  const headingId = `dac-profile-${title.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, '')}`
```
rushtong: Can we end up with duplicate ids after the regex? Would it be safer to use an ID?

rushtong (approved):
Couple questions inline, but still 👍🏽

src/pages/manage_dac/DacProfileSection.tsx
kevinmarete: Good call — extracted. `.profile-card` and `.profile-card-heading` now live in `src/styles/profile-card.css` (imported from `index.css`), moved verbatim out of `UserProfile.css` including the 700px padding override. The user profile page and its child components use those class names, and `DacProfileSection` uses them instead of the `sx` literals, so there is one definition rather than two kept in sync by a comment.

One side effect worth flagging: `DacProfileSection` now renders a plain `<section>`/`<h1>` instead of MUI `Paper`/`Typography`. Emotion injects into `<head>` after `index.css` at equal specificity, so `Paper`'s own `border-radius`/`box-shadow` and `Typography`'s `margin: 0` would have beaten the shared class. Plain elements also match how `UserProfile` builds the same card.

src/pages/manage_dac/DacProfileSection.tsx
kevinmarete: Yes, that was reachable — any two titles that normalize to the same slug collide, and `EditDac` in profile mode renders inside `DacProfile`, so both components' sections share one page. Switched to React's `useId()`, which is guaranteed unique per instance and does not depend on the title at all.

PR #3895, opened 2026-08-27
[DT-4008] Remove the client-side legacy match suppression from duos-ui

kevinmarete:
### Addresses

[DT-4008](https://broadworkbench.atlassian.net/browse/DT-4008). The duos-ui half of the same cleanup as [DT-4009](https://github.com/DataBiosphere/consent/pull/3038), which retires the backend migration surface. Neither depends on the other.

Security risk: no.

### Summary

duos-ui suppressed the system match client-side for rows computed before Consent decided abstention itself. [DT-3866](https://broadworkbench.atlassian.net/browse/DT-3866) scoped that suppression to legacy algorithm versions rather than removing it, since the [DT-3861](https://broadworkbench.atlassian.net/browse/DT-3861) production audit showed the rows were real.

Two commits, gated on different preconditions and revertible independently:

- **`4ff7d14d` — the match-version suppression.** Removes `CLIENT_SUPPRESSING_VERSIONS`, the `client-suppresses` arm of `AlgorithmVersionHandling`, the absent-version fallback, `shouldAbstain`, `isOther`, and `AbstainDataUseCodes`. Anything below the backend-abstaining floor is now `unrecognized`, so a legacy or missing version reports the uninterpretable result rather than a hidden one.
- **`96e4cd5f` — the multi-primary collapse** in `processDefinedLimitations`. The function stays; `processRestrictionStatements` calls it for every non-`diseaseRestrictions` key.

### Do not merge yet — and the first commit may not be mergeable at all

**Corrected 2026-08-27.** An earlier version of this PR said the first commit was waiting on DT-3863's production recompute. That was wrong, and the measurement is [recorded on the ticket](https://broadworkbench.atlassian.net/browse/DT-4008?focusedCommentId=116591).

`LegacyDataUseService.recomputeAbstainingMatches` selects on `abstainsWhenMatched() && darCount > 0`, so it only reaches DARs that include an abstaining dataset. Legacy rows on canonical datasets are never restamped — a property of the code, not of any environment's data. A complete zero-failure run in dev left 393 legacy rows standing.

Production, measured before DT-3863 ships: **529 legacy rows, 48% of the match corpus**, of which only **3** sit under an open DataAccess election. No null `algorithm_version` rows anywhere. And `v1` is exactly the unresolved-identifier population — 410 rows, a perfect overlap — so 78% of the legacy corpus is what [DT-3942](https://broadworkbench.atlassian.net/browse/DT-3942) exists to reconcile.

So `4ff7d14d` is blocked on DT-3942, as [the epic](https://broadworkbench.atlassian.net/browse/DT-3860) always said. There is a further catch: DT-3942's scope backfills `dataset_id` and preserves rationales but never touches `algorithm_version`, so those 410 rows would still read `v1` after it completes. An explicit recompute step has to be added somewhere, or this commit stays blocked indefinitely. Raised on DT-3942.

`96e4cd5f` is unaffected. Its precondition is that no persisted record carries multiple primaries, and production has exactly one such dataset, awaiting its domain-approved disposition.

### Review notes

Three places where the code did not match the ticket's instructions:

**`isOther` is removed entirely, not just its secondary-Other arm.** `shouldAbstain` was its only caller, so scoping the edit as written would have orphaned it.

**The `isEmpty(matchVals)` → `N/A` branch is gone.** That branch *was* the suppression — `matchVals` was only ever empty because the legacy-and-unmatchable path deliberately emptied it. Derived unconditionally from a non-empty `matchResults`, it cannot be empty, so the branch is unreachable rather than merely unused.

**`processDefinedLimitations` loses its `dataUse` parameter.** With the collapse gone it is a pure translation lookup, and `processRestrictionStatements` already guards on the key being set — which is what the collapse was reading. That collapsed the per-key specs into one `it.each`, since they had become the same assertion, and made the empty-`diseaseRestrictions` case an exact duplicate.

The ticket also refers to `BACKEND_ABSTAINING_VERSIONS`; DT-3866 replaced that set with the `FIRST_BACKEND_ABSTAINING_VERSION` floor, which is what the change is written against. The floor stays a floor, so a matcher newer than this build still reads as `backend-abstains`.

Legacy versions in general-purpose fixtures move to `v5` — after this change a `v2` fixture silently exercises the uninterpretable path, which is not what those tests are about.

### Verification

`pnpm run lint` 0 errors, `pnpm run type-check` 0 errors, full suite 369 files / 4,321 tests green. E2E is unaffected — the Playwright specs cover liveness, about, auth, status, and home, none of which reach this code.

Coverage percentages dip slightly on two files, which is arithmetic rather than a regression — deleting fully-covered code shrinks the denominator while the pre-existing uncovered code stays. Absolute uncovered counts are flat or better:

| file | uncovered lines | uncovered branches |
|---|---|---|
| `BucketUtils.ts` | 2 → 2 | 6 → 5 |
| `dataUseTranslation.ts` | 36 → 36 | 27 → 26 |

----

- Verify all tests go green

[DT-4008]: https://broadworkbench.atlassian.net/browse/DT-4008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-4009]: https://broadworkbench.atlassian.net/browse/DT-4009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3866]: https://broadworkbench.atlassian.net/browse/DT-3866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3861]: https://broadworkbench.atlassian.net/browse/DT-3861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3942]: https://broadworkbench.atlassian.net/browse/DT-3942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useLibraryPageState.ts
+++ b/src/hooks/useLibraryPageState.ts
@@ -32,7 +32,7 @@ import { getFormattedName } from 'src/components/forms/SelectOptionInterface'
  * That class covers only the eight restrictions those forms collect, so the codes
  * below fill in the rest of the secondary vocabulary the corpus can hold: the
  * modifiers `consentTranslations` translates for the dataset views, plus the
- * `AbstainDataUseCodes` the voting flow recognizes. Without them a real, selectable
+ * population restrictions the index stores. Without them a real, selectable
  * checkbox would read as a bare abbreviation.
  *
  * `OTHER` is spelled out rather than taken from `SecondaryDataUseTerms.OTH`: the form's
--- a/src/libs/dataUseTranslation.ts
+++ b/src/libs/dataUseTranslation.ts
@@ -353,7 +353,7 @@ export const processRestrictionStatements = async (
   }
 
   if (key !== 'diseaseRestrictions') {
-    return processDefinedLimitations(key, dataUse, consentTranslations)
+    return processDefinedLimitations(key, consentTranslations)
   }
 
   const diseaseValue = value as (string | { label: string })[]
@@ -368,32 +368,13 @@ export const processRestrictionStatements = async (
   return processDiseaseRestrictionsWithUrls(diseaseValue as string[])
 }
 
-/**
- * Shows only the narrowest permission along the DS > HMB > GRU chain, since a broad statement beside
- * a narrow one misleads. POA sits outside that chain and still renders alongside DS.
- */
+/** Only `diseaseRestrictions` is function-valued, and it resolves through its own path. */
 export const processDefinedLimitations = (
   key: string,
-  dataUse: DataUse,
   translations: ConsentTranslationsMap,
 ): TranslationEntry | undefined => {
-  const targetKeys = ['hmbResearch', 'populationOriginsAncestry', 'generalUse']
-  const isHMBActive = !!dataUse.hmbResearch && isEmpty(dataUse.diseaseRestrictions)
-  const isPOAActive = !!dataUse.populationOriginsAncestry
-  const isGeneralUseActive = !!dataUse.generalUse && !isHMBActive && !isPOAActive && isEmpty(dataUse.diseaseRestrictions)
-  let statement: TranslationEntry | undefined
-  if (
-    !targetKeys.includes(key)
-    || (key === 'hmbResearch' && isHMBActive)
-    || (key === 'populationOriginsAncestry' && isPOAActive)
-    || (key === 'generalUse' && isGeneralUseActive)
-  ) {
-    const translation = translations[key]
-    if (translation && typeof translation !== 'function') {
-      statement = translation
-    }
-  }
-  return statement
+  const translation = translations[key]
+  return translation && typeof translation !== 'function' ? translation : undefined
 }
 
 // Extend DataUse to include otherRestrictions for runtime compatibility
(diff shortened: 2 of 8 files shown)


kevinmarete:
Superseded by #3896. This PR carried two removals gated on unrelated preconditions; the client-side legacy match suppression moved to DT-4032 and is parked on `km-dt-4032-legacy-match-suppression`, so nothing is lost. #3896 carries the translation collapse alone, whose precondition is met.

PR #3896, opened 2026-08-27
[DT-4008] Drop the multi-primary collapse from the Data Use translation

kevinmarete:
### Addresses

[DT-4008](https://broadworkbench.atlassian.net/browse/DT-4008).

Security risk: no.

### Summary

`processDefinedLimitations` hid a broader primary whenever a narrower one was set, along the DS > HMB > GRU chain. That only ever mattered for records carrying more than one primary, and production now has none — verified: datasets carrying more than one of GRU, HMB, POA, non-empty DS, or nonblank Other returns 0. The collapse can now only disagree with what Consent classifies.

The function stays, since `processRestrictionStatements` calls it for every non-`diseaseRestrictions` key, but it reduces to the translation lookup and no longer takes the data use — the caller already guards on the key being set, which is what the collapse was reading.

The per-key specs collapse into one `it.each`: with the branch gone they were the same assertion, and the empty-`diseaseRestrictions` case became a duplicate. A case for the function-valued entry is added, since that guard is the function's remaining behaviour.

----

- Verify all tests go green

[DT-4008]: https://broadworkbench.atlassian.net/browse/DT-4008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/libs/dataUseTranslation.ts
+++ b/src/libs/dataUseTranslation.ts
@@ -353,7 +353,7 @@ export const processRestrictionStatements = async (
   }
 
   if (key !== 'diseaseRestrictions') {
-    return processDefinedLimitations(key, dataUse, consentTranslations)
+    return processDefinedLimitations(key, consentTranslations)
   }
 
   const diseaseValue = value as (string | { label: string })[]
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
👍

rushtong (approved):
Very nice to see this cleanup, thank you 👍🏽

PR #3897, opened 2026-08-27
[DT-3873] Cover template validation over a real request

kevinmarete:
### Addresses

[DT-3873](https://broadworkbench.atlassian.net/browse/DT-3873).

Security risk: no.

### Summary

`StudyTemplateUpload.spec.tsx` mocks the ajax layer, so nothing proved a real multipart upload reaches Consent and comes back in the shape the page expects. Adds an e2e covering a rejected template, its errors clearing on remove, a valid template creating a draft, and a replacement after a rejection.

It stops at draft navigation and deletes the draft it creates. A study cannot be un-created, so going further would leave a permanent record in dev on every run, and `DataSubmissionFormV2.modes.spec.tsx` already covers past that point.

Also reports a failed source-draft cleanup through `ErrorReporter`. It showed a toast only, which is transient, so the rate of these was not measurable.

### What was blocking it

The automation accounts had never accepted the DUOS Terms of Service. `useSessionReconciler` redirects a user with `tosAccepted === false`, so the app rendered the ToS page at the requested URL — which presents as a missing element, not a redirect. ADMIN and CHAIR have now accepted it in dev; MEMBER, RESEARCHER and SIGNING_OFFICIAL have not.

Worth knowing separately: `auth.spec.ts` asserts the `Admin Console` / `DAC Console` tab is visible, and those are nav chrome that render on the ToS page too. It passed throughout, so it never proved the app was usable past sign-in. Tightening it to assert something only the destination page renders would catch this class of drift.

CHAIR is used because no automation account holds `DataSubmitter`, the real persona here. Chairperson is admitted by both the route's `RoleBAC` and the endpoint's `@RolesAllowed`.

### Note

The invalid fixture is copied from consent byte-identical with a provenance row added, as the fixture README requires.

----

- Verify all tests go green

[DT-3873]: https://broadworkbench.atlassian.net/browse/DT-3873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/e2e/studyTemplate.spec.ts
+++ b/test/e2e/studyTemplate.spec.ts
@@ -42,7 +42,11 @@ test.describe('study template validation', () => {
   test.beforeEach(async ({ page, signInAs }) => {
     await signInAs(ROLE)
     await page.goto(UPLOAD_PATH)
-    await expect(page.getByRole('heading', { name: 'Upload Study Template' })).toBeVisible()
+    // RoleBAC redirects rather than erroring, so a role that cannot reach the page shows up here
+    // as a changed URL rather than as a missing element further down.
+    await expect(page).toHaveURL(new RegExp(`${UPLOAD_PATH}$`))
+    // TableHeaderSection renders its title in a div, so the page's first real heading is this one.
+    await expect(page.getByRole('heading', { name: '1. Start from the blank template' })).toBeVisible()
   })
 
   test('reports the errors Consent found and creates no draft', async ({ page }) => {
--- a/test/e2e/studyTemplate.spec.ts
+++ b/test/e2e/studyTemplate.spec.ts
@@ -1,0 +1,100 @@
+import { test, expect } from './support/auth'
+import type { Page } from '@playwright/test'
+
+/*
+ * The only coverage that crosses the HTTP boundary: StudyTemplateUpload.spec.tsx mocks the ajax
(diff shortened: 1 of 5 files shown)


fboulnois (approved):
👍

test/e2e/studyTemplate.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+const deleteDraft = async (page: Page, draftUuid: string) => {
+  const status = await page.evaluate(async (uuid) => {
+    const stored = localStorage.getItem('OidcUser')
+    const user = stored ? JSON.parse(stored) : {}
+    const token = user?.profile?.idp_access_token ?? user?.id_token
+    const config = await (await fetch('/config.json')).json()
+    const response = await fetch(`${config.apiUrl}/api/draft/v1/${uuid}`, {
+      method: 'DELETE',
+      headers: { Authorization: `Bearer ${token}` },
+    })
```
rushtong, line 34: The BFF cutover will break this so we'll need another ticket to cover rewriting it. See https://broadworkbench.atlassian.net/browse/DT-3980 for what this might look like - it covers the background sign in feature which will be similarly broken.

rushtong (approved):
Looks good pending new ticket to cover the BFF rewrite 👍🏽

PR #3898, opened 2026-08-27
[DT-4000] Shift Data Library default to Card view of Studies

kevinmarete:
### Addresses

[DT-4000](https://broadworkbench.atlassian.net/browse/DT-4000).

Security risk: no.

### Summary

Studies becomes the Data Library's default tab and renders as a card grid instead of a DataGrid table. Scoped to the public Data Library through the `renderGrid` override on `LibraryPageShell`; every other tab, and the Studies table in the researcher console, is unchanged.

Each card carries the study name, PI, data types, species, phenotype, participant and dataset stats, and a row of access-type and DUO data-use pills. Selection, sorting, pagination and filters carry over from the table view.

### Review notes

The branch already contained the seam — the `renderGrid` override, the Studies default, the new `studyAsset` aggregations — but imported `accessManagementDisplay` and `StudyCardGrid`, neither of which existed, so it did not compile. This adds those, rebased onto develop.

Selection reuses `studyAsset.computeRowSelection` and `selectionToDatasetIds` rather than reimplementing them, so a study means the same set of datasets in either view.

No backend change. The two new aggregations read `dataUse.primary.code.keyword` and `accessManagement.keyword`, both already indexed and already aggregated by `datasetAsset`.

After

----

- Verify all tests go green

[DT-4000]: https://broadworkbench.atlassian.net/browse/DT-4000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/data_library/LibraryDataGrid.tsx
+++ b/src/components/data_library/LibraryDataGrid.tsx
@@ -6,7 +6,7 @@ import {
 } from '@mui/x-data-grid'
 import { Box, Typography, CircularProgress } from '@mui/material'
 import { isEmpty } from 'src/utils/NodashUtil'
-import { LibraryDataGridProps } from 'src/types/library'
+import { LibraryDataGridProps, PAGE_SIZE_OPTIONS } from 'src/types/library'
 import { assetRegistry, LibraryRow } from 'src/components/data_library/assets'
 
 const EMPTY_RADAR_IDS = new Set<number>()
@@ -71,13 +71,16 @@ export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
     ),
   }), [asset, data, selectedDatasetIds])
 
+  // Applied as a delta over this page's rows: replacing the whole selection would discard
+  // anything selected on another page or another tab.
   const handleSelectionChange = (newSelection: GridRowSelectionModel) => {
-    onSelectionChange(
-      asset.selectionToDatasetIds(
-        Array.isArray(data) ? data as LibraryRow[] : [],
-        Array.from(newSelection.ids),
-      ),
-    )
+    const rows = Array.isArray(data) ? data as LibraryRow[] : []
+    const pageDatasetIds = asset.selectionToDatasetIds(rows, rows.map(row => asset.getRowId(row)))
+    const stillSelected = new Set(asset.selectionToDatasetIds(rows, Array.from(newSelection.ids)))
+
+    const next = new Set(selectedDatasetIds)
+    pageDatasetIds.forEach(id => (stillSelected.has(id) ? next.add(id) : next.delete(id)))
+    onSelectionChange(Array.from(next))
   }
 
   const isRowSelectable = (params: { row: LibraryRow }) =>
@@ -126,7 +129,7 @@ export const LibraryDataGrid: React.FC<LibraryDataGridExtendedProps> = ({
         columns={columns}
         rowCount={total}
         loading={loading}
-        pageSizeOptions={[25, 50, 100]}
+        pageSizeOptions={PAGE_SIZE_OPTIONS}
         paginationModel={paginationModel}
         paginationMode="server"
         onPaginationModelChange={onPaginationChange}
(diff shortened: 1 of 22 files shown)


fboulnois (approved):
👍

rushtong:
Findings:

I confirmed this locally, sorting has no visual effect:
> StudyCardGrid.tsx:127 — the sort dropdown does not reorder the cards.
studyAsset uses sortingMode: 'client', and its buildQuery drops the sort argument. I confirmed the emitted Elasticsearch body is byte-identical with and without a sort. The table worked because LibraryDataGrid hands sortingMode='client' to MUI's DataGrid, which sorts the rows itself. StudyCardGrid renders studies.map(...) in raw response order and uses sortModel only to highlight the dropdown. Every option except "Relevance" does nothing. Fix: sort studies in the useMemo before rendering.

Codex and Claude both found this:
> useLibraryUrlState.ts:359 — the default-tab flip also hits "My Data Submissions".
DatasetSubmissions.tsx shares useLibraryPageState, includes AssetType.STUDIES in SUBMISSION_TABS, and never overrides the tab. It supplies extraColumns only when the tab is Datasets (line 186). A researcher who opens /dataset_submissions with no tab parameter now lands on Studies, so the Submitter, Custodian, Status, Edit and Delete columns are gone until they click Datasets. The PR description states the researcher console is unchanged. No test covers this.

Tested locally and confirmed. Choose 10 on the study tab, then switch to the dataset tab and the page controls drop
> StudyCardGrid.tsx:17 — page size 10 removes the table's rows-per-page control.
PAGE_SIZES offers 10; LibraryDataGrid sets pageSizeOptions={[25, 50, 100]}. handleTabChange keeps pageSize, and parsePageSizeParam accepts any positive integer. I rendered LibraryDataGrid with pageSize: 10: the footer became "1–1 of 1" and the "Rows per page" selector disappeared. The user is then stuck at 10 rows on every table tab with no control to change it.

I was unable to verify this, but it could be real:
> StudyCardGrid.tsx:140 — jump-to-last-page pulls every earlier study bucket (plausible, not measured).
The DataGrid footer only offered next/prev, so page depth grew one step at a time. Pagination renders the last page number directly, and buildQuery sizes its composite aggregation at (page + 1) * pageSize. On 5,000 studies at 100 per page, one click asks Elasticsearch for 5,000 buckets — each with a top_hits and a 10,000-term dataset_ids sub-aggregation — and the client keeps 100. Severity depends on the real corpus size.

kevinmarete:
All addressed in `fe883ec5`, verified on dev (314 studies).

- **Sorting** — moved into the aggregation. Composite ordered by `study.studyName.keyword` for the name sorts (study id trailing for uniqueness); `terms` ordered by the metric sub-agg for participants/datasets, which a composite can't express. Not sorted in the `useMemo` as suggested: that data is already page-sliced, so it would have reordered 25 arbitrary buckets while looking corpus-wide. Page 1 ends `…JPNFKA_GRU_WES`, page 2 opens `…KENKIL_GRU_WES`.
- **Default tab** — now a parameter; only `DataLibrary` passes `STUDIES`. `/dataset_submissions` lands on Datasets.
- **Page size 10** — `PAGE_SIZE_OPTIONS` shared by both grids and enforced in `parsePageSizeParam`, which also closes the URL path.
- **Deep paging** — capped at 10000, matching `STUDIES_AGG`. At ~4.1k studies it never fires. Worth noting the cost was the same at page N however you got there; offset paging over an aggregation is inherently O(page × pageSize).
- **Selection** — both grids apply a delta instead of rebuilding from the page (`LibraryDataGrid` had this too). `computeRowSelection` is `.every()` now, with an indeterminate state, so a part-selected study can't be promoted.
- **shard_size** — added, exact under the cap.
- **Relevance** — renamed to Default; true relevance isn't expressible over a composite.
- **Footer undercount** — resolved per dataset via a `dataset → study` map (`selectionStudyMap.ts`). Select DUOS-000771 (p1) + DUOS-000802 (p2) → "2 datasets from 1 study"; deselect the first → "1 dataset from 1 study".

rushtong (approved):
Thank you for the updates 👍🏽

PR #3899, opened 2026-08-31
[DT-4019] Harden cookie attributes: Phase 5, story 5-D

rushtong:
### Addresses

[DT-4019](https://broadworkbench.atlassian.net/browse/DT-4019) — Phase 5, story 5-D: verify strict cookie attributes on all cookies.

Security risk: low.

### Summary

The application sets exactly two cookies. This story hardens one and adds verification for the other.

**`cookie_control` (client, `src/utils/CookieUtils.ts`)** — the GDPR banner preference cookie. It was written with only `path` and `max-age`:

- Adds `SameSite=Strict`, not `Lax`. This cookie takes no part in the OAuth callback, so nothing needs it on a cross-site navigation.
- Adds `Secure` only when `location.protocol` is `https:`. Unconditional `Secure` would make plain-HTTP dev setups drop the cookie silently.
- URL-encodes the JSON value. `getCookiePairs()` already decodes, but production wrote raw JSON, so the read and write paths disagreed.
- Guards the `JSON.parse` in `getAcknowledged()`. The value is user-controlled and must not throw during render.

`HttpOnly` does not apply — the client reads this cookie with `document.cookie`.

**`sessionId` (server, `server/src/index.ts`)** — verification only. It already sets `HttpOnly`, `Secure` in production, `SameSite=Lax` and `Path=/`. Added tests in `server/test/index.test.ts` that assert the options `buildApp()` hands `@fastify/session`, including the production `Secure` branch. `session.test.ts` and the load harness stand the plugin up with their own options, so they cannot catch a drift in the production configuration.

Later hardening noted for Phase 6, out of scope here: migrate the session cookie to a `__Host-` prefixed name so the browser itself enforces `Secure`, host-only scope and `Path=/`.

### Test plan

- `pnpm --filter duos-server test` — 370 pass, 17 files.
- `pnpm exec vitest run test/utils/CookieUtils.spec.ts` — 11 pass. The spec now asserts the real attribute string: `path=/`, `max-age`, `SameSite=Strict`, `Secure` present over HTTPS and absent over HTTP, an encoded value, a round trip, and a corrupt value that reads `false` instead of throwing.
- `pnpm exec vitest run test/components/CookieBanner.spec.tsx test/components/DuosFooter.spec.tsx` — 6 pass (the two consumers).
- `tsc --noEmit` clean for client and server; `oxlint` and `eslint` clean on the changed files.
- Verified the new production-`Secure` test fails when the assertion is flipped, so it is not passing trivially.

----

- Verify all tests go green

[DT-4019]: https://broadworkbench.atlassian.net/browse/DT-4019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/test/index.test.ts
+++ b/server/test/index.test.ts
@@ -173,11 +173,6 @@ describe('plugin registration order', () => {
   })
 })
 
-// Story 5-D: the session cookie is configured in Epic 1; this block is the
-// hardening-phase verification that the PRODUCTION options are the strict set.
-// Asserted against the mocked @fastify/session registration because that is the
-// only place buildApp()'s own options are visible — session.test.ts and the load
-// harness stand the plugin up themselves and so cannot catch a drift here.
 describe('session cookie attributes', () => {
   async function sessionCookieOptions() {
     const { default: sessionPlugin } = await import('@fastify/session')
@@ -190,9 +185,6 @@ describe('session cookie attributes', () => {
   it('sets HttpOnly, SameSite=Lax and Path=/', async () => {
     const cookie = await sessionCookieOptions()
     expect(cookie.httpOnly).toBe(true)
-    // Lax, not Strict: Strict would strip the cookie from the top-level
-    // redirect B2C makes back to /auth/callback, which would arrive sessionless
-    // and lose the PKCE verifier and state.
     expect(cookie.sameSite).toBe('lax')
     expect(cookie.path).toBe('/')
   })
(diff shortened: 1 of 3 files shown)


otchet-broad (approved):
👍

fboulnois (approved):
👍

PR #3900, opened 2026-08-31
[DT-3996] Record the SameSite=Lax decision and make it a single source of truth

rushtong:
## Addresses
[DT-3996](https://broadworkbench.atlassian.net/browse/DT-3996)

**Security risk: no** — this PR documents and tests the existing session-cookie behavior. No production behavior is intentionally changed.

## Summary

Documents why the DUOS session cookie uses `SameSite=Lax` instead of `Strict` and adds regression coverage for that decision.

- Adds ADR-012 and an Infosec-focused explanation of the session-cookie threat model.
- Explains that `SameSite=Strict` would prevent the session cookie from being sent on Azure B2C's cross-site redirect to `/auth/callback`, causing sign-in to fail.
- Records the remaining limitations of `Lax`, including sibling `*.broadinstitute.org` subdomains, and the controls that address them:
  - session-bound CSRF tokens for unsafe routes;
  - PKCE state binding for the OAuth flow;
  - Fetch Metadata protections for state-changing GET requests.
- Centralizes the production session configuration in `sessionOptions.ts` and reuses it in all session-related test harnesses, preventing configuration drift.
- Adds integration tests demonstrating that:
  - `Lax` preserves the session through the B2C callback;
  - `Strict` withholds the cookie and causes the callback to fail.

The decision assumes the authorization response continues to use `response_mode=query`. A future change to `form_post` must revisit the cookie configuration because `Lax` cookies are not sent on cross-site POST callbacks.

[DT-3996]: https://broadworkbench.atlassian.net/browse/DT-3996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rushtong:
Review feedback addressed in 32a4045d. All three blocking points were correct; I verified each against the source before acting.

**1. `response_mode` is now an enforced invariant.** You were right that a bare "not `form_post`" assertion would still leave the test simulating a GET, so I went further than the minimal fix: `login.ts` passes `response_mode: 'query'` explicitly (the code-flow default, so behavior-preserving), and the test's browser model now derives the return trip's method from the value production actually chose. It models both axes — the mode decides GET or POST, `SameSite` then decides whether the cookie rides along, since a cross-site POST withholds a `Lax` cookie exactly as `Strict` does. `login.test.ts` pins the value independently.

Proven both ways: with the direct assertion temporarily removed, `form_post` still fails the Lax test at the browser model (`expected undefined to be 'sessionId=…'`), so the derived method is a real guard rather than decoration.

I also took the point about the "executable" claim. The ADR now states plainly that a **B2C-side** policy change is invisible to this repository and its CI, and can only be caught by a human reading the ADR.

**2. The split-cookie alternative is corrected.** "Buys nothing against the live threat" was wrong and contradicted ADR-009 in the same directory. The ADR now separates what the design *would* achieve — it narrows the ADR-009 GET residual, on **every** browser including those sending no Fetch Metadata headers — from why it is still rejected: it leaves the higher-priority sibling-subdomain threat untouched, and the GET residual belongs upstream in DT-3945. A revisit trigger is recorded: if DT-3945 stalls and pre-Fetch-Metadata browsers still matter.

**3. CSRF coverage is qualified.** Now "every *cookie-authenticated* unsafe route", naming both exemptions and why they are safe. I verified the reasoning myself rather than restating it: both paths are in `UNAUTHENTICATED_PATHS`, so no `Authorization` is injected; `rewriteRequestHeaders` strips the caller's own `cookie` and `authorization`; the exemption is keyed on method *and* path; and drift fails closed with `MissingCSRFSecretError`. ECM, TDR and Bard exempt nothing.

**Non-blocking cookie name — fixed rather than deferred.** You were right that the constant was decorative. Since this PR exists to end exactly that kind of fake centralization, leaving it half-wired would have undercut the change. `cookieName: SESSION_COOKIE_NAME` now goes to the plugin, and all five production `clearCookie` sites import it. Test files keep their literals deliberately: they assert the observable wire name, so a rename has to be asserted rather than assumed to be self-consistent.

**Also consolidated** on the repo owner's request: the separate infosec note is folded into ADR-012, which now opens with a self-contained "The infosec question, answered" section for that audience.

**Verification:** 385 tests pass, typecheck and lint clean. Mutations still fail as required — `form_post` fails 2 tests, `sameSite: 'strict'` fails 5, `rolling: true` fails 3 (with the `ERR_HTTP_HEADERS_SENT` flood the config comment predicts).

rushtong:
Copilot review addressed in 83af4946 — one accepted in full, one accepted in part.

**1. `SameSite=None` without `Secure` — valid, fixed by narrowing the type.** Correct catch. `secure` defaults to production-only, so `sameSiteOverride: 'none'` would have built a cookie no browser accepts, and because `app.inject()` has no cookie jar, no test would have failed. Worth adding that `'none'` was also unreachable: nothing passed it, and its branch in the browser model was dead code.

Rather than guard the pairing, I made it unrepresentable — the override is now typed `'strict'` only, and the dead branch is gone. That is the same reasoning that excluded `'lax'` from the type in the first place: an option that must never be combined a certain way is better removed than policed. If `None` is ever genuinely needed, it should arrive together with a `secure` guard.

**2. Duplicated cookie lookup — accepted; importing the constant — declined, with reasons.**

The duplication was real, and worse than it looked: `cookieAfterB2cRedirect` used `find` where `sessionCookieHeader` uses last-wins plus a cleared-cookie check. Since the callback can emit two `sessionId` headers after session rotation (story 5-C), the browser model had subtly wrong semantics. Both now share one `sessionCookie()` finder, so the model inherits the careful version.

I did not switch it to `SESSION_COOKIE_NAME`, and I want to be explicit that this is a decision rather than an oversight — it is now documented at the finder. Tests here pin the *observable wire name*. ADR-012 § Alternatives actively contemplates renaming this cookie to `__Host-sessionId` as later hardening; if the suite imported the constant, that rename would stay green everywhere and the suite would assert nothing about the name a browser actually sees. Pinning the literal forces a rename to be an asserted, deliberate change. The single-source-of-truth goal applies to production code, which does now import the constant at all five `clearCookie` sites and passes it to the plugin as `cookieName`.

Happy to revisit if you would rather have the constant imported and a separate explicit assertion on the literal name — that is the other coherent shape, and it costs one more test.

**Verification:** 385 tests pass, typecheck and lint clean. Because these changes touched the guard helpers themselves, I re-ran all three mutations rather than assuming they still bite: `response_mode: 'form_post'` fails 2 tests, `sameSite: 'strict'` fails 5, `rolling: true` fails 3.

fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3904, opened 2026-08-31
[DT-4021] Add @fastify/helmet with a strict Content Security Policy (BFF Phase 5, story 5-F)

rushtong:
### Addresses

[DT-4021](https://broadworkbench.atlassian.net/browse/DT-4021) — BFF Phase 5, story 5-F. Security risk: **low** (the policy ships report-only; enforcement is a per-environment env var).

### Summary

The server sent no security headers. This registers `@fastify/helmet` with a strict Content Security Policy, plus a hardened violation-report sink.

**`connect-src` is derived from `config.json`, never hardcoded.** Only inventoried, active fields are read, each reduced to its origin; blanks and unparseable values drop out. Two literals remain — `'self'` and the banner bucket.

The allowlist is mode-specific:

| Mode | `connect-src` |
|---|---|
| BFF (`bffEnabled: true`) | `'self'`, `apiUrl` (unauthenticated feature flags, read pre-login), `bardApiUrl` (anonymous metrics), the banner bucket |
| Legacy | the above plus `ecmApiUrl` and `tdrApiUrl` |

ECM and TDR are omitted in BFF mode **on purpose**: those calls are same-origin after cutover. `terraUrl` is never allowlisted — it is navigated to, not fetched. `DUOS_API_URL` overrides the file's `apiUrl`, and the policy follows the override.

**Report-only first, with real collection.** `DUOS_CSP_REPORT_ONLY` defaults to `true`. `POST /csp-report` accepts reports in both browser shapes and is bounded five ways: an 8 KB body limit, exactly two accepted media types (everything else 415s before a handler runs), a per-request cap on how many reports one body may contribute, a logged-field allowlist with truncation and control-character flattening, and a fixed-window log budget charged once per request. It always answers 204 and is registered outside both cutover switches, so a legacy deployment gets it too.

**Helmet defaults that would have broken things:**

- `Cross-Origin-Opener-Policy` is **off in legacy mode**, `same-origin-allow-popups` only under `bffEnabled`. `-allow-popups` is not a safe middle ground for the legacy popup — see the note below.
- `Cross-Origin-Embedder-Policy` stays off — the banner bucket and the two direct upstreams send no CORP header.
- HSTS and `upgrade-insecure-requests` are production only, so plain-HTTP dev and docker-compose setups keep working.
- Vite HMR gets `'unsafe-inline'` on `script-src` and `ws:`/`wss:` on `connect-src`, both behind the existing `isDev` flag.

**`script-src` stays `'self'`.** Story 5-A removed `react-google-charts`, the only third-party script. Neither the root `index.html` nor the built `build/index.html` carries an inline script or style, so no hash or nonce is needed.

**The proxy's download sandbox survives.** `upstreamProxy.ts` sets `content-security-policy: sandbox` per reply. Both writes reach the raw response through `setHeader` — Fastify's stream path deliberately avoids `writeHead` — and the proxy's wins by running second. Ordering is the whole mechanism, so it is fragile: registering helmet later would silently strip the sandbox. Asserted against the real proxy, alongside a sibling route proving helmet is live in the same app.

**COOP was the dangerous one, and self-review caught it.** I first shipped `same-origin-allow-popups`, believing it protected the legacy B2C popup sign-in. It does not: that token spares a popup only on its *initial* navigation, while the popup's own document is `unsafe-none`. The return leg from B2C compares `unsafe-none` against our COOP — a mismatch — so the browser swaps browsing context groups and `window.opener` goes null. `oidc-client-ts` then throws on `postMessage` and `signinPopup()` never resolves. Measured in Chromium with two local origins and a cross-origin hop between:

```
COOP on callback  : openerNull=true,  postMessage threw, opener saw nothing
COOP off (control): openerNull=false, postMessage fine,  opener got the message
```

**COOP is not part of the CSP and has no report-only mode**, so `DUOS_CSP_REPORT_ONLY` would not have caught this — it breaks sign-in on the first deploy, in every environment, since none has `bffEnabled` on yet. Legacy deployments now get no COOP until Epic 6 retires that flow.

Rationale, the full inventory, and the consequences are in `docs/plans/bff_adrs/ADR-013-content-security-policy.md`, now linked and summarised in `docs/plans/BFF_Overview.md` alongside the other ADRs.

### Review round

`develop` is merged in (one conflict, imports only). A full review against the story, plus the outstanding Copilot comment, turned up seven defects. All are fixed in `8f0c455f`.

**The report sink charged its log budget per *report*, not per request.** 174 report entries fit inside the 8 KB body limit against a 60-per-minute cap. Neither content-type parser checks the body's shape, so one request per minute — from anyone on the internet — exhausted the window and silently dropped every genuine violation. That is precisely the collection run this story exists to perform, switched off for the cost of one POST a minute. Now capped at 8 reports per request and charged one unit per request.

**Fastify's own request logging was unbounded on the sink.** The route set no log level, so every POST emitted an `incoming request` / `request completed` pair — including the 415s and 413s that never reach a handler. None of the documented controls touched it, so log volume tracked attacker request rate directly. The route is now `logLevel: 'silent'`, with the report lines written through `app.log` carrying `reqId` by hand. `'warn'` was the obvious fix and the wrong one: a route level *pins* rather than raises, so it would have overridden a deployment that deliberately configured a lower level.

**Four smaller ones:**

- The dropped-report count only flushed when a later request crossed the window boundary, so a flood followed by silence left no trace, and when it did flush it borrowed an unrelated request's `reqId`. Now flushed on a heartbeat and on close, through the app logger.
- `index.ts` installs the app error handler *after* this plugin registers, so 415 and 413 answered with `FST_ERR_CTP_INVALID_MEDIA_TYPE` / `FST_ERR_CTP_BODY_TOO_LARGE` from the one route reachable without a session, where every other route generalises. The plugin scope now has its own handler returning a bare status.
- The e2e spec ended on fixed 1500 ms sleeps and `networkidle`, so a violation raised by a late request went unrecorded and the case passed green on a policy that would break the page — the exact failure the harness self-check exists to prevent. Every wait is now on a named response, set up before the navigation that triggers it. This also clears the 6 SonarCloud MAJOR issues (S2925 ×3, S9332 ×3).
- `attachReportOnlyPolicy` copied `content-encoding` and `content-length` onto a body Playwright had already decoded. Harmless today, an opaque navigation failure the moment this harness moves behind the Fastify server.

**Copilot's comment was still valid.** The BFF connect-src case set `DUOS_API_URL` and the `security headers` afterEach never restored it, leaking the override to any suite sharing the Vitest worker. Fixed to match the sibling `GET /config.json` block.

**Two audit results recorded rather than acted on**, both in ADR-013:

1. **B2C is navigated to, never fetched.** The authority URL is not a `config.json` field at all — it arrives in Consent's `/oauth2/configuration` response — so the "every origin derives from config" rule has nothing to derive from if that changes. It does not need to today: `oidcBroker.ts` passes an explicit `metadata` object, which short-circuits discovery, and JWKS, userinfo, session monitoring, and the silent-renew iframe are all dead in this configuration. Four one-line edits to that file would each break sign-in with no config field to repair it from.
2. **`FeatureFlag.ts` has no caller anywhere in `src/`** — only its own unit test. Nothing else calls `getUpstreamApiUrl` outside the legacy-only broker, so BFF-mode `connect-src` allowlists `apiUrl` for a flow that does not run. Kept deliberately: the entry costs nothing while the policy is report-only, and dropping it would greet whoever wires the module up with a blocked request and no obvious cause. It does mean `/public/features/*` may have no consumer, which the follow-up below should settle before it is built.

**Two follow-ups this PR does not do:**

1. **Dedicated public BFF endpoints** — `/public/notifications`, `/public/features/*`, `/public/metrics/event`, each omitting session-token injection and carrying response validation, body limits, and abuse controls. That collapses BFF-mode `connect-src` to `'self'`. Required deliverable of this story; **ticket still to be filed** — DT-4021's only issue link is "split from DT-3610", so nothing has been created for it yet. Text below.
2. **The e2e collector runs locally, not in CI.** `pnpm run serve` is `vite preview`, which sends no headers, so `test/e2e/support/csp.ts` fulfils the document itself to attach the policy. Chrome then treats that document as coming from an unknown address space, making every same-origin subresource a public-to-loopback Private Network Access transition — allowed from a secure context, blocked without one. Locally the preview server speaks HTTPS and the page loads; in CI it speaks plain HTTP and nothing loads. Disabling the browser's PNA checks would be the wrong fix, so the spec is `test.skip`ped under `CI`. The right fix is to serve the e2e run through the Fastify server, which sends the real header and needs no interception — Epic 6 harness work.

<details>
<summary>Follow-up ticket text</summary>

**Title:** Move the three direct browser flows onto dedicated public BFF endpoints

Under `bffEnabled`, three browser connections still bypass the BFF and force exact-origin entries into the CSP `connect-src`: banner notifications (`storage.googleapis.com/broad-duos-banners`), unauthenticated feature flags (`/feature`, `/feature/:key` on Consent), and anonymous Bard metrics.

They cannot be folded into the existing proxies as `unauthenticatedPaths`: that list is an **exact** set match (`server/src/proxy/upstreamProxy.ts`), so `/feature` alone leaves `/feature/:key` returning 401; and anonymous and identified metrics share `POST /api/event`, where marking the path unauthenticated would silently turn *identified* metrics anonymous.

Add `/public/notifications`, `/public/features/*`, and `/public/metrics/event` as dedicated endpoints that explicitly omit session-token injection and carry response validation, body limits, and abuse controls. Then drop the three origins from `BFF_CONNECT_FIELDS`/`BANNER_ORIGIN` in `server/src/security/csp.ts`, leaving BFF-mode `connect-src` at `'self'`.

**Settle first:** `src/libs/ajax/FeatureFlag.ts` currently has no caller in `src/`. `/public/features/*` may have no consumer to build for — decide whether the module is being wired up or removed before building the endpoint.

See `docs/plans/bff_adrs/ADR-013-content-security-policy.md`.
</details>

### Test plan

- `pnpm --filter duos-server test` — **475 pass**. 82 of those are this story's: 71 across `server/test/csp.test.ts` and `server/test/cspReport.test.ts`, plus 11 in the `security headers` block of `index.test.ts`. Covers the mode split, blank/unparseable config fields, the dev allowances, the serialized header in both report-only and enforcing modes, the `DUOS_API_URL` override reaching the policy, the proxy sandbox precedence against the real proxy, the COOP mode split, and the headers reaching the SPA document route as well as the report sink. The sink's own bounds now include: one oversized batch is charged a single unit and a later client is still logged; the per-request report cap; 415 and 413 naming no framework error code; the dropped count written on close; Fastify's per-request pair suppressed while a sibling route still emits it; and a logger configured below `warn` staying silent.
- `pnpm run test` (client) — **4509 pass** across 379 files. No client source is touched, and vitest excludes `test/e2e/**`.
- `pnpm run lint`, `pnpm run type-check`, `pnpm --filter duos-server typecheck` — clean, zero errors.
- **Not yet re-run: the Playwright spec.** It passed locally before the review round, but the waits were rewritten since and the spec has not been exercised against the built app again. It needs `pnpm run build && pnpm run serve` in one shell and `pnpm exec playwright test test/e2e/csp.spec.ts` in another; it is skipped under `CI`. The signed-in case additionally needs `DUOS_AUTOMATION_RESEARCHER_SA` and remains unrun.
- Not yet done: the report-only run on a deployed environment. That is the gate for flipping `DUOS_CSP_REPORT_ONLY=false` per environment.

----

- Verify all tests go green


rushtong:
Superseded by a three-PR stack, so each piece can be reviewed and deployed on its own:

- #3907 — story 5-F1, helmet's non-CSP headers (~340 lines). **COOP ships alone**: it can sever legacy sign-in, has no report-only mode, and every environment is legacy today. Deploy and verify sign-in before anything rides on it.
- #3908 — story 5-F2, the CSP violation report sink (~915).
- #3909 — story 5-F3, the policy itself, report-only (~720).

The e2e collector is not in the stack. It cannot run in CI until the e2e run is served through Fastify, and a spec that never runs is not coverage — it is now story 6-K in Epic 6.

The proxy side is a separate terra-helmfile PR (story 5-F4). Without it the enforcing header never reaches the browser.

Closing this in favour of the stack.

PR #3905, opened 2026-08-31
[DT-4020] End the B2C session at sign-out (BFF Phase 5, story 5-E)

rushtong:
## Addresses

Partially addresses [DT-4020](https://broadworkbench.atlassian.net/browse/DT-4020): BFF Phase 5, story 5-E — front-channel logout.
Requires https://github.com/broadinstitute/terra-helmfile/pull/6495 which adds a post-logout URL to DUOS' BFF configurations.

## Why

`POST /auth/logout` destroyed the DUOS BFF session but left the browser’s Azure B2C SSO session active. This change adds BFF initiated B2C logout while preserving a reliable local-only fallback.

### Server

- Builds the B2C end-session URL from OIDC discovery using the session’s ID token.
- Returns:
  - `200 { redirectUrl }` when B2C logout is available.
  - `204` when only the local BFF session can be terminated.
- Always destroys the BFF session and clears its cookie, even if discovery, revocation, audit logging, or end-session URL construction fails.
- Adds `DUOS_POST_LOGOUT_REDIRECT_URI` to the documented and generated environment configuration.

### Client

- Makes `Auth.signOut` responsible for the complete logout flow and navigation.
- Treats logout as confirmed only when:
  - `/auth/logout` returns `200` with a valid redirect URL;
  - it returns `204`; or
  - a follow-up `/auth/me` probe confirms that the session is gone.
- Retries once when the session remains authenticated and leaves local state intact when logout cannot be confirmed.
- Fixes the stale-CSRF-token path to process the retry response.
- Coalesces concurrent sign-out requests to prevent competing logout flows.
- Displays a persistent warning with a Retry action when logout cannot be confirmed.

### Post-logout routing

- Adds the public `/post-logout` route.
- Stores the requested local destination in `sessionStorage` across the B2C round trip.
- Validates the destination on both write and read, removes it after use, and defaults to `/`.
- I have manually updated all [B2C app registrations](https://portal.azure.com/) (`duos-dev`, `duos-staging`, `duos-bee`, and `duos` (prod)) to handle the `/post-logout` routes.

## Scope note

Automatic logout following a terminal API `401` remains local-only. The proxy has already destroyed the BFF session at that point, including the ID token required for B2C logout. The existing `prompt: 'login'` behavior continues to require the login screen on the next sign-in.
 

[DT-4020]: https://broadworkbench.atlassian.net/browse/DT-4020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/PostLogout.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * replaces the current history entry with it — defaulting to '/' when the
+ * value is missing or is not a same-origin path.
+ *
+ * The page performs NO state-changing request: a state-changing GET here would
+ * reintroduce the CSRF exposure the BFF closed.
+ */
+export default function PostLogout() {
+  useEffect(() => {
+    Redirect.replace(takePostLogoutTarget())
+  }, [])
```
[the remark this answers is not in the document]
rushtong: Checked this, and no change is needed: the rule does not apply here.

`Redirect` and `takePostLogoutTarget` are module-level imports, not component-scope values. `exhaustive-deps` tracks only values from inside the component — props, state, context, and locals. Module bindings are stable, so the rule ignores them.

Verified rather than assumed:

| Check | Result |
| --- | --- |
| `pnpm exec oxlint src/pages/PostLogout.tsx` | exit 0, no output |
| `pnpm exec eslint src/pages/PostLogout.tsx` | exit 0, no output |
| CI `lint` job on this branch | pass |

A clean run proves nothing unless the rule is live, so I confirmed that too. Temporarily adding a component-scope local to this same effect produced:

```
src/pages/PostLogout.tsx:22:48: warning react-hooks(exhaustive-deps):
React Hook useEffect has a missing dependency: 'sanityLocal.value'
```

So the rule is enabled and fires for a component-scope value, and stays silent for these two imports. That probe was reverted.

The second half of the claim also fails on its own: the rule is configured as `"warn"` in `.oxlintrc.json`, and oxlint exited 0 even while emitting the warning above — so it could not fail CI. The branch already carries pre-existing `react(set-state-in-effect)` warnings in other files while `pnpm lint` passes.

Adding an `oxlint-disable-next-line` here would leave a dead suppression for a rule that never fires, which would mislead the next reader. The empty dependency array is also the intent: the effect runs once on mount and then navigates away.

src/libs/auth/auth.ts
```diff
@@ -35,6 +45,204 @@ export const Redirect = {
...
+ * first caller's validated target wins. The promise is cleared once the
+ * attempt settles — especially on 'unconfirmed', so Retry starts a fresh
+ * attempt instead of re-awaiting the failed one.
+ */
+let signOutInFlight: Promise<SignOutResult> | null = null
+
+const coalescedBffSignOut = (redirectTo: string): Promise<SignOutResult> => {
+  signOutInFlight ??= bffSignOut(redirectTo)
+    .catch(() => UNCONFIRMED)
+    .finally(() => {
```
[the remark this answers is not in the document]
rushtong: Good catch — fixed in ade3b296.

The path is reachable: `Storage.clearStorage()` calls `localStorage.clear()` unguarded, so blocked site data makes `localCleanup()` throw and `bffSignOut` reject.

Confirmed with a test before changing anything — it left one `sessionStorage` entry behind:

```
AssertionError: expected Storage{…} to have a length of +0 but got 1
```

The `catch` now calls `clearPostLogoutTarget()`, so the invariant "an unconfirmed sign-out leaves no stored target" holds unconditionally rather than only on the classified path. I also verified the new test fails again when the fix is reverted, so it is a real guard.

On severity, for anyone reading later: the exposure was minor, not a redirect risk. `takePostLogoutTarget()` validates as a same-origin path and deletes on read, and any later sign-out overwrites the value. The worst case was a direct visit to `/post-logout` landing on a stale local path instead of `/`.

fboulnois (approved):
👍

server/src/auth/logout.ts
```diff
@@ -1,23 +1,31 @@
...
+}
+
+async function revokeTokens(request: FastifyRequest): Promise<void> {
   try {
     const config = await getOidcConfig()
 
-    // B2C does not reliably expose a revocation_endpoint in its discovery
-    // document, so revocation is typically skipped here. v6: the endpoint is
-    // read from config.serverMetadata(), and revocation is tokenRevocation().
     if (config.serverMetadata().revocation_endpoint) {
```
otchet-broad, line 35: Doesn't this imply that we really need revocation_endpoints based on the provider that issued the token - and if one is configured for that provider, that we handle the revocation?  I'm wondering if this needs to be combined with information from the request.session to actually make this determiniation.

src/libs/auth/signOutNotice.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ToastNotifications.showError({
+    timeout: null,
+    onDismiss: () => {
+      if (activeNotice === notice) {
+        activeNotice = null
+      }
+    },
+    text: (
+      <span data-cy="unconfirmed-sign-out-notice">
+        We could not confirm that you were signed out. You may still be signed in.
```
otchet-broad, line 26: If we expect the user to click the retry button, we should ask them to do that here.

server/src/auth/logout.ts
rushtong: Good question, and the short answer is that there is only ever one issuer here, so one revocation endpoint is the right check. Checked against the code rather than from memory:

- The BFF has exactly one OIDC client. `getOidcConfig()` discovers `DUOS_AZURE_ISSUER_URL` once per process; there is no per-provider client to look up an endpoint on.
- Every session token comes from that single B2C exchange — `callback.ts` sets `accessToken`, `refreshToken`, and `idToken` from one `tokens` response.
- `request.session.idp` records which sub-provider the user chose **at B2C**, and the type comment says so: "All tokens are B2C-issued regardless — this field exists for the audit trail and observability, not client selection."

So combining with `request.session` would not give us anything to act on: DUOS never holds a Google or Microsoft token, so there is no upstream credential those providers would accept for revocation. Provider selection happens on the B2C-hosted login page and stays inside B2C.

If the underlying worry is "does the user's Google/Microsoft session survive sign-out?" — that one is real, but revocation is not the lever. Federated single sign-out is: B2C attempts it at the upstream provider, and provider behavior varies, so we treat that session as possibly surviving. `prompt: 'login'` on the authorization request (`login.ts`) is what guarantees the login screen always appears on the next sign-in, regardless of any surviving cookie.

I have added that reasoning as a comment above `revokeTokens` in 20f84fa3, since the question is a fair one for the next reader to have.

src/libs/auth/signOutNotice.tsx
rushtong: Agreed — fixed in 20f84fa3. The notice now reads:

> We could not confirm that you were signed out. You may still be signed in. Select Retry to try again.

A Retry button with no ask left the user to work out what to do about it. The spec asserts the instruction too, so it cannot quietly disappear.

otchet-broad (approved):
👍

PR #3907, opened 2026-09-04
[DT-4021] Register @fastify/helmet for the non-CSP security headers (BFF Phase 5, story 5-F1)

rushtong:
### Addresses

[DT-4021](https://broadworkbench.atlassian.net/browse/DT-4021) — BFF Phase 5, **story 5-F1**. Security risk: **medium** (COOP can sever legacy sign-in; that is why this ships alone).

This is the first of several PRs. The full implementation of DT-4021 is much larger and riskier than I had hoped so it is broken down here:
- **5-F1 security headers (this PR)** → 
- [5-F2 report sink](https://github.com/DataBiosphere/duos-ui/pull/3908) → 
- [5-F3 the policy](https://github.com/DataBiosphere/duos-ui/pull/3909)  → 
- [5-F4 terra-helmfile proxy](https://github.com/broadinstitute/terra-helmfile/pull/6497)  → 
- 5-F5 run report - no code, just a report execution → 
- [5-F6 public BFF endpoints](https://github.com/DataBiosphere/duos-ui/pull/3914)

### Summary

The server sent no security headers and relies on the standard Terra proxy. This registers `@fastify/helmet` with the Content Security Policy **deliberately off** — that is [5-F3](https://github.com/DataBiosphere/duos-ui/pull/3909). `Cross-Origin-Opener-Policy` is off entirely in legacy mode since legacy sign-in needs to open a popup window. Rationale in `docs/plans/bff_adrs/ADR-013-content-security-policy.md`, which is written across the stack and grows with it.

### Test notes
I have tested both legacy and BFF sign in with the current proxy settings and both work - that is the primary risk in this PR.

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.\n\n- Label PR with a Jira ticket number and include a link to the ticket\n- Label PR with a security risk modifier [no, low, medium, high]\n- PR describes scope of changes\n- Get a minimum of one thumbs worth of review, preferably two if enough team members are available\n- Get PO sign-off for all non-trivial UI or workflow changes\n- Verify all tests go green\n- Test this change deployed correctly and works on dev environment after deployment\n

[DT-4021]: https://broadworkbench.atlassian.net/browse/DT-4021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3908, opened 2026-09-04
[DT-4021] Add the CSP violation report sink (BFF Phase 5, story 5-F2)

rushtong:
## Addresses

[DT-4021](https://broadworkbench.atlassian.net/browse/DT-4021) — BFF Phase 5, **story 5-F2**. Security risk: **medium** (a new endpoint on the public internet).

Part of a chain of PRs addressing this ticket:
- [5-F1 security headers](https://github.com/DataBiosphere/duos-ui/pull/3907) → 
- **5-F2 report sink (this PR)** → 
- [5-F3 the policy](https://github.com/DataBiosphere/duos-ui/pull/3909) → 
- [5-F4 terra-helmfile proxy](https://github.com/broadinstitute/terra-helmfile/pull/6497)  → 
- 5-F5 run report - no code, just a report execution → 
- [5-F6 public BFF endpoints](https://github.com/DataBiosphere/duos-ui/pull/3914)

## Summary

`POST /csp-report`, accepting reports in both browser body shapes. It lands **before** the policy so 5-F3 changes a flag rather than adding an endpoint and a policy at once, and it is inert until something reports to it.

### Controls

To safely expose this unauthenticated public endpoint to pod logs, six complementary controls are enforced:

1. **8 KB Body Limit** – Rejects large payloads via Fastify before parsing (`413`).
2. **Strict Content-Type Allowlist** – Rejects non-CSP media types before handlers run (`415`).
3. **Max Reports per Request** – Caps entries per payload.
4. **Sanitized Field Allowlist** – Truncates fields and strips control characters.
5. **Per-Request Log Budget** – Charged once per *request* rather than per *report* (preventing a single 8 KB payload from exhausting the logging window for genuine reports).
6. **30 req/min Rate Limit** – Applied per client IP via `@fastify/rate-limit` (`global: false` to avoid blocking SPA asset loading).

**Additional Hardening Details:**
* **Custom Route Error Handler:** All error responses (`413`, `415`, `429`) return status codes with empty bodies to prevent leaking internal `FST_ERR_*` codes.
* **Custom Request Logging (`logLevel: 'silent'`):** Standard Fastify request logging is disabled so unauthenticated rejections don't bypass log budgets. Valid report logs are emitted via `app.log` with the request ID manually attached.

### DIFF
--- a/server/src/security/cspReport.ts
+++ b/server/src/security/cspReport.ts
@@ -12,8 +12,8 @@ import type { RateLimitOptions } from '@fastify/rate-limit'
  *  2. Exactly two accepted media types; everything else gets 415 without
  *     reaching a handler.
  *  3. A cap on how many reports one request may contribute. Neither parser
- *     checks the body's shape, so an 8 KB array of minimal entries — around
- *     170 of them fit — is one request that would otherwise write 170 lines.
+ *     checks the body's shape, so an 8 KB array of minimal entries — 174 of
+ *     them fit — is one request that would otherwise write 174 lines.
  *  4. A field allowlist — only the report fields the spec defines survive,
  *     each truncated. Unknown keys never reach the log.
  *  5. A fixed-window budget charged once per *request*, not once per report,
@@ -35,8 +35,11 @@ import type { RateLimitOptions } from '@fastify/rate-limit'
  * MAX_LOGGED_PER_WINDOW × MAX_REPORTS_PER_REQUEST must multiply by the replica
  * count; the constants below are one pod's ceiling, not the deployment's.
  *
- * Story 5-G adds real per-route rate limiting at the edge and through
- * `@fastify/rate-limit`; the counter here is the log-volume floor beneath it.
+ * The per-client limit is enforced by `@fastify/rate-limit`, which index.ts
+ * registers once with `global: false`; story 5-G attaches the auth-route
+ * limits to that same registration. Its store is per process too, which is why
+ * the epic puts flood protection at the ingress and treats both counters here
+ * as the backstop beneath it.
  */
 
 /** Where the browser posts violation reports. Story 5-F3's policy points here. */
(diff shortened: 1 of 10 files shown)


server/src/security/cspReport.ts
```diff
@@ -0,0 +1,305 @@
...
+ * pod, so a per-client limit above it would let one caller spend the whole
+ * thing. Thirty a minute still lets a genuinely broken page report freely —
+ * a violation that repeats past that adds nothing the first thirty did not
+ * already say.
+ *
+ * `request.ip` reads X-Forwarded-For because the app sets `trustProxy`, so
+ * this counts real clients rather than the sidecar. The store is per process:
+ * see the note in index.ts on why the edge is where flood protection belongs.
+ */
+export const MAX_REQUESTS_PER_WINDOW = 30
```
otchet-broad, line 86: No change needed now, but I'm thinking about how we should bring this into the UI code as well so that the frontend might benefit from this.  Any ideas?  

I'm also concerned we might need more than 30 requests per minute for our initial launch. Some of our pages are starting to make multiple calls to consent for rendering results and fast navigation in under a minute might lead good users to 429 responses.  Have you tested locally with this across things like the data library (with searching and filtering) to see if we start getting into 429 land sooner than expected?

rushtong: I have done local testing and have not hit that limit yet, but I agree, this does look pretty tight. I'll take another round of heavy filtering, navigation, etc and maybe raise that limit.

rushtong: And just to clarify, the 30-per-minute limit applies to only one route: `POST /csp-report` so it is not global. We are adding global rate limiting in my current batch of PRs, but in this case, the effect is limited to this path.

otchet-broad (approved):
Looks good and has me thinking lots of thoughts about expected UI load.
Thank you!

fboulnois (approved):
👍

PR #3910, opened 2026-09-04
[DT-4022] Bind the app error handler to the routes buildApp registers (BFF Phase 5, story 5-G1)

rushtong:
### Addresses

[DT-4022](https://broadworkbench.atlassian.net/browse/DT-4022) — BFF Phase 5, story 5-G1. 
Security risk: **medium** (this is an information-disclosure fix, and it changes the error body of every route `buildApp` registers).

**PR 1 of 3.** Story 5-G was split to reduce PR complexity:
1. **this PR** — bind the app error handler to the routes `buildApp` registers
2. `gr-dt-4022-login-rate-limit` — rate-limit `POST /auth/login` ([5-G2](https://github.com/DataBiosphere/duos-ui/pull/3911))
3. `gr-dt-4022-callback-rate-limit` — rate-limit `GET /auth/callback` ([5-G3](https://github.com/DataBiosphere/duos-ui/pull/3912))

### Summary

**Fastify binds a route's error handler when the route is registered, not at `ready()`.** `fastify.setErrorHandler(...)` sat at the *end* of `buildApp()`, so it reached no route built inside that function. Every one of these fell back to Fastify's default handler, which serialises `err.message` to the client:

`/auth/login`, `/auth/callback`, `/auth/csrf-token`, `/auth/logout`, `/auth/me`, `/health`

Confirmed empirically against the installed Fastify 5.12.1, not inferred from the docs. The mechanism is `lib/route.js`, which assigns `context.errorHandler = this[kErrorHandler]` inside an `after(...)` callback — and every `await fastify.register(...)` in `buildApp()` drains that queue, freezing each route's handler at whatever was set when the *next* awaited register ran.

Three leaks that reached a browser with a 500:

| Source | What leaked |
|---|---|
| `auth/callback.ts` re-throws non-`AuthorizationResponseError` failures from `oidc.authorizationCodeGrant` | token-endpoint response detail, JWT validation specifics |
| `auth/oidcClient.ts` `requireEnv` | the name of the missing deployment variable |
| `auth/logout.ts` | the discovered `end_session_endpoint` |

The fix moves the call ahead of every route registration. The encapsulated proxy declares its own error handler (ADR-010) and is unaffected either way.

----

- Verify all tests go green

[DT-4022]: https://broadworkbench.atlassian.net/browse/DT-4022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/test/index.test.ts
+++ b/server/test/index.test.ts
@@ -159,69 +159,72 @@ describe('error handler', () => {
 })
 
 describe('handleServerError', () => {
+  // The whole contract of the handler. The status varies with the error; the
+  // body never does. Nothing a client reads is derived from err.message.
+  const GENERIC_BODY = { error: 'An unexpected error occurred.' }
+
   function fakeRequest() {
     return { ip: '203.0.113.1', url: '/auth/login', log: { warn: vi.fn(), error: vi.fn() } }
   }
 
   function fakeReply() {
-    const reply = { status: vi.fn(() => reply), send: vi.fn(() => reply) }
+    const reply = {
+      sentStatus: 0,
+      sentBody: undefined as unknown,
+      status: vi.fn((code: number) => {
+        reply.sentStatus = code
+        return reply
+      }),
+      send: vi.fn((body: unknown) => {
+        reply.sentBody = body
+        return reply
+      }),
+    }
     return reply
   }
 
-  it('logs the failure at error and hides its message', async () => {
+  async function run(err: Error) {
     const { handleServerError } = await import('../src/index.js')
-    const err = Object.assign(new Error('internal secret data'), { statusCode: 502 })
     const request = fakeRequest()
     const reply = fakeReply()
-
     // oxlint-disable-next-line @typescript-eslint/no-explicit-any
     handleServerError(err as any, request as any, reply as any)
+    return { request, reply }
+  }
 
-    expect(request.log.error).toHaveBeenCalled()
-    expect(reply.status).toHaveBeenCalledWith(502)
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
👍

server/test/index.test.ts
```diff
@@ -158,6 +158,54 @@ describe('error handler', () => {
...
+
+  it('honors err.status when the error carries no statusCode', async () => {
+    const { handleServerError } = await import('../src/index.js')
+    const err = Object.assign(new Error('bad input'), { status: 400 })
+    const reply = fakeReply()
+
+    // oxlint-disable-next-line @typescript-eslint/no-explicit-any
+    handleServerError(err as any, fakeRequest() as any, reply as any)
+
+    expect(reply.status).toHaveBeenCalledWith(400)
```
otchet-broad: What is the expectation for the 'bad input' text?  Can you modify this to make it clear on what's expected to be returned from handleServerError for that?

rushtong: Will make this more explicit, thank you 👍🏽

otchet-broad: If I understand correctly, in all cases it should be 'An unexpected error occurred.'

rushtong: Tests are updated, thank you 👍🏽

otchet-broad (approved):
Thanks for the clarification!

PR #3909, opened 2026-09-04
[DT-4021] Derive the Content Security Policy from runtime config (BFF Phase 5, story 5-F3)

rushtong:
### Addresses

[DT-4021](https://broadworkbench.atlassian.net/browse/DT-4021) — BFF Phase 5, story 5-F3  
Security risk: **low** — CSP is report-only by default, with enforcement controlled per environment.

Part of a chain of PRs addressing this ticket:
- [5-F1 security headers](https://github.com/DataBiosphere/duos-ui/pull/3907) → 
- [5-F2 report sink](https://github.com/DataBiosphere/duos-ui/pull/3908) → 
- **5-F3 the policy (this PR)** → 
- [5-F4 terra-helmfile proxy](https://github.com/broadinstitute/terra-helmfile/pull/6497)  → 
- 5-F5 run report - no code, just a report execution → 
- [5-F6 public BFF endpoints](https://github.com/DataBiosphere/duos-ui/pull/3914)

### Summary

Adds a Content Security Policy in report-only mode, controlled by `DUOS_CSP_REPORT_ONLY` (default: `true`).

- Builds `connect-src` from active URLs in `config.json` rather than hardcoding environment-specific origins.
- Uses a mode-specific allowlist:

  | Mode | `connect-src` |
  |---|---|
  | BFF | `'self'`, `apiUrl`, `bardApiUrl`, banner bucket |
  | Legacy | BFF sources plus `ecmApiUrl` and `tdrApiUrl` |

- Honors the `DUOS_API_URL` override.
- Restricts the GCS banner source to `https://storage.googleapis.com/broad-duos-banners/` rather than the shared GCS origin.
- Keeps `script-src` limited to `'self'`; no active inline or third-party scripts were found.
- Does not add `blob:` to `img-src`; no active blob-backed images were found.
- Preserves the stricter `content-security-policy: sandbox` header on proxied downloads.

Blank or invalid config values are ignored. ECM and TDR are intentionally omitted in BFF mode because those requests become same-origin through the BFF. `terraUrl` is excluded because it is used for navigation, not fetching.

### Known limitations

- The deployed httpd sidecar currently replaces enforcing CSP headers, while allowing report-only headers through. The `terra-helmfile` update is tracked in story 5-F4 and must land before enforcement.
- Browser-level CSP coverage is deferred to story 6-K, when e2e tests are served through Fastify. The current Vite preview setup does not expose these headers in CI.
- The deployed report-only validation and enforcement rollout are tracked in story 5-F5.

[...]

PR #3913, opened 2026-09-04
[DT-4021] Move feature flags and anonymous metrics onto public BFF endpoints (BFF Phase 5, story 5-F6)

rushtong:
### Addresses

[DT-4021](https://broadworkbench.atlassian.net/browse/DT-4021) — BFF Phase 5, **story 5-F6**. Security risk: **low**.

**Fourth of the 5-F stack, on [#3909](https://github.com/DataBiosphere/duos-ui/pull/3909).** Review [#3907](https://github.com/DataBiosphere/duos-ui/pull/3907) → [#3908](https://github.com/DataBiosphere/duos-ui/pull/3908) → #3909 first; this diff is against #3909.

### Summary

Two browser connections still bypassed the BFF under `bffEnabled` and forced exact-origin entries into `connect-src`. They now go through dedicated public endpoints:

| Endpoint | Upstream | Drops from `connect-src` |
|---|---|---|
| `GET /public/features`, `/public/features/:key` | Consent | `apiUrl` |
| `POST /public/metrics/event` | Bard | `bardApiUrl` |

They could not be folded into the existing proxies as unauthenticated paths: `unauthenticatedPaths` is an **exact** set match, so `/feature` alone leaves `/feature/:key` returning 401, and anonymous and identified metrics share `POST /api/event` — marking it unauthenticated would silently turn *identified* metrics anonymous.

**These do not reuse `registerUpstreamProxy`, and that is the main design decision to review.** It throws unless `@fastify/csrf-protection` is registered first, which only happens inside the `bffEnabled` block — and these must register **outside both switches**, since they serve pre-login and legacy traffic. More importantly its entire purpose is attaching the session's access token. A separate module makes omitting the token **structural rather than configured**: a future edit to the shared proxy cannot start injecting one here. The module contains no reference to `request.session` at all.

The trust-boundary behaviour is duplicated rather than shared, deliberately — `cookie`, `authorization` and `x-csrf-token` stripped on the way out, `set-cookie` and `www-authenticate` stripped on the way back, `x-app-id` injected. Only the pure URL plumbing (`upstreamBase`, `upstreamPath`) is imported.

**Two departures from the session proxies, both commented:**

- **Bodies are parsed, not streamed.** The shared proxy installs a wildcard pass-through parser because DUOS uploads multipart and binary. Here only `application/json` is registered, with a `bodyLimit` — an unread stream is never measured against `bodyLimit`, so buffering is what makes the 413 exist at all, and it means only well-formed JSON reaches Bard.
- **Feature-flag responses are validated.** A response carrying a body that is not JSON is logged and answered 502 rather than streamed from a public endpoint onto our origin. Bodiless responses (a 204, or a 404 with no content) pass through — requiring JSON unconditionally would turn "flag not found" into a fabricated outage.

No CSRF and no Fetch Metadata guard, both deliberate and both commented. These carry no credentials, so there is nothing to forge with, and the Fetch Metadata guard exists to protect state-changing GETs that carry the session cookie.

### `/public/notifications` is deliberately not built

A separate backlog item repoints `notificationService.ts` at environment-specific buckets. Proxying GCS is the one endpoint of the three with no existing pattern to copy, and building it against a URL shape that is about to change would mean writing it twice. **BFF-mode `connect-src` therefore ends this story at `'self'` plus the banner bucket**, not `'self'` alone. Revisit when that lands.

### The tests are the only proof this works

`src/libs/ajax/FeatureFlag.ts` has **no caller anywhere in `src/`** — only its own unit test. Keeping the module was a deliberate decision (recorded in ADR-013), so `/public/features/*` is built for a consumer that does not exist yet. No amount of exercising the app will show whether it works. That is called out in the module, in its spec, and in the ADR.

### Test plan

- `pnpm --filter duos-server test` — **514 pass**, 33 new. The central case is *"sends no Authorization header upstream even when the request carries a session holding an access token"*, run against a real seeded, refreshable session. Also: path routing and query passthrough, `cookie`/`x-csrf-token` stripping, `x-app-id`, `x-forwarded-for` chaining, `set-cookie` suppression, bare 413/415/400/429 responses, the non-JSON 502, that no CSRF token is required, a cross-site shape the Fetch Metadata guard would have rejected, and the partial-configuration registration matrix (a missing upstream var disables one endpoint and warns; a malformed one fails startup naming itself, so a legacy deployment setting neither still boots).
- `pnpm run test` (client) — **4,515 pass** across 379 files. `FeatureFlag.spec.ts` gains a URL-selection block for both modes; `Metrics.spec.ts` asserts the anonymous BFF path carries no `Authorization` and that identify/syncProfile never touch `/public/metrics`.
- `pnpm run type-check`, `pnpm --filter duos-server typecheck`, `pnpm run lint` — clean, zero errors.

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.\n\n- Label PR with a Jira ticket number and include a link to the ticket\n- Label PR with a security risk modifier [no, low, medium, high]\n- PR describes scope of changes\n- Get a minimum of one thumbs worth of review, preferably two if enough team members are available\n- Get PO sign-off for all non-trivial UI or workflow changes\n- Verify all tests go green\n- Test this change deployed correctly and works on dev environment after deployment\n

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -62,6 +62,11 @@ DUOS_OAUTH_REDIRECT_URI=http://local.dsde-dev.broadinstitute.org:3000/auth/callb
 DUOS_POST_LOGOUT_REDIRECT_URI=http://local.dsde-dev.broadinstitute.org:3000/post-logout
 
 # --- Downstream API --------------------------------------------------------
+# Both DUOS_API_URL and DUOS_BARD_URL are read twice: by the session-carrying
+# proxies inside the bffEnabled block, and by the public endpoints (story
+# 5-F6), which register outside both cutover switches. Unset, the matching
+# public endpoint is simply not registered and the server logs a warning; set
+# but malformed, startup fails naming the variable.
 DUOS_API_URL=https://consent.dsde-dev.broadinstitute.org
 # ECM (externalcreds), proxied at /ecm-api for RAS/eRA Commons linking.
 # Optional: when unset, the BFF boots without the ECM proxy and logs a warning.
@@ -69,7 +74,8 @@ DUOS_ECM_URL=https://externalcreds.dsde-dev.broadinstitute.org
 # TDR, proxied at /tdr-api for snapshot enumeration on dataset pages.
 # Optional, same behavior as DUOS_ECM_URL when unset.
 DUOS_TDR_URL=https://jade.datarepo-dev.broadinstitute.org
-# Bard (Mixpanel gateway), proxied at /bard-api for identified usage metrics.
+# Bard (Mixpanel gateway), proxied at /bard-api for identified usage metrics
+# and at /public/metrics/event for anonymous ones.
 # Optional, same behavior as DUOS_ECM_URL when unset.
 DUOS_BARD_URL=https://terra-bard-dev.appspot.com
 
--- a/docs/plans/bff_adrs/ADR-013-content-security-policy.md
+++ b/docs/plans/bff_adrs/ADR-013-content-security-policy.md
@@ -61,7 +61,8 @@ Under `bffEnabled`, sign-in has no popup and no opener, so it keeps
 
 COEP demands a CORP or CORS header on every cross-origin subresource. The banner
 bucket and the two direct upstreams send neither, so enabling it would block
-them. It stays off until those become same-origin (5-F6).
+them. Two of the three became same-origin in 5-F6; COEP stays off while the
+banner bucket is still fetched directly.
 
 ### HSTS is production-only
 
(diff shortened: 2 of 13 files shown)


rushtong:
Superseded by #3914. The head branch was renamed from `gr-dt-4022-public-bff-endpoints` to `gr-dt-4021-public-bff-endpoints` (this is DT-4021, story 5-F6); the rename closed this PR, so it was reopened as #3914 with the same base and body.

PR #3914, opened 2026-09-04
[DT-4021] Move feature flags and anonymous metrics onto public BFF endpoints (BFF Phase 5, story 5-F6)

rushtong:
## Addresses

[DT-4021](https://broadworkbench.atlassian.net/browse/DT-4021) — BFF Phase 5, **story 5-F6**.  
Security risk: **low**.

Part of a chain of PRs addressing this ticket:
- [5-F1 security headers](https://github.com/DataBiosphere/duos-ui/pull/3907) (merged) → 
- [5-F2 report sink](https://github.com/DataBiosphere/duos-ui/pull/3908) (merged) → 
- [5-F3 the policy](https://github.com/DataBiosphere/duos-ui/pull/3909) (merged) → 
- [5-F4 terra-helmfile proxy](https://github.com/broadinstitute/terra-helmfile/pull/6497)  → 
- 5-F5 CSP enforcement reports → 
  - Config PRs for enabling app-level CSP enforcement
    - https://github.com/broadinstitute/terra-helmfile/pull/6498 (merged)
    - https://github.com/broadinstitute/terra-helmfile/pull/6499 (merged)
    - https://github.com/broadinstitute/terra-helmfile/pull/6500 (post-audit)
- **5-F6 public BFF endpoints (this PR)**

## Summary

This PR moves **two of the three remaining direct browser paths** onto public BFF endpoints when `bffEnabled` is enabled:

| Flow | Public BFF endpoint | Upstream | Removed from `connect-src` |
|---|---|---|---|
| Feature flags | `GET /public/features`, `/public/features/:key` | Consent | `apiUrl` |
| Anonymous metrics | `POST /public/metrics/event` | Bard | `bardApiUrl` |

Identified metrics continue through the authenticated Bard proxy. Legacy-mode clients retain their existing direct upstream URLs.

**Added a line to `.gitignore`** so that AI CI runs don't accidentally commit an unnecessary `server/pnpm-lock.yaml` file.

### Known and accepted gap: notifications

Notification banners remain a direct browser fetch to `https://storage.googleapis.com/duos-banners-<env>/`, the bucket named by `bannersUrl`. BFF-mode `connect-src` is therefore `'self'` plus that one path-scoped bucket entry, not `'self'` alone.

DT-4063 (#3942) has since moved the feed to per-environment buckets, so the URL shape is settled. It did not put the fetch behind the BFF, and no ticket covers that yet. Proxying a public GCS object is the one of the three paths with no existing pattern to copy, so `/public/notifications` stays out of this PR.

### Design

The public endpoints use a separate proxy module rather than `registerUpstreamProxy`. They do not read session credentials, forward cookies or authorization headers, or inject a session access token.

The routes register independently of `DUOS_DB_HOST` and `bffEnabled`, allowing them to operate without session infrastructure. Client routing switches to these endpoints only when `bffEnabled` is enabled.

The existing proxies cannot accommodate these flows through their current unauthenticated-path configuration:

- Exact path matching would require separate handling for `/feature` and `/feature/:key`.
- Anonymous and identified metrics share the upstream `POST /api/event` route. Making that route unauthenticated would also remove authentication from identified events.

### Safeguards

- Strip outbound `cookie`, `authorization`, and `x-csrf-token` headers.
- Strip upstream `set-cookie`, `clear-site-data`, and `www-authenticate` response headers.
- Inject `x-app-id` and apply response hardening.
- Apply per-route rate limits and an 8 KiB limit to metrics request bodies.
- Parse metrics bodies as JSON, rejecting malformed JSON and unsupported content types.
- Reject feature-flag responses with non-JSON content types when they contain a body; preserve bodiless responses.

CSRF and Fetch Metadata checks are intentionally omitted because these endpoints do not use the caller’s session authority.

### Validation

Added tests cover client URL selection, credential stripping, response handling, request limits, upstream failures, and registration with missing or invalid upstream configuration.

`FeatureFlag.ts` currently has no application callers. It is retained for future use, as documented in ADR-013, so ordinary UI testing does not exercise the feature-flag flow. Automated client and proxy tests provide coverage for that path.

[DT-4063]: https://broadworkbench.atlassian.net/browse/DT-4063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fboulnois (approved):
A few issues I found with Claude and Codex:

## High priority: telemetry failures

1. **An empty `bardApiUrl` suppresses all error reporting.**
  `src/libs/ajax/fetchAdapter.ts:59` The guard uses `url.startsWith(bardApiUrl)`, which matches every URL when `bardApiUrl` is empty. The base config already uses an empty value, and this PR makes leaving it empty more likely in BFF mode. The guard predates this PR. **Fix:** only match a nonempty value, or share one metrics-URL predicate with `Metrics.ts`.

2. **Missing `DUOS_BARD_URL` makes anonymous metrics appear successful while being lost.**
  `server/src/index.ts:274`; `server/src/proxy/publicProxy.ts:242` BFF startup only warns when this variable is missing. The metrics route is then absent, so requests fall through to the SPA handler and receive `200 text/html`. The client treats that as success. **Fix:** require the variable in BFF mode or register a route that returns 503.

3. **A shared timeout disables metrics after 30 seconds. (pre-existing)**
  `src/libs/ajax/Metrics.ts:9` The module creates one `AbortSignal.timeout(30000)` at load time. Once it aborts, later calls using it fail immediately and their errors are swallowed, including error-reporting calls. Introduced by DT-979 (#2722). **Fix:** create a fresh timeout signal for each call.

## Medium priority

4. **Metrics URL detection can throw and suppress unrelated errors.**  
   `src/libs/ajax/fetchAdapter.ts:53` `new URL()` throws on invalid input, and callers do not await `reportError`, creating an unhandled-rejection risk. Matching only the pathname also suppresses failures from foreign origins with the same path. **Fix:** handle invalid URLs and require the same origin.

5. **The custom JSON parser changes valid request bodies.**  
   `server/src/proxy/publicProxy.ts:217` It duplicates Fastify's parsing and size checks, while omitting the built-in secure JSON parser's protections. Passing parsed values to reply-from silently forwards `null`, `false`, `0`, and `""` as empty bodies. Objects are unnecessarily parsed and serialized again. **Fix:** validate JSON while retaining the raw string, then forward it explicitly through `opts.body` and `contentType`, with content length recomputed.

6. **The CSRF exemption is too easy to miss for future public routes.**  
   `src/libs/ajax/fetchAdapter.ts:79` Only `POST /public/metrics/event` is exempted. A future public write route could request an auth-gated CSRF token and send anonymous visitors through logout after a 401. **Fix:** exempt same-origin `/public/` paths in `needsCsrfToken`.

7. **The public error handler hides useful logs.**  
   `server/src/proxy/publicProxy.ts:214` The handler returns minimal responses but bypasses server error logging, hiding rate-limit events and genuine server failures. **Fix:** keep the minimal responses and add appropriate logging.

8. **Rate limiting can silently disappear after a registration refactor.**  
   `server/src/proxy/publicProxy.ts:212` Route limits do nothing unless the rate-limit plugin is registered in an ancestor scope. Current tests register it themselves, so they would miss a startup regression. **Fix:** fail startup if `app.hasDecorator('rateLimit')` is false, following the existing CSRF guard pattern.

rushtong:
@fboulnois - Thanks for the findings, all should now be addressed except for bullet **3: A shared timeout disables metrics after 30 seconds. (pre-existing)**. I'm filing that as a follow up ticket in the epic in [DT-4086](https://broadworkbench.atlassian.net/browse/DT-4086)

[DT-4086]: https://broadworkbench.atlassian.net/browse/DT-4086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

server/src/proxy/publicProxy.ts
```diff
@@ -0,0 +1,289 @@
...
+export const PUBLIC_METRICS_EVENT_PATH = '/public/metrics/event'
+
+/** Bound anonymous JSON requests while allowing room for event properties. */
+export const METRICS_BODY_LIMIT = 8 * 1024
+
+const RATE_LIMIT_WINDOW_MS = 60_000
+
+export const FEATURES_MAX_PER_WINDOW = 60
+
+export const METRICS_MAX_PER_WINDOW = 120
```
otchet-broad, line 37: Does Bard have any kind of rate limiting?  Could this number possibly be greater than that?  Would it be better to just delegate rate limiting to that service instead of adding it here?

server/src/proxy/publicProxy.ts
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Bare bodies, like the session proxies' error shape under ADR-010: a parser
+ * or rate-limit error message can carry framework detail, and no client reads
+ * it. Logging still happens here, because this handler replaces the app-level
+ * handleServerError for the whole /public scope.
+ */
+function handlePublicError(error: FastifyError, request: FastifyRequest, reply: FastifyReply): FastifyReply {
+  const statusCode = error.statusCode ?? 500
+  if (statusCode === 429) {
+    request.log.warn({ ip: request.ip, url: request.url }, '[public-proxy] rate limit exceeded')
```
otchet-broad, line 219: We should probably check to see if a `Retry-After` header is set and if not, set it appropriately.

otchet-broad (commented):
Setting the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) header when we're returning a 429 is probably the biggest thing we should do at this point.  It will allow callers to know when they can resume calling the service.

I'm also wondering if we should really be doing rate limiting on Bard ourselves.  It seems like we should make sure that we're honoring the rate limiting that Bard does itself.  

The other concern with Bard:  will they honor the x-forwarded-for header?  If not, it may look like every request is coming from the proxy instead of the remote client.  Probably worth confirming.

server/src/proxy/publicProxy.ts
rushtong: `@fastify/rate-limit` adds `Retry-After` (and the `X-RateLimit-*` headers) on every 429. [Adding a test](https://github.com/DataBiosphere/duos-ui/pull/3914/commits/5cd84381f9fc3ddfdf556229ae1b624779aa1c2c) for continued enforcement.

server/src/proxy/publicProxy.ts
rushtong: Checked the Bard source (https://github.com/DataBiosphere/bard): it has no rate limiting. The only throttle is a 10s delay on the `request:failed` event. So there is nothing on the Bard side. The 120/min per-client limit here is the only guard on an anonymous call to DUOS' path. The legacy path had none because browsers called Bard directly.

rushtong:
> The other concern with Bard: will they honor the x-forwarded-for header? If not, it may look like every request is coming from the proxy instead of the remote client. Probably worth confirming.

Bard doesn't read `X-Forwarded-For`. It forbids an `ip` property in the body and calls Mixpanel from its own server, so Mixpanel only sees Bard's address before our BFF code took over the call. The header we forward is correct but has no consumer today, so this change is neutral.

kevinmarete (approved):

otchet-broad (approved):
👍

PR #3916, opened 2026-09-08
[DT-4002] Improve the SO Data Access Approvals help text and fix header text overflow

kevinmarete:
### Addresses

Completes [DT-4002](https://broadworkbench.atlassian.net/browse/DT-4002). Security risk: **no**. [DT-3321](https://broadworkbench.atlassian.net/browse/DT-3321) asked for the same change and is now closed as a duplicate of DT-4002.

### Summary

The DAR Approvals description in the Signing Official console read "Your Institution's Data Access Approvals: Records from all current data access approvals", which communicated neither of the two things an SO can actually do on the page. It now reads:

> Review all approved Data Access Requests (DARs) submitted by researchers at your institution. When required, your approval must happen before a DAR can proceed to Data Access Committee (DAC) review — you can approve those requests directly from this page.

That longer text exposed a pre-existing layout bug in the shared `TableHeaderSection`: the title and description containers carried hardcoded `width: 150%` / `width: 120%` inline overrides. With short text this went unnoticed, but at realistic length the box inflated well past the visible page width, so the text only wrapped off-screen and appeared to run off the page. Replaced the fixed percentages with `flex: 1` / `minWidth: 0` on the header container so text wraps within the real page width. `minWidth: 0` is the load-bearing half — the container sits inside a `flex-direction: row` parent that would otherwise refuse to shrink it below min-content.

The title ("My Institution's Data Access Approvals") is unchanged.

After

### Testing

`TableHeaderSection` is used by 21 files, so this ran against the full suite rather than the touched specs: 3895 tests / 351 files pass, with `pnpm run lint` and `pnpm run type-check` both clean. Neither change had test coverage before, so this adds an assertion for the new description text and a regression test that fails if the fixed widths come back.

----

- Verify all tests go green

[DT-4002]: https://broadworkbench.atlassian.net/browse/DT-4002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3321]: https://broadworkbench.atlassian.net/browse/DT-3321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rushtong (approved):
👍🏽

otchet-broad (approved):

PR #3920, opened 2026-09-09
[DT-4003] Clean up and simplify the DAR vote page UI

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-4003

Security risk: **no** — presentation-only changes to an existing review page; no new data access, routes or controls.

### Summary

Cleans up the DAR review/vote page: the read-only warning moves to the top of the page as a MUI `Alert` matching `ManualReviewWarningBanner`, the Voting History tab loses its redundant "Votes" title, and the Full DAR tab loses the DAR ID/project title heading, the divider above the summary and the embedded voting history table (now only on the page's own Voting History tab) along with the orphaned `VotingHistoryOverview` component.

The Full DAR tab's step navigation moves from a vertical side panel to a horizontal tab bar above full-width content, so it lines up with the page's other tabs. It reuses the review page's `reviewTabsSx` via a new `stepTabsSx` rather than restating those styles as CSS overrides on MUI internals.

`DataAccessRequestApplication` renders in two places under `existingDarsReadOnlyMode` — inside this tab, and standalone at `/dar_application_review/:collectionId` and `/progress_report_application/:collectionId` (the "Review", "Review Closeout" and progress-report actions in the DAR collection table). Only the embedded case has a header above it, so the heading and layout changes are keyed on a new `embedded` prop set by `DarCollectionReview`; the standalone pages keep their own heading and side panel.

After

### Testing

Added coverage for the read-only warning and its position relative to the review header, `ScrollableTabs` orientation, and the embedded-vs-standalone heading and step-tab layout. `pnpm run lint` and `pnpm run type-check` are clean, and the full Vitest suite passes (377 files, 4505 tests).

----

- Verify all tests go green


src/pages/dar_application/DataAccessRequestApplication.css
```diff
@@ -199,6 +207,13 @@ fieldset[disabled] .regular-checkbox:before {
   background-color: rgba(43, 189, 155, .3);
 }
 
+/* Layout only; the tab bar itself is styled by stepTabsSx so it matches the review
+   page's other tab bars. */
+.step-tabs-container--horizontal {
+  width: 100%;
+  margin-bottom: 1rem;
+}
```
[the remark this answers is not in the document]
kevinmarete: Leaving this as-is — the divergence from the vertical variant is the intent of the ticket, and the label styling isn't actually lost.

`.step` never renders in this mode: every tab reachable under `existingDarsReadOnlyMode` is built with `showStep: false` (`DataAccessRequestApplication.tsx:417`, `:424`, `:433`), so there is no "Step N" line to style.

`.title` keeps its font size and colour. `stepTabsSx` sets both on `.MuiTab-root` and the label div inherits them, and no other `.title` rule exists in the app to compete. Montserrat comes from the global MUI theme (`muiThemeFix`), and `textTransform: 'none'` is in `stepTabsSx` because a bare `Tabs` doesn't inherit it from `TabControl`'s `baseTabsSx` the way the review page's other tab bars do.

Matching the side panel's bold 17px styling would undo the change DT-4003 asks for: the Full DAR tab bar is meant to line up with the review page's other tabs, which is why it now derives from `reviewTabsSx` rather than restating overrides on MUI internals.

fboulnois (approved):
A few suggestions developed with Codex:

### [P2] Preserve voting history on standalone researcher pages

`src/pages/dar_application/DataAccessRequestApplication.tsx:403` removes voting history for every read-only use, including `/dar_application_review/:collectionId`. Researchers lose the dataset decisions, dates, and rationales previously available there; the collection review’s replacement Voting History tab is restricted to DAC users/admins. Suppress the duplicate history only when embedded in the voting page.

### [P2] Keep the heading outside the embedded Full DAR view

`src/pages/dar_application/DataAccessRequestApplication.tsx:663` hides the DAR/project heading whenever `existingDarsReadOnlyMode` is true. Both standalone review and progress-report creation use that flag and have no surrounding `ReviewHeader`. Use an explicit embedded-view flag to preserve their heading.

## Material UI assessment

The new `Alert`, orientation prop, and indicator `slotProps` are appropriate uses of `MUI Alert` and `Tabs`. Two improvements are advisable:

- Narrow `.MuiButtonBase-root` styling in `src/pages/dar_application/DataAccessRequestApplication.css:226` to `.MuiTab-root`; it currently also styles scroll buttons.
- The existing `ScrollableTabs` uses tabs as scroll navigation without tab-panel associations or a tablist label. This predates the branch, but remains an accessibility gap in the redesigned horizontal navigation.

kevinmarete:
Thanks @fboulnois — all four addressed.

**Voting history on standalone researcher pages** — real regression, fixed. `VotingHistoryOverview` is restored and the tab is now dropped only when `embedded`, so `/dar_application_review/:collectionId` and `/progress_report_application/:collectionId` keep it and only the Full DAR tab defers to the review page's own Voting History tab. The vote shaping moved to `src/pages/dar_application/votingHistoryData.ts` so the component keeps the size reduction, and `buildVoteRecords` is skipped entirely when embedded. Its spec came back with it, and the embedded/standalone test now asserts the tab either way.

**Heading outside the embedded view** — already the case on this branch: the heading is gated on `!embedded`, not `existingDarsReadOnlyMode`, so standalone review and progress-report creation keep theirs. `DataAccessRequestApplication.spec.tsx` covers both.

**`.MuiButtonBase-root`** — narrowed to `.MuiTab-root` at all three sites, so the scroll buttons are no longer caught, and the two duplicate blocks are merged.

**Tablist accessibility** — added an `aria-label` to the `Tabs`. The missing tab-panel associations need the tablist role dropped for a nav/list across all three consumers, which is more than this branch should carry; happy to file it as a follow-up.

src/pages/dar_application/votingHistoryData.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const formatDate = (dateString: string | number): string => {
+    return new Date(dateString).toLocaleDateString('en-US', {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+    })
+  }
+
+  const isElectionClosed = !hasFinalVote && (election?.status === ElectionStatus.CLOSED || election?.status === 'Canceled')
```
otchet-broad: Is there a better way to deal with these string constants? 'Canceled', 'Approved', 'Denied' or are these the only uses of them throughout the code?  I worry that we don't have consistency in terminology across the app.

otchet-broad (approved):
Nice work.  See my one small comment about constants.  Not a deal breaker, but it would be nice to get these consolidated so that we can use them in a consistent manner across code.

src/pages/dar_application/votingHistoryData.ts
kevinmarete: Good catch — they weren't the only uses, and the duplicate was the one that mattered.

`ElectionStatus` in `src/utils/DarUtils.ts` now carries `CANCELED`, and a sibling `VoteDecision` holds `APPROVED`/`DENIED`/`PENDING`. This file uses both, and so does `processElectionStatus` (`src/libs/utils.ts:314`), which was independently spelling out `'Approved'`/`'Denied'` for every election-status table in the app — that was the real terminology risk, since the two could have drifted apart.

I left the dashboard tile labels (`ResearcherDashboard`, `SigningOfficialDashboard`, `DACDashboard`) as literals: those are display strings sitting alongside `'Total'`, `'In Process'` and `'Active'`, so importing constants for two of four entries would read worse than it does now. Happy to pull those in too if you'd rather have the words in exactly one place.

83aba008

PR #3921, opened 2026-09-10
[DT-3314] Add Registration Date and Pre-Auth columns to the Manage Users table

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3314

Security risk: **low** — read-only columns on an admin-only table.

### Summary

Adds sortable Registration Date and Pre-Auth columns to the Manage Users table.

Registration Date is read in UTC rather than through `formatDate`, whose local-time getters render a UTC-midnight date as the previous day west of UTC. Pre-Auth lists the DAAs a user's library card is pre-authorized against; a failed DAA fetch falls back to `DAA-<id>` rather than blocking the page.

`daaLabel` moves to the shared `src/libs/daaHelpers` now that two features need it.

Out of scope: "Last Login" — nothing in the stack stores it.

### Screenshot

shot-manage-users

----

- Verify all tests go green


fboulnois (approved):
👍

src/components/manage_users_table/manageUsersTableUtils.ts
```diff
@@ -43,3 +43,22 @@ export const userDacs = (roles: UserRole[] | undefined, dacNameById: Map<number,
 }
 
 export const formatUserDacs = (dacs: UserDac[]): string => dacs.map(dac => dac.name).join(', ') || 'None'
+
+// Read in UTC: formatDate's local getters report the previous day west of UTC.
+export const formatRegistrationDate = (createDate: Date | string | undefined): string => {
+  if (isNil(createDate)) {
+    return '- -'
+  }
+  const date = new Date(createDate)
```
otchet-broad: FYSA: We have DayJS in the project.

otchet-broad (approved):
Looks good.  One note about date formatting is that we have DayJS in the project that could help with the work we're doing in the formatting space, but not required.

src/components/manage_users_table/manageUsersTableUtils.ts
kevinmarete: Good call — switched to `dayjs.utc(createDate).format('YYYY-MM-DD')`.

Worth flagging for anyone reaching for dayjs here: plain `dayjs(createDate)` parses in local time and reports the previous day west of UTC (`2022-01-01T00:30:00.000Z` renders as 2021-12-31 in ET), so this needs the `utc` plugin to keep the calendar date the API sent. The specs cover that case and fail without it.

`LibraryCardTable` had that local-time bug already and its create date column was uncovered, so I fixed it the same way and added a test — the two columns now agree.

PR #3922, opened 2026-09-10
[DT-981] Let a user dismiss a system notification banner

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-981

Security risk: **low** — a per-user boolean in localStorage hiding an already-public announcement.

### Summary

Ops banners from GCS had no dismiss control, so they reappeared on every visit.

`notificationService` gains `dismissBanner` / `isBannerDismissed`, persisted per user under `dismissedBanner_<id>`, plus `visibleBanner` for the "should this show?" question. `DismissibleBanner` pairs a banner with recording its dismissal; the header, user profile and DAR application just use it. A dismissal survives a reload, leaves other banners showing, and does not suppress a later banner.

`Notification` moves to MUI `Alert` per copilot-instructions, which supplies the close button, severity icon and colours — the hand-rolled div, the icon switch and the CSS module go with it, along with three `!important` declarations. `CloseIconComponent` gains the `aria-label` it was missing at all nine call sites.

The DAR coverage work that was on this branch is #3924, split out at review request.

### Screenshot

shot-banners

----

- Verify all tests go green


src/components/Notification.module.css
fboulnois, line 1: 👍

fboulnois (commented):
Can we split this out into smaller tickets?

1. introducing the new system notification banner interface
2. split out the individual pages / components that are affected

kevinmarete:
Good call — split done.

Most of the size wasn't the banner feature at all: the DAR application coverage and Sonar work had nothing to do with dismissing banners, and is now #3924.

What's left here is the interface and its three call sites:

- `notificationService` — `isBannerDismissed` / `dismissBanner`
- `Notification` — `onDismiss`, plus the move to MUI `Alert`
- `CloseIconComponent` — `aria-label`
- wiring in `DuosHeader`, `UserProfile`, `DataAccessRequestApplication` (12 lines in the last)

Happy to split the three call sites out per your point 2 as well if you'd still like that — I left them together since they're a few lines each and the feature isn't observable without them. #3924 touches the same DAR file, so whichever lands second needs a small rebase.

kevinmarete:
Correction on the split: the DAR coverage PR is #3924, not #3923 — renaming its branch to match the new ticket (DT-4093) closed the original, so it was reopened under a new number. Body updated.

test/components/App.spec.tsx
```diff
@@ -31,6 +31,7 @@ vi.mock('src/libs/notificationService', () => ({
   NotificationService: {
     getActiveBanners: vi.fn().mockResolvedValue([]),
   },
+  isBannerVisible: vi.fn().mockReturnValue(true),
```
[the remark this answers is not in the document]
kevinmarete: Fixed in 91ff4fd1 — `dismissBanner` added to the mock. Audited the other specs that stub this module too; each now provides what its subject imports.

test/components/Notification.spec.tsx
```diff
@@ -19,41 +17,63 @@ const makeBanner = (overrides: Partial<Banner> = {}): Banner => ({
...
 
-  it('merges customStyle into the container', () => {
+  // The feed is ops-authored, so a level outside the four known ones can reach the component.
+  it('falls back to info when the banner carries an unknown level', () => {
     const { container } = render(
-      <Notification
-        notificationData={makeBanner()}
-        customStyle={{ backgroundColor: 'rgb(255, 0, 0)' }}
-      />,
+      <Notification notificationData={{ message: 'odd level', level: 'critical' as Banner['level'] }} />,
```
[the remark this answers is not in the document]
kevinmarete: This one type-checks as written. `npx tsc --noEmit` is clean, and CI's type-check job passes — tsconfig includes `**/*`, so the spec is in scope. I confirmed tsc really is checking this file by planting a deliberate `const deliberate: number = 'not a number'`, which it caught at line 64, then restoring it: zero errors reported here.

kevinmarete:
Both fixed in 1e7023c7.

**Cross-copy dismissal** — confirmed at the source: `getActiveBanners` returns everything active and `getBannerObjectById` filters `{active: true, id}`, so an active `eRACommonsOutage` really does render in the header and inline on the page at once. `dismissBanner` now notifies subscribers and `DismissibleBanner` drops its banner when any copy of that id is dismissed, so the wiring sits in one component rather than each page. Two tests cover it directly, and I checked they fail when the subscription is broken.

**Severity fallback** — right, `SEVERITY` is a plain object so `toString` and friends returned an inherited member past the `?? 'info'`. Guarded with `Object.hasOwn`, with a test that fails against the old lookup.

kevinmarete:
All three fixed in 40c4bf2e.

- **Hover underline** — a real regression from the Alert migration. The deleted CSS module had an `a:hover` rule as well as `a`, and `index.css:65` strips both with `!important`. The `sx` now covers `& a, & a:hover`.
- **Header re-filtering** — keyed the banner effect on the current user id. Dismissals are per user and App reconciles identity while the header stays mounted, so a stale filter could cross users.
- **Service-level pub/sub cover** — added; the component specs swap both functions for the test bus, so the real `forEach` and unsubscribe had none.

Each verified by mutation: removing the fix fails its test.

kevinmarete:
**Scope note (dbb4f413)** — this PR grew past DT-981 and has been trimmed back.

Removed: the dedicated dismissal store, `clearStorage` preservation, and the `isLogged` threading through the service and three components. DT-981 keys dismissals off `Storage.getCurrentUserSettings` and its testing note says *"clear localStorage or switch user, confirm it reappears"* — so preserving dismissals through sign-out was the opposite of what the ticket asks. The auth plumbing that came with it also introduced races, including the DAR page reinitialising its form over user edits on an auth transition.

Kept: the close button, per-user keying, per-banner isolation (the AC), the MUI `Alert` migration and its regressions, and the cross-copy sync that "hides that specific banner immediately" needs.

fboulnois (approved):
👍

src/libs/notificationService.ts
```diff
@@ -13,6 +14,43 @@ export interface Banner {
...
+export const onBannerDismissed = (listener: (id: string) => void): (() => void) => {
+  dismissalListeners.add(listener)
+  return () => {
+    dismissalListeners.delete(listener)
+  }
+}
+
+/** Record that the current user (or anonymous browser) has dismissed this banner. */
+export const dismissBanner = (id: string): void => {
+  Storage.setCurrentUserSettings<boolean>(dismissedBannerKey(id), true)
```
otchet-broad, line 44: By policy this means that whenever banner text needs t be shown, we need to change the key to one that hasn't been used before.  Has the banner posting documentation been updated to reflect this change?

otchet-broad (approved):
This change looks good.  It just needs to be clear to people who update the banner text that the must change the banner IDs if the text should be reviewed by all users.

src/libs/notificationService.ts
kevinmarete: There was no banner-posting doc to update — nothing in this repo, and no DUOS-specific page in Confluence (the only "Banner Instructions" page is Terra/Firecloud's). So I wrote one: `docs/notification-banners.md`, covering the bucket path, the JSON shape and the id rule — new text everyone must see goes out under a new, previously unused id, with the old entry retired via `active: false`. `notificationService.ts` now points at it where the key is derived.

PR #3925, opened 2026-09-11
[DT-3621] Update Voting History table styling to match the Data Library

kevinmarete:
### Addresses

[DT-3621](https://broadworkbench.atlassian.net/browse/DT-3621) — Update Voting History Table Styling

Security risk: no

### Summary

Restyles the Voting History tables (Chair Votes / Member Votes on DAR Collection Review, and the voting history table on the researcher's DAR Application page) to match the flat DataGrid presentation the Data Library established: white rows, thin grey rules, non-uppercase bold headers, bordered/rounded container. Chair and Member votes move from two stacked always-visible sections to a two-tab layout styled like the Data Library's tab bar, with count badges. Presentation only — no columns, data, or sort behaviour changed.

The branch as it stood did not compile: its only commit imported `voteHistoryTableStyles` and `VoteHistoryTabs` but neither file was ever committed. This PR adds both, rebases the work onto develop (it was ~40 commits behind and conflicted with #3920, which removed the "Votes" heading this branch still rendered), and completes the remaining acceptance criteria.

This did **not** migrate to a real MUI DataGrid. The installed `@mui/x-data-grid` is Community edition, which has no expandable detail rows (Pro only), and `ElectionWithMemberVotesTable` relies on an expand/collapse row for member vote details. The existing `SimpleTable` engine is kept and restyled instead. `SimpleTable.tsx` and the shared `Styles.TABLE.*` tokens are untouched, since many unrelated tables (InstitutionTable, DarDatasetTable, ManageUsersTable) depend on them — all styling is scoped through the `styles` prop these components already pass in.

`VoteHistoryTabs` deliberately mirrors `LibraryTabs` rather than reusing it: `LibraryTabsProps` is typed to `AssetType`, and widening it to plain strings would break every Data Library call site under `strictFunctionTypes`. It reuses `COUNT_BADGE_SX` for the count pills so the badge shape and brand colour stay in one place.

`VoteSummaryTable` gains an optional `styles` prop defaulting to its current internal styles, so the nested member-vote detail table can be restyled inside Voting History without affecting `MemberVoteSummary` on the active-voting page.

Verified: `pnpm run type-check` and `pnpm run lint` clean, `pnpm test` 4533/4533 passing. Tab switching, count badges, sorting, and the member-vote expand/collapse were confirmed against a headless Chromium render with fixture data. New tests were mutation-checked — rendering both tables at once, wiring `onChange` to a fixed key, and dropping the count badge each fail the suite.

After

[DT-3621]: https://broadworkbench.atlassian.net/browse/DT-3621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/vote_history_table/ChairVoteHistoryTable.tsx
+++ b/src/components/vote_history_table/ChairVoteHistoryTable.tsx
@@ -1,8 +1,8 @@
 import React, { useMemo, useState } from 'react'
 import SimpleTable from '../SimpleTable'
 import { VoteHistoryRow } from 'src/types/model'
-import { Styles } from 'src/libs/theme'
 import { formatDate, sortVisibleTable } from 'src/libs/utils'
+import { voteHistoryTableStyles } from './voteHistoryTableStyles'
 
 interface ChairVoteHistoryTableProps {
   voteHistory: VoteHistoryRow[]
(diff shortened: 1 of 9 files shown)


fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3927, opened 2026-09-14
DT-3206 - Replace fuzzy study description matches with tags

ncalvanese1:
### Addresses
DT-3206 -- Replacing data library study description logic with tags.

### Summary
Data library search definitions previously used search terms in the study description field, which had the potential for turning up a large number of false positives. Now that study tags have been implemented, we have gone through the data library logic and switched most data libraries to use tags instead of study descriptions, and have added tags where appropriate to ensure the libraries that shouldn't change have remained consistent.

Also, as part of this work, refactored `libraryVersions.spec.ts` to use constraint-based testing rather than structure-specific assertions. Tests now validate Elasticsearch best practices (e.g., `.keyword` suffix usage, valid query structures) instead of expecting specific library configurations. This makes tests future-proof—library query structures can change freely as long as architectural constraints are maintained, eliminating test breakage from routine refactors.

----

- Verify all tests go green


kevinmarete (approved):
Looks good 👍🏾

fboulnois (approved):
👍

PR #3928, opened 2026-09-15
[DT-4062] Share the access-eligibility predicate

otchet-broad:
### Addresses

### Summary

Extracts the Active Researcher Status predicate into `useApplyForAccessEligibility`, so
`LibraryFooter`, `RequestAccessButton` and the study page's new sidebar cannot disagree about
who may apply. The predicate is a library card and nothing else — a chairperson, signing
official or admin holding a card has it too — so it never folds in a role check.

Corrects row-level request gating. 'Request Now' was disabled whenever anything at all was
selected; the grid now receives the selected ids rather than a boolean and disables a row only
when the selection reaches beyond it, so a row that *is* the whole selection stays actionable.
That matters for the study page, which selects every controlled dataset by default and would
otherwise arrive with every row's button disabled.

Also adds a Data Location column (sorting disabled — the cell maps the stored enum to a friendly
label, so sorting the raw value would disagree with the screen) and renames the trailing Actions
column to Export, which is all it holds.

Separately, `.gitignore` now covers crash dumps (`core`, `core.[0-9]*`, `hs_err_pid*`,
`*.hprof`).

----

- Verify all tests go green


kevinmarete (commented):
The new predicate is documented as governing every Apply for Access path, but the public `DatasetStatistics` route still offers the controlled-dataset action without checking Active Researcher Status, while `DatasetSearchFooter` retains its own copy. Please route these entry points through the shared predicate so users receive consistent gating everywhere.

fboulnois:
Some minor suggestions:

**1. `src/components/data_library/columns/datasetColumns.tsx:171`: renaming `actions` → `export` breaks the `extraColumns` override**

`LibraryDataGrid` overrides base columns by `field`. Renaming the base column's `field` from `actions` to `export` means `makeSubmissionColumns`, which still emits `field: 'actions'`, no longer replaces the base column — both survive. My Data Submissions therefore renders an extra, permanently-empty "Export" column (that page never passes `exportableDatasets`).

**2. `src/components/data_library/LibraryDataGrid.tsx:59`: columns memo no longer memoizes**

Replacing the `hasSelection` boolean memo key with the `selectedDatasetIds` array reference defeats the memo. `LibraryPageShell.tsx:70` defaults it to a fresh `[]` literal on every render for consumers that omit it (DatasetSubmissions), so all column definitions are rebuilt and handed to MUI DataGrid on every parent render.

**3. `.gitignore:23`: unanchored `core` pattern is broad**

`core` matches any file or directory named `core` at any depth, not just a root-level core dump; a future `src/core/` would be silently untracked. Use `/core` to match the stated intent.

fboulnois (approved):
Some minor suggestions:

**1. `src/components/data_library/columns/datasetColumns.tsx:171`: renaming `actions` → `export` breaks the `extraColumns` override**

`LibraryDataGrid` overrides base columns by `field`. Renaming the base column's `field` from `actions` to `export` means `makeSubmissionColumns`, which still emits `field: 'actions'`, no longer replaces the base column — both survive. My Data Submissions therefore renders an extra, permanently-empty "Export" column (that page never passes `exportableDatasets`).

**2. `src/components/data_library/LibraryDataGrid.tsx:59`: columns memo no longer memoizes**

Replacing the `hasSelection` boolean memo key with the `selectedDatasetIds` array reference defeats the memo. `LibraryPageShell.tsx:70` defaults it to a fresh `[]` literal on every render for consumers that omit it (DatasetSubmissions), so all column definitions are rebuilt and handed to MUI DataGrid on every parent render.

**3. `.gitignore:23`: unanchored `core` pattern is broad**

`core` matches any file or directory named `core` at any depth, not just a root-level core dump; a future `src/core/` would be silently untracked. Use `/core` to match the stated intent.

kevinmarete (commented):
The study page intentionally displays controlled datasets that may still be awaiting DAC approval. Although `isRowSelectable` now excludes those rows, the Request Path column checks only accessManagement, so an unapproved dataset still gets an enabled “Request Now” button.
Please hide or disable this action when `dacApproval !== true`. If it remains visible but disabled, it should have a specific explanation such as “This dataset is awaiting DAC approval,” rather than reusing the selection-related tooltip.

otchet-broad:
@kevinmarete - The current behavior matches today's production behavior in that if I'm someone with extra privileges to see a row from the library that isn't already approved, we disable the row from selection.  This is adding that back in for these cases.  Some users need to be able to see things in the library because it's necessary for the user to be able to get to the actions column.  If there are additional requirements around this, could we make a ticket and review that outside of these changes?

kevinmarete (commented):
Thanks for the clarification. My concern is specifically the **Request Now** action, not whether privileged users can see the row or access its management actions.  

`isRowSelectable` now excludes datasets awaiting DAC approval, but the Request Path column does not check `dacApproval`. When no other dataset is selected, an active researcher still sees an enabled **Request Now** button that attempts to create a DAR for that unapproved dataset. This is inconsistent with the new selection behavior.  

Please hide or disable **Request Now** when `dacApproval !== true`, with an appropriate explanation. If retaining the current behavior is an explicit product requirement, please link the follow-up ticket or product decision so this can be tracked separately.

otchet-broad:
@kevinmarete - I'll implement your request for this new feature on a branch so that it can be reviewed without impacting the current work.  I'll put together a PR for it so it lands with this larger body of work.

kevinmarete (approved):
This is good  👍🏾 after your clarification on my request moving to a new branch.

fboulnois (approved):
👍

PR #3929, opened 2026-09-15
[DT-4062]  Rebuild the study page frame

otchet-broad:
### Addresses

### Summary

Replaces the study page's single-column layout with the redesign's frame: a sectioned page, a
table of contents built from the sections that register themselves, and a sticky sidebar carrying
the selection summary and 'Apply for Access'. The sidebar is the wide-viewport counterpart of
`LibraryFooter`; both take the shared eligibility predicate. `StudyQueryResult` centralizes the
loading, error and empty states every later section uses. The contents are plain anchors —
nothing highlights the section you are currently in.

The datasets grid moves to server-side paging and sorting against the search index, holding the
previous page's rows while the next loads so paging does not collapse the grid to its empty state
and back. Grid height derives from the row count, so a small study does not reserve a full page
of blank space.

The bit worth the closest read is the default selection. It covers every controlled dataset in
the study, not just the visible page — otherwise 'Apply for Access' would silently submit a
subset of a study larger than one page — which costs an extra query only when the study does not
fit on one page. It is seeded during render rather than from an effect, so the grid never paints
an empty selection it immediately replaces. A person can also interact with the visible page
before the study-wide request completes; that counts as initialization too, so a late response
cannot replace a deliberate choice with every controlled dataset in the study.

Study metadata no longer depends on the dataset index alone. A valid study may have no datasets
and therefore no matching index document, so the name, description, data types and PI name fall
back to the study record.

`/studies/:studyId` is a live route, so this is the PR that makes the redesign visible. Everything
it renders comes from data available today, so it can merge without a backend counterpart.

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

fboulnois (approved):

**1. Column rename breaks extra-column dedupe: `src/components/data_library/columns/datasetColumns.tsx:171`**

Renaming the base column `field: 'actions'` → `'export'` breaks LibraryDataGrid's extra-column dedupe (`extraFields`, `LibraryDataGrid.tsx:54-57`). `makeSubmissionColumns` still emits `field: 'actions'`, so the My Data Submissions → Datasets grid now renders both an Export column and an Actions column instead of Actions alone.

**2. `null` dataTypes can blank the page: `src/components/study_details/StudyDetails.tsx:61`**

`studyDataTypes = study?.dataTypes ?? studyRecord?.dataTypes` can be `null`. `StudyTitleBadges`'s `dataTypes = []` default only fires on `undefined`, so `[...new Set(null)]` throws and blanks the page.

**3. Doc comment improvement: `src/hooks/useStudyDetailsData.ts:124`**

Comment claims "the caller falls back to selecting the visible page" on failure. The caller deliberately does not: the `selectableDatasetIds !== undefined` guard blocks seeding, and the branch's own test asserts the empty selection.

**4. Controlled datasets pre-selected on load: `src/hooks/useStudyDetailsData.ts:130`**

The new default selection filters only on `isRowSelectable` (open/external), while the study query uses `showAllControlled: true`. Controlled datasets still awaiting DAC approval are now pre-checked on page load and go into a DAR with one click. Worth confirming against product intent.

**5. Columns memo rebuilds every render: `src/components/data_library/LibraryPageShell.tsx:70`**

`selectedDatasetIds = []` default allocates a new array each render, and LibraryDataGrid's `columns` memo now depends on that array identity (was a stable boolean), so columns rebuild every render for DatasetSubmissions.

kevinmarete (commented):
👍🏾

kevinmarete (approved):

fboulnois (approved):
👍

PR #3051, opened 2025-09-17
DT-2240: Replace select menu for LC Add User modal

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2240

## Summary

- Update the select component so that it doesn't allow for creating new entities.
- Updated the styling to match what we're doing in the edit DAC page.
- Minor refactoring to simplify the complicated TS form field construction.

### New screen

Screenshot 2025-09-17 at 4 58 00 PM

 Screenshot 2025-09-17 at 4 56 09 PM

----
Have you read [Terra's Contributing Guide](https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #3930, opened 2026-09-15
[DT-4062] Show PI institution and profile links

otchet-broad:
### Addresses

### Summary

Adds the PI institution row and ORCID / LinkedIn / website icons to the overview.

The page's primary `study` comes from the Elasticsearch-backed search index, which does not carry
these fields, so they are read from the relational store through a new `Study.getById`.
`DataSet.getStudyById` now delegates to it instead of duplicating the request, while still typing
the payload as the data-submission form's editable `Study` shape — the two callers of one endpoint
stay honest about wanting different views of it.

That relational response is also the fallback for study metadata when a study has no datasets and
so has no search-index document, which is why the two changes sit together.

The profile links live inside the PI Name row, and `StudyInfoTable` drops rows whose value is
falsy, so that row's presence cannot hinge on `piName` alone: the index sometimes has no name for
a study whose profile links are populated. Only plain http(s) urls are linked.

`GET api/dataset/study/{studyId}` already exists on consent `develop`, so until #3051 deploys the
request still returns 200 — only the new fields are absent. The institution row is omitted and no
icons render. This is the one branch in the stack that degrades quietly rather than showing an
error state: the endpoint is not new, only the fields on it are.

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

fboulnois (approved):
Some minor findings identified with Claude and Codex:

#### 1. Profile link validation checks the scheme but never the host (security):

`src/components/study_details/piProfileLinks.ts:39`: `validateHttpUrl` only verifies the URL scheme. Any submitter-supplied `https://` URL in `piOrcid` or `piLinkedinUrl` is rendered on the public study page under the "ORCID profile" / "LinkedIn profile" label and the iD icon.

This is a link-spoofing surface on a page visible to unauthenticated visitors. Suggested fix: constrain the host to `orcid.org` / `linkedin.com` and demote anything else to the generic `website` kind.

#### 2. ORCID scheme check is case-sensitive

`src/components/study_details/piProfileLinks.ts:28`: `normalized.startsWith('http')` misclassifies `HTTPS://orcid.org/0000-...` as a bare identifier, producing `https://orcid.org/HTTPS://orcid.org/0000-...`.

#### 3. Relational fallbacks use `??`, so empty index values win

`src/components/study_details/StudyDetails.tsx:64`: `??` does not fire on `''` or `[]`, so a stale or empty index document suppresses exactly the relational `piName` / `dataTypes` the fallback exists to supply.

kevinmarete (approved):
👍🏾

PR #3931, opened 2026-09-15
[DT-4062] Add study ratings and comments

otchet-broad:
### Addresses

### Summary

Adds the Comments & Ratings section: the average rating, existing comments with their authors and
institutions, and a composer for posting or editing your own. `StudyComments.deleteComment`
covers the DELETE endpoint for a follow-up; no delete control is rendered yet.

Posting is gated on the Researcher role *and* an active library card, because that is what
`StudyCommentService` checks when the comment is saved. Gating on the card alone would offer the
composer to a chairperson or signing official whose post the backend would reject. The card half
comes from the shared predicate in PR 1.

One comment per user per study is a database constraint, so posting again edits in place rather
than adding a row. The composer says so before the first post, together with the disclosure that
the comment appears publicly under the author's name and institution, and stops at the 2000
characters the backend accepts.

The endpoint is paged, so comments load through `useInfiniteQuery` and 'show more' fetches the
next page and appends it — the page size is capped, so asking for a bigger page is not an option.
Offset paging can repeat a boundary item if another reader posts mid-sequence, so the pages
already on screen are refreshed before the next offset is calculated and the list is de-duplicated
by id.

The composer reads the reader's own comment from the response's `yourComment` rather than by
scanning the loaded page: paging puts it on an unpredictable page, and scanning would offer to add
a comment to someone who already has one and then silently revise it on save.

The section renders through `StudyQueryResult`, as its siblings do, so until consent #3052 deploys
the failed list request reports itself rather than presenting as a study nobody has commented on.

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

fboulnois (approved):

**1. A failed background refetch wipes loaded comments and the unsaved draft**: `src/components/study_details/StudyCommentsSection.tsx:86`

Gating the whole section on `error` means a failed *background* refetch, as the `refetch()` in "Show more", or the post-success `invalidateQueries`, discards already-loaded comments and the composer along with the reader's unsaved draft, replacing everything with "Unable to load comments and ratings." With `refetchOnWindowFocus: false`, a 5-minute `staleTime`, and no retry control, the section stays stuck there.

**Fix:** only show the error state when there is no cached data (e.g. `error && !data`); surface a background-refetch failure as a non-destructive inline message with a retry.

**2. "Show more" refetches every previously loaded page**: `src/components/study_details/StudyCommentsSection.tsx:79`

`showMoreComments` refetches all loaded pages before fetching the next, so a full walk of a 200-comment study costs ~44 GETs instead of 8, and widens the failure window for finding 1.

**Fix:** call `fetchNextPage()` directly instead of `refetch()` + fetch.

**3. Unencoded `studyId` interpolated into the API path**: `src/libs/ajax/StudyComments.ts:5`

The route-derived `studyId` is interpolated into the API path without encoding. `/studies/1%2F..%2F..%2Fsomething` decodes to `1/../../something` and sends an authenticated GET to a normalized, unintended path. `FileStorageObject.ts:23` and `FeatureFlag.ts:25` already use `encodeURIComponent` for exactly this.

**Fix:** wrap the id in `encodeURIComponent(...)` (or validate it as numeric).

**4. Text typed while the POST is in flight is silently discarded**: `src/components/study_details/StudyCommentsSection.tsx:53`

Inputs stay enabled during the mutation and `onSuccess` re-seeds state from the server copy, so anything typed mid-flight is lost without warning.

**Fix:** disable the inputs while `isPending`, or skip re-seeding when the local value diverges from the submitted one.

**5. Pagination counts raw items but the list de-duplicates**: `src/hooks/useStudyDetailsData.ts:45`

`getNextPageParam` counts raw fetched items while the list de-duplicates by `studyCommentId`. A repeated boundary item makes `loaded` reach `total` early, hiding "Show more" while fewer than `total` distinct comments are on screen.

**Fix:** derive `loaded` from the de-duplicated set, or page by a stable cursor rather than an offset count.

kevinmarete (approved):
👍🏾

PR #3933, opened 2026-09-15
[DT-4062] Add study recommendation sections

otchet-broad:
### Addresses

### Summary

Adds two recommendation sections below the research outputs: studies often requested in the same
data access request as this one, and studies recommended by shared data type or PI. Each card
carries the study name, PI and description, and links through to that study.

These are responsive grids rather than carousels — every recommendation renders, wrapping onto
further rows as needed, with no horizontal scrolling or next/previous affordance. The PR
previously called them carousels; a grid is what is built and what this describes. Building real
carousel interaction on a page already this long is worth doing deliberately rather than as a
review fix, so it is left as a follow-up.

Each card's action area is a router link with a `to` value rather than a button that navigates
imperatively, matching how study destinations work elsewhere (`StudyCard`). That restores the
native link actions a button swallows: open in a new tab, copy link address, middle-click.

Both sections stay on the page when they come back empty, saying so, for the same reason as the
other sections — an absent section on a study page reads as a failure to load.

The response is typed as a new `StudyRecommendation` rather than the index-shaped
`StudyAggregation`. The endpoints return only studyId, studyName, studyDescription, piName,
datasetCount and datasetIds, so the richer type would let a reader of `study.dataTypes` compile
and then fail at runtime. `datasetCount` therefore comes typed for whoever wants it on the card.

The sections sit near the bottom of a long page, and BrowserRouter preserves the document offset
during an in-app navigation, so the scroll position resets whenever the route points at a
different study — otherwise following a recommendation opens the next study at the same deep
scroll position.

Worth knowing for review: the backend restricts candidates to publicly visible studies, so a
recommendation never reveals a private study to someone who could not otherwise see it.

----

- Verify all tests go green


kevinmarete (commented):
After an initial successful request, React Query can retain `recommendations` while setting `error` when a background refetch fails. Because `error` is passed unconditionally to `StudyQueryResult`, the component replaces valid cached cards with the error state. This also contradicts the comment that background refetches should keep cards on screen.
Please make the error blocking only when no recommendations are available—for example, `error={recommendations.length === 0 ? error : undefined}`—and add a regression test covering cached data plus a refetch error.

kevinmarete (approved):
👍🏾 

Non-blocking: the similar-studies endpoint considers both overlapping data types and a shared PI, but the heading says “Recommended Studies based on Data Type.” Could we rename it to “Recommended Studies” or mention both signals?

fboulnois (approved):

**1. Unencoded route parameter in API path:** `src/libs/ajax/StudyRecommendations.ts:6`

`studyId` from `useParams` is interpolated raw. `useParams` decodes the URL, so `/studies/1%2F..%2Fdar-summaries%2F5` becomes `/api/metrics/study-recommendations/1/../dar-summaries/5/similar`, which the browser normalizes into an authenticated GET at a different endpoint. Sibling `StudyComments.ts:5-9` wraps the same id in `encodeURIComponent` for exactly this reason.

**2. Scroll reset fires on mount, defeating Back/Forward restoration:** `src/components/study_details/StudyDetails.tsx:269`

The `[studyId]` effect runs on initial mount, not only when the route changes to a different study. Scroll deep into `/studies/5`, follow a dataset link, press Back: the browser restores the offset, then the effect snaps to top. A ref that skips the first run, or a pathname-driven reset in the router layer, matches the comment's stated intent.

**3. Error guard keys on list length, not cached-ness:** `src/components/study_details/StudyRecommendationCarousel.tsx:28`

A study with zero recommendations shows "No study recommendations yet."; if a background refetch then fails, React Query keeps `data=[]` and sets `error`, and `recommendations.length === 0 ? error : undefined` flips the section to the error banner. `StudySecondaryResearchOutputs.tsx:50` uses `data ? undefined : error`; use `recommendations === undefined ? error : undefined` here.

**4. Recommendation queries hoisted into the page component:** `src/components/study_details/StudyDetails.tsx:89`

`useSimilarStudies` and `useFrequentlyRequestedWithStudies` live in `StudyDetailsContent` and are threaded through props, so every pending→success→refetch transition re-renders the whole page including the DataGrid. Siblings `StudyDarHistory.tsx:49` and `StudySecondaryResearchOutputs.tsx:41` take `studyId` and own their hook. A thin wrapper removes 4 imports, 2 hooks and 6 props.

**5. Scroll-to-top is a router-level concern patched into one page:** `src/components/study_details/StudyDetails.tsx:266`

The root cause the comment names — `BrowserRouter` preserves document offset on in-app navigation — applies app-wide: "Back to library" from a deep study page lands on `/datalibrary` at the same offset. A single pathname-keyed reset next to `BrowserRouter` in `src/index.tsx`/`AppRoutes.tsx` (or a data router with `<ScrollRestoration>`) fixes every route once.

**6. Current study not filtered from its own recommendations:** `src/components/study_details/StudyRecommendationCarousel.tsx:34`

If `/similar` or `/frequently-requested-with` returns the viewed study (natural for a data-type similarity query), a card links back to the current page; `useParams` is unchanged, so neither the key nor the scroll effect fires and the click appears to do nothing. Filter `String(r.studyId) !== studyId` in the hook, or pass the current id to the carousel.

fboulnois (approved):
👍

PR #3934, opened 2026-09-15
[DT-4062] Add study registration asset sections

otchet-broad:
### Addresses

### Summary

Adds the registration assets: AI models, workspaces, presentations, clinical trials, intellectual
property and funding resources as tables, publications as cards, and the overview count badges.
Each table reuses the data library's existing column definitions for that asset type, so a study
page and the library render the same asset the same way.

The library's Study column is dropped from these tables: it exists for the cross-study view, and
these endpoints return the registration JSON, which carries no study name — in production data
every asset has `studyId: ""` and no name, so the column rendered blank with an invisible link.

Rows are keyed by position rather than by the asset's own id, which is submitter-supplied and
frequently blank (`modelId: ""` in real data). Two equal ids are one id to the grid, which rendered
a single row for two distinct assets. Nothing here needs the asset's identity — the tables have no
selection, editing or paging. The publication cards key the same way.

Publications get cards rather than a table because title, authors and journal do not read as
tabular data; only plain http(s) urls are linked. `Study.getPublications` normalizes author names,
which the payload may carry as `authorNames` or as `authors[].name`, so the card does not need to
know which it was given.

The six column factories are module constants rather than being called on every render.

Every section stays on the page when empty and reports a failed fetch in place. An asset section
that silently disappears reads as 'this study has none', which is the wrong thing to tell a
researcher when the request failed.

----

- Verify all tests go green


kevinmarete (approved):
Looks good overall. 

One follow-up: `StudyPublicationCards` passes `error` unconditionally, so a failed background refetch replaces cached publication cards with the error state. Consider matching `StudyAssetTable` by passing the error only when `data.length === 0`, with a test covering a successful load followed by a failed refetch.

fboulnois (approved):
A few suggestions identified with Claude and Codex:

**1. Intellectual property endpoint may use the wrong collection name:** `src/libs/ajax/Study.ts:64-65`

The route uses `intellectualProperty`, while other collection references use `intellectualProperties`. This could cause 404s and prevent the section from loading. Confirm the backend contract before changing it; this is not a proven defect.

**2. Failed background refetch hides publication cards:** `src/components/study_details/StudyPublicationCards.tsx:30,37`

A refetch error replaces already-loaded publication cards with an error message despite cached data remaining available. Match `StudyAssetTable` by showing the error only when there is no data.

**3. Count badges use plural labels for single assets:** `src/components/study_details/StudyAssetCountBadges.tsx:9,15`

Badges display “1 Datasets”, “1 Models”, and similar labels because they always use plural names. Supply singular and plural labels and select the appropriate form based on the count.

**4. Eight requests start on page mount:** `src/components/study_details/StudyDetails.tsx:110-117`

The PI query and seven asset queries run immediately, including for sections users may never view. These requests fetch needed data and are not redundant. Consider loading lower sections when visible or adding a combined assets endpoint.

kevinmarete (approved):
👍🏾 

Non-blocking: the count badges always use plural labels, producing text such as “1 Datasets,” “1 Models,” and “1 Workspaces.” Could the badge input provide singular and plural forms, or select the label based on the count?

fboulnois (approved):
👍

PR #3936, opened 2026-09-15
[DT-3313] [2/6] Derive filter options from the full corpus, not the current page

kevinmarete:
### Addresses

[DT-3313](https://broadworkbench.atlassian.net/browse/DT-3313) — 2 of 6, splitting #3926 per review. Stacked on #3935. Security risk: no.

### Summary

This is the one carrying the risk, so it lands against the existing filter set only — no new filter keys — as suggested in review.

Workspace Tools, Workspace Platform, Registry and Optional Data Use derived their options from `data.items`, which only holds the active tab's current page. They listed whatever happened to be on screen, and went empty the moment the user switched tabs. That's the bug this fixes.

They now come from the whole matching corpus. `tabCountsResponse` already carries every matching study via `STUDIES_AGG`'s `terms.size: 10000`, so re-running an asset's own `transformResponse` with an unbounded page size yields its full flattened item list whichever tab is rendering. This mirrors `libraryCounts.ts`, which already calls `transformResponse` a second time with different pagination purely to read `.total`. To be explicit, since it came up in review: that unbounded page size is client-side pagination over a response already in memory, not an Elasticsearch request.

The visible tab additionally reads a corpus with its own filter keys cleared, which is the part worth the attention. The filter clauses match whole studies, so selecting one value drops every study without it, and a sibling value living only in one of those studies could never be offered again — no client-side pass can recover a study the response never carried. Filters owned by other tabs stay applied, so the options stay scoped the way the grid is. The clearing is a no-op until the tab actually has one of its own filters set, so the second `useLibraryTabCounts` shares the counts query's key, and its request, until it does.

A value excluded by *another* tab's filter is absent from that corpus too, since that filter stays applied. `withSelectedValues` re-adds anything currently selected, or the checkbox disappears while the filter stays active — and the external chips deliberately skip keys the current tab renders itself, so nothing would be left to clear it with short of resetting every filter.

Eight new tests. Mutation-checked rather than trusted on coverage: collapsing the corpus map to the unfiltered response fails the self-exclusion test, and short-circuiting `withSelectedValues` fails its own. The first pass of tests missed `withSelectedValues` entirely — every case I'd written was already satisfied by self-exclusion — so the test that isolates it uses a filter owned by another tab, which is the only situation where self-exclusion cannot help.


fboulnois (approved):
Looks reasonable, two findings identified with Claude:

**1. Option corpus clears all visible filters, not just the corpus-derived ones:** `src/hooks/useLibraryPageState.ts:124`

The option corpus clears *every* filter in `assetFilterRegistry[tab].visibleFilters`, not just the checkbox filters whose options come from that corpus.

- On **Clinical Trials**, `clinicalTrialDates` gets cleared, so `clinicalTrialRegistry` options are derived from trials outside the date window. `fullCorpusItems` also passes `EMPTY_FILTERS`, so the client-side date pass is skipped too. Checking such a registry makes `matchesClinicalTrialFilters` reject every row → empty grid, `0` badge.
- Same on **Biospecimens** for `biospecimenCollectionDate`, `biospecimenPostMortemInterval`, and the shared study-level `participantCount` vs. `biospecimenDataUse`.
- **Workspaces** is unaffected — its only visible filters are the two option filters.

This is a regression: options previously came from `data.items`, which was filtered by the full filter set, so an offered value could never match nothing.

**Fix:** clear only the corpus-derived checkbox keys.

**2. `withSelectedValues` appends instead of merging, breaking alphabetical order:** `src/components/data_library/filterRegistry.ts:645`

`withSelectedValues` concatenates missing selected values onto the end of the option array, breaking the alphabetical order the lists are built with. `LibraryFilters.renderCheckboxSection` renders in array order with no sorting or show-more, so a still-checked value re-added this way appears at the bottom of a long list.

**Fix:** merge into the sort instead of appending.

kevinmarete:
Both fixed, thanks.

1. `4c88b20f` — clears only the keys whose options that corpus supplies, and re-applies the rest in the transform instead of `EMPTY_FILTERS`, so the client-side date/range pass still runs. A date-only edit now widens nothing and mounts no second aggregation.
2. `b25f2bed` — merges the re-added value into the order, but only where the list is already alphabetical; the enum lists keep their deliberate order.

otchet-broad (approved):
👍

PR #3937, opened 2026-09-15
[DT-3313] [6/6] Replace the Datasets Cited filters with a column on both grids

kevinmarete:
### Addresses

[DT-3313](https://broadworkbench.atlassian.net/browse/DT-3313) — 6 of 6, splitting #3926 per review. Stacked on #3939. Security risk: no.

### Summary

**This is the one that needs a product call**, which is why it's isolated and last — it removes a filter users have today. Nothing else in the series depends on it, so a hold here blocks nothing.

Presentations and Publications each offered a Yes/No "Datasets Cited" filter. A per-row fact reads better as a column: the Elasticsearch clause matches whole studies, so selecting "Yes" kept every presentation of any study that had at least one citing presentation, including the non-citing ones. The row-level pass then hid them, which meant the filter's answer depended on which other presentations happened to share the study.

Both grids now carry a Datasets Cited column. A row indexed without the field reads as "No", matching how both transforms already default it (`citation ?? false`), and the citation text goes in the tooltip where the index carries it. Because this lands after #3939, both tabs keep a filter panel throughout — they just trade this one boolean for Event/Format/Access/date or Journal/Access/date.

Retired URL params (`datasetsCited`, `presentationsDatasetsCited`, `publicationsDatasetsCited`) are now deleted on write. I moved that delete from `serializeFiltersToUrl` into `updateState`: serialization only runs when an update carries `filters`, so an old link's param survived indefinitely for anyone who never touched the filter panel — the case the cleanup exists for. A test covers it via a query-only update.

Mutation-checked: forcing the column to always read "Yes" fails 7 tests, and dropping the retired-param delete fails its own.


kevinmarete:
@fboulnois this bundles two separable changes: adding the Datasets Cited column (additive), and removing the Yes/No filter (a user-visible removal).

@jlaw-codes — the removal needs your nod. Today's filter can't be a clean per-row filter: the Elasticsearch clause matches whole studies, so "Yes" returns every presentation of any study with at least one citing presentation, and the client-side pass then hides the rest. The answer depends on which siblings share the study.

If that isn't a quick yes, I'll split the column out so it can merge and the removal waits.

Asking before #3939 merges, since this is based on it.

fboulnois (approved):
**1. Citation tooltip shows on rows labelled "No":** `src/components/data_library/columns/sharedColumns.tsx:78`

The tooltip shows a dataset citation on rows labelled "No". `datasetCitation` is required on the publication form (`src/utils/darFormUtils.ts:396`) regardless of the `citation` answer, so uncited publications routinely carry text; the cell reads "No" while its tooltip/accessible description names a citation.

The new spec only covers `citation: false` with empty text, so this combination is untested.

**Fix:** guard the tooltip on `params.row.citation`.

**2. Hardcoded `tabIndex={0}` breaks the DataGrid roving tabindex:** `src/components/data_library/columns/sharedColumns.tsx:85`

The hardcoded `tabIndex={0}` bypasses the DataGrid's roving tabindex: every visible row adds a page-level tab stop (50 extra Tab presses at pageSize 50), and Tab from outside enters the first citation cell rather than the grid's focused cell — with no focus ring, since `LibraryDataGrid` sets `:focus-within { outline: none }`.

The repo's own pattern at `src/components/manage_users_table/ManageUsersTable.tsx:70-72` uses `tabIndex={params.tabIndex}` for exactly this reason.

**Fix:** use `tabIndex={params.tabIndex}`.

kevinmarete:
Both fixed in `fdfc33ad`, thanks — checked each premise and they hold.

1. `datasetCitation` is required independently of the `citation` answer (`darFormUtils.ts:396` is its own check), so uncited publications do carry text, and `describeChild` was handing it to a cell reading "No" as its accessible description. Tooltip now guarded on `params.row.citation`.

2. Confirmed the full chain — `ManageUsersTable.tsx:70-72` uses `tabIndex={params.tabIndex}` for exactly this, and `LibraryDataGrid.tsx:155` does set `:focus-within { outline: none }`, so the stolen focus would have been invisible too. Now takes the grid's roving value.

Worth flagging on the test side: the existing `tabindex` assertion passed against either implementation, because the spec harness never passed `tabIndex` at all and the hardcoded `0` matched by accident. It now drives both `0` and `-1`, plus a case for the uncited-row-with-text combination from your first point.

otchet-broad (commented):
👍

otchet-broad (approved):
👍

PR #3938, opened 2026-09-15
[DT-3313] [3/6] Add filters to the AI Models and Workspaces tabs

kevinmarete:
### Addresses

[DT-3313](https://broadworkbench.atlassian.net/browse/DT-3313) — 3 of 6, splitting #3926 per review. Stacked on #3936. Security risk: no.

### Summary

Mechanical, following the pattern #3936 establishes. AI Models gains Format, License, Cloud and Tags; Workspaces gains Cloud and Access alongside the Tools and Platform it already had. Both grids gain a Cloud column, so the new filter acts on something the user can see.

Each new filter is a checkbox group whose options come from the full corpus (#3936), a `matchAny` clause, and a client-side row predicate. The predicate is not redundant: the Elasticsearch clauses only decide which *studies* enter the shared aggregation, so without it a study with one matching model contributes all of its models to the grid — and to the tab-count badge, which is derived from the same function.

Also carries the fix for the blank page reported on #3926. Studies indexed before Cloud became a creatable multi-select hold it as a bare string rather than an array, and the new chip column called `.join` on it, which threw and blanked the whole grid — only these two tabs, since they're the two with a Cloud column. `cloud`, `tags`, `tools` and `trainedOnDatasets` are now normalized through `toStringArray`, and the raw ES types say `cloud?: string[] | string` so the normalization can't be skipped. Verified against the dev corpus.

Mutation-checked: short-circuiting the `modelCloud` predicate fails 2 tests, and making `toStringArray` drop the bare-string case — i.e. reintroducing the blank-page bug — fails 4.


fboulnois (approved):

**1. New model filters use case-sensitive exact matching on free-text fields:** `src/components/data_library/assets/modelAsset.ts:14`

The new model filters (`modelFormat`, `modelLicense`, `modelCloud`, `modelTags`) re-check rows with exact, case-sensitive `includes`, but all four are free-text fields — `AiModelAddEdit` renders format/license as plain text inputs, and tags/cloud as creatable selects.

Consequence: `"PyTorch"` and `"pytorch"` become two separate checkboxes; selecting one drops the other's rows, even though the analyzed `match_phrase` clause already admitted both studies.

Every other free-text/corpus-derived filter in the repo uses `includesIgnoreCase` (`workspacePlatform`, `clinicalTrialRegistry`, `clinicalTrialInterventionType`, `biospecimenDataUse`); only enum-backed keys use exact `includes`.

**Fix:** use `includesIgnoreCase` for these four keys.

**2. `workspaceAccess` uses exact matching on an unvalidated free-text field:** `src/components/data_library/assets/workspaceAsset.ts:40`

Same class as above, but sharper: `workspaceAccess` uses exact `includes` on the line immediately after `workspacePlatform` uses `includesIgnoreCase`. `access` is a bare `FormField` with no options and no validator at all — the most variance-prone field of the set.

**Fix:** use `includesIgnoreCase` here too.

**3. `modelCloud` and `workspaceCloud` share the label 'Cloud':** `src/components/data_library/filterRegistry.ts:534`

Both keys label as `'Cloud'`. On the Workspaces tab, a carried-over model filter renders as "Cloud: AWS" under "Filters from other views", visually identical to the tab's own unchecked "Cloud" section.

The repo already disambiguates its one other label collision ('Datasets Cited (Presentations)?' vs '(Publications)?').

**Fix:** disambiguate one of the two labels.

kevinmarete:
All three fixed in `ec18dc3e`, thanks.

1 & 2. `includesIgnoreCase` for `modelFormat`, `modelLicense`, `modelCloud`, `modelTags` and `workspaceAccess`. Added a local copy of the helper rather than hoisting it — that would have reached into `clinicalTrialAsset`/`biospecimenAsset`, which belong to an earlier PR in the stack. Four tests pin the casing behaviour.
3. Now `Cloud (AI Models)` and `Cloud (Workspaces)`; I qualified both, since leaving either bare still reads ambiguously on the other tab.

The same collision class turned up twice more downstream — `Access`/`Format` on #3939 and `Status` on #3940, both fixed there, with a test that fails on any new duplicate label.

otchet-broad (approved):
👍

PR #3939, opened 2026-09-15
[DT-3313] [4/6] Add filters to the Presentations and Publications tabs

kevinmarete:
### Addresses

[DT-3313](https://broadworkbench.atlassian.net/browse/DT-3313) — 4 of 6, splitting #3926 per review. Stacked on #3936. Security risk: no.

### Summary

Presentations gains Event, Format, Access and a Presentation Date range; Publications gains Journal, Access and a Published Date range. Both grids gain an Access column so the new filter acts on something the user can see. The existing Datasets Cited filter is untouched here — #3937 handles that swap, and it lands after this so neither tab is ever left without a filter panel.

The checkbox filters follow the template from #3936: options from the full corpus, a `matchAny` clause, a client-side row predicate.

The two date ranges are the part worth a look. Both reuse `isInvertedDateRange`, so crossed bounds read as inactive consistently in the panel's validation message, `isFilterActive`, the badge, the chips and the query clause. The row predicate defers to `isFilterActive` rather than testing the bounds itself for exactly that reason — otherwise an inverted range would build no Elasticsearch clause while the client-side pass still excluded every row, emptying the grid while the panel flagged the range as invalid. A missing date matches neither bound, which is how the Elasticsearch range clause already behaves.

Both use the existing `dateRangeClause` helper rather than the inline `as unknown as QueryClause` cast the older date filters use.

Mutation-checked: short-circuiting the `presentationAccess` predicate fails 1 test, and ignoring the date lower bound fails 2.


fboulnois (approved):
A few suggestions:

**1. Free-text clauses use `match_phrase`, so ES substring-matches while the client re-check is exact:** `src/components/data_library/filterRegistry.ts:556`

The new free-text clauses (`presentationEvent`, `presentationFormat`, `presentationAccess`, `publicationJournal`, `publicationAccess`) use `matchAny`, i.e. `match_phrase`, which matches any value *containing* the phrase.

Selecting Journal "Nature" also matches a study whose only publication is in "Nature Genetics". That study then enters the shared `studies` aggregation, so the Studies/Datasets badges and every other tab's grid include it, while only the Publications grid hides the row via the exact-equality client re-check in `matchesPublicationFilters`. Same for "ASHG" vs. "ASHG Annual Meeting", "Oral" vs. "Oral Presentation".

`soApprovalModel`/`accessManagement` already use an exact `.keyword` term match.

**Fix:** use an exact `.keyword` term match here too, so the query and the client predicate agree.

**2. `presentationAccess` and `publicationAccess` share the label 'Access':** `src/components/data_library/filterRegistry.ts:540`

Both keys carry `label: 'Access'`. Because external chips are labelled from `FILTER_DEFINITIONS[key].label`, a presentation Access filter carried onto the Publications tab renders "Access: open" right next to that tab's own "Access" section with the same option, with no way to tell them apart.

**Fix:** disambiguate one of the two labels.

kevinmarete:
Thanks — good catch on the first one.

1. Fixed in `4021a3cf`: the five free-text keys now use an exact `.keyword` term match instead of `match_phrase`, so the clause and the row predicate agree. Worth a sanity check before merge: nothing in the repo references a `.keyword` subfield under `study.assets.*`, and there is no mapping file here, so I am relying on ES default dynamic mapping — the same thing `study.studyName.keyword` and `study.dataTypes.keyword` already depend on. If those fields turn out to be explicitly mapped without a keyword subfield, these filters would match nothing rather than over-match.

2. Already fixed in `103f240b`, pushed ~45 min before your review — the labels are now `Access (Workspaces)`, `Access (Presentations)` and `Access (Publications)`. `Format` collided the same way (`modelFormat` vs `presentationFormat`) and is qualified too, and there is now a test that fails on any new duplicate label. It caught a third one, `Status`, on #3940.

One thing to decide: this is the same defect shape as finding 1 on #3938, but the two fixes point opposite ways — there the client was loosened to `includesIgnoreCase` to match the analyzed clause, here the clause was tightened to match the exact client. `.keyword` looks like the more correct direction, since the option lists are built from exact corpus values. `workspacePlatform`, `clinicalTrialRegistry` and `biospecimenDataUse` are all on the loose side today and predate this stack, so I have left them alone.

kevinmarete:
Checked the `.keyword` caveat against the dev index — it holds, so `4021a3cf` is safe.

All five subfields exist and are populated:

```
study.assets.presentations.event.keyword     ["", "The Event"]
study.assets.presentations.format.keyword    ["", "PPTX"]
study.assets.presentations.access.keyword    ["", "Private"]
study.assets.publications.journal.keyword    ["Journal of My Stuff", "Test Journal"]
study.assets.publications.access.keyword     ["", "Private"]
```

And the over-match reproduces on real data. Dev has two journals, so the bare token `Journal` is the `Nature` / `Nature Genetics` case:

```
term  journal.keyword = "Test Journal"   -> 1 study
term  journal.keyword = "Journal"        -> 0 studies
match_phrase journal  = "Journal"        -> 2 studies   <- the bug
match_phrase journal  = "Test Journal"   -> 1 study
```

The extra study lands in the shared aggregation every tab's badge counts, while only the Publications grid drops the row.

Worth noting this also settles the direction question from my last comment: `workspacePlatform`, `clinicalTrialRegistry` and `biospecimenDataUse`, plus the five keys on #3938, all still use `match_phrase` and have this same over-match today. Happy to fold a fix into the epic if you want it.

otchet-broad (approved):
👍

PR #3940, opened 2026-09-15
[DT-3313] [5/6] Add filters to the Intellectual Property and Funding Resources tabs

kevinmarete:
### Addresses

[DT-3313](https://broadworkbench.atlassian.net/browse/DT-3313) — 5 of 6, splitting #3926 per review. Stacked on #3936. Security risk: no.

### Summary

The smallest of the asset PRs. Intellectual Property gains Type and Status alongside the Filed Date it already had; Funding Resources gains Funder Name alongside Funding Dates. Same template as #3938 and #3939: options from the full corpus, a `matchAny` clause, a client-side row predicate.

One thing that isn't purely additive, and is the reason to read the predicates rather than skim them: both previously ended with the date check and returned `true` early when the range was inactive. That was correct when the date was the tab's only filter. With a second filter on the tab, an inactive range has to fall through to the remaining checks instead of passing the row outright, so both predicates are restructured to run the date check as a block and let the array checks decide.

Mutation-checked: short-circuiting either the `ipStatus` or the `fundingFunderName` predicate fails a test.


fboulnois (approved):
One suggestion:

**1. New URL params have no round-trip test coverage:** `test/hooks/useLibraryUrlState.spec.tsx`

The three new URL params (`ipType`, `ipStatus`, `fundingFunderName`) have no parse/serialize round-trip coverage, unlike every sibling filter added earlier in this stack.

A wrong or duplicated `param` string in `ARRAY_FILTER_PARAM_CONFIG` type-checks fine (only `key` is typed) and would silently clobber another filter's query param on share/restore.

**Fix:** add round-trip cases for the three new params, matching the existing sibling coverage.

kevinmarete:
Fixed in `19170c5c`, thanks.

Added parse/serialize/clear cases for `ipType`, `ipStatus` and `fundingFunderName`, matching the sibling coverage.

Also added a pass for the collision specifically: no per-key test can catch a duplicated `param`, since each key still round-trips fine in isolation while quietly clobbering its sibling. That one writes every array filter at once and reads them all back. Pointing `fundingFunderName` at `ipType`'s param fails it; a plain typo fails the per-key cases.

src/components/data_library/filterRegistry.ts
```diff
@@ -667,8 +685,29 @@ const FILTER_DEFINITIONS: Record<FilterKey, FilterDefinition> = {
       return dateRangeClause('study.assets.publications.publishedDate', { gte: after, lte: before })
     },
   },
+  ipType: {
+    label: 'Type',
+    buildClause: filters =>
+      filters.ipType.length > 0
+        ? matchAny('study.assets.intellectualProperties.type', filters.ipType)
```
otchet-broad: Should this be using the .keyword termAny helper from #3939 ?  Same for IP Type and Status.  These are a controlled vocabulary that should be exact matched against, right?

kevinmarete: Yes — you're right on all three. Fixed in `ea019e8c`.

They are controlled vocabularies, their options are read off the corpus, and each asset's row predicate is already an exact `includes`. `match_phrase` is the looser side of that pair: selecting IP Type "Provisional Patent" also matched a study holding only a "Provisional Patent Application". That study entered the shared `studies` aggregation, which the Studies (`total_studies` cardinality) and Datasets (`datasets_count` filter) badges count straight from with no row re-check, so those badges over-counted while the IP grid itself dropped the row — the same divergence `termAny` was added for in #3939.

One caveat that applies equally to the #3939 fields: the option lists are trimmed by `uniqueValues` while `.keyword` indexes the raw value, so a document with `"  Patent "` yields the option `Patent` and the `term` clause won't match it back. Genuinely fixing that wants a normalizer on the keyword subfield in the index mapping rather than anything on this side.

`clinicalTrialStatus`, `clinicalTrialPhase` and `clinicalTrialRegistry` have the identical shape but predate this stack, so I left them out of scope here — happy to pick them up separately if you'd like.

otchet-broad (approved):
👍

PR #3943, opened 2026-09-17
[DT-4062] Don't show DAR requester identity, but name the institution.

otchet-broad:
### Addresses

### Summary

The dataset page's request cards showed only the code, title, dates and research
use statement, so there was no way to tell where a grant had gone without
opening the study page. They now name the requesting institution, matching the
cards in `StudyDarHistory`.

Not the PI. The name identifies the person holding the grant, which neither page
publishes; the backend stops sending it either way, so it would render
"Not provided" for every row. The `Not provided` fallback stays for the
institution, which is optional on the payload and would otherwise render a bare
label.

The study's own PI, shown further up this page, is untouched — that is study
metadata rather than a requester, and not what this section describes.

Two tests: one expands a card and asserts the institution appears while the
requester's name does not, one asserts the fallback. Both fail with the block
removed.

----

- Verify all tests go green


kevinmarete (approved):

kevinmarete (approved):
👍🏾

fboulnois (approved):
<ignore>

PR #3944, opened 2026-09-18
[DT-3940] Remove the client-side legacy match suppression

kevinmarete:
### Addresses

[DT-3940](https://broadworkbench.atlassian.net/browse/DT-3940) — the duos-ui half, carved out of DT-4008. Independent of the Consent PRs (DataBiosphere/consent#3073–#3076), but gated on the same rollout.

**Security risk: no** — client-side display logic for a match result the DAC already sees; no request, credential or authorization path is touched.

### Summary

`BucketUtils` suppressed v1 and v2 match results because Consent did not decide abstention before v3. The `dataset_id` migration has restamped or removed every such row, so the suppression can now only misread a version it does not recognize.

Legacy and absent versions report the uninterpretable result, the same as any other unrecognized stamp. `CLIENT_SUPPRESSING_VERSIONS`, the `client-suppresses` arm, `shouldAbstain`, `isOther` and `AbstainDataUseCodes` are gone, along with the `isEmpty(matchVals)` branch that was the suppression itself — derived unconditionally from a non-empty `matchResults`, it is unreachable rather than merely unused.

Specs drop the `shouldAbstain` cases and assert the uninterpretable result where they asserted a suppressed one. General-purpose `v2` fixtures move to `v5` so they keep exercising the path they were written for; `test/libs/ajax/Match.spec.ts` keeps its stamp because it never reaches classification.

Lint, type-check and the full suite are clean.

[DT-3940]: https://broadworkbench.atlassian.net/browse/DT-3940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fboulnois (approved):
Suggestions:

**1. Absent `algorithmVersion` now silently blanks the DAC's suggestion:** `src/utils/BucketUtils.ts:189`

Before this change a row with `algorithmVersion` undefined was read as backfilled-v1 and, for a matchable data use, rendered Yes/No. Now it renders "System match unavailable for this algorithm version". `MatchResult.algorithmVersion` is still declared optional at `src/types/model.ts:962`, so nothing in the type or code enforces the commit's assumption that Consent never omits the stamp anymore. If any Consent code path, or an un-migrated environment, returns a match without a version, every bucket containing it loses its system suggestion with no error surfaced to anyone. Either make `algorithmVersion` required in `MatchResult` so a missing stamp fails the contract visibly, or add a comment here recording that the `dataset_id` migration is the guarantee being relied on.

**2. Abstain plus true match is reported as "matched both true and false":** `src/utils/BucketUtils.ts:275`

A bucket with `matchResults` of `[{match:true, abstain:false, algorithmVersion:'v5'}, {match:false, abstain:true, algorithmVersion:'v5', rationales:['The dataset has an Other primary Data Use and requires manual review.']}]` produces `matchVals=[true,false]`, falls to the else branch, and yields "Unable to determine a system match" with a rationale claiming the algorithm matched both true and false. It did not match false; it abstained. `abstain` (line 243) is computed before the branch but only consulted inside the `length === 1` case, so the DAC never sees the real reason and the abstention rationale Consent supplied is discarded. Pre-existing, but this PR reworks exactly this block and deleting the `isEmpty(matchVals)` guard leaves this as the only remaining punt path.

**3. Unrecognized-version rationale names interpretable versions too:** `src/utils/BucketUtils.ts:229`

`bucketForVersions('v5','v2')` (already a test case) produces "This match was recorded by algorithm version v5, v2, which DUOS cannot interpret." A DAC member reads that v5 is uninterpretable, which is false and will generate support questions once v1/v2 are the only rows hitting this path. Filter `versions` to those where `classifyOneVersion(v) === 'unrecognized'` before joining, or reword to "includes results from version(s) X that DUOS cannot interpret".

**4. `AlgorithmVersionHandling` collapses to a boolean predicate:** `src/utils/BucketUtils.ts:187`

With only `'backend-abstains' | 'unrecognized'` left, `new Set(versions.map(classifyOneVersion)).size === 1 ? [...handlings][0] : 'unrecognized'` is exactly `versions.every(isBackendAbstaining)`. The doc comment on line 196 ("Versions mixed across match runs matter only when they disagree on how the results are read") is now vacuous since there is only one accepted reading. Replace the type, both functions, and the comment with `const isInterpretableVersion = (v?: string) => ...` and `if (!versions.every(isInterpretableVersion))` at the call site. `handling` is used exactly once.

**5. Reworded provenance comment is ungrammatical and mislabels the codes:** `src/hooks/useLibraryPageState.ts:47`

The sentence "the population and consent restrictions the corpus carries without them" does not parse (what is "without them"?). The entries outside `SecondaryDataUseTerms` are NCTRL/NAGR/NCU/RS-G/RS-PD (all in `consentTranslations`, `src/libs/dataUseTranslation.ts:218+`) and POP-M/POP-F/POP-PD (population). Nothing in the list is a consent restriction; IRB/COL/etc. were never labeled here. Previously POP-* pointed at `AbstainDataUseCodes`, a code-level list a reader could grep; now "the corpus carries" is unverifiable from the repo. Name the actual source (e.g. the dbGaP/GA4GH DUO population codes the dataset index stores) so the next person can check whether POP-* is still real vocabulary.

**6. One fixture still uses `algorithmVersion: 'v2'`:** `test/libs/ajax/Match.spec.ts:31`

The commit message says "general-purpose fixtures move to v5 so they keep exercising the path they were written for", and the other three specs were updated, but this one was missed. It is an ajax pass-through test so nothing fails today. The cost is that the canonical `MatchResult` example in the test tree now models data the app deliberately refuses to interpret, which will mislead the next person who copies it into a `BucketUtils` or `DarCollectionReview` test and gets "System match unavailable" instead of Yes/No.

kevinmarete:
Thanks — took 1, 3, 4 and 6; 5 went in on the last push.

**1** Added the comment rather than tightening the type. `MatchResult` mirrors what Consent returns, so making `algorithmVersion` required would move the fiction to the deserialization boundary instead of removing it. The predicate now records that the `dataset_id` migration is the guarantee being relied on.

**3** The rationale lists only the versions that failed the check, so `v5, v2` now reads "…includes results from algorithm version v2…". Asserted that v5 is absent.

**4** Replaced with `isInterpretableVersion` and `versions.filter(...)` at the call site; the type, both classify functions and the vacuous comment are gone.

**5** Fixed, with one correction: NCU is not in `consentTranslations` — that map carries NPU (`src/libs/dataUseTranslation.ts:218+`). The comment now credits `consentTranslations` for NCTRL/NAGR/RS-G/RS-PD and describes NCU and POP-* as codes neither source carries. I did not name a source for POP-*: after the deletion there is no in-repo list, and I would rather not record a provenance claim I cannot verify here.

**2** Left as is. It is pre-existing and changes what the DAC sees for abstain-plus-match buckets, which is past removing the suppression — happy to fold it into a follow-up if you would rather it ship with this.

fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3946, opened 2026-09-18
[DT-4126] Correct two labels on the study page

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4126](https://broadworkbench.atlassian.net/browse/DT-4126)

### Summary

Two non-blocking review comments from kevinmarete, landed together.

**Heading.** The similar-studies section said "based on Data Type", but the query
matches on either signal:

    AND (NULLIF(s.pi_name, '') = source.pi_name OR s.data_types && source.data_types)

so a study could appear there for sharing a PI with no data-type overlap at all,
and the heading could not explain why. It names both now.

**Badges.** `StudyAssetCountBadges` built every label as `${count} ${label}`, so
a single item read "1 Datasets". The caller supplies singular and plural rather
than the component trimming an "s" — these four pluralize regularly, but an asset
that does not would read as a bug. The component had no spec at all; it has one
now, covering the singular, the plural and the zero cases it already handled.

----

- Verify all tests go green


kevinmarete (approved):
Nice 👍🏾

fboulnois (approved):
👍

PR #3947, opened 2026-09-18
[DT-4127] Let the recommendation sections own their queries

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4127](https://broadworkbench.atlassian.net/browse/DT-4127)

### Summary

The two recommendation components had their data fetched by
`StudyDetailsContent` and threaded in as six props, so every pending, success
and refetch on either query re-rendered the whole page — the datasets `DataGrid`
included. Every other section already takes a `studyId` and owns its hook; these
two now match, and the page loses four imports, two hooks and six props.

No behaviour change, and **no test needed editing to keep passing**, which was
the point: the sections render the same headings under the same anchors from the
same queries.

Two things deliberately preserved:

- `StudyRecommendationCarousel` still has no `= []` default on `recommendations`.
  `undefined` means nothing has loaded and `[]` means loaded and empty, which is
  what its error guard reads.
- Both hooks keep the `select` filter that stops a study recommending itself.

Addresses @fboulnois' fourth suggestion on #3933, split out as agreed since it is
a refactor rather than a defect.

Based on DT-4126 rather than the recommendations branch, so the corrected
"Recommended Studies by Data Type or PI" heading moves into the new section
component already fixed — the two would otherwise conflict on that line.

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

fboulnois (approved):
👍

rushtong (approved):
Nice cleanup 👍🏽

PR #3949, opened 2026-09-21
[DT-4176] Create the default metrics `AbortSignal` per call instead of once per module

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-4176

### Summary
Discovered while reviewing and updating the BFF Phase 6 Rollout stories.

- `src/libs/ajax/Metrics.ts` created one `AbortSignal.timeout(30000)` at module evaluation and used it as the default `signal` parameter of `captureEvent`, `syncProfile` and `identify`. `AbortSignal.timeout` fires once, so 30 s after page load the shared signal was permanently aborted and every later default-signal call aborted immediately.
- `captureEvent` swallows errors with `.catch(() => {})`, so this was silent: every Bard event fired more than 30 s after page load was lost.
- The shared signal is now a `defaultSignal()` factory that returns a fresh 30 s signal per call. The 30 s timeout is unchanged. Callers that pass their own signal are unaffected.
- `test/libs/ajax/Metrics.spec.ts` gains a `default timeout signal` block. Vitest's fake timers do not drive Node's internal `AbortSignal.timeout`, so the tests stand in a `setTimeout`-backed equivalent, load `Metrics` fresh under the fake clock (`vi.resetModules` + `vi.doMock` sharing one hoisted `retryFetchPost` mock), and then:
  - advance 30 001 ms and assert `captureEvent` still POSTs with a non-aborted signal (fails on `develop`: `expected true to be false`)
  - assert two calls get distinct signals and each times out 30 s after its own call (fails on `develop`: same signal object)
  - assert a caller-supplied signal is passed through unchanged for `captureEvent` and `identify`

---

- Verify all tests go green


otchet-broad (approved):
👍

fboulnois (approved):
looks good 👍

